PR #2, opened 2015-10-15
tests wip

v-stepanov:
@jkliff @tmuehl @valgog welcome to review PR :)


src/test_hack.py
```diff
@@ -0,0 +1,116 @@
...
+import json
+import unittest
+
+import test_common
+from test_common import TEST_TOPIC, TEST_PARTITIONS_NUM
+
+
+# this test case doesn't care if there are any events present or not
+# for api version 0.3
+class EventstoreDataIndependentTestCase(unittest.TestCase):
```
hjacobs: you should check out py.test --- it's much nicer to use than the old-style TestCase stuff : http://pytest.org/latest/

Examples: https://github.com/zalando-stups/senza/tree/master/tests (not the best test code :smirk: )

v-stepanov: Ok, thanks for the hint. Will check that.

PR #10, opened 2015-10-23
Stream from multiple partitions

v-stepanov:
@valgog @tmuehl Please take a look

### DIFF
--- a/nakadi/event_stream.py
+++ b/nakadi/event_stream.py
@@ -9,70 +9,93 @@ import sys
 
 def create_stream_generator(kafka_pool, topic, partition, start_from, opts):
 
+    partitions_offsets = [(0, 0), (1, 0), (2, 0)]
+
+    topics = {}
+    for partition_offset in partitions_offsets:
+        topics[(topic, partition_offset[0])] = partition_offset[1]
+
+    partitions = [partition_offset[0] for partition_offset in partitions_offsets]
+
     start = datetime.datetime.now()
 
     # stream
     def generator():
-        messages_read = 0
-        keep_alive_in_a_row = 0
-        current_batch = []
-        batch_start_time = datetime.datetime.now()
 
+        keep_alive_in_a_row = 0
+
+        messages_read =  0
+        current_batch = dict(zip(partitions, [[] for partition in partitions]))
+
+        batch_start_time = datetime.datetime.now()
         with kafka_pool.kafka_client() as client:
 
-            consumer = KafkaConsumer((topic, partition, start_from),
-                         kafka_client = client,
-                         auto_commit_enable = False,
-                         consumer_timeout_ms = 200)
+            consumer = KafkaConsumer(topics,
+                                     kafka_client=client,
+                                     auto_commit_enable=False,
+                                     consumer_timeout_ms=200,
+                                     fetch_message_max_bytes= 1024) # 128 kb max batch size for topic/partition
 
             while True:
                 try:
                     message = consumer.next()
-
                     # if we read the message - reset keep alive counter
                     keep_alive_in_a_row = 0
 
                     # put message to batch
                     messages_read += 1
                     message_json = json.loads(message.value.decode('utf-8'))
-                    current_batch.append(message_json)
+                    current_batch[message.partition].append(message_json)
 
                 except ConsumerTimeout:
                     pass
 
                 # check if it's time to send the batch
                 time_since_batch_start = datetime.datetime.now() - batch_start_time
-                if time_since_batch_start.total_seconds() >= opts['batch_flush_timeout'] != 0 or len(current_batch) >= opts['batch_limit']:
+                max_batch_size = max([len(batch_per_partition) for batch_per_partition in current_batch.values()])
+                latest_offsets = consumer.offsets("fetch")
+                if time_since_batch_start.total_seconds() >= opts['batch_flush_timeout'] != 0 or max_batch_size >= opts['batch_limit']:
+                    for partition in partitions:
+                        topic_partition = (topic.encode('UTF-8'), partition)
+                        # send the messages we could read so far
+                        if len(current_batch[partition]) > 0:
+                            stream_message = __create_stream_message(partition, latest_offsets[topic_partition], current_batch[partition])
+                            with __measure_time(current_batch[partition], stream_message):
+                                yield stream_message
 
-                    # send the messages we could read so far
-                    if len(current_batch) > 0:
-                        stream_message = __create_stream_message(partition, start_from + messages_read, current_batch)
-                        with __measure_time(current_batch, stream_message):
-                            yield stream_message
+                        # just send the keep alive
+                        else:
+                            yield __create_stream_message(partition, latest_offsets[topic_partition])
 
-                    # just send the keep alive
-                    else:
-                        # if we hit keep alive count limit - close the stream
+                    # if we hit keep alive count limit - close the stream
+                    if max_batch_size == 0:
                         if keep_alive_in_a_row >= opts['batch_keep_alive_limit'] != -1:
                             break
-
-                        # send keep alive message
                         keep_alive_in_a_row += 1
-                        yield __create_stream_message(partition, start_from + messages_read)
 
                     # init new batch
-                    current_batch = []
+                    current_batch = dict(zip(partitions, [[] for _ in partitions]))
                     batch_start_time = datetime.datetime.now()
+
+                    yield '\n\n\n'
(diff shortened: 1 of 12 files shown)


v-stepanov:
please don't merge it yet, as I have to fix/extend tests.

nakadi/event_stream.py
```diff
@@ -7,72 +7,94 @@
 import sys
 
 
-def create_stream_generator(kafka_pool, topic, partition, start_from, opts):
+def create_stream_generator(kafka_pool, topic, cursors, opts):
+
+    topics = {}
+    for cursor in cursors:
+        topic_partition = (topic, int(cursor['partition']))
+        topics[topic_partition] = int(cursor['offset'])
```
valgog: I would suggest an explicit check for the format of the partition and offset, though it could be ok for the hack

valgog: Another question, do we want to scramble the offsets so that they do not look like numbers?

Just base64 encoding of the numeric value of the offset would be sufficient and easy enough.

v-stepanov: @valgog cursors format validation already happens here https://github.com/zalando/nakadi/pull/10/files#diff-c2048a92d3dde8d2820ec40ed201fb19R33

nakadi/event_stream.py
```diff
@@ -7,72 +7,94 @@
...
     # stream
     def generator():
-        messages_read = 0
+
         keep_alive_in_a_row = 0
-        current_batch = []
-        batch_start_time = datetime.datetime.now()
 
+        messages_read = 0
+        current_batch = dict(zip(partitions, [[] for _ in partitions]))
```
valgog: my brain explodes! :D do use comments for initialisations like this :speak_no_evil:

nakadi/event_stream.py
```diff
@@ -7,72 +7,94 @@
...
-                    if len(current_batch) > 0:
-                        stream_message = __create_stream_message(partition, start_from + messages_read, current_batch)
-                        with __measure_time(current_batch, stream_message):
-                            yield stream_message
-
-                    # just send the keep alive
-                    else:
-                        # if we hit keep alive count limit - close the stream
+                batch_sum = sum([len(batch_per_partition) for batch_per_partition in current_batch.values()])
+                latest_offsets = consumer.offsets("fetch")
```
valgog: Is this offset pointing to the next event already?

nakadi/event_stream.py
```diff
@@ -7,72 +7,94 @@
...
-                    # send the messages we could read so far
-                    if len(current_batch) > 0:
-                        stream_message = __create_stream_message(partition, start_from + messages_read, current_batch)
-                        with __measure_time(current_batch, stream_message):
-                            yield stream_message
-
-                    # just send the keep alive
-                    else:
-                        # if we hit keep alive count limit - close the stream
+                batch_sum = sum([len(batch_per_partition) for batch_per_partition in current_batch.values()])
```
valgog: Do we need to keep recalculating the batch sum over and over? Do not we have a possibility just have a counter for that?

nakadi/event_stream.py
```diff
@@ -7,72 +7,94 @@
...
+                latest_offsets = consumer.offsets("fetch")
+
+                if time_since_batch_start.total_seconds() >= opts['batch_flush_timeout'] != 0 or batch_sum >= opts['batch_limit']:
```
valgog: are you sure you do not need to convert the int value of opts map to ints?

nakadi/event_stream.py
v-stepanov: @valgog the comment will not help :) it's just too complicated, I will rewrite that.

nakadi/event_stream.py
v-stepanov: @valgog yes

nakadi/event_stream.py
v-stepanov: @valgog +, will refactor

nakadi/event_stream.py
v-stepanov: @valgog I think I don't quite get you. The values of opts map are ints already.

nakadi/hack.py
```diff
@@ -120,71 +122,44 @@ def __get_partitions_offsets(topic):
...
+        return {'detail': 'missing required query parameter "%s"' % e.parameter}, 400
+
+    # check that topic exists
     if not __topic_exists(topic):
         return {'detail': 'topic not found'}, 404
-    if not __partition_exists(topic, partition):
-        return {'detail': 'partition not found'}, 404
+
+    # get cursors to start reading from
+    cursors_str = flask.request.headers.get('x-nakadi-cursors')
```
valgog: I would use a constant here :D

nakadi/hack.py
```diff
@@ -120,71 +122,44 @@ def __get_partitions_offsets(topic):
...
+    except NotIntegerParameterException as e:
+        return {'detail': '"%s" query parameter should be an integer number' % e.parameter}, 400
+    except RequiredParameterNotFoundException as e:
+        return {'detail': 'missing required query parameter "%s"' % e.parameter}, 400
+
+    # check that topic exists
     if not __topic_exists(topic):
         return {'detail': 'topic not found'}, 404
-    if not __partition_exists(topic, partition):
-        return {'detail': 'partition not found'}, 404
```
valgog: I cannot find the code that would check if the partitions, that you fetch from `x-nakadi-cursors` header, are valid,

nakadi/hack.py
```diff
@@ -120,71 +122,44 @@ def __get_partitions_offsets(topic):
...
+    else:
+        try:
+            cursors = get_cursors(cursors_str)
+        except WrongCursorsFormatException:
+            return {'detail': '"x-nakadi-cursors" header has wrong format'}, 400
 
     # returning generator in response will create a stream
-    stream_generator = event_stream.create_stream_generator(kafka_pool, topic, partition, start_from, stream_opts)
+    stream_generator = event_stream.create_stream_generator(kafka_pool, topic, cursors, stream_opts)
     return flask.Response(stream_generator, mimetype = 'text/plain', status = 200)
```
valgog: oh, you sure this mimetype is `text/plain`?

nakadi/hack.py
v-stepanov: @valgog right, this check is still missing. Will add that.

nakadi/kafka_pool.py
```diff
@@ -59,6 +60,8 @@ def kafka_client(self):
                 logging.warning("Exception when closing kafka client: %s", e)
             raise kafka_error
         except:
+            logging.warning("Exception occured while kafka client was out of pool:")
+            traceback.print_exc()
```
valgog: oh, what are you trying to do here? Your could catch the exception using the following syntax:

```python
try:
  ...
except Exception as e:
  ...
```

nakadi/hack.py
```diff
@@ -120,71 +122,44 @@ def __get_partitions_offsets(topic):
...
-    else:
-        partition = int(partition)
-
-    # check that topic and partition exist
+    try:
+        stream_opts['batch_limit'] = get_int_parameter('batch_limit', flask.request, False, 1)
+        stream_opts['batch_flush_timeout'] = get_int_parameter('batch_flush_timeout', flask.request, False, 0)
+        stream_opts['batch_keep_alive_limit'] = get_int_parameter('batch_keep_alive_limit', flask.request, False, -1)
+        stream_opts['stream_limit'] = get_int_parameter('stream_limit', flask.request, False, 0)
+        stream_opts['stream_timeout'] = get_int_parameter('stream_timeout', flask.request, False, 0)
```
valgog: Please have a look on the examples, that @jmcs pointed out

https://github.com/zalando/lizzy/blob/master/lizzy/swagger/lizzy.yaml#L74
https://github.com/zalando/lizzy/blob/master/lizzy/api.py#L74

nakadi/hack.py
v-stepanov: @valgog Well, it's definitely not json (as we have \n separators), so for now text/plain would be fine I think

nakadi/event_stream.py
valgog: Yes, `connexion` takes care or that types based on the Swagger definition of the options type. Did not know about that when writing the comment :D

nakadi/hack.py
valgog: I personally would try it with JSON and some additional suffix... :) or some compleatly own content type, including the version of the serialization...

PR #13, opened 2015-10-27
Update EventSchema.md

unlaut:
Some ideas: 
1 We should have a fixed place where the information determining the Kafka topic goes. I call it content-type here. The Create/Update/Delete OP goes to a separate field. (data_op_type).
2 Resource Events have three extra mandatory fields (data, data_key and data_op_type) that business events are not allowed to use, which allows for more effective validation.
3 Business Events are not allowed to create sub-objects.

### DIFF
--- a/docs/EventSchema.md
+++ b/docs/EventSchema.md
@@ -6,36 +6,56 @@ of that format is to have a transparent way to exchange events between distribut
 
 DRAFT JSON-Schema definitions:
 ```yaml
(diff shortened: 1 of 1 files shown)


valgog:
We have now a common understanding, that event format in that details, is not of concern for Nakadi, it should be defined and owned by Saiki. Probably we will return to this when the validation plugin support is implemented in Nakadi

PR #15, opened 2015-11-03
(fixed #14) added logging of request and response for some cases; ext…

### LINKED ISSUE #14: Improve logging of errors of push-endpoint
We need to improve the logging of push-endpoint. Currently if error-response is created on connexion-level we don't have too much log data about that. We just have simple access log from gunicorn:

> [02/Nov/2015:13:03:10 +0000] "POST /topics/eventstore.article.16/events HTTP/1.1" 400 152 "-" "Java/1.8.0_05"

But in this case we need at least to see the request body.
First we need to check if it is possible to configure connexion to log request body in a case of creation of error-response.

v-stepanov:
@valgog @jkliff @tmuehl feedback welcomed

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,9 @@ FROM zalando/python:3.4.0-2
 COPY requirements.txt /
 RUN pip3 install -r /requirements.txt
 
-# copy project code
+# copy project code and gunicorn config
 ADD nakadi /nakadi
+COPY gunicorn_conf.py /
 # remove python cache files
 RUN find ./nakadi | grep -E "(__pycache__|\.pyc|\.pyo$)" | xargs rm -rf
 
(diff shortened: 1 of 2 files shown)


v-stepanov:
Of course the best way would be to dump request body only in a case of error status_code, but due to implementation of gunicorn, the request body is not available at the post_request hook.

tmuehl:
I like having the gunicorn config in a separate file.

v-stepanov:
@tmuehl yeah, mee too. But this branch never went to master as it doesn't actually work on AWS. So the separation of gunicorn didn't go to master as well :)

tmuehl:
Why does it now work on AWS?

v-stepanov:
I don't know. The logging of response body just didn't work there :( However it worked pretty fine locally.

valgog:
I will close the PR because we will not investigate on that further.

PR #21, opened 2015-11-18
Do not decode-encode json when building batch

CyberDem0n:
Besides that move duplicated code which creates batch into separate function

### DIFF
--- a/nakadi/event_stream.py
+++ b/nakadi/event_stream.py
@@ -1,6 +1,5 @@
 from contextlib import contextmanager
-import datetime
-import json
+import time
 import logging
 from kafka import KafkaConsumer
 from kafka.common import ConsumerTimeout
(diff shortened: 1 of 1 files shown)


nakadi/event_stream.py
```diff
@@ -80,47 +77,31 @@ def generator():
...
 
-def __create_stream_message(partition, offset, events = None, topology = None):
-    message = {
-        'cursor': {
-            'partition': str(partition),
-            'offset': str(offset)
-        }
-    }
+def __create_stream_message(partition, offset, events=None, topology=None):
+    message = ['{"cursor":{"partition":"', str(partition), '","offset":"', str(offset), '"}']
```
valgog: Is there a high performance method to escape a string safely as a JSON string when marshalling offsets?

PR #24, opened 2015-11-19
added endpoint to post batch of events

v-stepanov:
@CyberDem0n @jkliff @tmuehl @valgog


nakadi/swagger.yaml
```diff
@@ -206,6 +206,50 @@ paths:
           description: Not available
           schema:
             $ref: '#/definitions/Problem'
+  '/topics/{topic}/events/batch':
```
v-stepanov: Any ideas of endpoint URL ?

PR #41, opened 2015-12-11
Change the swagger to what is actually implemented

tmuehl:
Adopt to the swagger file what is actually implemented in the python version

@valgog @v-stepanov @jkliff @CyberDem0n

### DIFF
--- a/nakadi/swagger.yaml
+++ b/nakadi/swagger.yaml
@@ -216,10 +216,6 @@ paths:
         - name: event
           in: body
           description: Even being sent.
-          schema:
-            type: array
-            items:
-              $ref: '#/definitions/Event'
       responses:
         '201':
           description: Event submitted.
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/41" target="_blank">ReviewNinja

nakadi/swagger.yaml
```diff
@@ -227,6 +216,10 @@ paths:
         - name: event
           in: body
           description: Even being sent.
+          schema:
```
v-stepanov: This will make connexion validate the request body and it will fail as it is not a JSON

v-stepanov:
Maybe we can just have some thumbsup as a approval? Ninja for just approvals is a big overhead

v-stepanov:
:+1:

v-stepanov:
hmm, it seems that thumbsup automatically adds a ninja star

tmuehl:
Yea, that would be great

PR #43, opened 2015-12-11
Nakadi event bus low level api review

tmuehl:
This is a review request for the API Guild.

This is the first stage of the review, where we minimised the file to contain only Nakadi Low-Level API related endpoints.

This API provides:

* Standardised event format (see `#/definitions/Event`)

* Low-level API
  > (`LoLa`) allows the consumers to get event streams from the specified partitions
  > for specified cursor positions. That means, that consumers should implement a protocol, 
  > to distribute partitions, that are being consumed between themselves.
  > They also are responsible for storing current cursor positions per partition.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -2,18 +2,83 @@ swagger: '2.0'
 info:
   title: Nakadi Event Bus API Definition
   description: |
-    -------------------------------
-    Nakadi Event Bus API Definition
+    -----------
+    Definitions
+    -----------
+    Nakadi at its core aims at being a generic and content-agnostic event broker with a convenient API.
+    In doing this, Nakadi abstracts away, as much as possible, details of the backing messaging
+    infrastructure. The single currently supported messaging infrastructure is Kafka (Kinesis is planned
+    for the future).
+
+    In Nakadi every Event has an EventType, and a **stream** of Events is exposed for each registered
+    EventType.
+
+    An EventType defines properties relevant for the operation of its associated stream, namely:
+
+    * The **schema** of the Event of this EventType. The schema defines the accepted format of Events of an
+    EventType and will be, if so desired, enforced by Nakadi. Usually Nakadi will respect the schema for the
+    EventTypes in accordance to how an owning Application defines them.
+      **Note:** *Currently the specification of the schema must be pushed into Nakadi on EventType creation;
+      in the future, assuming that Applications will expose the schema for its owned resources, Nakadi might
+      support fetching the schema directly from them.*
+      Nakadi might define globally minimal schema complience requirements (i.e. some fields that must be
+      present on schemas of all EventTypes).
+
+    * The expected **validation** and **enrichment** procedures upon reception of an Event.
+      (*TODO: define validation and enrichment*)
+
+    * The **ordering** expectations of Events in this stream. (*TODO: define ordering not in terms of
+    partitioning*)
+
+    Except for defined enrichment rules, Nakadi will never manipulate the content of any Event.
+
+    Clients of Nakadi can be grouped in 2 categories: **Application owners** and **Clients** (clients in turn
+    are both **Producers** and **Consumers** of Events). Application owners interact with Nakadi via the
+    **Schema Registry API** for the definition of EventTypes, while Clients via the **Stream API** for
+    submission and reception of Events. A low level **Topic API** is available, providing full control and
+    responsibility of position tracking and partitioning (and therefore ordering) to the Clients.
+
+    The consumption of Events proceeds via the establishment of a named **Subscription** to an EventType.
+    Subscriptions are persitent relationships from an Application (which might have several instances) and the
+    stream of an EventType, whose consumption tracking is managed by Nakadi, freeing Consumers from any
+    responsibility in tracking the current position on a Stream.
+
+
+    Scope and status of this document
     -------------------------------
 
-    * Low-level API
-    * Stantardised event format (see #/definitions/Event)
-    * Partition strategy: event store will infer the partition from the `partitioning_key` in the message body
+    The present API specification is in **draft** state and is subject to change.
 
-  version: '0.4'
+    In this document, ready for review are included:
+    * The Schema Registry API, including configuration possibilities for the Schema, Validation, Enrichment
+    and Partitioning of Events, and their effects on reception of Events.
+
+    * The Stantardised event format (see #/definitions/Event) (Note: at a later moment this will be
+    configurable and not be inherent part of this API).
+
+    * Topic API (low level).
+
+    Other aspects of the Event Bus are at this moment to be defined and otherwise specificied, not included in
+    this version of this specification.
+
+    Notable omissions here are:
+    * The contract between Nakadi and clients in the context of the high-level API (i.e. the process for
+    clients to establish subscriptions).
+
+    * Submission of events to topics in batched manner.
+
+    * Validation and enrichment procedures.
+
+    * Security scopes (OAuth) for the different operations
+
+    * Validation and Enrichment Strategies available in Nakadi are out of this document an can be consulted in
+    an external document (TBD link to document).
+
+
+  version: '0.5'
   contact:
     name: Team Aruha @ Zalando
-    email: nakadi-users@zalando.de
+    email: team-aruha+nakadi-maintainers@zalando.de
 schemes:
   - https
 consumes:
@@ -34,22 +99,252 @@ paths:
       summary: Get monitoring metrics
       responses:
         '200':
-          description: Metrics data
+          description: Ok
           schema:
             $ref: '#/definitions/Metrics'
         '401':
-          description: User is not authenticated
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Server error
           schema:
             $ref: '#/definitions/Problem'
         '503':
           description: Not available
           schema:
             $ref: '#/definitions/Problem'
+  /event-types:
+    get:
+      tags:
+        - schema-registry-api
+      description: Returns a list of all registered `EventType`s
+      responses:
+        '200':
+          description: Ok
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/EventType'
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Server error
+          schema:
+            $ref: '#/definitions/Problem'
+
+    post:
+      tags:
+        - schema-registry-api
+      description: |
+        Creates a new `EventType`. Upon creation the name MUST be unique. Failing to do so will fail the
+        request with a 422 Unprocessable entity.
+
+        **Implementation note:** The creation of an EventType implicitly creates the structures in the backing
+        messaging implementation needed for the reception and persistency of the Events. Considering that at
+        this time only Kafka is used, this corresponds to the creation of a Topic. If so desired, clients can
+        interact directly with the topic using the low level API (for existing restrictions see the
+        corresponding methods on the topic-api).
+
+        The fields validation-strategies, enrichment-strategies and partition-resolution-strategy have all an
+        effect on the incoming Event of this EventType. For its impacts on the reception of events please
+        consult the Event submission API methods.
+
+        * Validation strategies define an array of `EventValidationStrategy` to be evaluated on reception of
+        an `Event` of this `EventType`. Details of usage can be found in an external document (TBD link to
+        document).
+        **Notes**: Some EventValidationStrategies might support additional parameters. The behaviour is
+        specific to each strategy and one is suggested to consult its `doc` field for detailes of operation.
+        The default `SchemaValidationStrategy` supports passing fields to be ignored if the qualifier matches
+        the input.
+
+        * TDB Enrichment strategy
+
+        * TDB Partition resolution strategy
+
+        * The schema of an `EventType` is defined as an `EventTypeSchema`. Currently only json-schema is
+        supported. Using `EventTypeSchema.type` other than json-schema or passing a EventTypeSchema.schema
+        that is invalid with respect tothe type will fail this creation with a 422 Unprocessable entity.
+        Detailes will be included in the response's Problem.
+
+        Nakadi MIGHT impose necessary schema, validation and enrichment minimal configurations that MUST be
+        followed by all EventTypes (examples include: validation rules to match the schema; enriching every
+        Event with the reception date-type; adhering to a set of schema fields that are mandatory for all
+        EventTypes). **The mechanism to set and inspect such rules is not defined at this time and might not
+        be exposed in the API.**
+
+      parameters:
+        - name: event-type
+          in: body
+          description: EventType to be created
+          schema:
+            type: object
+            $ref: '#/definitions/EventType'
+          required: true
+      responses:
+        '200':
+          description: Ok
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '422':
+          description: Unprocessable Entity
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Server error
+          schema:
+            $ref: '#/definitions/Problem'
+
+
+  /event-types/{name}:
+    get:
+      tags:
+        - schema-registry-api
+      description: |
+        Returns the `EventType` identified by its name.
+      parameters:
+        - name: name
+          in: path
+          description: Name of the EventType to load.
+          type: string
+          required: true
+      responses:
+        '200':
+          description: Ok
+          schema:
+            $ref: '#/definitions/EventType'
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '404':
+          description: EventType not found
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Server error
+          schema:
+            $ref: '#/definitions/Problem'
+
+    put:
+      tags:
+        - schema-registry-api
+      description: |
+        Updates the `EventType` identified by its name. Behaviour is the same as creation of `EventType` (POST
+        /event-type) except where noted below.
+
+        The name and the topic fields cannot be changed. Attempting to do so will result in a 422 failure.
+
+        At this moment changes in the schema are not supported and will produce a 422 failure. (TODO: define
+        conditions for backwards compatible extensions in the schema)
+      parameters:
+        - name: name
+          in: path
+          description: Name of the EventType to update.
+          type: string
+          required: true
+        - name: event-type
+          in: body
+          description: EventType to be updated.
+          schema:
+            type: object
+            $ref: '#/definitions/EventType'
+          required: true
+
+      responses:
+        '200':
+          description: Ok
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '404':
+          description: EventType not found.
+          schema:
+              $ref: '#/definitions/Problem'
+        '422':
+          description: Unprocessable Entity
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Server error
+          schema:
+            $ref: '#/definitions/Problem'
+
+  /event-types/{name}/events:
+    post:
+      description: |
+        Publishes one or more events of this EventType.
+
+        Reception of Events will always respect the configuration of its EventType w.r.t. validation,
+        enrichment and partitioning. The steps performed on reception of incoming message are:
+
+        1. Every validation rule specified in the `EventType` will be checked in order against the incoming
+        Events. Validation rules are evaluated in the order they are defined and the Event is **rejected**
+        in the first case of failure with a status of 400 Bad request. If the offending validation rule
+        provides information about the violation it will be included in the Problem response.
+        If the `EventType` defines schema validation it will be performed at this moment.
+        **TBD behaviour in case of violation within a stream of events**)
+
+        1. Once the validation succeeded, the content of the Event is updated according to the enrichment
+        rules in the order the rules are defined in the `EventType`.
+        No preexisting value might be changed (even if added by an enrichment rule). Violations on this will
+        force the immediate **rejection** of the Event with a 409 Conflict status. The invalid overwrite
+        attempt will be included in the respose's Problem object.
+
+        1. The incoming Event's relative ordering is evaluated according to the rule on the
+        `EventType`. Failure to evaluate the rule will **reject** the Event.
+        **Restriction**: POSTing Events based on its `EventType` requires it to have a
+        `PartitioningStrategy` defined. Violating this condition will produce a 405 Not allowed.
+
+      parameters:
+        - name: name
+          in: path
+          type: string
+          description: Name of the EventType
+          required: true
+      responses:
+        '200':
+          description: Ok
+        '400':
+          description: Bad request
+          schema:
+            $ref: '#/definitions/Problem'
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '404':
+          description: EventType not found.
+          schema:
+              $ref: '#/definitions/Problem'
+        '405':
+          description: Not allowed.
+          schema:
+              $ref: '#/definitions/Problem'
+        '409':
+          description: Conflict.
+          schema:
+              $ref: '#/definitions/Problem'
+        '422':
+          description: Unprocessable Entity
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Server error
+          schema:
+            $ref: '#/definitions/Problem'
+
   /topics:
     get:
       description: Lists all known `Topics` of this Nakadi cluster
       tags:
         - monitoring
+        - management-api
       responses:
         '200':
           description: The request was successful. Returns a list `Topic`s in the response body.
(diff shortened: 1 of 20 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/43" target="_blank">ReviewNinja

valgog:
@tfrauenstein please have a look on the pull request. Do you think it is ok to submit it like this to the API guild?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+          required: true
+        - name: event
+          in: body
+          description: The body contains an array of events being sent.
+          required: true
+          schema:
+            type: array
+            description: An array of events to be pushed to Nakadi
+            items:
+              $ref: '#/definitions/Event'
```
valgog: This is very controversial point here, as we are expecting `POST` to inject an array of events. This is needed to enable batching. So posting one event will be done as an array of one element.

valgog: Maybe the better way would be to use `Content-Type` here, that defines, what kind of an object is being passed in. Is there some standard, that would define an batch operation?

tfrauenstein:
yes, let's submit this pull request for review! Thx.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        $ref: '#/definitions/EventMetaData'
+  EventMetaData:
+    type: object
+    properties:
+      id:
+        type: string
+        format: uuid
+      created:
+        type: string
+        format: data-time
```
jhorstmann: Typo, should be `date-time`.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+      root_id:
```
jhorstmann: Some documentation for `root_id`, `parent_id` and `scopes` would be helpful.

bocytko: +1 especially for scopes

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+definitions:
+  Event:
+    type: object
+    description: |
+      This is the most general representation of an event, that can be processed
+      by Nakadi.
+
+      It should be used as a base definition for all events, that flow through
+      Nakadi by extending attributes of this object type.
+    properties:
```
bocytko: So further properties can be added on the top-level, right? Would be cool to include this in the spec as well.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+
+
+# ################################### #
+#                                     #
+#             Definitions             #
+#                                     #
+# ################################### #
+
+definitions:
+  Event:
```
bocytko: There is no information on required vs. optional fields in the `Event`.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+  - https
+consumes:
+  - application/json
+produces:
+  - application/json
+securityDefinitions:
+  oauth2:
+    type: oauth2
+    flow: implicit
+    authorizationUrl: 'https://auth.example.com/oauth2/tokeninfo'
```
tfrauenstein: I think we should also authorize event bus access using scopes.
Would expect 2 scopes:  <stups-appl-id>.read and <stups-appl-id>.write

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        '401':
+          description: User is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '503':
+          description: Not available
+          schema:
+            $ref: '#/definitions/Problem'
+  /topics:
+    get:
```
tfrauenstein: How topics are created? Would also expect API support here.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        '503':
+          description: Not available
+          schema:
+            $ref: '#/definitions/Problem'
+  '/topics/{topic}/events':
+    get:
+      tags:
+        - low-level-api
+      summary: |
+        Starts a stream delivery for the specified partitions of the given topic.
```
tfrauenstein: I don't understand this resource method. How is the "specified partition" info passed? 
Via the x-nakadi-cursor header? If yes, the header need to be mandatory, not optional.

What is the difference to the GET /topics/{topic}/partitions/{partition} resource method?
Do we really need both LL methods?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+      parameters:
+        - name: topic
+          in: path
+          description: Topic where to get events from
+          type: string
+          required: true
+        - name: x-nakadi-cursors
+          in: header
+          description: Cursors pointing to partitions/offsets to start read from
+          required: false
```
tfrauenstein: See comment above.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+            $ref: '#/definitions/Problem'
+  '/topics/{topic}/events':
+    get:
+      tags:
+        - low-level-api
+      summary: |
+        Starts a stream delivery for the specified partitions of the given topic.
+        The tracking of the current position in the partitions and of which
+        partitions is being read is in the responsibility of the client.
+        No commits are needed.
```
tfrauenstein: It's not clear how offsets are handled here? Is there an auto-commit provided, and how does it look like?
If not, I would expect API commit support for partitions.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+          in: query
+          description: |
+            Maximum number of `Event`s in this stream. If 0 or undefined, will
+            stream indefinately.
+
+            Stream initialization will fail if `stream_limit` is lower than
+            `batch_limit`.
+          type: integer
+          format: int32
+          required: false
```
tfrauenstein: pls. add info on default

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+          description: Maximum number of keep-alive messages to get in a row before closing the connection. Unlimited by default.
+          type: integer
+          format: int32
+          required: false
+      responses:
+        '200':
+          description: |
+            Starts streaming to the client.
+            Stream format is a continuous series of
+            `SimpleStreamEvent`s separated by `\n`
```
tfrauenstein: This is a sequence of JSON objects which is itself not JSON, isn't it?
Why not returning an array of SimpleStreamEvent's here?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        '401':
+          description: Not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '404':
+          description: Not found. Typically in case of requesting from a non existing topic of partition.
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Not found. Typically in case of requesting from a non existing topic of partition.
```
tfrauenstein: Unclear: "requesting from a non existing topic of partition"
I assume it is returned if either the topic or the partition does not exist.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+          in: header
+          description: Cursors pointing to partitions/offsets to start read from
+          required: false
+          type: array
+          items:
+            type: string
+            format: '#/definitions/Cursor'
+        - name: batch_limit
+          in: query
+          description: Maximum number of `SimpleStreamElement`s in each chunk of the stream. If unspecified assumes default value 1 (i.e. each event is individually submitted).
```
jhorstmann: Is it `SimpleStreamElement` or `SimpleStreamEvent` or are those two separate concepts?

api/nakadi-event-bus-api.yaml
jhorstmann: Line delimited json (https://en.wikipedia.org/wiki/JSON_Streaming#Line_delimited_JSON) might be easier to deal with in a streaming fashion compared to arrays, at least for some parser implementation. The problem I see is that there is no standard content type for it, but I saw several mentions of using `application/x-json-stream`.

api/nakadi-event-bus-api.yaml
tfrauenstein: I see passing of 1 event as special case of passing event batches, so I would be fine with current proposal.
It should be explained, whether on servicer side streaming is applied -- so 1st events already get be pushed into queue even if last event of batch has not yet been passed by client.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+          required: true
+          schema:
+            type: array
+            description: An array of events to be pushed to Nakadi
+            items:
+              $ref: '#/definitions/Event'
+        - name: Content-Encoding
+          in: header
+          required: false
+          description: '"Allow the publisher to gzip compress the payload by setting this header to "gzip". For Example "Content-Encoding: gzip"'
```
tfrauenstein: fine. Why we do not support gzip to for the consumer methods controlled via Accept-Encoding header?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        '503':
+          description: Not available
+          schema:
+            $ref: '#/definitions/Problem'
+  '/topics/{topic}/partitions':
+    get:
+      tags:
+        - low-level-api
+        - monitoring
+      summary: Lists the `Partition`s for the given topic
```
tfrauenstein: summary: returns all partitions including its current offsets for a given topic.

api/nakadi-event-bus-api.yaml
valgog: Oh, this a mistake in the description. It should be `SimpleStreamEvent`

api/nakadi-event-bus-api.yaml
valgog: Good point

api/nakadi-event-bus-api.yaml
valgog: You are right @jhorstmann, probably it makes sense to use that content type. 
@tfrauenstein we can event implement several content types that would serialise streams differently, actually serialising into the never ending JSON array could make sense as well.

api/nakadi-event-bus-api.yaml
valgog: Today I was thinking that maybe it will be a good idea to make it an Array of Array of Events, with inner arrays being kind of atomic batches, and the server would not accept the next batch until the current one is committed into the store.

What do you think guys? Would that make sense, or it is too complicated?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+          in: path
+          description: Topic to consume events from
+          type: string
+          required: true
+        - name: partition
+          in: path
+          description: Id of the partition where to get events from
+          type: string
+          required: true
+        - name: start_from
```
tfrauenstein: editorial: "start_offset" is a better name -- API handles offsets of partitions.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+    type: object
+    required:
+      - detail
+    properties:
+      detail:
+        type: string
+        description: Problem description
+        example: Topic does not exist
+  Metrics:
+    type: object
```
tfrauenstein: Missing object structure definition.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+  Topic:
+    type: object
+    required:
+      - name
+    properties:
+      name:
+        type: string
+        description: Topic name
```
tfrauenstein: makes sense, to define the topic name syntax here, e.g.
        pattern: '^[a-zA-Z0-9-_]+$'

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        type: string
+        format: uuid
+      parent_id:
+        type: string
+        format: uuid
+      scopes:
+        type: array
+        items:
+          type: string
+  Problem:
```
tfrauenstein: Problem object definition should be consistent with definition of API Guideline -- see https://docs.google.com/document/d/1Btz_xm3cpwG_N76qe82klQLFGjRW1e0ZqS60RTaw4Jw/edit#heading=h.q42qw9lw9paw

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+  '/topics/{topic}/events':
+    get:
+      tags:
+        - low-level-api
+      summary: |
+        Starts a stream delivery for the specified partitions of the given topic.
+        The tracking of the current position in the partitions and of which
+        partitions is being read is in the responsibility of the client.
+        No commits are needed.
+      parameters:
```
tfrauenstein: All methods on events should also support FlowId header.

api/nakadi-event-bus-api.yaml
tfrauenstein: +1 -- some are definitely required (like 'event_type') others are provided by Nakadi (like 'created').
Since you use the same structure for POST and GET required attributes are requested as input params provided by producer, whereas non-required and read-only attributes are computed by Nakadi.

This crucial structure definitely needs clearer description of the different attributes (semantics and pragmatics) and all attributes should also armed with good examples.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+
+definitions:
+  Event:
+    type: object
+    description: |
+      This is the most general representation of an event, that can be processed
+      by Nakadi.
+
+      It should be used as a base definition for all events, that flow through
+      Nakadi by extending attributes of this object type.
```
tfrauenstein: Stronger statements needed here: Defines the generic event structure valid for all events produced / and consumed by Nakadi client applications. 
Must also specify how producers can add event type specific attributes, how they define its scheme and provide this info as contract to any consumers.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+    type: object
+  Topic:
+    type: object
+    required:
+      - name
+    properties:
+      name:
+        type: string
+        description: Topic name
+        example: article.120
```
tfrauenstein: IMO topics are owned by applications that also define the application specific event scheme -- see comment above.
So owner_application_id could be an interesting additional topic attribute.
Also number of partitions is an interesting topic property that should not be hided.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+    description: |
+      This is the most general representation of an event, that can be processed
+      by Nakadi.
+
+      It should be used as a base definition for all events, that flow through
+      Nakadi by extending attributes of this object type.
+    properties:
+      event_type:
+        type: string
+        example: 'https://resource-events.zalando.com/ResourceCreated'
```
tfrauenstein: I find this example confusing: Why an url, why a resource with a noun? 
I would see '/sales-order created' as better example. 

Not as part of the API, but in its architecture context we should agree on basic understanding of the event categories (e.g. resource vs business events), how they typically look like. Perhaps it is also a good idea to have a event type naming convention as part of stanardization, for instance, of data change events.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+      partitioning_key:
```
tfrauenstein: editorial:  partitioned_by or partitioning_field are better names as it need not to be a key.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        type: string
+        example: 'ABC123XXX-001'
+      metadata:
+        $ref: '#/definitions/EventMetaData'
+  EventMetaData:
+    type: object
+    properties:
+      id:
```
tfrauenstein: provided by Nakadi, correct?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        type: string
+        format: uuid
+      created:
+        type: string
+        format: data-time
+      root_id:
+        type: string
+        format: uuid
+      parent_id:
```
tfrauenstein: editorial: parent_event_id is a better name

api/nakadi-event-bus-api.yaml
tfrauenstein: I'm not sure about concept of root_id.

What I'm missing here: flow_id as event attribute.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        type: string
+        format: uuid
+      scopes:
```
tfrauenstein: Unclear: Are this the scopes the producer requires to be assigned to the consumer in order to read the event?

I think the authorization concept for event access needs some further discussion -- ad-hoc I would expect it as part of the event scheme definition and not passed with every single event.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        type: array
+        items:
+          type: string
```
tfrauenstein: For multitenancy support we also need +tenant_id as generic event attribute.

api/nakadi-event-bus-api.yaml
valgog: We explicitly removed the scopes here, as I wanted to avoid discussions on that topic in the first stage. 

We could also go an easy way and start with two scopes here, that generally allows reading from the system or publishing to the system.

We could also a scope for "adding topics"

api/nakadi-event-bus-api.yaml
valgog: For now it is not implemented. And it is not so trivial actually. But we can add the endpoint indeed.

api/nakadi-event-bus-api.yaml
valgog: Yes, the idea is that `x-nakadi-cursor` head is used, but if it is not defined, the default could be trivial, just start reading everything from now on. 

Per partition extraction is much more low level then this one. But reading events from a specified partition (from `/topics/{topic}/partitions/{partition}/events` endpoint allows to read only from one partition. So client would need to open as many connections, as there are partitions.

I thinks that this second method is not needed actually. But it is used by MOVMNT already.

api/nakadi-event-bus-api.yaml
valgog: There is not auto-commit or any possibility to commit at all in Low Level API.  

Clients are responsible to keep track of the cursors (offsets) themselves.

api/nakadi-event-bus-api.yaml
tmuehl: I like that generic 3-scopes idea for the beginning.

api/nakadi-event-bus-api.yaml
tmuehl: Streaming of batches of events is not considered for producing events so far, but might be feasible to add. We should use the same approach (line delimited json or never-ending array) as for the consuming.
Streaming within one batch (e.g. already write the first event of the batch as soon as it arrives), needs to be discussed. However, maybe, we should point out that writing a batch is not atomar and transactional.

api/nakadi-event-bus-api.yaml
tmuehl: Actually we do. :)
It's a mistake that it is named "Content-Encoding" in the definition. Should be changed to "Accept-Encoding". Thank's for recognizing!

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+          in: path
+          description: Topic where to send events to.
+          type: string
+          required: true
+        - name: partition
+          in: path
+          description: Partition where which the events are stored.
+          type: string
+          required: true
+        - name: Content-Encoding
```
tmuehl: Should be changed to "Accept-Encoding"

api/nakadi-event-bus-api.yaml
tmuehl: True. Maybe we should have two definitions: EventRead and EventWrite?

api/nakadi-event-bus-api.yaml
tmuehl: Yes.

api/nakadi-event-bus-api.yaml
whiskeysierra: > What I'm missing here: flow_id as event attribute.

Shouldn't we stick to the header for this?

api/nakadi-event-bus-api.yaml
whiskeysierra: > True. Maybe we should have two definitions: EventRead and EventWrite?

Swagger allows to mark properties as `readOnly: true` for this purpose.

whiskeysierra: > others are provided by Nakadi (like 'created')

What if the client wants to pass in the timestamp when the even initially occured, not when it arrived in Nakadi (which is not so interesting, business-wise).

api/nakadi-event-bus-api.yaml
valgog: We could keep them in headers, but then we still will have to rewrite the events and add the flow_id into the event. As flow_id should be archived together with the event actually.

api/nakadi-event-bus-api.yaml
valgog: I think this is one of the required fields in the MetaData part...

api/nakadi-event-bus-api.yaml
whiskeysierra: I'm not sure if exposing our (internal) concept of "Flow-ID" to the outside by making it a first-class citizen here. Could this just be part of some general metadata?

api/nakadi-event-bus-api.yaml
whiskeysierra: > All methods on events should also support FlowId header.

I'm not sure whether we should expose this in our Open Source systems.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,486 @@
...
+        - name: topic
+          in: path
+          description: Topic where to get events from
+          type: string
+          required: true
+        - name: x-nakadi-cursors
+          in: header
+          description: Cursors pointing to partitions/offsets to start read from
+          required: false
+          type: array
```
whiskeysierra: I'm not sure whether Swagger supports JSON in headers. Arrays in headers only support primitive elements. See https://github.com/swagger-api/swagger-spec/blob/master/versions/2.0.md#headerObject.

> **Required.** The internal type of the array. The value MUST be one of `"string"`, `"number"`, `"integer"`, `"boolean"`, or `"array"`. Files and models are not allowed.

api/nakadi-event-bus-api.yaml
whiskeysierra: > actually serialising into the never ending JSON array could make sense as well.

Sounds weird to me.

api/nakadi-event-bus-api.yaml
whiskeysierra: Why not just offset?

api/nakadi-event-bus-api.yaml
whiskeysierra: I like a real URL that a developer can deference for further documentation. Self-describing events so to say.

api/nakadi-event-bus-api.yaml
whiskeysierra: > as it need not to be a key.

Can you explain what you understand as a *key* here?

api/nakadi-event-bus-api.yaml
whiskeysierra: > For multitenancy support we also need +tenant_id as generic event attribute.

Nakadi should support it, yes. But I don't think we want to expose this as a first-class citizen in our Open Source software.

api/nakadi-event-bus-api.yaml
whiskeysierra: Agreed, although to be honest that definition is actually compatible with the RFC.

api/nakadi-event-bus-api.yaml
whiskeysierra: I guess it's *too dynamic* to be statically defined here?

whiskeysierra: `additionalProperties` could be useful to describe this.

api/nakadi-event-bus-api.yaml
tfrauenstein: For me a 'key' is an attribute by which you can retrieve or identify a data record, i.e. event instance here.
The partitioning_field here is used as explicit input for computation of the event partition (and not as key in the sense above).

Additional hints wrt. partitioning here: there are 2 levels of client control:
a) control by passing the partitioning criteria:
Here Nakadi maps it to event partition range using a specific hash with low collision probability. 
b) full control by passing the partition value:
Here Nakadi directly uses the partition value and refuses out of range values. 
 
I think Nakadi should support both variants. 
Variant a) is covered by current proposal -- P.S: we should disclose here also the used hash function.
Variant b) should be covered by new partitionId input param.

P.S: If not explicitly provided, Nakadi will use a default partition computation which ensures equally distribution among available event topic partitions.

jkliff:
Cancelling PR since the target of the merge is not master, but nakadi-jvm

PR #46, opened 2015-12-21
Low level api review fixes

jkliff:
First step of API adjustments for review.


valgog:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/zalando/nakadi/pull/46'>Refresh</a>.<a href='#crh-update-none'></a>

api/nakadi-event-bus-api.yaml
```diff
@@ -150,11 +178,11 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
         '404':
-          description: Not found. Typically in case of requesting from a non existing topic of partition.
+          description: Not found. Typically in case of requesting from a non existing topic or partition. Details are provided on the returned `Problem`.
           schema:
             $ref: '#/definitions/Problem'
         '500':
```
v-stepanov: I know that it wasn't added by you, but maybe we should not define 500 in our API. For me throwing 500 means that there is an error in your app. 503 seems to be more appropriate code.

api/nakadi-event-bus-api.yaml
```diff
@@ -474,7 +501,7 @@ definitions:
         type: string
       offset:
         type: string
-  SimpleStreamEvent:
+  EventStreamAtom:
```
v-stepanov: Why EventStreamAtom? For me that looks more like an EventStreamBatch

valgog: The batch would not have the cursor information

valgog: ##### @tmuehl @CyberDem0n 

Please vote on the name:
* `EventStreamAtom` - as this is an atomic part of the stream (one cannot send less then this)
* `EventStreamBatch` - maybe not so pathetic as `atom` :D

api/nakadi-event-bus-api.yaml
```diff
@@ -317,7 +345,7 @@ paths:
           required: true
         - name: batch_limit
           in: query
-          description: Maximum number of `SimpleStreamElement`s in each chunk of the stream. If unspecified assumes default value 1 (i.e. each event is individually submitted).
+          description: Maximum number of `EventStreamAtom`s in each chunk of the stream. If unspecified assumes default value 1 (i.e. each event is individually submitted).
```
valgog: Actually I think we have some problem here. @v-stepanov what does this parameter do in python prototype? Are we talking about number of `Event`s in `EventStreamAtom.events` array?

v-stepanov: @valgog Yes, it's a number of events in a batch. And there's really a problem with description, because it's not a number of `EventStreamAtom`s. It's a number of events in `EventStreamAtom`.

api/nakadi-event-bus-api.yaml
```diff
@@ -2,18 +2,46 @@ swagger: '2.0'
...
+    Notable omissions here are:
+    * The process of creating a `Topic` and definition of the related `EventType`s is not fully defined and its details will be added as it becomes more clear.
+    * Certain aspects of the validation procedures of incoming events might not be possible to finalize until the preceding point is pending.
+    * The specific ways of defining the validation and enrichment of events are unspecified at this point.
+
   version: '0.4'
   contact:
     name: Team Aruha @ Zalando
-    email: nakadi-users@zalando.de
+    email: team-aruha+nakadi-maintainers@zalando.de
```
valgog: I suppose this email is better. Nakadi users are not necessarily responsible for the API

api/nakadi-event-bus-api.yaml
valgog: :+1:  changed the description

valgog: !resolved

api/nakadi-event-bus-api.yaml
jkliff: I vote for ...Batch.

api/nakadi-event-bus-api.yaml
jkliff: 503 Service Unavailable is somewhat more specific than 500 and implies that the condition is temporary. From the RFC:

> The 503 (Service Unavailable) status code indicates that the server
> is currently unable to handle the request due to a temporary overload
> or scheduled maintenance, which will likely be alleviated after some
> delay. 

At this moment one can't guarantee that any error that will happen will fit into this restriction, so 500 is preferable.

valgog: :+1:  !resolved

api/nakadi-event-bus-api.yaml
valgog: :+1: !resolved

v-stepanov:
:shipit:

api/nakadi-event-bus-api.yaml
jkliff: !resolved :+1:

api/nakadi-event-bus-api.yaml
tmuehl: +1 for Batch

api/nakadi-event-bus-api.yaml
```diff
@@ -2,18 +2,62 @@ swagger: '2.0'
...
   description: |
+    -----------
+    Definitions
+    -----------
+    Nakadi at its core aims at being a generic and content-agnostic event broker with a convenient API.
+    In doing this, Nakadi abstracts away, as much as possible, details of the backing messaging
+    infrastructure. The single currently supported messaging middleware is Kafka (Kinesis is planned
+    for the future).
+
+    In Nakadi every Event has a EventType, and a **stream** of Events is exposed for each registered
```
v-stepanov: typo. Should be "an"

api/nakadi-event-bus-api.yaml
```diff
@@ -38,18 +93,159 @@ paths:
...
           schema:
             $ref: '#/definitions/Problem'
+  /event-types:
+    get:
+      tags:
+        - schema-registry-api
+      description: Returns a list of all registered `EventType`s
+      responses:
+        '200':
+          description: Metrics data
```
v-stepanov: :-|

api/nakadi-event-bus-api.yaml
```diff
@@ -2,18 +2,73 @@ swagger: '2.0'
...
+
+
+
+    Scope and status of this document
     -------------------------------
 
-    * Low-level API
-    * Stantardised event format (see #/definitions/Event)
-    * Partition strategy: event store will infer the partition from the `partitioning_key` in the message body
+    The present API specification is in **draft** state and is subject to change.
```
v-stepanov: This point is not quite clear for me. If the API will change we should not put the first version on review then. If it will be extended (not changed) then it's ok.

api/nakadi-event-bus-api.yaml
```diff
@@ -38,18 +93,159 @@ paths:
           schema:
             $ref: '#/definitions/Metrics'
         '401':
-          description: User is not authenticated
+          description: Client is not authenticated
```
v-stepanov: In a case of metrics-endpoint I think it can be both user or client :)

api/nakadi-event-bus-api.yaml
```diff
@@ -38,18 +93,159 @@ paths:
...
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+
+    post:
+      tags:
+        - schema-registry-api
+      description: |
+        Creates a new `EventType`. Upon creation the name MUST be unique.
```
v-stepanov: What we return if it is not? 422?

api/nakadi-event-bus-api.yaml
```diff
@@ -38,18 +93,159 @@ paths:
...
+          description: Ok
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '422':
+          description: Unprocessable Entity
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
```
v-stepanov: I see that for some endpoints you have 500, for some - not. I think we should be consistent here. 500 can theoretically happen for any endpoint.

api/nakadi-event-bus-api.yaml
```diff
@@ -38,18 +93,159 @@ paths:
...
+        conditions for backwards compatible extensions in the schema)
+      parameters:
+        - name: name
+          in: path
+          description: Name of the EventType to update.
+          type: string
+          required: true
+        - name: event-type
+          in: body
+          description: EventType to be created
```
v-stepanov: *updated

api/nakadi-event-bus-api.yaml
```diff
@@ -484,3 +752,114 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/Event'
+
+  EventType:
+    description: An event type defines the schema and its runtime properties.
+    properties:
+      name:
+        type: string
+        description: Name of this EventType. The name for the EventType must follow the pattern
```
v-stepanov: I think it's worth mentioning here that it should be unique

api/nakadi-event-bus-api.yaml
```diff
@@ -2,18 +2,73 @@ swagger: '2.0'
...
+    In doing this, Nakadi abstracts away, as much as possible, details of the backing messaging
+    infrastructure. The single currently supported messaging middleware is Kafka (Kinesis is planned
+    for the future).
+
+    In Nakadi every Event has an EventType, and a **stream** of Events is exposed for each registered
+    EventType.
+
+    An EventType defines properties relevant for the operation of its associated stream, namely:
+
+    * The **schema** of the Event of this EventType (*TODO: define shcema*)
```
valgog: Should we mention JSON-schema here?

api/nakadi-event-bus-api.yaml
```diff
@@ -484,3 +752,114 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/Event'
+
+  EventType:
```
v-stepanov: "required" section is missing

api/nakadi-event-bus-api.yaml
```diff
@@ -484,3 +752,114 @@ definitions:
...
+        $ref: '#/definitions/PartitionResolutionStrategy'
+
+      schema:
+        type: object
+        $ref: '#/definitions/EventTypeSchema'
+        description: |
+          The schema for this EventType. This is expected to be a json schema in yaml format (other formats
+          might be added in the future).
+
+  EventTypeSchema:
```
v-stepanov: any fields required?

api/nakadi-event-bus-api.yaml
```diff
@@ -2,18 +2,73 @@ swagger: '2.0'
...
+    * The **ordering** expectations of Events in this stream. (*TODO: define ordering not in terms of
+    partitioning*)
+
+    Except for defined enrichment rules, Nakadi will never manipulate the content of any Event.
+
+    Clients of Nakadi can be grouped in 2 categories: **Application owners** and **Clients** (clients in turn
+    are both **Producers** and **Consumers** of Events). Application owners interact with Nakadi via the
+    **Schema Registry API** for the definition of EventTypes, while Clients via the **Stream API** for
+    submission and reception of Events. A low level **Topic API** is available, providing full control and
+    responsibility of position tracking and partitioning (and therefore ordering) of to the Clients.
```
valgog: ... ~~of~~ to the Clients.

api/nakadi-event-bus-api.yaml
```diff
@@ -484,3 +752,114 @@ definitions:
...
+          Usually is the same as the name property.
+      validation-strategies:
+        description: |
+          Determines the validation that has to be executed upon reception of Events of this type. Events are
+          rejected if any of the rules fail (see details of Problem response on the Event publishing methods).
+
+          Rule evaluation order is the same as in this array.
+        type: array
+        items:
+          $ref: '#/definitions/EventValidationStrategy'
```
v-stepanov: I currently don't understand how all these strategies are defined. Will there just be a set of predefined strategies like some enum ?

api/nakadi-event-bus-api.yaml
```diff
@@ -433,18 +658,56 @@ definitions:
         type: array
         items:
           type: string
+
   Problem:
```
v-stepanov: I think in the description we can reference https://github.com/zalando/problem

api/nakadi-event-bus-api.yaml
```diff
@@ -38,18 +93,159 @@ paths:
...
+        The fields validation-strategies, enrichment-strategies and partition-resolution-strategy have all an
+        effect on the incoming Event of this EventType. For its impacts on the reception of events please
+        consult the Event submission API methods.
+
+        * Validation strategies define an array of EventValidationStrategies to be evaluated on reception of
+        an Event of this EventType.
+
+        * TDB Enrichment strategy
+
+        * TDB Partition resulution strategy
```
valgog: resolution

api/nakadi-event-bus-api.yaml
```diff
@@ -38,18 +93,223 @@ paths:
...
+        enrichment and partitioning. The steps performed on reception of incoming message are:
+
+        1. Every validation rule specified in the EventType will be checked in order against the incoming
+        Events. Validation rules are evaluated in the order they are defined and the Event is **rejected**
+        in the first case of failure with a status of 400 Bad request. If the offending validation rule
+        provides information about the violation it will be included in the Problem response.
+        If the EventType defines schema validation it will be performed at this moment.
+        **TBD behaviour in case of violation within a stream of events**)
+
+        1. Once the validation succeeded, the content of the Event is updated according to the enrichment
```
v-stepanov: 2.

api/nakadi-event-bus-api.yaml
```diff
@@ -38,18 +93,223 @@ paths:
...
+        rules in the order the rules are defined in the EventType.
+        No preexisting value might be changed (even if added by an enrichment rule). Violations on this will
+        force the immediate **rejection** of the Event with a 409 Conflict status. The invalid overwrite
+        attempt will be included in the respose's Problem object.
+
+        1. The incoming Event's relative ordering it then Evaluated according to the according rule on the
```
v-stepanov: 3.

api/nakadi-event-bus-api.yaml
```diff
@@ -34,22 +89,239 @@ paths:
...
```
v-stepanov: I'm afraid I completely don't understand the first sentence here. What do you mean by order validation?

api/nakadi-event-bus-api.yaml
```diff
@@ -34,22 +89,239 @@ paths:
...
+          required: true
+      responses:
+        '200':
+          description: Ok
+        '400':
+          description: Bad request
+          schema:
+            $ref: '#/definitions/Problem'
+        '401':
+          description: User is not authenticated
```
v-stepanov: Client?

api/nakadi-event-bus-api.yaml
jkliff: See POST /event-types/
Currently is says:
The mechanism to set and inspect such rules is not defined at this time and might not be exposed in the API.

But there are some predefined rules indeed which I will find some place where to describe them.

api/nakadi-event-bus-api.yaml
```diff
@@ -34,22 +89,239 @@ paths:
...
+        No preexisting value might be changed (even if added by an enrichment rule). Violations on this will
+        force the immediate **rejection** of the Event with a 409 Conflict status. The invalid overwrite
+        attempt will be included in the respose's Problem object.
+
+        1. The incoming Event's relative ordering it then Evaluated according to the according rule on the
+        EventType. Failure to evaluate the rule will **reject** the Event.
+        **Restriction**: POSTing Events based on its EventType requires the EventType to have a
+        PartitioningStrategy defined. Violating this condition will produce a 405 Not allowed.
+
+      parameters:
```
v-stepanov: where's the event(s) itself?

api/nakadi-event-bus-api.yaml
```diff
@@ -34,22 +89,239 @@ paths:
...
+          schema:
+            type: object
+            $ref: '#/definitions/EventType'
+          required: true
+
+      responses:
+        '200':
+          description: Ok
+        '401':
+          description: User is not authenticated
```
valgog: Client

api/nakadi-event-bus-api.yaml
jkliff: Not yet here.

api/nakadi-event-bus-api.yaml
```diff
@@ -34,22 +89,239 @@ paths:
...
+        If the EventType defines schema validation it will be performed at this moment.
+        **TBD behaviour in case of violation within a stream of events**)
+
+        1. Once the validation succeeded, the content of the Event is updated according to the enrichment
+        rules in the order the rules are defined in the EventType.
+        No preexisting value might be changed (even if added by an enrichment rule). Violations on this will
+        force the immediate **rejection** of the Event with a 409 Conflict status. The invalid overwrite
+        attempt will be included in the respose's Problem object.
+
+        1. The incoming Event's relative ordering is Evaluated according to the rule on the
```
v-stepanov: I'm afraid I still don't get it. What exact validation is happening here? I actually thought that partitioning is happening on this step.

api/nakadi-event-bus-api.yaml
```diff
@@ -484,3 +860,119 @@ definitions:
...
           $ref: '#/definitions/Event'
+
+  EventType:
+    description: An event type defines the schema and its runtime properties.
+    properties:
+      name:
+        type: string
+        description: |
+          Name of this EventType. The name for the EventType must follow the pattern
+          'organization:team:application:event-type'
```
valgog: `event-type-name`

(and you can use the ` quotes)

api/nakadi-event-bus-api.yaml
```diff
@@ -484,3 +860,119 @@ definitions:
...
+      topic:
+        type: string
+        description: |
+          References the name of the low-level topic that is holding the stream of Events for this EventType.
+          Usually is the same as the name property.
```
valgog: Maybe we should say, that the topic field will be provided, only if it differs from the `EventType.name` value

api/nakadi-event-bus-api.yaml
```diff
@@ -263,7 +598,9 @@ paths:
         - name: Content-Encoding
           in: header
           required: false
-          description: '"Allow the publisher to gzip compress the payload by setting this header to "gzip". For Example "Content-Encoding: gzip"'
+          description: |
+            "Allow the publisher to gzip compress the payload by setting this header to "gzip". For Example
```
valgog: The first quite is not needed here

api/nakadi-event-bus-api.yaml
```diff
@@ -317,7 +659,9 @@ paths:
           required: true
         - name: batch_limit
           in: query
-          description: Maximum number of `SimpleStreamElement`s in each chunk of the stream. If unspecified assumes default value 1 (i.e. each event is individually submitted).
+          description: |
+            Maximum number of `EventStreamBatch`s in each chunk of the stream. If unspecified assumes default
```
valgog: Maximum number of `Event`s in each `EventStreamBatch`

api/nakadi-event-bus-api.yaml
```diff
@@ -317,7 +659,9 @@ paths:
...
+            value 1 (i.e. each event is individually submitted).
```
valgog: This is for sending events out of Nakadi, yes? In this case probably one should talk about _delivering_ events and not _submitting_ them.

api/nakadi-event-bus-api.yaml
```diff
@@ -333,7 +677,8 @@ paths:
         - name: batch_flush_timeout
           in: query
           description: |
-            Maximum time in seconds to wait for the flushing of each chunk; if the `batch_limit` is reached before this
+            Maximum time in seconds to wait for the flushing of each chunk; if the `batch_limit` is reached
+            before this
             time is reached the messages are immediately flushed to the client.
             A value of 0 will flush every batch events immediatly as soon as they come in.
```
valgog: This is more difficult than that. This timeout is the timeout, after with any batch, that is being collected by the system, is closed and flushed to the client. The default value could be defined differently for different settings of `batch_limit`, for example:

  | `batch_flush_timeout` undefined | `batch_flush_timeout > 0`
------------ | ------------- | -------------
**`batch_limit` undefined** | one-by-one | flush on timeout
**`batch_limit > 0`** | flush on size | flush on size OR timeout

api/nakadi-event-bus-api.yaml
```diff
@@ -362,16 +708,16 @@ paths:
         - name: Accept-Encoding
           in: header
           required: false
-          description: '"If set the client requests the body to be gzip compressed. For Example "Content-Encoding: gzip"'
+          description: |
+            "If set the client requests the body to be gzip compressed. For Example "Content-Encoding: gzip"
```
valgog: Unneeded quote

api/nakadi-event-bus-api.yaml
```diff
@@ -400,7 +746,7 @@ definitions:
   Event:
     type: object
     description: |
-      This is the most general representation of an event, that can be processed
+      **To be changed**. This is the most general representation of an event, that can be processed
```
valgog: looking on the line 757, the example of the `event_type` value should changed according to our `:` scoped suggested format

api/nakadi-event-bus-api.yaml
```diff
@@ -474,7 +859,12 @@ definitions:
         type: string
       offset:
         type: string
-  SimpleStreamEvent:
+  EventStreamBatch:
+    description: |
+      One chunk of events in a stream. A batch consists of an array of `Event`s plus a `Cursor` pointing to
+      the offset of the last Event in the stream.
```
valgog: To the offset of an event to be delivered directly after the last event in `EventStreamBatch.events` array?

Do I remember this correctly? Was the semantic inclusive or exclusive? We discussed it several times already and several times we decided on this already :D

api/nakadi-event-bus-api.yaml
v-stepanov: @valgog 
it's a little bit different (at least in current implementation)
if the batch_limit is not set - it's default value is 1. So the pure flush-on-timeout is not possible.

valgog: Good to, know, I would vote for the suggestion, that I provided :smile:

v-stepanov: @valgog then it will be not consistent
With current behavior batch_limit is by default always 1.
With your suggestion it will be by default 1 only when batch_flush_timeout is also not defined

api/nakadi-event-bus-api.yaml
```diff
@@ -333,7 +677,8 @@ paths:
         - name: batch_flush_timeout
           in: query
```
jkliff: Any opinions on changing all these parameters from quety to header?

api/nakadi-event-bus-api.yaml
valgog: Yes

api/nakadi-event-bus-api.yaml
valgog: So the `Cursor` is pointing on the last `Event` in `EventStreamBatch.events` array.

api/nakadi-event-bus-api.yaml
```diff
@@ -59,21 +347,42 @@ paths:
...
-        - low-level-api
-      summary: |
+        - topic-api
+        - draft
+      description: |
         Starts a stream delivery for the specified partitions of the given topic.
-        The tracking of the current position in the partitions and of which
-        partitions is being read is in the responsibility of the client.
+
+        The event stream is formated as a sequence of `EventStreamBatch`es separated by `\n`. Each
```
v-stepanov: I just want to clarify how current streaming from multiple partitions works (to what we agreed before implementing streaming from multiple partitions)
For example for batch_limit 5 you will get something like (this displays three batches)
```json
{"cursor":{"partition":"0","offset":"4555"}, "events":[{"thisIs":"dummyEvent"}, 
                                                       {"thisIs":"dummyEvent"}, 
                                                       {"thisIs":"dummyEvent"}]}\n
{"cursor":{"partition":"1","offset":"2365"}, "events":[{"thisIs":"dummyEvent"}]}\n
{"cursor":{"partition":"2","offset":"6544"}, "events":[{"thisIs":"dummyEvent"}]}\n
\n
{"cursor":{"partition":"2","offset":"6546"}, "events":[{"thisIs":"dummyEvent"}, 
                                                       {"thisIs":"dummyEvent"}]}\n
{"cursor":{"partition":"3","offset":"9875"}, "events":[{"thisIs":"dummyEvent"}, 
                                                       {"thisIs":"dummyEvent"}, 
                                                       {"thisIs":"dummyEvent"}]}\n
\n
{"cursor":{"partition":"0","offset":"4556"}, "events":[{"thisIs":"dummyEvent"}]}\n
{"cursor":{"partition":"1","offset":"2366"}, "events":[{"thisIs":"dummyEvent"}]}\n
{"cursor":{"partition":"2","offset":"6547"}, "events":[{"thisIs":"dummyEvent"}]}\n
{"cursor":{"partition":"3","offset":"9877"}, "events":[{"thisIs":"dummyEvent"}, 
                                                       {"thisIs":"dummyEvent"}]}\n
\n
...
```

jkliff: Hmm, doesn't this match the description?

api/nakadi-event-bus-api.yaml
jkliff: Yes.

api/nakadi-event-bus-api.yaml
```diff
@@ -59,21 +347,42 @@ paths:
...
+        EventStreamBatch constains a chunk of Events and a Cursor pointing to the end of the chunk (i.e. last
+        delivered Event).
+        If streaming for several distinct partitions, each one is a independent `EventStreamBatch`.
+
+        The streaming can be controlled in terms of size of each chunk, timeout for flushing each chunk,
+        total amount of delivered Events and total time for the duration of the stream.
+
+        Nakadi will keep a streming conneciton open even if there are no events to be delivered. In this case
```
v-stepanov: *streaming

v-stepanov: * connection

api/nakadi-event-bus-api.yaml
v-stepanov: Well, there's \n\n between "logical batches" (I don't actually know if it is good or bad)

v-stepanov: Also I think we should discuss again if we need all timeouts to be handled as specified per partition or not.
(with current implementation timeouts are not per partition)

api/nakadi-event-bus-api.yaml
```diff
@@ -34,22 +99,252 @@ paths:
...
+        an `Event` of this `EventType`. Details of usage can be found in an external document (TBD link to
+        document).
+        **Notes**: Some EventValidationStrategies might support additional parameters. The behaviour is
+        specific to each strategy and one is suggested to consult its `doc` field for detailes of operation.
+        The default `SchemaValidationStrategy` supports passing fields to be ignored if the qualifier matches
+        the input.
+
+        * TDB Enrichment strategy
+
+        * TDB Partition resolution strategy
```
tmuehl: How do validation-, enrichment- and partition-resolution-strategies will be registered and implemented? Are they registered implicitly when a new EventType is posted?

api/nakadi-event-bus-api.yaml
```diff
@@ -34,22 +99,252 @@ paths:
...
+        - schema-registry-api
+      description: |
+        Creates a new `EventType`. Upon creation the name MUST be unique. Failing to do so will fail the
+        request with a 422 Unprocessable entity.
+
+        **Implementation note:** The creation of an EventType implicitly creates the structures in the backing
+        messaging implementation needed for the reception and persistency of the Events. Considering that at
+        this time only Kafka is used, this corresponds to the creation of a Topic. If so desired, clients can
+        interact directly with the topic using the low level API (for existing restrictions see the
+        corresponding methods on the topic-api).
```
tmuehl: I think I would maybe separated these two steps, because of SLAs and keeping control over the guaranteed scalability for one Topic (EventType, respectively).

api/nakadi-event-bus-api.yaml
jkliff: In the current idea, they are not registered. They simply exist and one can use then referring by name and adding some parameters. 
There will be a way of listing the ones that exist. I'll add this as TBD.

api/nakadi-event-bus-api.yaml
jkliff: At first we'll not be supporting any refined way of managing this. 
There is a field "cardinality" that is to be provided on creation of the EventType that will work as a hint of the load this EventType/Topic might face and will influence this. 

Just a couple notes about the possibilities that we "lose" by not having 2 separate steps, and why they are not problematic:
- we're being consistent with the hiding of topics and putting the EventType as the first class citizen;
- if one wants a "raw topic" where he can put anything in or use in the "low level way", simply as point to point communication, he simply needs to create an EventType that does not specify anything related to validation or any other enforcement. The behaviour is the same at the end.

api/nakadi-event-bus-api.yaml
tmuehl: Yes, I see. How do we deal with the case that someone specifies a *-strategy "xyz" that does not exist? If we refer existing strategies, why is the doc field (of the *-strategies) required from the client?

api/nakadi-event-bus-api.yaml
```diff
@@ -484,3 +851,123 @@ definitions:
...
+
+        type: object
+        $ref: '#/definitions/PartitionResolutionStrategy'
+
+      schema:
+        type: object
+        $ref: '#/definitions/EventTypeSchema'
+        description: |
+          The schema for this EventType. This is expected to be a json schema in yaml format (other formats
+          might be added in the future).
```
tmuehl: What is the purpose of this? Isn't that only a different representation of the same EventType?

api/nakadi-event-bus-api.yaml
jkliff: - not existing: should be a 422. Will add in the doc.
- doc being required from the client? i don't get this.

api/nakadi-event-bus-api.yaml
tmuehl: Discussed offline. It's clear.

PR #55, opened 2016-01-04
Event validation

jkliff:


valgog:
This pull request is now managed with CodeReviewHub. <a href='https://github.com/zalando/nakadi/pull/55'>Refresh</a>.<a href='#crh-update-none'></a>

v-stepanov:
@jkliff Tests failed because of compilation problem

> Execution failed for task ':compileJava'.

> Compilation failed; see the compiler error output for details.
> /home/travis/build/zalando/nakadi/src/main/java/de/zalando/aruha/nakadi/validation/ValidationStrategy.java:16: error: cannot find symbol

> STRATEGIES.put(EventBodyMustRespectSchema.NAME, new EventBodyMustRespectSchema());

src/main/java/de/zalando/aruha/nakadi/domain/ValidationStrategyConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+
+    public String getStrategyName() {
+        return strategyName;
+    }
+
+    public void setStrategyName(final String strategyName) {
+        this.strategyName = strategyName;
+    }
+
+    public JSONObject getAdditionalConfiguration() {
```
v-stepanov: I think it can be at least marked as @Nullable (or to be Optional in a perfect case :-D )

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -0,0 +1,83 @@
...
+            final EventTypeValidator etsv = new EventTypeValidator(eventType);
+            eventTypeValidators.put(eventType.getName(), etsv);
+
+        }
+
+        return eventTypeValidators.get(eventType.getName());
+
+    }
+
+    public static EventTypeValidator lookup(final EventType eventType) {
```
v-stepanov: @Nullable ?

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -0,0 +1,83 @@
...
+    private final List<EventValidator> validators = Lists.newArrayList();
+
+    public EventTypeValidator(final EventType eventType) {
+        this.eventType = eventType;
+    }
+
+    public void validate(final JSONObject event) {
+
+        // FIXME: return information on the validation place that failed
+        final boolean allItemsPass = validators.stream().anyMatch(new Predicate<EventValidator>() {
```
v-stepanov: I believe it should be allMatch here instead of anyMatch

v-stepanov: also, a lamda instead of new Predicate will simplify the code a lot.

src/test/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -0,0 +1,233 @@
...
+        }
+
+        final ObjectMapper mapper = new ObjectMapper();
+        mapper.registerModule(new JsonOrgModule());
+
+        final Function<OverrideDefinition, OverrideDefinition> enhanceWithQualifiedSchema = definition -> {
+            final JSONObject schema = eventType.getEventTypeSchema().getSchema();
+            final JSONObject copy = new JSONObject(schema, JSONObject.getNames(schema));
+
+            definition.getIgnoredProperties().forEach(new Consumer<String>() {
```
v-stepanov: lambda?

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
jkliff: it's indeed anyMatch: we stop at the first that does not pass.

src/test/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -0,0 +1,233 @@
...
+                                                                                 SchemaValidationOverrides.class)
+                                                                             .getOverrides().stream()
+                                                                             .map(enhanceWithQualifiedSchema)
+                                                                             .map(toQualifiedJSONSchemaValidator)
+                                                                             .collect(Collectors.toList());
+
+        return new QualifiedJSONSchemaValidationChain(qualifiedValidators, defaultSchemaValidator);
+    }
+
+    public static int findElement(final JSONArray array, final String element) {
```
v-stepanov: But the function actually finds element's index, not the element itself. The more appropriate name can be indexOfElement

src/test/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -0,0 +1,233 @@
...
+
+        for (int i = 0; i < array.length(); i++) {
+            if (element.equals(array.getString(i))) {
+                return i;
+
+            }
+        }
+
+        return -1;
```
v-stepanov: I think it's actually a perfect place to use Optional<Integer> as a return type to avoid returning -1

src/test/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+}
+
+class SchemaValidationOverrides {
```
v-stepanov: What is the purpose of this class? It seems to be just a list wrapper.

src/test/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -0,0 +1,233 @@
...
+        this.defaultSchemaValidator = defaultSchemaValidator;
+    }
+
+    @Override
+    public boolean accepts(final JSONObject event) {
+        final EventValidator validator = qualifiedValidators.stream().filter(it ->
+                                                                    it.getQualifier().getMatch().equals(
+                                                                        event.getString(it.getQualifier().getField())))
+                                                            .map(EventValidator.class::cast).findFirst().orElse(
+                                                                defaultSchemaValidator);
```
v-stepanov: You sold your soul to Jalopy :-P

src/test/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -0,0 +1,233 @@
...
+        schema = SchemaLoader.load(effectiveSchema);
+    }
+
+    @Override
+    public boolean accepts(final JSONObject event) {
+        try {
+            schema.validate(event);
+            return true;
+        } catch (final ValidationException e) {
+            e.printStackTrace();
```
v-stepanov: I believe this is not needed as you pass the exception to logger any way.

src/test/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -0,0 +1,233 @@
...
+            LOG.error("Validation failed: {}. Schema was: {}",
+                new Object[] {Joiner.on(";").join(e.getCausingExceptions()), effectiveSchema, e});
```
v-stepanov: this can be simplified as the method accepts varargs

jkliff: Don't worry now about this exception handling/logging at this time. It is fully bogus.

src/main/java/de/zalando/aruha/nakadi/validation/ValidationStrategy.java
```diff
@@ -0,0 +1,27 @@
...
+public abstract class ValidationStrategy {
+
+    private static final Map<String, ValidationStrategy> STRATEGIES = Maps.newHashMap();
+
+    static {
+        STRATEGIES.put(EventBodyMustRespectSchema.NAME, new EventBodyMustRespectSchema());
+        STRATEGIES.put(FieldNameMustBeSet.NAME, new FieldNameMustBeSet());
+    }
+
+    public abstract EventValidator materialize(EventType eventType, ValidationStrategyConfiguration vsc);
```
v-stepanov: I'm afraid method name `materialize` doesn't really tell me much.

jkliff: With it a ValidationStrategy (something abstract) is materialized (made concrete) into a Validator.

src/test/java/de/zalando/aruha/nakadi/validation/JSONSchemaValidationTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+        final EventTypeValidator validator = EventValidation.lookup(et);
+
+        final JSONObject event = new JSONObject(
+                "{\"foo\": \"bar\", \"extra\": \"i should be no problem\", \"name\": \"12345\"}");
+
+        validator.validate(event);
+    }
+
+    private EventType buildAndRegisterEventType(final String name, final JSONObject schema) {
```
v-stepanov: can be static

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -0,0 +1,73 @@
...
+
+public class EventValidation {
+
+    private static final Map<String, EventTypeValidator> eventTypeValidators = Maps.newConcurrentMap();
+
+    public static EventTypeValidator forType(final EventType eventType) {
+        return forType(eventType, false);
+    }
+
+    public static EventTypeValidator forType(final EventType eventType, final boolean incrementalEdit) {
```
v-stepanov: I don't really understand what this method does, especially the `incrementalEdit` parameter

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
v-stepanov: Ah ok. I didn't notice that you use negative predicate. And that can actually confuse other people as well.

src/test/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
valgog: We should find some better solution here. But this is a nice peace of code to check the formatters if we decide to have a look on some other possibilities like `CheckStyle` :D

valgog: I think this comment is already resolved :D :+1:

jkliff: This is not even readable. I'll adjust this soon.

jkliff: actually, looking at this now, i don't see what the problem is. 
the code is correctly indented if you look closely.

v-stepanov: @jkliff Yes, of course it's indented correctly, Jalopy knows its job good. But even correctly indented it still looks ugly after what Jalopy did to it.

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -0,0 +1,73 @@
...
+    private final List<EventValidator> validators = Lists.newArrayList();
+
+    public EventTypeValidator(final EventType eventType) {
+        this.eventType = eventType;
+    }
+
+    public void validate(final JSONObject event) {
+
+        // FIXME: return information on the validation place that failed
+        final boolean allItemsPass = validators.stream().anyMatch(ev -> !ev.accepts(event));
```
v-stepanov: I still not see any good reason to have that double negative instead of just:
```java
final boolean allItemsPass = validators.stream().allMatch(ev -> ev.accepts(event));
```

src/test/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
v-stepanov: What I mean is that I like a lot how the code is written, but this formatting makes it completely not readable.
I would prefer to have something like:
```java
final EventValidator validator = qualifiedValidators
                .stream()
                .filter(it -> it
                        .getQualifier()
                        .getMatch()
                        .equals(event.getString(it.getQualifier().getField())))
                .map(EventValidator.class::cast)
                .findFirst()
                .orElse(defaultSchemaValidator);
```

v-stepanov: But it's a separate discussion actually.

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
jkliff: evaluate all vs evaluate one.

src/main/java/de/zalando/aruha/nakadi/validation/ValidationStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+import de.zalando.aruha.nakadi.domain.EventType;
+import de.zalando.aruha.nakadi.domain.ValidationStrategyConfiguration;
+
+public abstract class ValidationStrategy {
+
+    private static final Map<String, ValidationStrategy> STRATEGIES = Maps.newHashMap();
+
+    public abstract EventValidator materialize(EventType eventType, ValidationStrategyConfiguration vsc);
+
+    public static final void register(final String name, final ValidationStrategy strategy) {
```
v-stepanov: This method is static, I don't see much sense to make it final
(the same for the next method)

v-stepanov:
:shipit:

src/main/java/de/zalando/aruha/nakadi/validation/ValidationStrategy.java
jkliff: not allow redefinition on eventual subclasses

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
v-stepanov: It will work exactly the same.
I'm pretty sure allMatch is implemented in a "clever" way and quits after the first not matching.
So there will be no advantage.

PR #56, opened 2016-01-04
Nakadi low level api review

jkliff:
This is a review request for the API Guild.

This is the second stage of the review, that considers a lot of the input from the first review done in #43 

For the moment swagger definition does not contain correct definition and examples of the proposed `Event` schema.


api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1144 @@
...
+
+    An EventType defines properties relevant for the operation of its associated stream, namely:
+
+    * The **schema** of the Event of this EventType. The schema defines the accepted format of Events of an
+    EventType and will be, if so desired, enforced by Nakadi. Usually Nakadi will respect the schema for the
+    EventTypes in accordance to how an owning Application defines them.
+      **Note:** *Currently the specification of the schema must be pushed into Nakadi on EventType creation;
+      in the future, assuming that Applications will expose the schema for its owned resources, Nakadi might
+      support fetching the schema directly from them.*
+      Nakadi might define globally minimal schema complience requirements (i.e. some fields that must be
```
MALPI: Typo: complience -> compliance

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1144 @@
...
+      (*TODO: define validation and enrichment*)
+
+    * The **ordering** expectations of Events in this stream. (*TODO: define ordering not in terms of
+    partitioning*)
+
+    Except for defined enrichment rules, Nakadi will never manipulate the content of any Event.
+
+    Clients of Nakadi can be grouped in 2 categories: **Application owners** and **Clients** (clients in turn
+    are both **Producers** and **Consumers** of Events). Application owners interact with Nakadi via the
+    **Schema Registry API** for the definition of EventTypes, while Clients via the **Stream API** for
```
MALPI: while Clients via the **Stream API** for

Probably something like "communicate" is missing?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1144 @@
...
+    submission and reception of Events. A low level **Topic API** is available, providing full control and
+    responsibility of position tracking and partitioning (and therefore ordering) to the Clients.
+
+    The consumption of Events proceeds via the establishment of a named **Subscription** to an EventType.
+    Subscriptions are persitent relationships from an Application (which might have several instances) and the
```
MALPI: Typo: persitent -> persistent

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1144 @@
...
+        '500':
+          description: Internal Server Error. Details are provided on the returned `Problem`.
+          schema:
+            $ref: '#/definitions/Problem'
+    post:
+      tags:
+        - topic-api
+        - draft
+      description: |
+        Post a one or more `Event`s to a given `Topic` (of an `EventType`). Upon recpetion the `EventType`'s
```
MALPI: Post one ore more, right?

MALPI:
In general I have the following questions in mind:

* How long are events available when querying a topic
* Why are events not sepperated by Teams? For example it might be interesting to get all even types for a certain team?

api/nakadi-event-bus-api.yaml
jkliff: "Applications owners interact with ..., while Clients via."
So, "interact".

jkliff:
@MALPI 
- the retention time for events are not decided at the moment; it will certainly be set on a per event type basis and MIGHT in the future be exposed in the API. At the moment we make no statements about it.
- in our model an application owns the event type; given that a team owns an application, one can transitively find the event types for the team; we are however in fact missing the reference to the owning application in the EventType. This will be added.

MALPI:
@jkliff Since you don't want to make a statement about the retention time, is it then possible to keep track if a client which has registered for a topic, received the event? Or on the other hand to fetch the history of events for a topic?

jkliff:
@MALPI Tracking subscriptions is definitily possible and needed, but is for the high level api, not the low level, so not precisely defined atm.
The point you raise is however valid and I can now explain at least the following:
- up to the retention time, get the full history of events is possible (in the low level)
- the contract we'll define for subscriptions MUST make clear that subscribers have to consume the event in a shorter time interval than the retention time, otherwise he'll lose the event.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1145 @@
...
+          schema:
+            $ref: '#/definitions/Problem'
+
+        '503':
+          description: Not available
+          schema:
+            $ref: '#/definitions/Problem'
+
+  '/topics/{topic}/events':
+    get:
```
twz123: Since it's stressed that the Topic API is "low-level", is it also planned to have a "high-level" way of fetching events like a `GET /event-types/{name}/events` or does it make no sense?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1145 @@
...
+      tags:
+        - topic-api
+        - draft
+      description: |
+        Starts a stream delivery for the specified partitions of the given topic.
+
+        The event stream is formated as a sequence of `EventStreamBatch`es separated by `\n`. Each
+        `EventStreamBatch` constains a chunk of Events and a `Cursor` pointing to the end of the chunk (i.e.
```
twz123: Theres a standard `text/event-stream` [format](https://www.w3.org/TR/eventsource/#parsing-an-event-stream) around. Could it make sense to support that?

api/nakadi-event-bus-api.yaml
valgog: !resolved

api/nakadi-event-bus-api.yaml
jkliff: Yes, we are aware of this and considered this during the design. 
At some point we will support SSE (given adequate content negotiation), but, as the stream contains information on the cursor (tracking position of what was delivered), encoding the it in a SSE compliant fashion posed some tradeoffs that we opted not to take. Given this, we opted for this enveloped format at this moment for simplicity.

api/nakadi-event-bus-api.yaml
valgog: !resolved

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1218 @@
...
+
+  EventType:
+    description: An event type defines the schema and its runtime properties.
+    properties:
+      name:
+        type: string
+        description: |
+          Name of this EventType. The name for the EventType must follow the pattern
+          'organization:team:application:event-type-name'
+      owning_application:
```
Unisay: The difference between `application` embedded in the `EventType.name` URI and `owning_application` is not clear.
Also would be good to have an example.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1218 @@
...
+          References the name of the low-level topic that is holding the stream of Events for this EventType.
+          Usually is the same as the name property.
+
+      effective_schema:
+        type: string
+        #read-only: true
+        description: |
+          The effective schema of this `EventType`. The predefined schema validator will use the value of
+          this field as base for the validation.
+          The generation of thie value happens on creation of the `EventType` and is influenced by the
```
Unisay: Typo: `thie`

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1218 @@
...
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Server error
+          schema:
+            $ref: '#/definitions/Problem'
+
+    post:
```
Unisay: It is possible to create a new `EventType`, however there is no corresponding endpoint do remove it. 
In case if type was created with errors (e.g. wrong ordering or schema) and API client wants to fix them by recreating an `EventType`, how can client do this?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1218 @@
...
+    required:
+      - type
+      - schema
+
+  EventValidationStrategy:
+    description: |
+      Defines a rule for validation of an incoming Event.
+
+      Rules might require additional parameters; see the `doc` field of the existing rules for details. See
+      GET /registry/validation-rules for a list of available rules.
```
Unisay: The actual endpoint is `GET /registry/validation-rule` not `GET /registry/validation-rules`

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1218 @@
...
+    required:
+      - name
+
+
+  PartitionResolutionStrategy:
+    description: |
+      Defines a rule for the resolution of incoming Events into partitions of a Topic.
+
+      Rules might require additional parameters; see the `doc` field of the existing rules for details. See
+      GET /registry/partition-rules for a list of available rules.
```
Unisay: ` GET /registry/partition-rules`: no such endpoint declared yet.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1218 @@
...
+      - name
+
+  EventEnrichmentStrategy:
+    description: |
+      Defines a rule for transformation of an incoming Event. No existing fields might be modified. In
+      practice this is used to set automatically values in the Event upon reception (i.e. set a reception
+      timestamp on the Event).
+
+      Rules might require additional parameters; see the `doc` field of the existing rules for details. See
+      GET /registry/enrichment-rules for a list of available rules.
```
Unisay: ` GET /registry/enrichment-rules`: no such endpoint declared yet.

api/nakadi-event-bus-api.yaml
tmuehl: it's /registry/validation-strategies :)

api/nakadi-event-bus-api.yaml
aikej: There is also WebSockets – which seems to be a great match for the requirements. Are there any reasons not to use it?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+        $ref: '#/definitions/Cursor'
+      events:
+        type: array
+        items:
+          $ref: '#/definitions/Event'
+
+  EventType:
+    description: An event type defines the schema and its runtime properties.
+    properties:
+      name:
```
aikej: It seems to be a little dangerous to allow the creator to set the name – Why don't you have required fields for the respective parts of the name and construct it server-side?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+        1. Once the validation succeeded, the content of the Event is updated according to the enrichment
+        rules in the order the rules are defined in the `EventType`.
+        No preexisting value might be changed (even if added by an enrichment rule). Violations on this will
+        force the immediate **rejection** of the Event with a 409 Conflict status. The invalid overwrite
+        attempt will be included in the respose's Problem object.
+
+        1. The incoming Event's relative ordering is evaluated according to the rule on the
+        `EventType`. Failure to evaluate the rule will **reject** the Event.
+        **Restriction**: POSTing Events based on its `EventType` requires it to have a
+        `PartitioningStrategy` defined. Violating this condition will produce a 405 Not allowed.
```
aikej: Validation and enrichment are forced to be implemented as synchronous actions with this API. While this makes usage easier it does pose some additional performance penalty on publishing an event. Unless I want to duplicate the partitioning in my publisher I have to publish all events sequentially – this feels like it might lead to bottlenecks.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+        that is invalid with respect to the schema's type.
+
+        * Referring any of Validation, Enrichment or Partitioning strategies that does not exist or whose
+        parameterization is deemed invalid.
+
+        Nakadi MIGHT impose necessary schema, validation and enrichment minimal configurations that MUST be
+        followed by all EventTypes (examples include: validation rules to match the schema; enriching every
+        Event with the reception date-type; adhering to a set of schema fields that are mandatory for all
+        EventTypes). **The mechanism to set and inspect such rules is not defined at this time and might not
+        be exposed in the API.**
```
aikej: Would I have to create multiple data-change-event-types for create, update and delete of 1 type of ressource? Or is it just 1?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+          The type of operation executed on the entity.
+          * C: Creation
+          * U: Update
+          * D: Deletion
+          * S: Snapshot
+    required:
+      - data_type
+      - data_op
+
+  DataChangeEvent:
```
aikej: Why don't you include URIs pointing to the changed resource? I think a strict symmetry between REST endpoints and change events make a very easy to understand way of interacting with all systems in the same manner – solely based on resources.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+    type: object
+    description: |
+      A Business Event.
+
+      Usually represents a status transition in a Business process.
+    allOf:
+      - $ref: '#/definitions/EventMetadata'
+    additionalProperties: true
+
+  DataChangeEventQualifier:
```
aikej: Why "DataChange" and not "Resource" as we have used during platform prototyping and also in early versions of Nakadi? Are we not talking about RESTful resources when some data changes?

https://lists.w3.org/Archives/Public/www-rdf-comments/2002JanMar/0148.html

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+    type: object
+    description: |
+      Indicators of a `DataChangeEvent`'s referred data type and the type of operations done on them.
+
+    properties:
+      data_type:
+        type: string
+        example: 'pennybags:order'
+      data_op:
```
aikej: Why did you opt to shorten some names (and in this case also the values) in the JSON structure? This makes things harder to read and has no real benefit if you enable gzip for the transmission.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+            type: string
+            format: date-time
+          received_at:
+            type: string
+            description: |
+              Timestamp of the reception of the Event by Nakadi. This is enriched upon reception of the
+              Event.
+              If set by the producer Event will be rejected.
+            format: date-time
+          parent_eid:
```
aikej: Do you have any concept for referencing events from other installations of Nakadi?

api/nakadi-event-bus-api.yaml
jkliff: 2 points here: 
- All steps of publishing are synchronous indeed. We need this to provide the synchronous answer that the event was indeed accepted. At some point in the future, depending on the needs, we might offer a asynchronous publishing API with modified semantics, but is not in scope now.
- About the second, maybe we left something unclear. (Please bear in mind that the bulk publishing is not specified in this version of the API)
Can you explain why you think you'd have to duplicate the partitioning in the publisher to avoid bottlenecks?

api/nakadi-event-bus-api.yaml
jkliff: Just one. 
For a resource Foo, its accompanying FooDataChangeEvent represents all change operations (See DataChangeEvent definition).

api/nakadi-event-bus-api.yaml
jkliff: No. 
In the platform, Nakadi will be operated as a centralized service.

api/nakadi-event-bus-api.yaml
aikej: Let's say I want to partition by product-id – I could in theory post events for different product-ids in parallel requests without causing any issues. But what I can't do is do parallel requests for the same product-id since I don't know which one actually will be processed first.
Therefor I will have to either duplicate the partitioning by product-id in my publisher to ensure this or publish all events strictly sequentially.

Or am I missing something?

api/nakadi-event-bus-api.yaml
jkliff: DataChange is more general than Resource and is in line with the fact that we are not imposing that the *ChangeEvents are of actual resources (i.e. with an URI pointing to it).
Nevertheless the point you raise and terminology that is indicated by it is very interesting and we'll take into consideration.

api/nakadi-event-bus-api.yaml
jkliff: The reason was storage and, given that these properties are all common, the readability problem would be offset by the familiarity with them. 
I agree on the gzip aspect for the transmission.

api/nakadi-event-bus-api.yaml
jkliff: At first, for simplicity. But the point you raise is valid.
We might

api/nakadi-event-bus-api.yaml
jkliff: I agree with your statement that you can't do parallel requests for the same product-id for obvious reasons (our synchronous semantics, as explained, forces you to only be sure that the next one is indeed "after" the first one if you received confirmation of the successful reception of the first one). I fail however to see how from this follows that you would have to duplicate the partitioning.

If you are posting events for different product-ids in parallel you are implicitly saying that the same product-ids are sent sequentially and the relative ordering over the same product-id is therefore guaranteed. This is all independent of the actually partitioning strategy for this I agree with your statement EventType that could be mapping 2 product-ids to the same or different partitions; this should be transparent (and to some degree irrelevant) to the Producer.

api/nakadi-event-bus-api.yaml
aikej: Allowing to have meaningful data that is not addressable by RESTful resources means we are not allowing to build applications that only use REST for certain functionality. I'd strongly suggest to avoid that situation. ;-)

api/nakadi-event-bus-api.yaml
aikej: What I mean is that you have to implement the logic to actually sent different product-ids in parallel but not the same. This is reimplementing the partitioning decision in the producer.
I agree that batching will help here though, so that should work for most requirements.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+
+    post:
+      tags:
+        - schema-registry-api
+      description: |
+        Creates a new `EventType`.
+
+        **Implementation note:** The creation of an EventType implicitly creates the structures in the
+        backing messaging implementation needed for the reception and persistency of the Events. Considering
+        that at this time only Kafka is used, this corresponds to the creation of a Topic. If so desired,
```
regispl: If I understand correctly, it means that there's a 1:1 relation between EventType and Topic. It means that I can't put multiple EventTypes into a single topic to guarantee the ordering, which is a common approach in EventSourcing world - how should that use case be handled? Should I have an EventType thats effectively a "wrapper" for other messages and have a schema defined in a way that handles all the possible "nested" messages? That would work, but it means that requires consumer to know about that wrapping and understand the "internal" message types to handle them. And the schemas may be huge

Wouldn't it be a good idea to not tie EventTypes with Topics in a 1:1 relationship but make it configurable 1:n (actually: n:1) with a default of 1:1? Everyone would know how many (and what) EventTypes are in a given topic and would have access to schemas required to read them.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+
+    An EventType defines properties relevant for the operation of its associated stream, namely:
+
+    * The **schema** of the Event of this EventType. The schema defines the accepted format of Events of an
+    EventType and will be, if so desired, enforced by Nakadi. Usually Nakadi will respect the schema for
+    the EventTypes in accordance to how an owning Application defines them.
+    **Note:** *Currently the specification of the schema must be pushed into Nakadi on EventType creation;
+    in the future, assuming that Applications will expose the schema for its owned resources, Nakadi might
+    support fetching the schema directly from them.*
+    Nakadi might define globally minimal schema compliance requirements (i.e. some fields that must be
```
tfrauenstein: "might define"? Confusing -- it's not clear what is meant by schema here.
Event structure consists of 2 parts:
a) generic part -- common structure valid for all events (of a certain class like BusinessEvent)
The scheme of the generic part is defined in this API.
b) dedicated part -- specific for each event type defining the structure of the specific event payload.
The payload scheme of the dedicated part is defined as part of the EventType definition. 
Here you use "schema" in the sense of "playload schema", do you? I would than call it so.
Nakadi restrictions on the "payload schema" are indicated here -- I would not expect it. Which restrictions?

The payload schema is an explicit event interface contract between producers and consumers. 
And Nakadi enforces payload schema definition, correct?
However, when schema validation is activated we will at least support "compatible extension match" validation mode (and perhaps additionally "exact match" mode), do we? Validation rule config options are not yet defined in this API spec.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+    present on schemas of all EventTypes).
+
+    * The expected **validation** and **enrichment** procedures upon reception of an Event.
+    (*TODO: define validation and enrichment*)
+
+    * The **ordering** expectations of Events in this stream. (*TODO: define ordering not in terms of
+    partitioning*)
```
tfrauenstein: The notion of "partition" is important element of this API -- and this is fine. 
Some statements about ordering properties of event streams would be helpful -- and it should be related to partitioning.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+
+    Except for defined enrichment rules, Nakadi will never manipulate the content of any Event.
+
+    Clients of Nakadi can be grouped in 2 categories: **Application owners** and **Clients** (clients in turn
```
tfrauenstein: editorial:  "application owners" --> "event type owners" because it is an API about events.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+    are both **Producers** and **Consumers** of Events). Application owners interact with Nakadi via the
+    **Schema Registry API** for the definition of EventTypes, while Clients via the **Stream API** for
+    submission and reception of Events. A low level **Topic API** is available, providing full control and
+    responsibility of position tracking and partitioning (and therefore ordering) to the Clients.
+
+    The consumption of Events proceeds via the establishment of a named **Subscription** to an EventType.
```
tfrauenstein: confusing. Would not mention subscriptions here -- part of the high level API and not (yet) subject of current API spec.
However, it is worth to mention that there is a high level and low level API and that current spec addresses the later.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+    Other aspects of the Event Bus are at this moment to be defined and otherwise specificied, not included
+    in this version of this specification.
+
+    Notable omissions here are:
+    * The contract between Nakadi and clients in the context of the high-level API (i.e. the process for
+    clients to establish subscriptions).
+
+    * Submission of events to topics in batched manner.
+
+    * Security scopes (OAuth) for the different operations
```
tfrauenstein: Why not? It's very basic and I would not expect exotic event api access rights, simply:
<event bus appl id>.read and <event bus appl id>.write

api/nakadi-event-bus-api.yaml
jkliff: > "might define"? Confusing -- it's not clear what is meant by schema here.
> Event structure consists of 2 parts:
> a) generic part -- common structure valid for all events (of a certain class like BusinessEvent)
> The scheme of the generic part is defined in this API.

Correct. There are slight differences when it's a BusinessEvent and when it's a DataChangeEvent: both share metadata in the generic part, but DataChangeEvent define further data_op and data_entity.
In this terminology, these are the generic part.

> b) dedicated part -- specific for each event type defining the structure of the specific event payload.
> The payload scheme of the dedicated part is defined as part of the EventType definition.
> Here you use "schema" in the sense of "playload schema", do you? I would than call it so.
> Nakadi restrictions on the "payload schema" are indicated here -- I would not expect it. Which restrictions?
> The payload schema is an explicit event interface contract between producers and consumers.

You are right. This statement is previous to the detailed specification of Business and DataChangeEvent formats. This is superfluous.

> And Nakadi enforces payload schema definition, correct?

For Business and DataChangeEvents this is definitely a logical step, although one could define an empty schema. 

> However, when schema validation is activated we will at least support "compatible extension match" validation mode (and perhaps additionally "exact match" mode), do we? Validation rule config options are not yet defined in this API spec.

These are definitions that are probably better set by the schema itself, not? There the owner of the schema can define directly what is allowd (with adequate use of additional properties, allOf, etc..).

api/nakadi-event-bus-api.yaml
jkliff: The understanding of 1-1 relationship is correct. 
Having the possibility of 1:n relationship was considered for a very long time but it posed several complex issues. This MIGHT be extended in the future, but the consequences need to be well evaluated.

What we do have in mind is definitely support streaming FROM several EventTypes. This does not solve the issue you raise, off course. 
The guarantees of "cross-entity ordering" are, in our vision, solved by a "correlation engine" that would bind on such a stream and order the events for delivery. For this we can't provide any detailed ideas at this time.

Maybew @valgog can give you more insights about the problematic and possible solutions here.

api/nakadi-event-bus-api.yaml
jkliff: Correct.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+            $ref: '#/definitions/EventType'
+          required: true
+      responses:
+        '200':
+          description: Ok
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '422':
```
tfrauenstein: The request should be rejected, if the event type resource instance already exists, shouldn't it?
Because than I have to use PUT.
It would be adequate to return 422 here and should be mentioned in the spec.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+      tags:
+        - schema-registry-api
+      description: |
+        Updates the `EventType` identified by its name. Behaviour is the same as creation of `EventType` (See
+        POST /event-type) except where noted below.
+
+        The name and the topic fields cannot be changed. Attempting to do so will result in a 422 failure.
+
+        At this moment changes in the schema are not supported and will produce a 422 failure. (TODO: define
+        conditions for backwards compatible extensions in the schema)
```
tfrauenstein: Sounds too restrictive -- I would describe it like follows:
Changes of the schema are supported. However, Nakadi checks that the passed schema is backward compatible to resource schema. If not, 422 error will be returned. (TODO: define notion of compatible extension on json schema)

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+          type: string
+          description: Name of the EventType
+          required: true
+        - name: X-Flow-Id
+          in: header
+          description: |
+            The flow id of the request, which is written into the logs and passed to called services. Helpful
+            for operational troubleshooting and log analysis.
+          type: string
+          format: flow-id
```
tfrauenstein: I'm missing the event payload in the body.

Below I see the POST /topics/{topic}/events for producing events.
So why we need this POST /event-types/{name}/events anyway? Why not delete it?

I'm also confused by the double identification of EventTypes -- via name and topic.
Why not simply use topic and single event-type identifier? Pls. also see comments below.
The complete further API is based on topic event-type identification.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+
+          - `business`: Events of this type will be BusinessEvents. `effective_schema` contains `metadata`
+            and any additionally defined properties passed in the `EventTypeSchema` directly on top level of
+            the Event. If name conflicts arise, creation of this EventType will be rejected.
+            Default validation_strategy for this `EventType` is `[{name: 'schema-validation'}]`.
+      topic:
+        type: string
+        description: |
+          References the name of the low-level topic that is holding the stream of Events for this EventType.
+          Usually is the same as the name property.
```
tfrauenstein: Both, name and topic are must be unique ids. Why we do not simply go with only topic (or name)?
I don't see the value 2 unique ids -- and find it confusing from API design pov.
I also find it strange, to have the topics as single resources -- it is an id attribute of the event type.
Why not simply and consistently use event-types as root resources for event type management and event creation?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+    description: An event type defines the schema and its runtime properties.
+    properties:
+      name:
+        type: string
+        description: |
+          Name of this EventType. Encodes the owner/responsible for this EventType.
+
+          The name for the EventType SHOULD follow the pattern, but is not enforced
+          'organization:team:owning_application:event-type', for example
+          'zalando:bazaar:price-service:price-change'.
```
tfrauenstein: I would prefer here simply:  STUPS_application_id:event_type_name

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+          * Owning application: SHOULD match the field owning_application; indicates
+          * EventType name: name of this EventType; SHOULD end in ChangeEvent for DataChangeEvents; MUST be
+          in the past tense for BusinessEvents.
+
+          (TBD: how to deal with organizational changes? Should be reflected on the name of the EventType?
+          Isn t it better to omit [organization:team] completely?)
+      owning_application:
+        type: string
+        description: |
+          Indicator of the Application owning this `EventType`.
```
tfrauenstein: +usually, should be STUPS-application-id

api/nakadi-event-bus-api.yaml
akauppi: And [Reactive Streams](http://www.reactive-streams.org/), which is an implementation agnostic standard between different streaming systems.

Note: I am just coming to nakadi (started looking yesterday) and am purely trying to understand what it is, what it tries to do, and how ready it is. Also, I am about at the same stage with regard to Reactive Streams. Just asking..

api/nakadi-event-bus-api.yaml
aikej: Do we really need that at all? It sounds like it will really hurt us evolving our systems and will cause then different semantically identical event-types because of incompatible schema changes!
Let's be more like the web instead of SOAPier than SOAP... ;-)

api/nakadi-event-bus-api.yaml
aikej: +1

api/nakadi-event-bus-api.yaml
jkliff: To be clear: schema evolution IS needed and WILL be supported. Although backwards compatible changes are definitely recommended, they are not always possible and this is clear. The mechanism with which this will be supported in Nakadi (basically we're talking here about versioning of EventTypes/schemas) is what was not defined until now. 

I'll adjust the wording in the API to make this situation more clear.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+        in the first case of failure with a status of 400 Bad request. If the offending validation rule
+        provides information about the violation it will be included in the Problem response.
+        If the `EventType` defines schema validation it will be performed at this moment.
+        **TBD behaviour in case of violation within a stream of events**)
+
+        1. Once the validation succeeded, the content of the Event is updated according to the enrichment
+        rules in the order the rules are defined in the `EventType`.
+        No preexisting value might be changed (even if added by an enrichment rule). Violations on this will
+        force the immediate **rejection** of the Event with a 409 Conflict status. The invalid overwrite
+        attempt will be included in the respose's Problem object.
```
tfrauenstein: You have described here, how validation works for one event.
However, API supports publishing of multiple events via event array. 
Just give hint, how validation works here -- I assume it works "atomic", i.e. first validation of all passed events before creation of any event. So you will not create duplicates by post request repetition, correct?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Server error
+          schema:
+            $ref: '#/definitions/Problem'
+
+  /topics:
+    get:
+      description: Lists all known `Topics` of this Nakadi cluster
```
tfrauenstein: Basically corresponds to GET /EventTypes?fields=topic i.e. filter topic field from event types.
Why not doing like this -- instead of inventing new resource (see also comment below)?

jkliff: On this topic please wait. 
There was a commit that messed up exactly this point. There is no duplicate method. Will be updated shortly.

api/nakadi-event-bus-api.yaml
jkliff: @tfrauenstein Can you please share your position on this here?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+        - name: topic
+          in: path
+          description: Topic where to get events from
+          type: string
+          required: true
+        - name: x-nakadi-cursors
+          in: header
+          description: Array of cursors indicating the partitions to read from and the starting point on
+            each. Assumes the offset on each cursor is not inclusive (i.e., first delivered Event is the
+            **first one after** the one pointed to in the cursor).
```
tfrauenstein: I would prefer having the cursors as part of the url, i.e. passed via url query string.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+            If unspecified will stream indefinately.
+
+            If this timeout is reached, any pending messages (in the sense of `stream_limit`) will be flushed
+            to the client.
+
+            Stream initialization will fail if `stream_timeout` is lower than
+            `batch_flush_timeout`.
+          type: integer
+          format: int32
+          required: false
```
tfrauenstein: just define the default here

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+        - name: batch_keep_alive_limit
+          in: query
+          description: |
+            Maximum number of keep-alive messages to get in a row before closing the connection. Unlimited by
+            default.
+          type: integer
+          format: int32
+          required: false
```
tfrauenstein: dito -- default definition

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+        '500':
+          description: Internal Server Error. Details are provided on the returned `Problem`.
+          schema:
+            $ref: '#/definitions/Problem'
+    post:
+      tags:
+        - topic-api
+        - draft
+      description: |
+        Post one `Event`s to a given `Topic` (of an `EventType`). Upon reception the `EventType`'s
```
tfrauenstein: editorial: not 'one event' but 'list of events'

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+        validation and enrichment rules are applied. For details on their behaviour please see POST /even
+        t-types/{name}/events.
+
+        (TBD: submission of batches of events)
+
+        Restriction: To avoid inconsistencies in the relative ordering of Events, **this operation cannot be
+        used if this Topic's defining EventType defines a `PartitionResolutionStrategy`.** Doing so will
+        produce a 405 Not allowed.
```
tfrauenstein: editorial: it's worth to hint that that post request repetition creates event duplicates.

api/nakadi-event-bus-api.yaml
jkliff: Adjusted wording to make this clear.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+            $ref: '#/definitions/Problem'
+        '404':
+          description: |
+            Topic not found.
+          schema:
+            $ref: '#/definitions/Problem'
+        '405':
+          description: |
+            Method not allowed. This topic does not support direct POSTing of Events. Use equivalent POST
+            /event-types/{name}/events instead.
```
tfrauenstein: Unclear -- also see comments above.

api/nakadi-event-bus-api.yaml
aikej: Ahh... Perfect! ;-)

api/nakadi-event-bus-api.yaml
jkliff: Isn't this the case already?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+          description: Server error
+          schema:
+            $ref: '#/definitions/Problem'
+
+  '/topics/{topic}/partitions/{partition}':
+    get:
+      tags:
+        - topic-api
+        - management-api
+      description: Returns the given `Partition` of this topic
```
tfrauenstein: editorial: Returns the given Partition (including its management cursors) of this topic.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+      - partition
+      - oldest_available_offset
+      - newest_available_offset
+    properties:
+      partition:
+        type: string
+      oldest_available_offset:
+        type: string
+      newest_available_offset:
+        type: string
```
tfrauenstein: just make clear:  
- oldest_available_offset point to the oldest event in the partition (queue)
- newest_available_offset point to the most recently produced event in the partition (queue)
Both are partition "storage" cursors -- not related related to consumer cursors that will be supported in future HL API extension.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+        - draft
+      description: |
+        Post one `Event`s to a given `Topic` (of an `EventType`). Upon reception the `EventType`'s
+        validation and enrichment rules are applied. For details on their behaviour please see POST /even
+        t-types/{name}/events.
+
+        (TBD: submission of batches of events)
+
+        Restriction: To avoid inconsistencies in the relative ordering of Events, **this operation cannot be
+        used if this Topic's defining EventType defines a `PartitionResolutionStrategy`.** Doing so will
```
tfrauenstein: Partition resolution strategies ist not that clear for me -- also not really explained below.
My simple view:
- event type defines "how partitioning should be done"
- I see basically 3 ways how to do it:

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+          eid:
+            description: |
+              Identifier of this Event.
+
+              Clients are allowed to generate this and this SHOULD be guaranteed to be unique from the
+              perspective of the producer. Consumers MIGHT use this value to assert uniqueness of reception
+              of the Event.
+
+              If not present upon reception, Nakadi will generate and fill this with a unique value.
+              (TBD: enforce UUID formatif set by the client?)
```
tfrauenstein: IMO it must be provided by producer -- and producer has to take care that it is unique, e.g. using UUIDs.
There always is the (low) risk that event duplicates may be created.
And generally clients should have the chance to detect that duplicates and handle it.
If eid is generated by Nakadi than producer event POST retries (e.g. after connection failure) will generate duplicates and client has no chance to do the de-duplication.
Should be described here too.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+            description: |
+              Event identifier of the Event that caused the generation of this Event.
+              Set by the procuder.
+            example: '105a76d8-db49-4144-ace7-e683e8f4ba46'
+          root_eid:
+            type: string
+            format: uuid
+            description: |
+              Event identifier of the root of the tree of Events (from the relation constructed from
+              parent-eid)
```
tfrauenstein: I'm sceptical about this root-id idea. 
It's not clearly defined, and will only work, if consistently applied by all event processing components involved.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+
+  BusinessEvent:
+    type: object
+    description: |
+      A Business Event.
+
+      Usually represents a status transition in a Business process.
+    allOf:
+      - $ref: '#/definitions/EventMetadata'
+    additionalProperties: true
```
tfrauenstein: I'm missing here the payload object.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+
+          (TBD: how to deal with organizational changes? Should be reflected on the name of the EventType?
+          Isn t it better to omit [organization:team] completely?)
+      owning_application:
+        type: string
+        description: |
+          Indicator of the Application owning this `EventType`.
+        example: price-service
+      type:
+        type: string
```
tfrauenstein: editorial: event_type has a type attribute -- a bit overloaded. Use "class" or "category" instead.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+          Usually is the same as the name property.
+
+      effective_schema:
+        type: string
+        #read-only: true
+        description: |
+          The effective schema of this `EventType`. The predefined schema validator will use the value of
+          this field as base for the validation.
+          The generation of this value happens on creation of the `EventType` and is influenced by the
+          `EventType.type`.
```
tfrauenstein: Event Type type (or class) is defined above and schema too below.
Unclear why now "effective_schema" is needed?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+          within this Event.
+          If set MUST be a valid required field as defined in the schema.
+          (TBD should be required? Is applicable only to both Business and DataChange Events?)
+
+      ordering_key:
+        type: string
+        description: |
+          Indicator of the field used for guaranteeing the ordering of Events of this type (used by the
+          PartitionResolutionStrategy).
+          If set MUST be a valid required field as defined in the schema.
```
tfrauenstein: see my comment about partitioning above.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+
+    required:
+      - name
+
+  EventTypeSchema:
+    properties:
+      type:
+        type: string
+        example: 'json-schema'
```
tfrauenstein: Should be an enum

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+        description: The type of schema definition (avro, json-schema, etc).
+      schema:
+        type: string
+        description: |
+          The schema as string in the syntax defined in the field type. Failure to respect the syntax will
+          fail any operation on an EventType.
+    required:
+      - type
+      - schema
```
tfrauenstein: A version and perhaps a version_date field make also sense.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1232 @@
...
+      - name
+
+  EventEnrichmentStrategy:
+    description: |
+      Defines a rule for transformation of an incoming Event. No existing fields might be modified. In
+      practice this is used to set automatically values in the Event upon reception (i.e. set a reception
+      timestamp on the Event).
+
+      Rules might require additional parameters; see the `doc` field of the existing rules for details. See
+      GET /registry/enrichment-strategies for a list of available rules.
```
tfrauenstein: here it is unclear what enrichment modes you want support here.
Do we need this for 1st API version? Why not starting without enrichment? Can be easily added later.

api/nakadi-event-bus-api.yaml
tmuehl: Maybe that is only a theoretical problem. Since an application is an owner of an EventType, you probably know as a producer for which entities you want to guarantee ordering. Since you maybe can sent events without being aware of the server side partitioning strategy.

api/nakadi-event-bus-api.yaml
jkliff: Without the notion of enrichment we can't provide some of the guarantees in metadata that Saiki expects. This is more or less the "technical means" with which we provide this.

api/nakadi-event-bus-api.yaml
jkliff: I would add this once we have some more clarity of how we want to deal with versioning of schemas.

api/nakadi-event-bus-api.yaml
jkliff: Agreed.

api/nakadi-event-bus-api.yaml
valgog: @aikej WebSockets and HTTP2 are very sexy indeed. And nakadi will support both these protocols in the future. Our goal now is to define a pure HTTH1.1 protocol.

api/nakadi-event-bus-api.yaml
valgog: !resolved

api/nakadi-event-bus-api.yaml
valgog: !resolved

api/nakadi-event-bus-api.yaml
valgog: @regispl Thank you for the comment. You are right, that 1:1 relationship of EventType to an ordering tool like Topic could be not sufficient.

Our considersations here are mostly related to the following issues:
* the only reason to push more then one event-type into the topic is the need to retain ordering. That also means, that these event-types should have the same ordering key (partitioning key). We even considered the possibility of mapping partitioning keys to topics
* applications should be made responsible of synchronisation of the the flow in such a unified topic and this is much more difficult to achieve then for one event-type. Especially if entities are produced by separate applications (that do not have the means of strong synchronisation usually), so the more correct way would have been not the mapping (`partitioning key` -> `topic`), but (`application`,`partitioning key` -> `topic`) as a partitioning tool, that reflects the real ordering guarantee for the consumer. But such a construct is much more difficult to understand and keep track of.

As @jkliff already mentioned, we definitely need a correlation engine (probably on the consumer side), that would reorder events coming from different topics. But such an engine is out of scope for us for the current milestone.

api/nakadi-event-bus-api.yaml
valgog: @akauppi Reactive Streams is more a software protocol, then an infrastructure, I am sure, that eventually we will have a _Reactive Client_ that publishes and consumes from Nakadi using _Reactive Streams_ principles

api/nakadi-event-bus-api.yaml
valgog: @aikej there are a lot of ideas, that we discussed a lot. In the essence, they boil down to the _Schema Registry_ that would know, where the event-types can be fetched from. If a standalone instance of the Nakadi cluster will join such a  _Schema Registry_, the clients will be able to find the address, where one should subscribe to.

For now it is out of scope of this API.

api/nakadi-event-bus-api.yaml
valgog: The complexity of usage is directly bound to the complexity of the usage of the `parent-eid`. 

Also `root-eid` has the same semantics as `flow-id` any, perhaps, having both these numbers does not make sense, because the only reason of having a `root-eid` is to simplify the lookup of the root of the chain of events.

api/nakadi-event-bus-api.yaml
valgog: @aikej I personally also support your point of view on the Resources. But there are many voices, that claim, that binding us _only_ to the RESTful resources could limit our flexibility, for example to send raw table row change events into the bus.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+
+          See documentation for the write operation for details on behaviour in case of unsuccessful
+          enrichment.
+        type: array
+        items:
+          $ref: '#/definitions/EventEnrichmentStrategy'
+        default: [{name: 'metadata-enrichment'}]
+
+
+      partition_resolution_strategy:
```
tfrauenstein: editorial: 'strategy' is more vage and related to mid/long term activities. Would prefer 'mode' or something similar here. 
P.S: Also related to other usage of 'strategy' in the API.

api/nakadi-event-bus-api.yaml
valgog: @aikej For now it is not quite clear to me, where one could fetch the URI for all possible change events. 

Theoretically, if everybody will be playing nicely, one would be able to reconstruct such an URI on the fly using the (`application`:`entity`):`primary key`

But questions arise of course, especially for the cases of _composite_ primary keys.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+            type: string
+            example: '105a76d8-db49-4144-ace7-e683e8f4ba46'
+          event_type:
+            description: |
+              The EventType of this Event. This is enriched by Nakadi on reception of the Event based on the
+              endpoint where the Producer sent the Event to.
+
+              If provided MUST match the endpoint. Failure to do so will cause rejection of the Event.
+            type: string
+            example: 'pennybags:payment-business-event'
```
valgog: Now I am not sure, why we did not choose the `.` as a separator for the scopes of the entity types :hammer:

api/nakadi-event-bus-api.yaml
valgog: Does anything say against using `.` as a separator for the scopes? (I have already commented on that higher)

api/nakadi-event-bus-api.yaml
valgog: :+1:

api/nakadi-event-bus-api.yaml
tfrauenstein: The DataChangeEvents have been invented for a specific problem: replication of persistent data stores from microservice applications via Saiki into the analytical data bases / DWH. Of course, not all data need to be published for BI/DS analysis. But the data that has to be published must fulfill certain properties, for instance, backward compatibility (contract via schemes), primary key, CUD and snapshot operations. There may be also a difference between API exposed resources and data stored in DB. For this specific purpose we introduced the DataChangeEvents. Likely, we additionally will have resource events or other event classes in future.

aikej: This turns internal data schemes into APIs – making it even harder to incrementally build software. I especially want to point out that for example the stock data will not evolve in a backwards compatible way, most probably even completely switching technologies for storage. But on the other hand we will of course try to be as backward as possible on API level.

I strongly urge to accept our RESTful resources as "the only truth" and base all other means of data access on these.
If any data is needed for analytical use cases it might also be needed for operational use cases at some point – so let's just have it in our resources.

api/nakadi-event-bus-api.yaml
aikej: Could be a required field when posting a resource event. And the partitioning key could default to this URI – guaranteeing all operations on one resource to be in order.

api/nakadi-event-bus-api.yaml
regispl: @valgog Thanks for reply! Let me refer to your answers in order:

> That also means, that these event-types should have the same ordering key (partitioning key). 

Yes, that's true and I don't see that being an issue or a complexity - producers and consumers should understand this semantics and should adhere to it. Was that considered to be a problem?

> applications should be made responsible of synchronisation of the the flow in such a unified topic and this is much more difficult to achieve then for one event-type. Especially if entities are produced by separate applications (that do not have the means of strong synchronisation usually)

What do you mean by "synchronisation" of the flow? When I think about Nakadi or any other generic "event bus" / dumping ground for time-ordered data, I'm assuming that you just want to push the events there in order they appear in your system. I believe that in a microservice architecture it should be consumer's role to deal with issues like this **if they need it** (producers are dumb, whatever happens, they "record" it as an event) and by consumer here I mean the client application - the one that handles the data (e.g. the app that's processing the time-ordered data) - not the queue consumer (not e.g. Kafka consumer).

Still, I might be missing something and the use-cases I'm thinking about might be somehow limited, so could you please give me an example which you thought about when discussing that problem? I mean - when is it unacceptable for different, non-synchronised services to write events out of order, while it's still OK for them to possibly write them with out-of-order timestamps to their local databases for example? Let's say you want to replay all the events to the queue because of an error in the processing that caused data corruption - if your sources of truth (local DBs) are possibly out of sync in terms of ordering, you still have a problem that you can't fix. 

> we definitely need a correlation engine (probably on the consumer side

Could you give me more details on how you think it should work? By consumer side do you mean something on Nakadi side that happens when client (consumer app) connects for consumption? Would that be in memory? How would you guarantee the ordering between low and high velocity topics then? Imagine an analytics engine for e.g. online store - you have topics with events like PageView (thousands / millions of events per minute) and ProductPurchased (tens of events per minute). How would you define the consumption time window / batch to make sure that you can properly order e.g. 2M PageView events produced over a minute and 10 ProductPurchased events produced in the same time? What if the "large" topic consumption can't catch up? You're reading ProductPurchase events as soon as they appear in the topic, but you're still little behind with PageViewed ones - do you stop consumption of the ProductPurchased and wait for PageViews to catch up? I think you're still not guaranteed to get all the messages in a given time window - even if the consumption time window is much larger than the "time window" used for returning data to the client app to allow you for accepting messages that are "late", it's still not protecting you from all the possible cases. I'm also a bit worried about consumption performance and resource requirements in that case (re-ordering time, memory, consumption delay) as an extra post-processing step is required and it's definitely non-trivial. 

Sorry for the huge number of questions, I hope I haven't lost / bored you somewhere on the way here ( ;-) ), but I'm quite concerned with this approach - I probably need some clarification on what you have in mind because at the moment (as I make a lot of assumptions :-) ) it seems to me like we're picking the right tool to do the job (Kafka), we remove from it what it's really good at (time-ordering, fast consumption) and then add another layer that's doing the same thing again but in less performant way and still allowing for out-of-order messages ;-) 

If I misunderstood the idea of correlation engine, please let me know how you see it as that's probably the most confusing part for me :-)

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+  description: |
+    -----------
+    Definitions
+    -----------
+    Nakadi at its core aims at being a generic and content-agnostic event broker with a convenient API.
+    In doing this, Nakadi abstracts away, as much as possible, details of the backing messaging
+    infrastructure. The single currently supported messaging infrastructure is Kafka (Kinesis is planned
+    for the future).
+
+    In Nakadi every Event has an EventType, and a **stream** of Events is exposed for each registered
```
igoryankov: For people who are not familiar with Kafka it's not clear the relation between EventType, **stream**. Might it be that one eventType goes to many streams? or several eventTypes reside in one **stream**?
What if I have one eventType which is fired once a day. Do we really need to create a whole **stream** (topic) for that?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+          description: Not available
+          schema:
+            $ref: '#/definitions/Problem'
+  /event-types:
+    get:
+      tags:
+        - schema-registry-api
+      description: Returns a list of all registered `EventType`s
+      parameters:
+        - name: X-Flow-Id
```
igoryankov: Is usage of 'X-Flow-Id' common way to trace application? You use it widely for Topic API and registry API. In shop we use exactly this header but for another purpose.

api/nakadi-event-bus-api.yaml
igoryankov: To my mind "bulk publishing" is must-have feature. Without that keeping the order of events fired from even one instance might be very slow.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+
+            * Stream initialization will fail if `stream_limit` is lower than `batch_limit`.
+          type: integer
+          format: int32
+          required: false
+          default: 0
+        - name: batch_flush_timeout
+          in: query
+          description: |
+            Maximum time in seconds to wait for the flushing of each chunk (per partition).
```
igoryankov: Are seconds enough? may be ms?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+
+            Stream initialization will fail if `stream_timeout` is lower than
+            `batch_flush_timeout`.
+          type: integer
+          format: int32
+          required: false
+        - name: batch_keep_alive_limit
+          in: query
+          description: |
+            Maximum number of keep-alive messages to get in a row before closing the connection. Unlimited by
```
igoryankov: Am I correct that for ALL partitions you will fire keep-alive message? It might be too much for keeping connection live, provided my consumer instance consumes many partitions.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+
+              Clients are allowed to generate this and this SHOULD be guaranteed to be unique from the
+              perspective of the producer. Consumers MIGHT use this value to assert uniqueness of reception
+              of the Event.
+
+              If not present upon reception, Nakadi will generate and fill this with a unique value.
+              (TBD: enforce UUID formatif set by the client?)
+            type: string
+            example: '105a76d8-db49-4144-ace7-e683e8f4ba46'
+          event_type:
```
igoryankov: Why do you prefer snake_case rather than camelCase?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+      - metadata
+
+  BusinessEvent:
+    type: object
+    description: |
+      A Business Event.
+
+      Usually represents a status transition in a Business process.
+    allOf:
+      - $ref: '#/definitions/EventMetadata'
```
igoryankov: Why you did not create special fields like 'payload' and 'metadata'? Why do you put all fields in one bucket? It might lead to naming clash.

api/nakadi-event-bus-api.yaml
tmuehl: @igoryankov Bulk Publishing will come later. This one is the first Milestone of the api. You are right, that we will probably need that in the future due to performance reasons.

api/nakadi-event-bus-api.yaml
tmuehl: @igoryankov https://pages.github.bus.zalan.do/ApiGuild/RestfulApiGuidelines/naming/Naming.html

api/nakadi-event-bus-api.yaml
valgog: @igoryankov The amount of discussions and time that we spent fighting about the way, how such a batch API should look like (especially the reporting of errors), showed us that it is not a trivial point. So batch API is the next in the backlog after the first Milestone is reached.

Also one by one submission could be not so bad, if one uses HTTP in keep-alive mode. The difference will be even less, if the size of the messages is much bigger then the size of the HTTP header block

api/nakadi-event-bus-api.yaml
tmuehl: I like the idea of the URI pointing to the changed resource if it's a REST resource. Can we add it as optional field?

api/nakadi-event-bus-api.yaml
tmuehl: Just curious. Why do you prefer the `.`?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+            description: |
+              Identifier of this Event.
+
+              Clients are allowed to generate this and this SHOULD be guaranteed to be unique from the
+              perspective of the producer. Consumers MIGHT use this value to assert uniqueness of reception
+              of the Event.
+
+              If not present upon reception, Nakadi will generate and fill this with a unique value.
+              (TBD: enforce UUID formatif set by the client?)
+            type: string
```
igoryankov: Do you use this 'eid' to prevent event's duplication? What if try to send it, get 500, and try to send it again?

api/nakadi-event-bus-api.yaml
tmuehl: @igoryankov Afaik the X-Flow-Id is used in the shop for exactly the same. What purpose do you mean in the shop?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+        entity (details are provided in the Problem object):
+
+        * Event name on creation not begin unique.
+
+        * Using `EventTypeSchema.type` other than json-schema or passing a `EventTypeSchema.schema`
+        that is invalid with respect to the schema's type.
+
+        * Referring any of Validation, Enrichment or Partitioning strategies that does not exist or whose
+        parameterization is deemed invalid.
+
```
igoryankov: I don't see any description on how to change number of partitions for existing topic. Is that the case for your system? Kinesis provides this feature.

api/nakadi-event-bus-api.yaml
tfrauenstein: Not really. Editorial -- just mention in your 422 violation list:
* EventType already exists (and PUT should be used instead of POST)

api/nakadi-event-bus-api.yaml
valgog: @igoryankov you are right. Rebalancing is on the roadmap. But this will come only after the Subscription API is there, that will make it possible to repartition using the same protocol, as during the rebalancing of partitions when new clients are added to the subscription.

api/nakadi-event-bus-api.yaml
valgog: @aikej  I would prefer to have some kind of a convention about the construction of the URI. What I am afraid of, is that making such an URI we would introduce quite a significant limitation on the payload format in generic case. 

But would not be a requirement from the RESTful resource schema prospective, to provide an URI of the primary key in resource itself? It would even most probably will be the value of the filed(s) that would be marked as a primary key for that resource?

api/nakadi-event-bus-api.yaml
valgog: @aikej :+1:

valgog: @aikej it is a very important and difficult discussion. I would compleatly support your point of view about the resources exposed over the RESTful APIs. The internal structures should be of no concern for anybody outside the team.

The problem that is quite difficult to solve though (and it is of cause out of scope for Nakadi actually), is a more generic problem: how would a service ensure, that what is sent over the bus, is actually what was persisted. If the queue is not the only (and the primary) persistence layer, then any `queue` <-> `persistence layer` operation should be performed in a 2PC (two phase commit), that would be very problematic, even if Nakadi would implement _prepared transactions_. Not every persistence solution implements transactions to support the 2PC. 

From the other hand, if we would use some kind of a _driver level_ persistence layer log extractor, that would eliminate the need to write to two systems in parallel, but would extract the results of the persistence operations in the form of the decoded log, that is automatically delivered into the Nakadi system. One could give much higher consistency guarantees for the whole system.

I would be really glad to discuss this important topic in a more structured way, then using this API review.

api/nakadi-event-bus-api.yaml
aikej: Ordering in such a system is a highly complex – and just as interesting – topic. @valgog knows I have my own ideas about that – maybe slightly "crazy" ones.
In short: I would love to see all events be actually just resource events (creates, updates, deletes of RESTful resources) and ordering only guaranteed for events on the same resource. And then implement a variation of eg Lamperts Logical Clocks to be able to order those events that are actually in a relation.
Even shorter: I wasn't able to convince anybody in the platform prototyping to go for this! ;-)

But thinking of this makes me wonder: Do we maybe need an array as parent-eid(s)? If I do any sort of corelation logic between different events I will actually be reacting to multiple "parent" events?!

api/nakadi-event-bus-api.yaml
aikej: @valgog What limitation do you mean?

Having a required field for all resource events would allow for clients to actually use that URI in a generic way. Without it I can imagine a pretty chaotic mix of wether and how different teams handle this. We could even go further: Knowing from which endpoint a resource originates already has all information about owning team, application and respective schema.
Everything would be so simple. Also it would mean a big step forward for my plans to take over the world with resource events! ;-)

api/nakadi-event-bus-api.yaml
aikej: Very happy to join that discussion – Can you schedule a workshop with interested parties @valgog ?

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+        * TDB Partition resolution strategy
+
+        * The schema of an `EventType` is defined as an `EventTypeSchema`. Currently only json-schema is
+        supported.
+
+        Following violations are enforced. Not meeting them will fail the request with a 422 Unprocessable
+        entity (details are provided in the Problem object):
+
+        * Event name on creation not begin unique.
```
rcillo: Wouldn't it be clearer to say "Event name must be unique".

api/nakadi-event-bus-api.yaml
valgog: !resolved

api/nakadi-event-bus-api.yaml
valgog: There was a huge discussion about this, and DataChangeEvents are providing the `data` container to hold the payload. Business Events are per design simplified and can be mapped directly to the event representation in memory of your application without the need to remap (at least it was the intention).

api/nakadi-event-bus-api.yaml
valgog: @igoryankov yes, the `eid` is indeed though of as a tool to detect duplications.

We request and trust the producers to created unique `eid`s. @tfrauenstein even insists, that nakadi should not generate any ids by itself and just reject events that do not contain `eid` and I now also support him.

We plan to implement a service, that will be showing alerts in case some of the `event_type`s are getting `eid` duplications, but the semantics should be discussed when we start working on it.

api/nakadi-event-bus-api.yaml
valgog: :+1:

api/nakadi-event-bus-api.yaml
valgog: @igoryankov this is a good point. 

We could introduce some optimisations in case we really see that it is a bottleneck. Usually there should not be too many partitions, we are not talking about hundreds of them.

This topic will be more of a concern when we start discussing the Subscription API. But there I hope that an application will need to keep not more then 1 connection to get all the events for the subscription (actually 1 connection for receiving and 1 connection for committing in HTTP1.1)

api/nakadi-event-bus-api.yaml
valgog: @igoryankov hm... I though that in such a situation you probably will not need batching at all and would like to get everything as fast as possible. 

If we see that somebody wants to created batches every half a second for example, we can extend this interface to accept number with a floating point and not int32 value.

Somehow seconds are easier to think about usually.

Please fill free to insist on milliseconds if you have some nice examples

api/nakadi-event-bus-api.yaml
valgog: @igoryankov Flow ID is a global concept and should be used everywhere actually :) nakadi is also an application like any other.

api/nakadi-event-bus-api.yaml
valgog: @igoryankov we want to abstract and hide the idea of Kafka topic actually. How the streams are implemented, using kafka topics or some partitions in a special kafka topic is an implementation detail. 

My idea for the future would be that nakadi could automatically detect highly loaded event-types and repartition them automatically if needed, or move them to "slow" topics mixing with other slow event-types.

api/nakadi-event-bus-api.yaml
valgog: Actually the `.` would make it possible to use it as a part of the URI without the need to rewrite it somehow. But following this logic one could say that the resource should be like

`team`/`application`.`event-type-name`

:scream:

@tfrauenstein what do you think? We talked yesterday about the unified URI for all the resources. Do you have some input here?

api/nakadi-event-bus-api.yaml
valgog: @tfrauenstein yes, probably we can use the enum with a single value for now. 

I suppose, that the reason, why @jkliff did not use enum here, is that this value is a parameter to the validation strategy.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1207 @@
...
+      parameters:
+        - name: event-type
+          in: body
+          description: EventType to be created
+          schema:
+            type: object
+            $ref: '#/definitions/EventType'
+          required: true
+      responses:
+        '200':
```
rcillo: Why not 201 created?

api/nakadi-event-bus-api.yaml
igoryankov: Do you have any concept of forwarding/duplicating some percentage of event-type traffic to another stream so that we can build independent staging infrastracture? Or you will push idea to have another consumer for this purpose?

api/nakadi-event-bus-api.yaml
igoryankov: In new shop it's already more than optional header for troubleshooting. It's only way to bind requests to different fragments related to to one page. We use it widely in shop events payload.

api/nakadi-event-bus-api.yaml
jkliff: I added a note to make clear this is out of scope for now.

api/nakadi-event-bus-api.yaml
jkliff: Corrected.

api/nakadi-event-bus-api.yaml
jkliff: Done.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1118 @@
...
+            $ref: '#/definitions/Problem'
+
+    put:
+      tags:
+        - schema-registry-api
+      description: |
+        Updates the `EventType` identified by its name. Behaviour is the same as creation of `EventType` (See
+        POST /event-type) except where noted below.
+
+        The name and the topic fields cannot be changed. Attempting to do so will result in a 422 failure.
```
rcillo: I think we should remove the reference to the `topic` since it doesn't exists.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1118 @@
...
+        type: array
+        items:
+          type: string
+        description: |
+          Indicators of the path of the properties that constitute the primary key (identifier) of the data
+          within this Event.
+          If set MUST be a valid required field as defined in the schema.
+          (TBD should be required? Is applicable only to both Business and DataChange Events?)
+
+      ordering_key_field:
```
valgog: @jkliff did we change this recently? 

Why only one field? Theoretically it can well be a composite key, the same way as `data_key_fields`. Also we have several times changed the name from `ordering` to `partitioning` and back. My last understanding was the we wanted to finalise on `partitioning_key` (and now seeing the composite key use-case: `partitioning_key_fields`).

Also later on we will have to formalise, what query language is used to extract the fields if they are not on the top level of the data.

api/nakadi-event-bus-api.yaml
```diff
@@ -0,0 +1,1118 @@
...
+      - name
+      - category
+
+  EventTypeSchema:
+    properties:
+      type:
+        type: string
+        example: 'json-schema'
+        description: The type of schema definition (avro, json-schema, etc).
+      schema:
```
tmuehl: I think it's a bit unfortunate to have `EventType.schema.schema`, i.e. an object `Schema` that exposes a property `schema`.

api/nakadi-event-bus-api.yaml
jkliff: (Y)

PR #58, opened 2016-01-05
Make nakadi jvm deployable

tmuehl:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/58" target="_blank">ReviewNinja

build.gradle
```diff
@@ -157,8 +157,9 @@ task buildDockerContainer {
 }
 
 task startDockerContainer(type: Exec) {
-    commandLine "bash", "-c", "docker run -e \"NAKADI_OAUTH2_ENABLED=false\" -e \"NAKADI_CONFIG=file:////envs/development.conf\" " +
-            "--net container:local-kafka -i -t -d " +
+    commandLine "bash", "-c", "docker run -e \"NAKADI_OAUTH2_ENABLED=false\" -e \"ZK_BROKERS=127.0.0.1:2181\" " +
```
v-stepanov: I think it would be better if default local configuration will be stored in application.yml. And that it can be overridden with environment variables.
And `@Value` annotation is a perfect tool for that. For instance if you have 
```java
@Value("${zookeeper.brokers}")
private String zookeeperBrokers;
```
Spring will first look for environment variable `ZOOKEEPER_BROKERS` if it doesn't find it - it will look for setting `zookeeper.brokers` in application.yml.
So for local config we will have everything in application.yml and for real deployment we will specify everything with env vars.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -48,15 +49,20 @@
...
     private static final long KAFKA_READ_TIMEOUT_MS = 0;
 
-    @Autowired
-    private KafkaFactory factory;
-
-    @Autowired
-    private ZooKeeperHolder zkFactory;
+    private final ZooKeeperHolder zkFactory;
+    private final Producer<String, String> kafkaProducer;
+    private KafkaFactory kafkaFactory;
```
v-stepanov: also can be final

v-stepanov:
Strange but the build for PR fails on travis. However it builds well locally for me.
```
:startDockerContainerdocker: "run" requires a minimum of 1 argument.

See 'docker run --help'.
```

I will look into that

v-stepanov:
I think the problem is that it can't find docker container to run with our magic expression
```
"\$(docker images | grep -m 1 \"^$dockerGroup/$dockerApplicationName\" | awk \'{print \$3}\')"
```
Maybe because you give container a name 'nakadi'

v-stepanov:
OK, now I found the reason. Docker container is not built :) Locally it uses the one that was built previously 

Step 4 : ADD envs envs
envs: no such file or directory

build.gradle
```diff
@@ -157,8 +157,9 @@ task buildDockerContainer {
...
+            "-e \"KAFKA_BROKERS=127.0.0.1:9092\" " +
```
jkliff: address of kafka and zookeeper should not be hardcoded in this command.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -48,15 +49,20 @@
...
 
     @Value("${nakadi.kafka.poll.timeoutMs}")
     private long kafkaPollTimeout;
```
jhorstmann: Could also be final if moved to constructor injection.

build.gradle
tmuehl: :+1:

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
tmuehl: :+1:

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
tmuehl: :+1:

src/main/java/de/zalando/aruha/nakadi/config/NakadiConfig.java
```diff
@@ -29,6 +33,11 @@
 
     public static final MetricRegistry METRIC_REGISTRY = new MetricRegistry();
     public static final HealthCheckRegistry HEALTH_CHECK_REGISTRY = new HealthCheckRegistry();
+    public static final String ZK_BROKERS_DEFAULT = "127.0.0.1:2181";
+    public static final String KAFKA_BROKERS_DEFAULT = "127.0.0.1:9092";
```
v-stepanov: Why don't you put these default properties to application.yml ?

tmuehl: Because then it's yet two other configuration parameters that will never be overridden. I don't think we need this to be configurable (it's the default value if the actual value is not set). Hence, I try to keep my configuration lean. :)
Imho, if you want to change it, you can change it in the code and redeploy. :)

v-stepanov: I don't suggest to put default values to application.yml. I suggest to put ZK_BROKERS and KAFKA_BROKERS to application.yml. And that will already work as a default mechanism. (If no env. vars set - the values from application.yml will be used as defaults)

tmuehl: Ah, ok. :) Done.

PR #66, opened 2016-01-12
Fetch list of kafka brokers from zookeeper

CyberDem0n:
Refactor code to use apache-curator library which is using exhibitor api
to track topology changes of zookeeper cluster. List of kafka nodes
periodically updated from zookeeper.

Initial list of exhibitor nodes and exhibitor port has to be configured.
Kafka namespace must be configured separately from zookeepe connection
string


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/66" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -0,0 +1,129 @@
...
+                return false;
+            return this.compareTo((Broker)other) == 0;
+        }
+
+        public String toString() {
+            return this.host + ":" + this.port;
+        }
+    }
+
+    class Brokers extends ArrayList<Broker> {
```
tmuehl: Why extra class?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -0,0 +1,129 @@
...
+    @Scheduled(fixedDelay = 30000)
+    private void updateBrokers() {
+        if (kafkaProperties != null) {
+            kafkaBrokerList = fetchBrokers();
+            kafkaProperties.setProperty("bootstrap.servers", buildBootstrapServers(kafkaBrokerList));
+        }
+    }
+
+    public Brokers getKafkaBrokers() {
+        return kafkaBrokerList;
```
tmuehl: Immutable copy of

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -0,0 +1,129 @@
...
+    private final String _BROKERS_IDS_PATH = "/brokers/ids";
+
+    @Autowired
+    private ZooKeeperHolder zkFactory;
+
+    private Brokers kafkaBrokerList;
+
+    private Properties kafkaProperties;
+
+    class Broker implements Comparable<Broker> {
```
tmuehl: public?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        @Override
+        public int compareTo(Broker other) {
+            final int result = host.compareTo(other.host);
+            return result == 0 ? port.compareTo(other.port) : result;
+        }
+
+        @Override
+        public boolean equals(Object other) {
```
tmuehl: If you override equals() always also override hashCode()

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
v-stepanov: I also think that this class can be removed now. It was created as a fix for DependencyInjection. But now the list of brokers is not a bean and not injected anywhere, so I think we don't need Brokers class any more.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -0,0 +1,129 @@
...
+        final String host;
+        final Integer port;
+
+        public Broker(final String host, final Integer port) {
+            this.host = host;
+            this.port = port;
+        }
+
+        @Override
+        public int compareTo(Broker other) {
```
tmuehl: For what do you need comparable?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -0,0 +1,129 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.scheduling.annotation.Scheduled;
+
+import de.zalando.aruha.nakadi.repository.zookeeper.ZooKeeperHolder;
+
+public class KafkaLocationManager {
```
tmuehl: Would love to see a KafkaLocationManagerTest. :)

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
CyberDem0n: My original idea was: if the list of brokers didn't changed in ZooKeeper we don't need to update it in the application.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -0,0 +1,126 @@
...
+    private List<Broker> fetchBrokers() {
+        final List<Broker> brokers = new ArrayList<Broker>();
+        try {
+            final CuratorFramework curator = zkFactory.get();
+            for (final String brokerId : curator.getChildren().forPath(_BROKERS_IDS_PATH)) {
+                try {
+                    final byte[] brokerData = curator.getData().forPath(_BROKERS_IDS_PATH + "/" + brokerId);
+                    brokers.add(getBroker(brokerData));
+                } catch (Exception e) {
+                    LOG.info("Failed to fetch connection string for broker {}: {}", brokerId, e);
```
tmuehl: I think it is not working that way. The interface of slf4j is:

```
info(String msg); 
info(String format, Object arg); 
info(String format, Object... args);           
info(String msg, Throwable t);
```

So if you pass in two parameters it will treat the Throwable as an object and thus calls e.toString() and loose the stacktrace.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -0,0 +1,126 @@
...
+                }
+            }
+        } catch (Exception e) {
+            LOG.error("Failed to fetch list of brokers from ZooKeeper: {}", e);
```
tmuehl: See above. If you pass in a throwable as a second parameter the first parameter won't be a format string but rather a fixed string. Format string is not working here. :(

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -0,0 +1,129 @@
...
+            List<Broker> brokers = fetchBrokers();
+            if (!brokers.isEmpty()) {
+                kafkaBrokerList = brokers;
+                kafkaProperties.setProperty("bootstrap.servers", buildBootstrapServers(brokers));
+            }
+        }
+    }
+
+    public List<Broker> getKafkaBrokers() {
+        return kafkaBrokerList;
```
tmuehl: Maybe it's worth to return an immutable copy here.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    public Properties getKafkaProperties() {
+        return kafkaProperties;
```
tmuehl: Maybe it's worth to return an immutable copy here.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationConfig.java
```diff
@@ -10,20 +10,40 @@
...
+
+    @Bean(name = "exhibitorAddress")
+    public String exhibitorAddress() {
+        return exhibitorAddress;
+    }
+
+    @Bean(name = "exhibitorPort")
+    public Integer exhibitorPort() {
+        return exhibitorPort;
+    }
```
tmuehl: I personally would prefer using constructor based injection instead having all those strings as beans for configuration. Makes it more straightforward, better testable and clear.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -54,12 +32,13 @@ private Properties getProps() {
...
     public Consumer<String, String> getConsumer() {
-        return new KafkaConsumer<>(getProps());
+        return new KafkaConsumer<>(kafkaLocationManager.getKafkaProperties());
     }
 
     public SimpleConsumer getSimpleConsumer() {
-        final String[] split = kafkaAddress.split(":");
-        return new SimpleConsumer(split[0], Integer.parseInt(split[1]), 1000, 64 * 1024, "leaderlookup");
+        for (Broker kafkaBroker : kafkaLocationManager.getKafkaBrokers()) {
+            return new SimpleConsumer(kafkaBroker.host, kafkaBroker.port, 1000, 64 * 1024, "leaderlookup");
```
v-stepanov: I believe in java it is considered a bad practice to directly access class fields outside of the class.

build.gradle
```diff
@@ -97,7 +97,7 @@ dependencies {
     // kafka & zookeeper
     compile 'org.apache.kafka:kafka-clients:0.9.0.0'
     compile 'org.apache.kafka:kafka_2.11:0.9.0.0'
-    compile 'org.apache.zookeeper:zookeeper:3.4.5'
+    compile 'org.apache.curator:curator-framework:2.9.1'
```
v-stepanov: @CyberDem0n  BTW, will it still be possible to use pure Zookeeper without Exhibitor?

CyberDem0n: Yes, it's possible. Locally it works without exhibitor. Curator configured in the way that if exhibitor is not accessible it will use static zookeeper connection string as a fallback.

v-stepanov: ok, cool

src/main/resources/application.yml
```diff
@@ -1,3 +1,4 @@
+ZK_BROKERS: 127.0.0.1:2181
```
v-stepanov: do you still need this one?

PR #68, opened 2016-01-14
Add reference to application owning an EventType; flowids

jkliff:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/68" target="_blank">ReviewNinja

jhorstmann:
Shouldn't the header be `X-Flow-ID`? I think you can also use something like

    parameters:
      - $ref: '#/parameters/Flow-ID'

In order to not repeat the description everywhere.

api/nakadi-event-bus-api.yaml
```diff
@@ -119,6 +119,15 @@ paths:
...
       description: Returns a list of all registered `EventType`s
+      parameters:
+        - name: X-Flow-Id
+          in: header
+          description: |
+            The flow id of the request, which is written into the logs and passed to called services. Helpful
+            for operational troubleshooting and log analysis.
+          type: string
+          format: flow-id
+          required: true
```
valgog: I do not think that these headers should be required.

PR #71, opened 2016-01-18
Move publishing (with resolution) under event-type, remove similar op…

jkliff:
- Removes pushing with resolution of partition out of /topic.
- Exposes all 3 types of strategies for information of possibilities of configuration of EventType.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/71" target="_blank">ReviewNinja

PR #72, opened 2016-01-19
Hila streaming preparations

v-stepanov:
@tmuehl @valgog @CyberDem0n @jkliff
Please take a look. The new functionality:
- creation of subscriptions
- HiLA streaming with dynamic partitions rebalancing
- committing consumed offsets
- added support for streaming from multiple topics
The API for subscriptions is pretty random, because we don't yet have that in our swagger.

What is not implemented yet:
- Currently there is no pause during rebalance, so we currently don't wait for commits of already sent messages (we need to discuss first how we want to implement that)
- When rebalancing happens we also need to send a new topology to a stream (we need to discuss its format)
- tests (will be added to this PR)

I believe there's a lot of space for improvements, so feel free to comment/criticize/advice


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/72" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -24,13 +24,16 @@
 
     private final String _BROKERS_IDS_PATH = "/brokers/ids";
 
-    @Autowired
     private ZooKeeperHolder zkFactory;
```
tmuehl: Can be final now.
(Furthermore, you could remove that PostConstruct and do it all in the Constructor.)

src/main/java/de/zalando/aruha/nakadi/repository/ZkSubscriptionRepository.java
```diff
@@ -0,0 +1,142 @@
...
+            for (final String topic : topics) {
+                createPersistentEmptyNode("/nakadi/subscriptions/%s/topics/%s", subscriptionId, topic);
+                createPersistentEmptyNode("/nakadi/subscriptions/%s/topics/%s/partitions", subscriptionId, topic);
+            }
+            for (final Cursor cursor: cursors) {
+                final String path = String.format("/nakadi/subscriptions/%s/topics/%s/partitions/%s", subscriptionId,
+                        cursor.getTopic(), cursor.getPartition());
+                zooKeeperHolder
+                        .get()
+                        .create()
```
CyberDem0n: .creatingParentsIfNeeded() will make unnecessary to create parent folders on lines 24 and [26-29]

v-stepanov: fixed

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
v-stepanov: @tmuehl fixed

v-stepanov:
@tmuehl @valgog @CyberDem0n @jkliff 
I created the acceptance and unit tests I wanted to have. So now I think it is finished. I will be happy if you take a second look.

src/main/java/de/zalando/aruha/nakadi/NakadiRuntimeException.java
```diff
@@ -0,0 +1,19 @@
+package de.zalando.aruha.nakadi;
+
+import de.zalando.aruha.nakadi.domain.Problem;
+
+public class NakadiRuntimeException extends RuntimeException {
```
tmuehl: Why RuntimeException?

v-stepanov: To use it in places that will be used in streams and lambdas :)
Because the checked exception needed to be declared in functional interfaces.

tmuehl: :+1:

src/main/java/de/zalando/aruha/nakadi/service/EventStreamCoordinator.java
```diff
@@ -0,0 +1,174 @@
...
+        subscriptionRepository.removeClient(eventStream.getSubscriptionId(), eventStream.getClientId());
+        rebalanceSubscriptionIfNeeded(eventStream.getSubscriptionId());
+    }
+
+    /**
+     * Checks if rebalance is needed for subscriptions running on this Nakadi instance
+     * todo: in general this scheduled task should be probably replaced with watching appropriate zookeeper nodes
+     */
+    @Scheduled(fixedRate = SUBSCRIPTION_REBALANCE_CHECK_PERIOD_IN_MS)
+    private void rebalanceSubscriptions() {
```
tmuehl: Maybe rename to `rebalanceAllSubscriptionsIfNeeded`? I had to read the code first to understand what the method actually does.

src/main/java/de/zalando/aruha/nakadi/service/EventStreamCoordinator.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    /**
+     * If new topology for this subscription found - rebalances according to new topology
+     *
+     * @param subscriptionId id of subscription to check for rebalance
+     */
+    private void rebalanceSubscriptionIfNeeded(final String subscriptionId) {
+        final Topology topology = subscriptionRepository.getTopology(subscriptionId);
```
tmuehl: I think it makes sense to have a mutex / lock here per subscription id.

tmuehl:
In general: Awesome code and approach! @v-stepanov :trophy:

src/main/java/de/zalando/aruha/nakadi/service/EventStream.java
```diff
@@ -55,23 +88,37 @@ public void streamEvents() {
...
+                    eventConsumer.setCursors(newCursors);
+                    partitions = cursorsToPartitions(newCursors);
+                    latestOffsets = cursorsToOffsetMap(newCursors);
+                    // forget about all messages not flushed
+                    messagesReadInBatch = 0;
+                    currentBatch = emptyBatch();
+                    batchStartTime = currentTimeMillis();
+                }
+
                 final Optional<ConsumedEvent> eventOrEmpty = eventConsumer.readEvent();
```
tmuehl: If you would pass in the timeout for `NakadiKafkaConsumer.pollFromKafka()` as optional parameter, you could use here the time-left-until-batch-timeout to have the optimal timeout. :)

src/main/java/de/zalando/aruha/nakadi/service/EventStreamCoordinator.java
v-stepanov: I'm not really sure what place exactly you mean.
If you mean to lock topology in subscription repository - it is not really needed as topology is just a list of client-ids in zookeeper. And it is not needed to be synchronized. (but maybe let's discuss that personally)

src/main/java/de/zalando/aruha/nakadi/service/EventStreamCoordinator.java
v-stepanov: It will sound like it will rebalance ALL subscriptions if needed :-D But in fact it will rebalance only needed. Maybe `rebalanceNeededSubscriptions` ?

src/main/java/de/zalando/aruha/nakadi/service/EventStream.java
v-stepanov: ah ok I see, that will make reading from kafka to be in-sync with batch timeout, right?

src/main/java/de/zalando/aruha/nakadi/service/EventStreamCoordinator.java
tmuehl: Yes, sounds good! :)

src/main/java/de/zalando/aruha/nakadi/service/EventStream.java
tmuehl: Exactly!

src/main/java/de/zalando/aruha/nakadi/service/EventStreamCoordinator.java
tmuehl: Yes. Discussed personally :white_check_mark:

src/main/java/de/zalando/aruha/nakadi/service/EventStream.java
v-stepanov: I again examined the possibility to implement that. And now I think that I would not like to do that at least for now. The long polls (when poll time equals the batch timeout) can bring problems. One of them is if we decide to switch to per-partition batch-timeout tracking - then the long polls will not make sense. (And as I understand from some discussion in the past - switching to per-partition timeouts is quite possible)
So for now I would like to leave this part as it is. (And maybe implement it later as an improvement)

src/main/java/de/zalando/aruha/nakadi/service/EventStreamCoordinator.java
v-stepanov: I checked again possible scenarios in my head and I think if even some clients start for subscription at the same time - it will not be a problem as a scheduled task will rebalance anyway to the latest version of topology.

src/main/java/de/zalando/aruha/nakadi/service/EventStreamCoordinator.java
v-stepanov: Ok, I renamed a method.

PR #73, opened 2016-01-26
Aruha 51 topics creation

v-stepanov:
@jkliff @valgog @rcillo @tmuehl @CyberDem0n 
Please review

### DIFF
--- a/src/acceptance-test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepositoryTest.java
+++ b/src/acceptance-test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepositoryTest.java
@@ -6,6 +6,7 @@ import org.apache.curator.CuratorZookeeperClient;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.common.PartitionInfo;
+import org.echocat.jomon.runtime.concurrent.RetryForSpecifiedTimeStrategy;
 import org.junit.Before;
 import org.junit.Test;
 
@@ -13,6 +14,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 
+import static org.echocat.jomon.runtime.concurrent.Retryer.executeWithRetry;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.arrayWithSize;
 import static org.hamcrest.Matchers.hasItem;
@@ -37,6 +39,7 @@ public class KafkaRepositoryTest {
     }
 
     @Test(timeout = 10000)
+    @SuppressWarnings("unchecked")
     public void whenCreateTopicThenTopicIsCreated() throws InterruptedException {
 
         // ARRANGE //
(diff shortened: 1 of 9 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/73" target="_blank">ReviewNinja

src/main/resources/application.yml
```diff
@@ -5,17 +5,27 @@ logging:
     ROOT: INFO
     org.springframework.web: INFO
 nakadi:
+  topic:
+    default:
+      partitionNum: 8
+      replicaFactor: 1
```
v-stepanov: In local configuration it is 1 instead of 3 because in local environment we have only one kafka broker. This should be overridden with env var when deploying to AWS.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -41,42 +45,68 @@
...
+                            final long retentionMs, final long rotationMs) {
+
+        final String connectionString = zkFactory.get().getZookeeperClient().getCurrentConnectionString();
+        final ZkClient zkClient = new ZkClient(connectionString, settings.getZkSessionTimeoutMs(),
+                settings.getZkConnectionTimeoutMs(), ZKStringSerializer$.MODULE$);
+        final ZkUtils zkUtils = ZkUtils.apply(zkClient, false);
+
+        final Properties topicConfig = new Properties();
+        topicConfig.setProperty("retention.ms", Long.toString(retentionMs));
+        topicConfig.setProperty("segment.ms", Long.toString(rotationMs));
```
v-stepanov: @CyberDem0n, it is a right setting for the rotation period?

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepositoryTest.java
```diff
@@ -0,0 +1,96 @@
...
+        when(zooKeeperHolder.get()).thenReturn(curatorFramework);
+
+        final KafkaRepository kafkaRepository = new KafkaRepository(zooKeeperHolder, mock(KafkaFactory.class),
+                repositorySettings);
+
+        // ACT //
+        kafkaRepository.createTopic(topicName);
+
+        // ASSERT //
+        Thread.sleep(2000);
```
rcillo: Is there any rational on why 2000 as the sleep time? I was thinking if it would be interesting to have a "retry strategy" for async assertions. So we could keep running the asserting until it passed **or** a certain timeout was reached.

Wdyt?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -41,42 +45,68 @@
...
+
+        AdminUtils.createTopic(zkUtils, topic, partitionsNum, replicaFactor, topicConfig);
+        zkClient.close();
```
rcillo: I was thinking if is there any chance of an error causing this connection to remain open. Do you think we should wrap this statement in a `finally` clause?

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepositoryTest.java
v-stepanov: The 2000 timeout is not rational it's just what feels to be enough :)

I really like your idea with retries and timeout. I believe it should be somehow implemented using JUnit TestRule. But for the beginning I would probably like to keep the tests as simple as possible, so I think this implementation with retries will be an overcomplication at this stage (unless there's already an implementation of that somewhere :-P)

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
v-stepanov: Yes, you are right. I wrapped it now as you suggested.

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepositoryTest.java
tmuehl: @v-stepanov You can find the `Retryer` here: :) https://github.com/echocat/jomon/tree/master/runtime/src/main/java/org/echocat/jomon/runtime/concurrent

Can be added as a dependency:
http://search.maven.org/#artifactdetails%7Corg.echocat.jomon%7Cruntime%7C1.6.2%7Cjar

v-stepanov: @rcillo @tmuehl 
Ok, you win, I added a retry strategy :)
That library is quite good, thanks for the hint.

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepositoryTest.java
```diff
@@ -0,0 +1,102 @@
...
+        when(zooKeeperHolder.get()).thenReturn(curatorFramework);
+
+        final KafkaRepository kafkaRepository = new KafkaRepository(zooKeeperHolder, mock(KafkaFactory.class),
+                repositorySettings);
+
+        // ACT //
+        kafkaRepository.createTopic(topicName);
+
+        // ASSERT //
+        executeWithRetry(() -> {
```
rcillo: This is a very nice :clap:

rcillo:
Looks great :+1: :shipit:

PR #76, opened 2016-01-27
Make nakadi jvm deployable

tmuehl:
Please have a look @v-stepanov @CyberDem0n @rcillo @valgog

### DIFF
--- a/senza.yaml
+++ b/senza.yaml
@@ -17,6 +17,9 @@ SenzaInfo:
     - ScalyrAccountKey:
         Description: "Key for Skalyr logging (optional)"
         Default: ""
+    - InstanceType:
+        Description: "The EC2 Instance type that is used for the Nakadi instances (optional, default: t2.micro)"
+        Default: "t2.micro"
     - InstancesMinumum:
         Description: "The minimum number of instances of the AutoScalingGroup (optional, default: 2)"
         Default: 2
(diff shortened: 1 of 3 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/76" target="_blank">ReviewNinja

senza.yaml
```diff
@@ -0,0 +1,162 @@
...
+        application_id: "{{Arguments.ApplicationId}}"
+        application_version: "{{SenzaInfo.StackVersion}}"
+        mint_bucket: "{{Arguments.MintBucket}}"
+        runtime: Docker
+        source: "{{Arguments.DockerImage}}"
+        health_check_path: /health
+        ports:
+          8080: 8080
+        scalyr_account_key: "{{Arguments.ScalyrAccountKey}}"
+        environment:
```
v-stepanov: some of these are legacy variables from python version which are not used in Nakadi-jvm

v-stepanov: Also I think it would be good to take a look at the latest list of properties https://github.com/zalando/nakadi/blob/nakadi-jvm/src/main/resources/application.yml and to check whether we want to expose some of that here. I believe at least the replica factor should be set here (because in default config it is 1)

v-stepanov:
Build fails for this branch, so it will make sense first to merge Alex's branch to this one (that will fix the build) and after that merge this branch to nakadi-jvm.

tmuehl:
@v-stepanov You're right, I thought it is already. Sorry.

PR #81, opened 2016-02-01
Aruha 78 read streams

v-stepanov:
@tmuehl @rcillo @CyberDem0n @jkliff @valgog 
Please take a look. It is an updated streaming and tests for streaming; actually quite a lot of things needed to be changed to align with latest API.
The only thing that is not there is the exclusive offsets. I faced a problem with that, which I would like to discuss when @jkliff and @valgog are back. So I will deliver (or not) that as a separate ticket/PR.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/81" target="_blank">ReviewNinja

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/utils/KafkaHelper.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.TopicPartition;
+
+import java.util.List;
+import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
+
+public class KafkaHelper {
```
jkliff: KafkaTestHelper

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -0,0 +1,170 @@
...
+                        .withBatchTimeout(batchTimeout)
+                        .withStreamTimeout(streamTimeout)
+                        .withBatchKeepAliveLimit(batchKeepAliveLimit)
+                        .build();
+
+                response.setStatus(HttpStatus.OK.value());
+
+                final String acceptEncoding = request.getHeader("Accept-Encoding");
+                final boolean gzipEnabled = acceptEncoding != null && acceptEncoding.contains("gzip");
+                final OutputStream output = gzipEnabled ? new FlushableGZIPOutputStream(outputStream) : outputStream;
```
jkliff: ok, this way of handling gzip is very messy: 
- supporting gzip controls the whole data transport and should not have to be dealt with explicitly this deep in the streaming). ARUHA-84 created for this.
- we end up passing this stream around and having to close optionally. this is made worse by the fact that it's being written to the stream in several different places.

src/main/java/de/zalando/aruha/nakadi/controller/TopicsController.java
```diff
@@ -72,8 +42,8 @@
     public ResponseEntity<?> listPartitions(@PathVariable("topicId") final String topicId) {
         try {
             return ok().body(topicRepository.listPartitions(topicId));
-        } catch (final NakadiException e) {
-            return status(503).body(e.getProblemMessage());
+        } catch (final Exception e) {
```
jkliff: This is not correct. NakadiException (or the ones that are known in a controlled manner) can be a source of 503, but not the general case like this. The catch all should be a 500, but 503 is a specific case.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -99,6 +94,30 @@ public void createTopic(final String topic, final int partitionsNum, final int r
...
     }
 
+    public boolean topicExists(final String topic) throws NakadiException {
+        return listTopics()
+                .stream()
+                .map(Topic::getName)
+                .anyMatch(t -> t.equals(topic));
+    }
+
+    public boolean areCursorsCorrect(final String topic, final List<Cursor> cursors) {
```
jkliff: What means "correct"?

src/main/java/de/zalando/aruha/nakadi/config/NakadiConfig.java
```diff
@@ -86,18 +79,36 @@ public ZooKeeperHolder zooKeeperHolder() {
...
     @Bean
-    public KafkaRepositorySettings topicRepositorySettings() {
+    @Profile("kafka")
+    public KafkaRepositorySettings kafkaRepositorySettings() {
         return new KafkaRepositorySettings();
     }
 
+    @Bean
+    public TopicRepository topicRepository() {
+        if (environment.acceptsProfiles("kafka")) {
```
jkliff: it'd be simpler now not having any of these profiles for this.
a) once we support other brokers nothing is telling us that the same nakadi instance will not be able to use more than one 
b) having this now forces us to specify this profile, which is not adding any value.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/utils/KafkaHelper.java
v-stepanov: Agree. I now renamed it within another PR that is based on this branch: https://github.com/zalando/nakadi/pull/84

src/main/java/de/zalando/aruha/nakadi/controller/TopicsController.java
v-stepanov: I fixed that now in a context of https://github.com/zalando/nakadi/pull/84 (that is base on this branch)

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
v-stepanov: Not out of bounds. 
Think it's better to rename it like that?

src/main/java/de/zalando/aruha/nakadi/config/NakadiConfig.java
v-stepanov: Can't agree with you at this point.

a) If we want one Nakadi instance to be able to use different brokers types we will then have to completely rework current architecture, as it will not be possible to achieve with current DI we have.
b) "forces us to specify this profile" sounds very scary but in fact this is just one property in application.yml and that's it.

Also, using spring-profiles opens a theoretical possibility to support other brokers (yes, only one per instance), and I believe this is better than nothing.

tmuehl: For me it also looks a bit like premature optimization. Until we really have other brokers I wouldn't introduce that profile. Having a lot of profiles and configurations (even if they have a usable default) makes the application more complex and leads to questions. I'd keep it as simple as possible.

Following this path we also might introduce profiles for Postgres and Zookeeper backends as well and maybe other things. That leads to a big mess up.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -0,0 +1,162 @@
...
+                    } catch (IOException e) {
+                        writeProblemResponse(response, outputStream, HttpStatus.BAD_REQUEST.value(),
+                                new Problem("incorrect syntax of X-nakadi-cursors header"));
+                        return;
+                    }
+                }
+
+                // check that offsets are not out of bounds
+                if (cursors.isPresent() && !topicRepository.areCursorsCorrect(topic, cursors.get())) {
+                    writeProblemResponse(response, outputStream, HttpStatus.UNPROCESSABLE_ENTITY.value(),
```
jkliff: Unprocessable entity would be more of a semantic issue (one can't make sense of the cursor, like missing data); looking up, I believe is more correct to say 412 Precondition Failed:

>    The 412 (Precondition Failed) status code indicates that one or more
>    conditions given in the request header fields evaluated to false when
>    tested on the server.  This response code allows the client to place
>    preconditions on the current resource state (its current
>    representations and metadata) and, thus, prevent the request method
>    from being applied if the target resource is in an unexpected state.

https://tools.ietf.org/html/rfc7232#section-4.2
Will include this information on the API too.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
jkliff: probably something in this direction would be indeed better, specially looking at the call site: 
```
// check that offsets are not out of bounds
if (cursors.isPresent() && !topicRepository.areCursorsCorrect(topic, cursors.get())) {
```

something like your suggestion would be immediately clear.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
v-stepanov: Changed it to 412

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
v-stepanov: Well, it appears it also checks if cursors' partitions exists, so areCursorsOutOfBounds will be misleading. 
So I just renamed to areCursorsValid, don't really see a possibility to have a more specific name.

src/main/java/de/zalando/aruha/nakadi/config/NakadiConfig.java
v-stepanov: Ok, if you all think it's not yet time for that - fine for me.
I removed "kafka" profile now

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -0,0 +1,170 @@
...
+            @RequestParam(value = "batch_limit", required = false, defaultValue = "1") final int batchLimit,
+            @RequestParam(value = "stream_limit", required = false, defaultValue = "0") final int streamLimit,
+            @RequestParam(value = "batch_flush_timeout", required = false, defaultValue = "30") final int batchTimeout,
+            @RequestParam(value = "stream_timeout", required = false, defaultValue = "0") final int streamTimeout,
+            @RequestParam(value = "stream_keep_alive_limit", required = false, defaultValue = "0") final int streamKeepAliveLimit,
+            @Nullable @RequestHeader(name = "X-Flow-Id", required = false) final String flowId,
+            @Nullable @RequestHeader(name = "X-nakadi-cursors", required = false) final String cursorsStr,
+            final NativeWebRequest request, final HttpServletResponse response) throws IOException {
+
+        LOG.trace("starting event stream for flow id: {}", flowId);
```
tmuehl: You shouldn't deal manually with FlowIds. It should be transparent to the business code and be handled globally (hopefully in an dedicated ticket).

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -0,0 +1,170 @@
...
+
+            }
+            catch (final NakadiException e) {
+                writeProblemResponse(response, outputStream, SERVICE_UNAVAILABLE, e.getProblemMessage());
+            }
+            catch (final Exception e) {
+                writeProblemResponse(response, outputStream, INTERNAL_SERVER_ERROR, e.getMessage());
+            }
+            finally {
+                outputStream.flush();
```
tmuehl: No need to flush. `close()` already flushes.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
v-stepanov: It was added before our discussion about flow-id :)
I removed it now

src/main/java/de/zalando/aruha/nakadi/domain/Cursor.java
```diff
@@ -0,0 +1,57 @@
+package de.zalando.aruha.nakadi.domain;
+
+public class Cursor {
```
tmuehl: Why not making this class immutable?

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
v-stepanov: How do you know that?
The documentation of OutputStream.close() does not say that. And the implementation can do whatever it wants.
So I would prefer to stay on the safe side.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -22,7 +22,7 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(KafkaLocationManager.class);
 
-    private final String _BROKERS_IDS_PATH = "/brokers/ids";
+    private static final String _BROKERS_IDS_PATH = "/brokers/ids";
```
tmuehl: Just curious: Why is there an underscore leading the constant name? Old C++ habit? :blush:

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -38,7 +35,6 @@
 
 import kafka.javaapi.consumer.SimpleConsumer;
 
-@Component
```
tmuehl: I like. :)

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -99,6 +94,30 @@ public void createTopic(final String topic, final int partitionsNum, final int r
         }
     }
 
+    public boolean topicExists(final String topic) throws NakadiException {
```
tmuehl: `final String topicName`?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
v-stepanov: don't know, I just added `static`

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -99,6 +94,30 @@ public void createTopic(final String topic, final int partitionsNum, final int r
...
+        return listTopics()
+                .stream()
+                .map(Topic::getName)
+                .anyMatch(t -> t.equals(topic));
+    }
+
+    public boolean areCursorsValid(final String topic, final List<Cursor> cursors) {
```
tmuehl: `final String topicName` ?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -99,6 +94,30 @@ public void createTopic(final String topic, final int partitionsNum, final int r
...
+        return cursors
+                .stream()
+                .allMatch(cursor -> partitions
+                        .stream()
+                        .filter(tp -> tp.getPartitionId().equals(cursor.getPartition()))
+                        .findFirst()
+                        .map(pInfo -> {
+                            final long newestOffset = Long.parseLong(pInfo.getNewestAvailableOffset());
+                            final long oldestOffset = Long.parseLong(pInfo.getOldestAvailableOffset());
+                            final long offset = Long.parseLong(cursor.getOffset());
```
tmuehl: This can potentially throw a `NumberFormatException` if the user specifies a non-number String in the header. A `NumberFormatException` here will result into a `500 Internal Server Error`. Instead it should be sth `4xx` (according to the api).

src/main/java/de/zalando/aruha/nakadi/domain/Cursor.java
v-stepanov: At first I wanted to do that but I faced deserialization problems (as there were no setters) which I was lazy to fix :)
Now I made it immutable and fixed deserialization.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
v-stepanov: I think just `topic` is better. The names like `topicId` or `topicName` are misleading as for me, because they assume there is something else than id or name (but topic doesn't have anything else)

And I think the Topic class is overcomplication, I don't use it anywhere.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
v-stepanov: please see my comment above

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
tmuehl: +1

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -0,0 +1,167 @@
...
+                final String acceptEncoding = request.getHeader("Accept-Encoding");
+                final boolean gzipEnabled = acceptEncoding != null && acceptEncoding.contains("gzip");
+                final OutputStream output = gzipEnabled ? new FlushableGZIPOutputStream(outputStream) : outputStream;
+
+                if (gzipEnabled) {
+                    response.addHeader("Content-Encoding", "gzip");
+                }
+
+                final EventConsumer eventConsumer = topicRepository.createEventConsumer(topic,
+                        streamConfig.getCursors());
```
tmuehl: What if the user provides foobar partition ids in the cursors? This should result in a `4xx`. Missing unit tests that cover all this.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
v-stepanov: Very good point, thanks. I fixed that now.

tmuehl:
@v-stepanov There is no single unit test (have I overseen?). Particularly the class `EventStreamController ` should be covered very well, imho. A lot is going on there and we have to be correct in what status code we respond to the client and validate the client's input. I think that should be unit tested.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
v-stepanov: If user provides partition `foobar` it will be detected by the check on line 106 
`topicRepository.areCursorsValid(topic, cursors)`

src/main/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumer.java
```diff
@@ -52,25 +54,26 @@ public NakadiKafkaConsumer(final KafkaFactory factory, final String topic, final
...
             final ConsumerRecords<String, String> records = kafkaConsumer.poll(pollTimeout);
-            eventQueue = StreamSupport.stream(records.spliterator(), false).map(record ->
-                                              new ConsumedEvent(record.value(), record.topic(),
-                                                  Integer.toString(record.partition()),
-                                                  Long.toString(record.offset() + 1))).collect(Collectors.toCollection(
-                                              Lists::newLinkedList));
+            eventQueue = StreamSupport
+                    .stream(records.spliterator(), false)
+                    .map(record -> new ConsumedEvent(record.value(), record.topic(),
+                            Integer.toString(record.partition()), Long.toString(record.offset() + 1)))
```
tmuehl: Can we somehow get rid of all there Integer-Long-String conversions? I makes it hard to read and very error-prone. For me it is already not clear anymore at some points if it should be a number or a string actually.
(I just write it here, but it addresses to the whole class.)

src/main/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumer.java
```diff
@@ -52,25 +54,26 @@ public NakadiKafkaConsumer(final KafkaFactory factory, final String topic, final
...
+                    .collect(Collectors.toCollection(Lists::newLinkedList));
         } catch (InvalidOffsetException e) {
             throw new NakadiException("Wrong offset provided", e);
```
tmuehl: How can both be mapped to the same `NakadiException`? It's actually a huge difference and on should result in a `5xx` and the other into a `4xx`.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -99,6 +94,35 @@ public void createTopic(final String topic, final int partitionsNum, final int r
...
+    public boolean areCursorsValid(final String topic, final List<Cursor> cursors) {
+        final List<TopicPartition> partitions = listPartitions(topic);
+        return cursors
+                .stream()
+                .allMatch(cursor -> partitions
+                        .stream()
+                        .filter(tp -> tp.getPartitionId().equals(cursor.getPartition()))
+                        .findFirst()
+                        .map(pInfo -> {
+                            final long newestOffset = Long.parseLong(pInfo.getNewestAvailableOffset());
```
tmuehl: Why don't you reuse `validateOffset`?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -99,6 +94,35 @@ public void createTopic(final String topic, final int partitionsNum, final int r
...
 
+    public boolean topicExists(final String topic) throws NakadiException {
+        return listTopics()
+                .stream()
+                .map(Topic::getName)
+                .anyMatch(t -> t.equals(topic));
+    }
+
+    public boolean areCursorsValid(final String topic, final List<Cursor> cursors) {
+        final List<TopicPartition> partitions = listPartitions(topic);
```
tmuehl: What if topic doesn't exist?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
tmuehl: Then we should delete the `Topic` class?
Anyway, for me it's still the name of a topic. A topic consists of the name, the partitions, lots of events, a configuration etc. It already confused me multiple times when reading the code.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
tmuehl: Yeah, I saw it. That's why it was just to satisfy my curiosity. :)

v-stepanov:
@tmuehl about unit tests for EventStreamController
There's no unit test, but there's an acceptance-test `EventStreamReadingAT` which covers quite a lot (but of course not all possible combination of parameters).
And at this stage I don't see a good reason to aim to 100% coverage, because the streaming was changed several times already and I had to rewrite tests almost completely.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumer.java
v-stepanov: I don't see a way to get rid of that. Strings should become numbers on kafka level (and vise versa).
Do you think that it will make sense to write some helper class that does the convertion?

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/EventStreamReadingAT.java
```diff
@@ -0,0 +1,328 @@
...
+
+        // validate amount of batches and structure of each batch
+        assertThat(batches, hasSize(PARTITIONS_NUM * keepAliveLimit));
+        batches.forEach(batch -> validateBatchStructure(batch, null));
+    }
+
+    @Test(timeout = 5000)
+    public void whenGetEventsWithUknownTopicThenTopicNotFound() {
+        given()
+                .param("batch_limit", "5")
```
rcillo: Is this required in order to test the desired behaviour?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumer.java
v-stepanov: The thing is that at this point the response status code is already sent.
Here only one reaction is possible - to close the stream.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
v-stepanov: I actually removed `validateOffset` as it wasn't used anymore. But merging with latest nakadi-jvm revived it somehow. I now removed it again

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
v-stepanov: it assumes that at this point you already checked that topic exists, and here you only validate cursors.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
tmuehl: I don't think so. Because in `areCursorsValid` you fetch all available topic and search for each of it in the cursors list provided by the user - not the other way around. If a user specifies `foobar` as partition id, the offset won't even validated at all in 106.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/EventStreamReadingAT.java
v-stepanov: Good point! I removed the params in this call.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumer.java
tmuehl: Hmm, would feel be better to have it in one place and not distributed.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumer.java
tmuehl: One could convert the `Cursor` into `KafkaCursor`. Or convert the values before the stream, I personally find it more readable.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
v-stepanov: there's `.orElse(false));` at the end, that will return false if the filter 
`.filter(tp -> tp.getPartitionId().equals(cursor.getPartition()))`
didn't find anything
I don't really see a way for unexisting partition to pass this check

src/main/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumer.java
v-stepanov: I collapsed exceptions handling as agreed

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -0,0 +1,167 @@
...
+            @RequestParam(value = "stream_limit", required = false, defaultValue = "0") final int streamLimit,
+            @RequestParam(value = "batch_flush_timeout", required = false, defaultValue = "30") final int batchTimeout,
+            @RequestParam(value = "stream_timeout", required = false, defaultValue = "0") final int streamTimeout,
+            @RequestParam(value = "stream_keep_alive_limit", required = false, defaultValue = "0") final int streamKeepAliveLimit,
+            @Nullable @RequestHeader(name = "X-nakadi-cursors", required = false) final String cursorsStr,
+            final NativeWebRequest request, final HttpServletResponse response) throws IOException {
+
+        return outputStream -> {
+            try {
+                // todo: we should get topic from EventType after persistence of EventType is implemented
```
rcillo: Actually with the deprecation of `topic` field, I'm not sure what this TODO is about.

v-stepanov: Oh, I didn't know that it is deprecated. 
How to get the topic for eventType now? Is it the same as eventType name?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
v-stepanov: I'm pretty fine with deleting that.
But I think that already goes out of this ticket scope.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
rcillo: That's an open question. I like the idea of using event type name, since it's immutable and enforces the 1:1 parity between kafka topics and event types :+1:

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
CyberDem0n: Luckily they'd fixed bug with flushing gzipped stream some years ago and we don't need to use FlushableGZIPOutputStream anymore.
Compression could be enabled in the `application.yml`:
```yaml
 server:
   port: 8080
+  compression:
+    enabled: true
```

By default it will compress only following content types: text/html, text/xml, text/plain
Since we want to have custom content type for the streaming endpoint, it should be added to the `application.yml` as well: server.compression.mime-types=custom/mime-type,text/html,text/xml,text/plain

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
CyberDem0n: @tmuehl you are absolutely right, it's C++ and python habits :В

src/main/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumer.java
v-stepanov: Ok, I created a separate class KafkaCursor that handles all conversions.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
v-stepanov: 1:1 is also fine for me. But since that is an open question I will keep the todo's.

rcillo: I think it's up to us to decide it. Since we are not exposing this anymore, it's more of a implementation detail. I think you're free to make the decision you think fits best here.

v-stepanov:
@tmuehl I created unit test for EventStream 
https://github.com/zalando/nakadi/blob/ARUHA-78-read-streams/src/test/java/de/zalando/aruha/nakadi/service/EventStreamTest.java

Unit test for controller is still to come

tmuehl:
@v-stepanov Great! I'm working on the `KafkaRepositoryTest`.

src/test/java/de/zalando/aruha/nakadi/controller/EventStreamControllerTest.java
```diff
@@ -0,0 +1,89 @@
...
+        final EventStreamController controller = new EventStreamController(topicRepositoryMock, objectMapper,
+                eventStreamFactoryMock);
+
+        mockMvc = standaloneSetup(controller)
+                .setMessageConverters(new StringHttpMessageConverter(), messageConverter)
+                .build();
+    }
+
+    @Test
+    public void whenNoParamsThenDefaultsAreUsed() throws Exception {
```
v-stepanov: This is basically all that I could test with mockMvc here.
The problem is that there's no a good way to write unit test for async endpoint. For example, mockMvc always finishes before the body is started to stream and before all checks are done, so it's not possible to check status code or response with it.

v-stepanov:
@tmuehl in general I have finished with unit tests, so you can take a look.
Let me know when you finish with tests for kafka-repo, I will take a look

src/main/java/de/zalando/aruha/nakadi/domain/Cursor.java
```diff
@@ -0,0 +1,52 @@
...
+    public String getPartition() {
+        return partition;
+    }
+
+    public String getOffset() {
+        return offset;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
rcillo: I've added a dependency that might make this easier to implement. Maybe you would like to use it here, just like this https://github.com/zalando/nakadi/blob/nakadi-jvm/src/main/java/de/zalando/aruha/nakadi/domain/EventTypeSchema.java#L35-L37

src/main/java/de/zalando/aruha/nakadi/domain/Topic.java
```diff
@@ -20,4 +20,19 @@ public String toString() {
         return "Topic [name=" + name + "]";
     }
 
+    @Override
+    public boolean equals(final Object o) {
```
rcillo: Same comment as https://github.com/zalando/nakadi/pull/81/files#r52461593

src/main/java/de/zalando/aruha/nakadi/domain/Cursor.java
v-stepanov: Well, I anyway generate `equals` and `hashCode` with IDE. So I would choose to avoid using reflection magic here :)

rcillo: Ok, so maybe I should remove this dependency altogether.

v-stepanov:
@rcillo @tmuehl @CyberDem0n @jkliff 
Ok, I finished unit tests.
https://github.com/zalando/nakadi/blob/ARUHA-78-read-streams/src/test/java/de/zalando/aruha/nakadi/service/EventStreamTest.java
https://github.com/zalando/nakadi/blob/ARUHA-78-read-streams/src/test/java/de/zalando/aruha/nakadi/controller/EventStreamControllerTest.java

Also, @tmuehl , I checked the tests you created for kafka repo. Looks good for me.

So I think that now everything is finished. Would be great to get some approvals :-D

src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepositoryTest.java
```diff
@@ -0,0 +1,227 @@
...
+    public static final List<Cursor> MY_TOPIC_INVALID_CURSORS = asList(cursor("0", "39"), cursor("1", "0"), cursor("1", "99"), cursor("1", "201"));
+
+    public static final long LATEST_TIME = kafka.api.OffsetRequest.LatestTime();
+    public static final long EARLIEST_TIME = kafka.api.OffsetRequest.EarliestTime();
+
+
+    private final KafkaRepository kafkaRepository;
+    private final KafkaProducer kafkaProducer;
+
+    public KafkaRepositoryTest() {
```
rcillo: I'm missing a topic creation test here.

v-stepanov: 1) It would be difficult to create unit test for that, as it calls static methods so we will need to bring PowerMock magic
2) We do have "integration" acceptance test for that :)

rcillo: :+1:

src/test/java/de/zalando/aruha/nakadi/controller/EventStreamControllerTest.java
```diff
@@ -0,0 +1,280 @@
...
+                        new MappingJackson2HttpMessageConverter(objectMapper))
+                .build();
+
+        mockMvc.perform(
+                get(String.format("/event-types/%s/events", TEST_EVENT_TYPE))
+                        .header("X-nakadi-cursors", "[{\"partition\":\"0\",\"offset\":\"0\"}]"))
+                .andExpect(status().isOk());
+
+        // we have to sleep here as mockMvc exits at the very beginning, before the body starts streaming
+        Thread.sleep(2000);
```
rcillo: Maybe we could use our retry strategy in order to assert this as fast as possible. 2000 timeout and 50 between retries. Wdyt?

tmuehl:
@v-stepanov Wow, awesome! :)

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
rcillo: I think we can just remove the comments :+1:

src/test/java/de/zalando/aruha/nakadi/controller/EventStreamControllerTest.java
v-stepanov: Very good point, thanks. 
I added that. Now the test is much faster.

v-stepanov:
@tmuehl @rcillo @CyberDem0n 
I merged from nakadi-jvm to resolve conflicts. Now the PR is ready to be approved again :)

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -0,0 +1,156 @@
...
+                @SuppressWarnings("UnnecessaryLocalVariable")
+                final String topic = eventTypeName;
+
+                // validate parameters
+                if (!topicRepository.topicExists(topic)) {
```
rcillo: I'm afraid that this operation is very costly and it's the first validation. It lists all topics to make sure that it exists. I think we should just check our database and try to stream. Maybe kafka is going to throw an error if we try to read from an unexisting topic? Wdyt?

rcillo: If kafka throws such an error (I didn't check de docs or the client), maybe we could remove this validation.

tmuehl: That's true. But I think it's not that costly, topics are read from Zookeeper. Furthermore, this check is only executed once when streaming starts.

rcillo: Yes, I don't think performance is a huge concern right now, but I was curious if is this really required. I thought that kafka would throw some specific error in case the topic wasn't found. But kafka and it's api is not that familiar for me yet.

v-stepanov: We fetch the list of kafka topics by a direct zookeeper request `zkFactory.get().getChildren().forPath("/brokers/topics")` i don't think that it is very costly.

But in general the topicExists method can really be improved by just checking in zookeeper that topic node exists. like `zkFactory.get().checkExists().forPath("/brokers/topics/my-topic")`

What do you think?

rcillo: Thats also a very good idea @v-stepanov I wasn't aware that it was possible.

v-stepanov: But anyway as the PR is closed and merged already, I suggest to postpone these optimisations till the moment when we really check which parts are time consuming and needed to be optimised. Because now we can only guess.

PR #85, opened 2016-02-02
ARUHA-56: created a draft of nakadi scopes schema;

v-stepanov:
This PR is only for discussion, don't merge please

### DIFF
--- a/nakadi-scopes.txt
+++ b/nakadi-scopes.txt
@@ -1,0 +1,7 @@
+Resource Type ID: nakadi_event
+Resource Type Name: Nakadi Event
+Owning team: aruha
+Application Scope: nakadi_event.read
+Description: Gives a general permission to read events from Nakadi (Zalando event bus).
+Application Scope: nakadi_event.write
+Description: Gives a general permission to write events to Nakadi (Zalando event bus).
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/85" target="_blank">ReviewNinja

nakadi-scopes.txt
```diff
@@ -0,0 +1,7 @@
+Resource Type ID: nakadi_event
+Resource Type Name: Nakadi Event
+Owning team: aruha
+Application Scope: nakadi_event.read
```
rcillo: I think we need a scope to create/update/list event types as of https://techjira.zalando.net/browse/ARUHA-50 right?

v-stepanov: Yeah, true. But that will be another resource. Can be something like `nakadi_event_type` with scopes also `read` and `write`.

jkliff: nakadi.event... not nakadi_event.
see https://github.com/zalando/nakadi/pull/86/files#diff-4492c6b26ee0a37518728ec1585420efR112

v-stepanov:
@jkliff @tmuehl @CyberDem0n @rcillo 
Updated the scopes as we agreed

PR #86, opened 2016-02-03
Add definition of partition offset with symbolic name OLDEST

jkliff:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/86" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -414,8 +414,10 @@ paths:
 
         The event stream is formated as a sequence of `EventStreamBatch`es separated by `\n`. Each
         `EventStreamBatch` constains a chunk of Events and a `Cursor` pointing to the **end** of the
-        chunk (i.e. last delivered Event). Currently this format is the only one supported by the system,
-        but in the future other MIME types will be suported (like `application/event-stream`)
+        chunk (i.e. last delivered Event). The cursor might specify the offset with the simbolic value
```
CyberDem0n: typo:
-simbolic
+symbolic

api/nakadi-event-bus-api.yaml
```diff
@@ -42,18 +42,18 @@ info:
...
 
     A low level **Unmanaged API** is available, providing full control and responsibility of position
     tracking and partitioning (and therefore ordering) to the Clients.
 
-    In the high level **Subscriptoin API** the consumption of Events proceeds via the establishment of a named
-    **Subscription** to an EventType. Subscriptions are persistent relationships from an Application (which
-    might have several instances) and the stream of one or more EventType's, whose consumption tracking is
-    managed by Nakadi, freeing Consumers from any responsibility in tracking of the current position on
-    a Stream.
+    In the high level **Subscription API** the consumption of Events proceeds via the establishment of a
```
tmuehl: Is the official naming now "Subscription API" and "Low Level API"?
I used "managed" and "unmanaged" API the past few days. Have we agreed on something here?

api/nakadi-event-bus-api.yaml
```diff
@@ -414,8 +423,10 @@ paths:
 
         The event stream is formated as a sequence of `EventStreamBatch`es separated by `\n`. Each
         `EventStreamBatch` constains a chunk of Events and a `Cursor` pointing to the **end** of the
-        chunk (i.e. last delivered Event). Currently this format is the only one supported by the system,
-        but in the future other MIME types will be suported (like `application/event-stream`)
+        chunk (i.e. last delivered Event). The cursor might specify the offset with the symbolic value
+        `OLDEST`, which will open the stream starting from the oldest available offset in the partition.
```
tmuehl: So, we stick to the term "OLDEST"? :)
(Just need to instigate a bit.)

tmuehl: Since we also have the term `oldest_available_offset` (see below, line 905) with a slightly different semantic.

api/nakadi-event-bus-api.yaml
```diff
@@ -892,12 +905,18 @@ definitions:
...
-          An offset of the newes available Event in that partition. This value will be changing upon
-          reception of new events for this partition by nakadi
+          An offset of the newest available Event in that partition. This value will be changing upon
+          reception of new events for this partition by Nakadi.
+
+          This value can be used to construct a cursor when opening streams (see
+          `GET /event-type/{name}/events` for details).
+
+          Might assume the special name OLDEST, meaning a pointer to the offset of the oldest available
+          event in the partition.
```
tmuehl: In my understanding it's one event BEFORE the oldest available. Hence, the term "OLDEST" is still misleading although the effect will be what you want. When consuming "OLDEST" makes sense, when exposing something to the client it leads to confusion.

valgog:
Guys, I would vote against `OLDEST` because technically it is not the oldest. Kynesis used different names. Oldest available differs from oldest. Oldest could be theoretically the records in the datalake.

tmuehl:
What about...
BEGIN
ZERO
GENESIS

tmuehl:
@valgog OLDEST in the way we discussed it yesterday really always points to the hypothetical event one before the first event in a partition. And first event really means "first element ever". Even if events expired already. That's why in my eyes OLDEST is misleading (see comments above).

api/nakadi-event-bus-api.yaml
jkliff: Not necessarily. Is more about having SOME name here and not XXX. The discussion on the name is to my understanding still open.

tmuehl: Ah, alright!

api/nakadi-event-bus-api.yaml
```diff
@@ -109,6 +109,15 @@ securityDefinitions:
     type: oauth2
     flow: implicit
     authorizationUrl: 'https://auth.example.com/oauth2/tokeninfo'
+    scopes:
```
v-stepanov: In our scopes do we have only one resourceId `nakadi` and four scopes for it?
Or do we have three resourceIds `nakadi`, `nakadi.event.type` and `nakadi.event.stream` ?

PR #88, opened 2016-02-05
Use Findbugs to detect potential bugs (and fix detected issues)

tmuehl:
@v-stepanov @rcillo @jkliff @CyberDem0n @valgog 
Please review

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -83,6 +83,10 @@ configurations {
     acceptanceTestRuntime.extendsFrom testRuntime
 }
 
+findbugs {
+    reportLevel = "high"
+    sourceSets = []
+}
 
 dependencies {
     // spring
--- a/build.gradle
+++ b/build.gradle
@@ -6,6 +6,7 @@ apply plugin: 'eclipse'
 apply plugin: 'application'
 apply plugin: 'spring-boot'
 apply plugin: "jacoco"
+apply plugin: 'findbugs'
 // if this is applied, include scala dependencies (see below)
 // apply plugin: 'scala'
 
@@ -145,6 +146,13 @@ dependencies {
 // tag::wrapper[]
 task wrapper(type: Wrapper) {
     gradleVersion = '2.3'
+}
+
+tasks.withType(FindBugs) {
+    reports {
+        xml.enabled = false
+        html.enabled = true
+    }
 }
 
 task stopAndRemoveStoragesInDocker {
--- a/src/main/java/de/zalando/aruha/nakadi/service/EventStream.java
+++ b/src/main/java/de/zalando/aruha/nakadi/service/EventStream.java
@@ -1,33 +1,31 @@
 package de.zalando.aruha.nakadi.service;
 
-import static java.lang.System.currentTimeMillis;
-
-import static java.util.function.Function.identity;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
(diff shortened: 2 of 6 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/88" target="_blank">ReviewNinja

build.gradle
```diff
@@ -147,6 +148,13 @@ task wrapper(type: Wrapper) {
     gradleVersion = '2.3'
 }
 
+tasks.withType(FindBugs) {
```
v-stepanov: where the report will be displayed?

v-stepanov:
I took my star back, the build fails :)

build.gradle
tmuehl: It is a single file `reports/findbugs/<env>.html`. I can be added as an individual tab in CI/CD servers such as Go.cd, Teamcity. I don't know if it's possible in Travis or Jenkins. Locally you can just point your browser to the file.

tmuehl:
@v-stepanov Damn, Findbugs rejects itself to be merged. :grinning: 

Fixed.

v-stepanov:
Hm, it actually fails the build if it finds some issues, so it will not be possible to build a project till they are fixed. 
I would prefer not to have that restriction. I would be more than happy to have it as a separate task that lives not in a build pipeline (and to be run once a week)
Also it adds quite some seconds to our build which is already quite slow.

tmuehl:
@v-stepanov I personally would like to habe it as an additional stage, which I don't know if and how to achieve in Travis CI.
Since the report of Findbugs cannot be displayed here in the PullRequest, I'm pretty sure nobody will care if we make it not-failing.

About the additional seconds: True. But unfortunately it's already slow and I think these additional seconds don't make a difference. :(

However, we can vote about having Findbugs in the build process and making the build failing or not.

v-stepanov:
@tmuehl 
+1 for voting

rcillo:
I think we should add it to the default build and fix all the issues if we believe it's relevant.

If we are in doubt about the usefulness of the tool, I think we should experiment for a defined extent of time.

src/main/java/de/zalando/aruha/nakadi/service/EventStream.java
```diff
@@ -101,7 +99,7 @@ public void streamEvents() {
                     currentBatch = emptyBatch();
                     batchStartTime = currentTimeMillis();
 
-                    outputStream.write(BATCH_SEPARATOR.getBytes());
+                    outputStream.write(BATCH_SEPARATOR.getBytes(UTF8));
```
rcillo: What was the issue here? is `"\n".getBytes(UTF8)` different from `"\n".getBytes()`?

rcillo:
:shipit:

src/main/java/de/zalando/aruha/nakadi/service/EventStream.java
tmuehl: `getBytes()` uses the default charset which is determined by the jvm i.e. the underlying os. Hence, if `byte[] bytes = s.getBytes()` is called on a windows machine, the `bytes` are send to a linux machine and `new String(bytes)` is called, it might fail.

rcillo: :+1: didn't knew it. Thanks

PR #90, opened 2016-02-09
Enable gzip compression globally

CyberDem0n:
and get rid from FlushableGZIPOutputStream

Compression could be enabled/disabled for certain content-types. As of
now we are simply streaming with content-type: text/plain and it should
be changed as soon as it will be changed in the API.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/90" target="_blank">ReviewNinja

v-stepanov:
I believe we need a test chat checks that gzip-compression works for streaming. At least a very simple one.

v-stepanov:
:shipit:

PR #91, opened 2016-02-09
Fixed link to API spec in README

whiskeysierra:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/91" target="_blank">ReviewNinja

tmuehl:
You are totally right, the link is broken! :(

The very correct one would be at the moment: https://github.com/zalando/nakadi/blob/nakadi-low-level-api-review/api/nakadi-event-bus-api.yaml

That should be fixed as soon as https://github.com/zalando/nakadi/pull/56 is merged. Which brings us to the point that we should eventually merge it. (@jkliff )

PR #93, opened 2016-02-10
ARUHA-89 Implement Flow-Ids

tmuehl:
@v-stepanov @CyberDem0n @rcillo @jkliff @valgog Please review

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/Application.java
+++ b/src/main/java/de/zalando/aruha/nakadi/Application.java
@@ -7,11 +7,6 @@ import org.springframework.boot.autoconfigure.SpringBootApplication;
 public class Application {
 
     public static void main(final String[] args) {
-        FlowIdUtils.push(FlowIdUtils.generateFlowId());
-        try {
-            SpringApplication.run(Application.class, args);
-        } finally {
-            FlowIdUtils.clear();
-        }
+        SpringApplication.run(Application.class, args);
     }
 }
--- a/src/main/java/de/zalando/aruha/nakadi/Application.java
+++ b/src/main/java/de/zalando/aruha/nakadi/Application.java
@@ -5,7 +5,13 @@ import org.springframework.boot.autoconfigure.SpringBootApplication;
 
 @SpringBootApplication
 public class Application {
+
     public static void main(final String[] args) {
-        SpringApplication.run(Application.class, args);
+        FlowIdUtils.push(FlowIdUtils.generateFlowId());
+        try {
+            SpringApplication.run(Application.class, args);
+        } finally {
+            FlowIdUtils.clear();
+        }
     }
 }
--- a/src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
+++ b/src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
@@ -1,0 +1,47 @@
+package de.zalando.aruha.nakadi;
+
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+public class FlowIdRequestFilter implements Filter {
+    public static final String X_FLOW_ID_HEADER = "x-flow-id";
+
+    @Override
+    public void init(final FilterConfig filterConfig) throws ServletException {
+
+    }
+
+    @Override
+    public void doFilter(final ServletRequest request, final ServletResponse response, final FilterChain chain) throws IOException, ServletException {
+        FlowIdUtils.clear();
+        String flowId = null;
+
+        if (request instanceof HttpServletRequest) {
+            final HttpServletRequest httpServletRequest = (HttpServletRequest) request;
+            flowId = httpServletRequest.getHeader(X_FLOW_ID_HEADER);
+        }
+
+        if (flowId == null) {
+            flowId = FlowIdUtils.generateFlowId();
+        }
+
+        FlowIdUtils.push(flowId);
+
+        try {
+            chain.doFilter(request, response);
+        } finally {
+            FlowIdUtils.clear();
+        }
+    }
+
+    @Override
+    public void destroy() {
+
+    }
+}
--- a/src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
+++ b/src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
@@ -10,7 +10,7 @@ import javax.servlet.http.HttpServletRequest;
 import java.io.IOException;
 
 public class FlowIdRequestFilter implements Filter {
-    public static final String X_FLOW_ID_HEADER = "x-flow-id";
+    public static final String X_FLOW_ID_HEADER = "X-Flow-Id";
 
     @Override
     public void init(final FilterConfig filterConfig) throws ServletException {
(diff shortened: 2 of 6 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/93" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+public class FlowIdRequestFilter implements Filter {
+    public static final String X_FLOW_ID_HEADER = "x-flow-id";
```
v-stepanov: in API it is `X-Flow-Id`

src/main/java/de/zalando/aruha/nakadi/config/WebConfig.java
```diff
@@ -38,4 +38,12 @@ public TimeoutCallableProcessingInterceptor timeoutInterceptor() {
     public EventPublishingController eventPublishingController() {
         return new EventPublishingController(kafkaRepository, new InMemoryEventTypeRepository());
     }
+
+    @Bean
+    public FilterRegistrationBean flowIfRequestFilter() {
```
v-stepanov: `flowIdRequestFilter` ?

tmuehl: Oh yes, sounds better :D

src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
tmuehl: `httpServletRequest.getHeader()` is case-insensitive.
But yes, to avoid confusion I should change it.

src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
```diff
@@ -0,0 +1,47 @@
...
+        FlowIdUtils.clear();
+        String flowId = null;
+
+        if (request instanceof HttpServletRequest) {
+            final HttpServletRequest httpServletRequest = (HttpServletRequest) request;
+            flowId = httpServletRequest.getHeader(X_FLOW_ID_HEADER);
+        }
+
+        if (flowId == null) {
+            flowId = FlowIdUtils.generateFlowId();
```
v-stepanov: I'm a little bit far from the flowId idea. 
What happens here? If flow Id is not present in request - we generate a new one?

v-stepanov:
Do you see any possibility to create a test for that?

src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
tmuehl: Yes

v-stepanov: why?

src/main/java/de/zalando/aruha/nakadi/FlowIdUtils.java
```diff
@@ -0,0 +1,44 @@
...
+        final byte[] bytes = new byte[18];
+        RANDOM.nextBytes(bytes);
+        return Base64.getEncoder().encodeToString(bytes);
+    }
+
+    public static void clear() {
+        while (!"".equals(NDC.pop()));
+    }
+
+    public static String peek() {
```
rcillo: Is this for future use?

tmuehl:
@v-stepanov Regrading test: Was thinking about that, too, but couldn't think of a valid test case. Unit-testing the Filter is more or less useless. It's rather a system test that checks that the FlowId ends up in the logs in two cases: `X-Flow-Id` header was set and header was not set.

I though that would be too much for now and is more a configuration test that we can include in the user-journey test in the end or as an AT.

Any ideas?

src/main/java/de/zalando/aruha/nakadi/FlowIdUtils.java
tmuehl: Yes.
If you create a new thread the FlowId stack from the parent thread is not passed to the child thread automatically. Hence we have to copy the stack to the child somehow. Or one hands over the parent thread's FlowId only.
There is a `copyStack` method missing. I will implement that as soon as Slava's branch is merged since it is a first and important case where we will need it (with the async streaming).

src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
tmuehl: To track one request in the logs.

src/main/java/de/zalando/aruha/nakadi/FlowIdUtils.java
rcillo: :+1: thanks for explaining

src/main/resources/log4j.properties
```diff
@@ -0,0 +1,23 @@
+log4j.rootCategory=INFO, CONSOLE
+
+PID=????
+LOG_LEVEL_PATTERN=%5p
+LOG_PATTERN=[%d{yyyy-MM-dd HH:mm:ss.SSS}] %x boot%X{context} - ${PID} ${LOG_LEVEL_PATTERN} [%t] --- %c{1}: %m%n
+
+# CONSOLE is set to be a ConsoleAppender using a PatternLayout.
+log4j.appender.CONSOLE=org.apache.log4j.ConsoleAppender
+log4j.appender.CONSOLE.layout=org.apache.log4j.PatternLayout
```
rcillo: I don't know if it really matters, but I was reading the docs and it seems that there are suggestions to use the `EnhancedPatternLayout` but I'm not sure why https://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/PatternLayout.html

src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
rcillo: Let me guess, I think that it makes it easier to group interleaved logs as belonging to a single request?

src/main/java/de/zalando/aruha/nakadi/Application.java
```diff
@@ -5,7 +5,13 @@
 
 @SpringBootApplication
 public class Application {
+
     public static void main(final String[] args) {
-        SpringApplication.run(Application.class, args);
+        FlowIdUtils.push(FlowIdUtils.generateFlowId());
```
rcillo: Why do we need to have an "application level" flow id? I thought it was something about the request, only. Does this comment make sense?

src/main/java/de/zalando/aruha/nakadi/FlowIdRequestFilter.java
tmuehl: @rcillo Exactly :)

src/main/resources/log4j.properties
tmuehl: To be honest, I just copied SpringBoot's default and only changed the `LOG_PATTERN` line to include the Flow-Id. :)
I'll look if it makes sense to change the PatternLayout for now. Thanks for the hint!

src/main/java/de/zalando/aruha/nakadi/Application.java
tmuehl: I don't know. Maybe not since it's already another thread that is no worker thread. I can remove it, don't care.

PR #94, opened 2016-02-11
ARUHA-49 Event Type registration

rcillo:
Creates a Kafka Topic together with Event Type persistence, since there
is a 1:1 relation between then.

If the topic creation fails, rollback Event Type persistence.

This implementation assumes that the Application is not going to
terminate unexpectedly.

But if it does suddenly terminates it might let the system in an
inconsistent state, where we have a database entry for the Event Type but
no Kafka Topic.

It would require a different approach in order to achieve guaranteed consistency.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/94" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepository.java
```diff
@@ -77,20 +74,25 @@ public void createTopic(final String topic) {
...
     public void createTopic(final String topic, final int partitionsNum, final int replicaFactor,
-                            final long retentionMs, final long rotationMs) {
-
-        final String connectionString = zkFactory.get().getZookeeperClient().getCurrentConnectionString();
-        final ZkUtils zkUtils = ZkUtils.apply(connectionString, settings.getZkSessionTimeoutMs(),
-                settings.getZkConnectionTimeoutMs(), false);
+                            final long retentionMs, final long rotationMs) throws TopicCreationException {
+        ZkUtils zkUtils = null;
         try {
+            final String connectionString = zkFactory.get().getZookeeperClient().getCurrentConnectionString();
```
rcillo: I just wrapped everything in a try/catch in order to be able to treat it properly in the controller. I thought about handling errors in a finer grained style, but I do need to know wether anything wrong happened with this operation in order to undo the persistency.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -4,9 +4,7 @@
 import de.zalando.aruha.nakadi.domain.EventType;
 import de.zalando.aruha.nakadi.problem.DuplicatedEventTypeNameProblem;
 import de.zalando.aruha.nakadi.problem.ValidationProblem;
-import de.zalando.aruha.nakadi.repository.DuplicatedEventTypeNameException;
-import de.zalando.aruha.nakadi.repository.EventTypeRepository;
-import de.zalando.aruha.nakadi.repository.NoSuchEventTypeException;
+import de.zalando.aruha.nakadi.repository.*;
```
v-stepanov: We agreed not to use wildcard imports

tmuehl:
@rcillo I think I would still do it the other way around. First create topic and then persist the event type to the db. Hence, it's only visible to users if a topic exists.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -32,16 +29,18 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(EventTypeController.class);
 
-    final private EventTypeRepository repository;
+    final private EventTypeRepository eventTypeRepository;
```
v-stepanov: I believe the order "private final" is an unofficial standard

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeDbRepository.java
```diff
@@ -84,4 +84,9 @@ public EventType mapRow(ResultSet rs, int rowNum) throws SQLException {
                 "SELECT et_event_type_object FROM zn_data.event_type",
                 new EventTypeMapper());
     }
+
+    @Override
+    public void removeEventType(String name) {
```
v-stepanov: final?

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeDbRepository.java
```diff
@@ -84,4 +84,9 @@ public EventType mapRow(ResultSet rs, int rowNum) throws SQLException {
...
+        jdbcTemplate.update("DELETE FROM zn_data.event_type WHERE et_name = ?", name);
```
tmuehl: I am no database guy, but I always feel more comfortable having a `LIMIT 1` in my delete.
@CyberDem0n Is that dumb nonsense?

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeDbRepositoryTest.java
```diff
@@ -146,6 +146,20 @@ public void whenListExistingEventTypesAreListed() throws NakadiException, Duplic
         assertThat(eventTypes, hasSize(2));
     }
 
+    @Test
+    public void whenRemoveThenDeleteFromDatabase() throws Exception {
+        EventType eventType = buildEventType();
+        String insertSQL = "INSERT INTO zn_data.event_type (et_name, et_event_type_object) VALUES (?, to_json(?::json))";
```
CyberDem0n: Why not use `saveEventType(eventType)` here? Just to avoid touching of this code if the table structure will change in a future.

rcillo:
@tmuehl my problem with the other way around is that Kafka is more complex to interact then Postgres.

For example, deletion of topics in kafka is asynchronous. So it would be possible to fail creating an Event Type and for some period of time have a Kafka marked for deletion. That would make it impossible to retry creating the Event Type, since it would not be allowed to have duplicated kafka topics.


I'm not 100% sure about that. I spent sometime learning about kafka yesterday but it's something new for me yet.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeDbRepository.java
rcillo: if it was possible to sql inject here, yes. Like ? been replaced by "anything OR 1=1" then we would erase the database :sweat_smile: but I think that jdbc does allow it.

CyberDem0n: RTFM: http://www.postgresql.org/docs/current/static/sql-delete.html
It's not possible to have LIMIT in DELETE statement.

tmuehl: Ah, damn.

tmuehl:
@rcillo Particularly, if Kafka topic creation is more likely to make trouble I would do it in a first step. If creating an event type fails, it's hopefully a real issue not not just something that is solved by "try again". Therefore, if it fails one anyhow has to troubleshoot. And if not, I would prefer having an unused topic in Kafka for some minutes that stops from trying again instead of exposing an event-type that cannot be used.

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeDbRepositoryTest.java
v-stepanov: It is considered to be a bad practice to use the system/class you test for setup of a test.

In two words, if you have a system A that can do get(X) and set(X) - you should not use set(X) for setup of test for get(X) (and vise versa). Because if a test of system A you can't yet trust system A. And also if you have bug in both set(X) and get(X) you test can be false positive.

rcillo:
@tmuehl actually, topic creation in kafka is supposed to work just fine, since uniqueness has already been checked by postgres. It uses only default parameters and the event name.

If it fails it would be probably due to something solvable with a retry. Maybe kafka disk is full os something.

But I do agree listing the event type that is not yet read to receive messages could be a problem. I'm gonna take a look at how to delete a topic from kafka and then change it.

PR #99, opened 2016-02-11
ARUHA-50 remove unnecessary dependencies

rcillo:
It's better to implement it explicitly than implicitly. Adding a huge
dependency just for this fucntionality doesnt seem resoanable.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/99" target="_blank">ReviewNinja

PR #102, opened 2016-02-12
ARUHA-50 Using snake case instead of camel case

rcillo:
Due to a spring boot problem, the wrong mapper bean is wired during application boot. That causes an issue where camel cases are not properly mapped to snake case.

So I'm explicitly mapping properties as snake case.

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/domain/EventType.java
+++ b/src/main/java/de/zalando/aruha/nakadi/domain/EventType.java
@@ -15,6 +15,7 @@ public class EventType {
     @Size(min = 1, message = "may not be empty")
     private String name;
 
+    @JsonProperty("owning_application")
     private String owningApplication;
 
     @NotNull


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/102" target="_blank">ReviewNinja

rcillo:
I had seem the docs http://docs.spring.io/spring-boot/docs/current/reference/html/howto-spring-mvc.html#howto-customize-the-jackson-objectmapper but it just doesn't work.

v-stepanov:
@rcillo 
In general we do have a global config for that here https://github.com/zalando/nakadi/blob/nakadi-jvm/src/main/java/de/zalando/aruha/nakadi/config/NakadiConfig.java#L72
Would be great if you can find a reason why it stopped working :)
Because owningApplication property is not the only place where it is broken. For class https://github.com/zalando/nakadi/blob/nakadi-jvm/src/main/java/de/zalando/aruha/nakadi/domain/TopicPartition.java
it also works wrong now (with camel-case)

rcillo:
@v-stepanov sure. I'm going to further investigate this issue :+1:

rcillo:
We are going to provide a general solution for this problem.

PR #106, opened 2016-02-15
ARUHA-95 Kafka offsets fetching reworked

v-stepanov:
Fetching of partitions offsets didn't work for the case when topic replica factor was lower than amount of kafka-brokers. I replaced the offsets fetching implementation with the one using only kafka consumer official possibilities. Checked that on AWS as well.
@tmuehl @jkliff @rcillo @CyberDem0n please review.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/106" target="_blank">ReviewNinja

PR #109, opened 2016-02-16
aruha-50 standard status code on error

rcillo:
When saving a new EventType and a JsonException is thrown, we return an
InternalServerError. But when updating we were returning a 422.

In order to make it standard and homogeneous we are changing it to
return a 500 as well.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/109" target="_blank">ReviewNinja

v-stepanov:
:shipit:

rcillo:
Solved by https://github.com/zalando/nakadi/pull/110

PR #110, opened 2016-02-16
Aruha 88 consistent exception handling

tmuehl:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/110" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/exceptions/NakadiException.java
```diff
@@ -26,4 +31,11 @@ public void setProblemMessage(final String problemMessage) {
         this.problemMessage = problemMessage;
     }
 
+    public Problem asProblem() {
+        final String detail = StringUtils.isNotBlank(problemMessage) ? problemMessage : getMessage();
+        return Problem.valueOf(getStatus(), detail);
+    }
+
+    protected abstract Response.StatusType getStatus();
```
rcillo: I think from a client code perspective this approach worked just perfectly. But I was just wondering if isn't it conceptually broken. The exceptions are "domain specific", e.g, they are not aware that this code is running in the context of an http request.

I don't think it's going to harm us in the future, but I just found it strange that a Domain exception knows about the Http stack.

rcillo: If we agree that all NakadiExceptions are mean to be used in a "controller", then I'm fine with this :+1:

rcillo: Another approach would be to have a "problem factory" that receives a "domain exception" and returns a "problem".

src/test/java/de/zalando/aruha/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -82,13 +82,13 @@ public void whenPostWithInvalidEventTypeThenReturn422() throws Exception {
 
     @Test
     public void whenPostDuplicatedEventTypeReturn409() throws Exception {
-        final DuplicatedEventTypeNameException e = mock(DuplicatedEventTypeNameException.class);
-        final Problem expectedProblem = new DuplicatedEventTypeNameProblem("some-name");
+//        final DuplicatedEventTypeNameException e = mock(DuplicatedEventTypeNameException.class);
```
rcillo: :scissors:

src/test/java/de/zalando/aruha/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -82,13 +82,13 @@ public void whenPostWithInvalidEventTypeThenReturn422() throws Exception {
...
+        final Problem expectedProblem = Problem.valueOf(Response.Status.CONFLICT, "some-name");
 
-        when(e.getName()).thenReturn("some-name");
+//        when(e.getName()).thenReturn("some-name");
```
rcillo: :scissors:

build.gradle
```diff
@@ -135,6 +135,7 @@ dependencies {
     compile 'io.dropwizard.metrics:metrics-core:3.1.0'
     compile 'com.ryantenney.metrics:metrics-spring:3.1.0'
     compile 'io.dropwizard.metrics:metrics-servlets:3.1.0'
+    compile 'org.apache.commons:commons-lang3:3.4'
```
rcillo: I've already added and then removed this dependency because I was using it only for a single method. Now I think we are using it just to compare blank strings. Maybe we could implement this comparison on a Utils class in order not to add another dependency. Wdyt? Maybe from your experience with Java projects you know that we end up using more and more stuff from apache commons.

tmuehl: I think, having apache commons in your classpath is always a good idea. :)

src/main/java/de/zalando/aruha/nakadi/exceptions/NakadiException.java
tmuehl: Yes, that is definitely a subject to discussion. Moreover, it was one main purpose of the ticket: https://techjira.zalando.net/browse/ARUHA-88.

The business Code shouldn't be aware of the HTTP stack, I agree with you. But it doesn't have to if the Exception itself provides a status code. We wanted to see how far we can push this idea.

rcillo: I think that it's going to work just fine. The only scenario that it would not work is that if we have different problems for the same exception, depending on the context. Then we would be forced to create a new Exception, just to have a different status code.

tmuehl: Yes. And you are always free to catch an Exception anywhere (e.g. in the Controller) and re-throw another one or translate it differently.

tmuehl:

src/main/java/de/zalando/aruha/nakadi/controller/ExceptionHandling.java
```diff
@@ -1,15 +1,40 @@
...
     public String formatFieldName(String fieldName) {
         return CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, fieldName);
     }
 
+    @Override
+    @ExceptionHandler
+    public ResponseEntity<Problem> handleThrowable(final Throwable throwable, final NativeWebRequest request) {
+        final String errorTraceId = generateErrorTraceId();
+        LOG.error("InternalServerError (" + errorTraceId + "):", throwable);
+        return Responses.create(Response.Status.INTERNAL_SERVER_ERROR, "An internal error happened. Please report it. (" + errorTraceId + ")", request);
```
rcillo: I was expecting that the first parameter would be an instance of `Problem`, right?

rcillo: Oh, I got it. I haven't used this functionality from the Problem spring library yet.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -49,7 +49,7 @@ public KafkaRepository(final ZooKeeperHolder zkFactory, final KafkaFactory kafka
     }
 
     @Override
-    public List<Topic> listTopics() throws NakadiException {
+    public List<Topic> listTopics() throws ServiceUnavailableException {
```
rcillo: I expected this to be a 500.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -58,7 +58,7 @@ public KafkaRepository(final ZooKeeperHolder zkFactory, final KafkaFactory kafka
                     .map(Topic::new)
                     .collect(Collectors.toList());
         } catch (Exception e) {
```
rcillo: Exception is too generic for unavailable. If there is a network problem I would agree with this. But if it's just a generic error, then it seems that 500 fits best.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumer.java
```diff
@@ -64,7 +60,7 @@ public NakadiKafkaConsumer(final KafkaFactory factory, final String topic, final
     }
 
     @Override
-    public Optional<ConsumedEvent> readEvent() throws NakadiException {
+    public Optional<ConsumedEvent> readEvent() {
```
v-stepanov: please remove `throws` from interface as well

src/main/java/de/zalando/aruha/nakadi/service/EventStream.java
```diff
@@ -118,7 +119,7 @@ public void streamEvents() {
             LOG.info("I/O error occurred when streaming events (possibly client closed connection)", e);
         } catch (IllegalStateException e) {
             LOG.info("Error occurred when streaming events (possibly server closed connection)", e);
-        } catch (NakadiException e) {
+        } catch (KafkaException | NakadiException e) {
```
v-stepanov: I believe there are no real conditions when NakadiException will be thrown here

src/test/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumerTest.java
```diff
@@ -162,7 +153,7 @@ public void whenReadEventsThenNakadiException() {
 
                 // ASSERT //
                 fail("The NakadiException should be thrown");
```
rcillo: /NakadiException/Exception/

v-stepanov:
@tmuehl  Great job, I think with what you implemented we don't need ARUHA-92 any more, right?

rcillo:
I'm not a 100% sure about the unavailable status, but I agree that it should be a 5xx.

So, :shipit:

src/test/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumerTest.java
v-stepanov: please adjust the fail message as well.

tmuehl:
@v-stepanov Thanks! Yes, I think ARUHA-92 gets obsolete by merging this branch.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/NakadiKafkaConsumer.java
tmuehl: done

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
tmuehl: I agree. But unfortunately the Curator Framework isn't more specific as well and doesn't expose something like a ZooKeeperException, which would be nice. :(
If you look into the `forPath()`, it only declares a generic exception in the signature - unfortunately.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
tmuehl: We had an offline discussion about this and came to the conclusion that we want to use a 503 if it doesn't work because an underlying resource is not available. It emphasises that it's probably a temporary problem and it might be worth to retry after a certain time frame.

src/main/java/de/zalando/aruha/nakadi/controller/ExceptionHandling.java
tmuehl: :)

v-stepanov:
:ok_hand:

PR #114, opened 2016-02-17
ARUHA-88 All endpoints can (with low probability) respond with a 503.…

tmuehl:
@v-stepanov @CyberDem0n @jkliff @rcillo Please review and discuss

All endpoints can (with low probability) respond with a 503 "Service Unavailable". This happens if an underlying service (e.g. Kafka, ZooKeeper, Postgres) is not working properly. A 503 is meant to emphasise that it is probably a temporary problem and not a misbehaviour in the application itself.

However it's still nothing that can be fixed on the client side.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -139,7 +139,7 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
         '503':
-          description: Service (temporary) not available
+          description: Service (temporarily) unavailable
           schema:
             $ref: '#/definitions/Problem'
   /event-types:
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/114" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -170,6 +170,10 @@ paths:
           description: Server error
           schema:
             $ref: '#/definitions/Problem'
+        '503':
+          description: Service (temporary) not available
```
rcillo: I have an English question. For an English speaker, would it sound better "Service (temporarily) unavailable"?

tmuehl: Uh, good question. Maybe, somebody with better English skills could tell. :confused:

tmuehl: Seems you are right, "temporarily" is the adverb of course, "temporary" the adjective.

rcillo:
:shipit:

PR #118, opened 2016-02-18
Limit Traffic into the LB to HTTPS

tmuehl:
There is no reason why we should expose more than 443.

### DIFF
--- a/senza.yaml
+++ b/senza.yaml
@@ -163,5 +163,5 @@ Resources:
       # https
       - IpProtocol: tcp
         FromPort: 443
-        ToPort: 8080
+        ToPort: 443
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/118" target="_blank">ReviewNinja

v-stepanov:
Just a comment not related to your changes.
I think we should add log-rotation from the very beginning, because logs filling the whole free space was for us the most common reason of nodes dying :)

> application_logrotate_interval: hourly
        application_logrotate_size: 50M
        application_logrotate_rotate: 4

rcillo:
:shipit: 

:+1: for the log rotate improvement as well.

PR #121, opened 2016-02-18
Aruha 61 validate persistency of schemas

rcillo:
In order to validate submitted events we need to make sure that each
`EventType` has a valid json-schema definition.

We are not to much strict on the schema data, e.g. if someone submits a
Swager definition, it should work just fine.

Examples of rejected schemas can be found in the added tests.

I also made a minor rename refactor in order to make it easier to compose error messages related to fields and to make our domain more similar to the API definition.

### DIFF
--- a/src/test/java/de/zalando/aruha/nakadi/controller/EventTypeControllerTest.java
+++ b/src/test/java/de/zalando/aruha/nakadi/controller/EventTypeControllerTest.java
@@ -278,6 +278,32 @@ public class EventTypeControllerTest {
 
     }
 
+    @Test
+    public void whenEventTypeSchemaJsonIsMalformedThen422() throws Exception {
+        EventType eventType = buildEventType();
+        eventType.getSchema().setSchema("invalid-json");
+
+        Problem expectedProblem = invalidProblem("schema.schema", "must be a valid json");
+
+        postEventType(eventType)
+                .andExpect(status().isUnprocessableEntity())
+                .andExpect((content().string(matchesProblem(expectedProblem))));
+    }
+
+    @Test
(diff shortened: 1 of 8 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/121" target="_blank">ReviewNinja

rcillo:
@v-stepanov @tmuehl @CyberDem0n I lost my :star: :sweat_smile: had to make another commit to please codacy bot.

rcillo:
I also would like to share with you that I did an alternative implementation using json-schema meta schema http://json-schema.org/schema

I experimented with some errors and both implementations behaved similarly. The main advantage of using a meta-schema to validate json-schemas is that it is more precise regarding the "path" that is invalid.

But I don't think we should be responsible to help the user find the error. I think he should use an external validation tool to find the error.

Wdyt ?

rcillo:
I'll merge this PR. If you have any considerations regarding the alternative implementation, I'll open a new one.

src/test/java/de/zalando/aruha/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -278,6 +280,32 @@ public void askingForANonExistingEventTypeResultsIn404() throws Exception {
 
     }
 
+    @Test
+    public void whenEventTypeSchemaJsonIsMalformedThen422() throws Exception {
```
v-stepanov: so with this test we validate the validation of validation schema :trollface:

rcillo: Yes hehe

recursive

PR #122, opened 2016-02-19
aruha-64 return validation error message

rcillo:
Returning a boolean is not sufficient to communicate the reason of the
validation failure, which is a system requirement.

This is the first PR in the aruha-64 feature.

I'll do it in the smallest usable chunks that I can in order to get early feedback.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/122" target="_blank">ReviewNinja

rcillo:
I had added a new commit improving tests https://github.com/zalando/nakadi/commit/29c172fb10c9e7051235037b912985e4ec759c62

Please, I need another :star: :smile:

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -52,11 +53,13 @@ public EventTypeValidator(final EventType eventType) {
...
-        // FIXME: return information on the validation place that failed
-        final boolean allItemsPass = validators.stream().anyMatch(ev -> !ev.accepts(event));
-        Preconditions.checkState(!allItemsPass, "Some validation failed");
+    public Optional<ValidationError> validate(final JSONObject event) {
+        return validators
+                .stream()
+                .map(validator -> validator.accepts(event))
+                .filter(errorOptional -> errorOptional.isPresent())
+                .findFirst()
+                .map(errorOptionalOptional -> errorOptionalOptional.get());
```
v-stepanov: can be simplified a little bit:
```
return validators
                .stream()
                .map(validator -> validator.accepts(event))
                .filter(Optional::isPresent)
                .findFirst()
                .orElse(Optional.empty());
```

src/main/java/de/zalando/aruha/nakadi/validation/FieldNameMustBeSet.java
```diff
@@ -5,16 +5,21 @@
...
 public class FieldNameMustBeSet extends ValidationStrategy {
 
     public static final String NAME = "field-name-must-be-set";
     private final EventValidator executor = new EventValidator() {
 
         @Override
-        public boolean accepts(final JSONObject event) {
-            return event.get("name") != null;
+        public Optional<ValidationError> accepts(final JSONObject event) {
+            if (event.has("name") && event.get("name") != null) {
```
v-stepanov: is it a possible case that `event.has("name") == true and event.get("name") == null` ?

rcillo: I think so. The first check `event.has("name")` checks if there is such a key in the object. But the value could still be `null`.

v-stepanov: ok, good.

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
rcillo: Nice :+1: done here https://github.com/zalando/nakadi/commit/5156215f37e5acb75fdb566d074f943f57099e46

PR #124, opened 2016-02-19
aruha-63 validate event schema on publish

rcillo:
The purpose of this PR is to provide the simplest implementation
possible to the validation of event schema. It's not focused on
performance, which should be addressed in a future ticket.

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
@@ -1,10 +1,18 @@
 package de.zalando.aruha.nakadi.controller;
 
 import com.codahale.metrics.annotation.Timed;
+import de.zalando.aruha.nakadi.domain.EventType;
+import de.zalando.aruha.nakadi.domain.ValidationStrategyConfiguration;
 import de.zalando.aruha.nakadi.exceptions.NakadiException;
 import de.zalando.aruha.nakadi.exceptions.NoSuchEventTypeException;
 import de.zalando.aruha.nakadi.repository.EventTypeRepository;
 import de.zalando.aruha.nakadi.repository.TopicRepository;
+import de.zalando.aruha.nakadi.validation.EventBodyMustRespectSchema;
+import de.zalando.aruha.nakadi.validation.EventValidator;
+import de.zalando.aruha.nakadi.validation.ValidationError;
+import de.zalando.aruha.nakadi.validation.ValidationStrategy;
+import org.json.JSONException;
+import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
@@ -14,6 +22,10 @@ import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.context.request.NativeWebRequest;
+import org.zalando.problem.MoreStatus;
+import org.zalando.problem.Problem;
+
+import java.util.Optional;
 
 import static org.springframework.http.ResponseEntity.status;
 import static org.springframework.web.bind.annotation.RequestMethod.POST;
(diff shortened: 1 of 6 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/124" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -55,4 +74,15 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName, @Reque
         }
     }
 
+    private Optional<ValidationError> validateSchema(@RequestBody String event, EventType eventType) {
```
v-stepanov: #classical comment goes here :)

v-stepanov: Also, why you have @RequestBody annotation here?

rcillo: Refactoring tool did it automatically. Fixed here https://github.com/zalando/nakadi/commit/969a3635cfcf53b6c40bbde008986b5135cd8b46

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -55,4 +74,15 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName, @Reque
...
+        try {
+            ValidationStrategy validationStrategy = new EventBodyMustRespectSchema();
+            ValidationStrategyConfiguration vsc = new ValidationStrategyConfiguration();
```
v-stepanov: for me it seems strange that we create all that every time. Is it possible to create that only one time, and every time call only validationStrategy.materialize(eventType, vsc) ?

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -55,4 +74,15 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName, @Reque
         }
     }
 
+    private Optional<ValidationError> validateSchema(final String event, final EventType eventType) {
+        try {
+            final ValidationStrategy validationStrategy = new EventBodyMustRespectSchema();
+            final ValidationStrategyConfiguration vsc = new ValidationStrategyConfiguration();
```
rcillo: This class plays no role in the validation, but I had to create it anyway since it's required.

I tried not to mess up with the existing validations code since it covers lots of future needs.

v-stepanov:
:shipit:

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
rcillo: good catch! done https://github.com/zalando/nakadi/commit/bad38659b1b301ab561796982eba15ff38c58afd

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -26,6 +38,8 @@
 
     private final TopicRepository topicRepository;
     private final EventTypeRepository eventTypeRepository;
+    private final ValidationStrategy validationStrategy = new EventBodyMustRespectSchema();
+    private final ValidationStrategyConfiguration vsc = new ValidationStrategyConfiguration();
```
v-stepanov: Is it a temporary solution for now? Because I believe we will need some validation strategies registry or something like that when we have more than 1 strategy :)

v-stepanov:
:star:

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
rcillo: Yes, definitely. I think it's also related to the caching problem that we discussed briefly on Friday. The optimal solution would reduce the number of objects to be instantiated in a request and also avoid hitting the database.

But by now, since we have a single validation strategy and we do hit the database, building a registry was left out of scope.

v-stepanov: ok, great.

PR #125, opened 2016-02-22
Aruha 56 oauth scopes

v-stepanov:
@rcillo @tmuehl @jkliff @CyberDem0n Please review

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/config/JsonConfig.java
+++ b/src/main/java/de/zalando/aruha/nakadi/config/JsonConfig.java
@@ -1,0 +1,35 @@
+package de.zalando.aruha.nakadi.config;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.module.SimpleModule;
+import com.fasterxml.jackson.datatype.jdk8.Jdk8Module;
+import com.fasterxml.jackson.datatype.jsonorg.JSONObjectDeserializer;
+import com.fasterxml.jackson.datatype.jsonorg.JSONObjectSerializer;
+import org.json.JSONObject;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
+import org.zalando.problem.ProblemModule;
+
+@Configuration
+public class JsonConfig {
+
+    @Bean
+    @Primary
+    public ObjectMapper jacksonObjectMapper() {
+        ObjectMapper objectMapper = new ObjectMapper().setPropertyNamingStrategy(
+                PropertyNamingStrategy.CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES);
+
+        SimpleModule jsonObjectModule = new SimpleModule();
+        jsonObjectModule.addSerializer(JSONObject.class, new JSONObjectSerializer());
+        jsonObjectModule.addDeserializer(JSONObject.class, new JSONObjectDeserializer());
+
+        objectMapper.registerModule(jsonObjectModule);
+        objectMapper.registerModule(new Jdk8Module());
+        objectMapper.registerModule(new ProblemModule());
+
+        return objectMapper;
+    }
+
+}
--- a/src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
+++ b/src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
@@ -1,0 +1,66 @@
+package de.zalando.aruha.nakadi.config;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import de.zalando.aruha.nakadi.repository.EventTypeRepository;
+import de.zalando.aruha.nakadi.repository.db.EventTypeDbRepository;
+import de.zalando.aruha.nakadi.repository.kafka.KafkaFactory;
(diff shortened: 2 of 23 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/125" target="_blank">ReviewNinja

v-stepanov:
I'm still not sure if the tests should go to integration or unit tests. They create the whole context of application but with repositories mocked.

src/main/java/de/zalando/aruha/nakadi/config/SecurityConfiguration.java
```diff
@@ -1,46 +1,75 @@
...
-        if (oauthEnabled) {
-            http.authorizeRequests().antMatchers("/health").permitAll().anyRequest().access(hasScope(UID));
-        } else {
-            http.authorizeRequests().anyRequest().permitAll();
+        LOG.info("Authentication mode: " + settings.getAuthMode());
+
+        if (settings.getAuthMode() == SecuritySettings.AuthMode.FULL) {
+            http.authorizeRequests()
+                    .antMatchers("/health").permitAll()
+                    .antMatchers(GET, "/metrics").access(hasScope(uidScope))
```
v-stepanov: I don't know if we should have the specific scope for accessing /metrics endpoint. I assumed that we don't.

src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
```diff
@@ -0,0 +1,66 @@
...
+public class RepositoriesConfig {
+
+    @Autowired
+    private Environment environment;
+
+    @Autowired
+    private JdbcTemplate jdbcTemplate;
+
+    @Autowired
+    private ObjectMapper objectMapper;
```
tmuehl: Just a feeling, I would be very careful that you end up using the correct ObjectMapper. You could avoid this by making it explicit and autowire the Configuration instead and call the factory method. Would be the more "java-config-ish".

tmuehl: Moreover, you have a circular dependency between the configs: `RepositoriesConfig` depends on `NakadiConfig` and vice versa.

v-stepanov: I didn't know that one can autowire Configuration. Spring is full of magic :)
Will do that, and will remove circular dependency.

v-stepanov: @tmuehl fixed

src/main/java/de/zalando/aruha/nakadi/config/SecurityConfiguration.java
tmuehl: I don't think, we have one, too

src/main/java/de/zalando/aruha/nakadi/config/SecurityConfiguration.java
```diff
@@ -1,46 +1,75 @@
...
-    private String clientId;
+    @Value("${nakadi.oauth2.scopes.uid}")
+    private String uidScope;
 
-    @Value("${nakadi.oauth2.enabled:true}")
-    private boolean oauthEnabled;
+    @Value("${nakadi.oauth2.scopes.nakadiRead}")
+    private String nakadiReadScope;
+
+    @Value("${nakadi.oauth2.scopes.eventTypeWrite}")
```
tmuehl: Why do you make the scopes configurable? Why no enum?

v-stepanov: @tmuehl in a case if someone will want to use nakadi as an appliance.

tmuehl: With other scope names? :) I think it's a bit overengineered.

v-stepanov: @tmuehl but they should have other scope names because current ones will be owned by our team, so the other team will not be able to manage access to their nakadi instance.
So for me it seems to be quite a realistic scenario that another team will want to have an instance and will want to have their own scopes.

src/main/java/de/zalando/aruha/nakadi/config/JsonConfig.java
```diff
@@ -0,0 +1,35 @@
...
+@Configuration
+public class JsonConfig {
+
+    @Bean
+    @Primary
+    public ObjectMapper jacksonObjectMapper() {
+        ObjectMapper objectMapper = new ObjectMapper().setPropertyNamingStrategy(
+                PropertyNamingStrategy.CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES);
+
+        SimpleModule jsonObjectModule = new SimpleModule();
```
rcillo: I'm not sure anymore, but I think I added this module when schema used to be a `JSONObject`. Maybe we don't need this anymore. I see that this is not in the scope of this task, but I just realised it now.

v-stepanov: I created a `technical-debt` ticket for that ARUHA-102

rcillo: :+1:

src/main/java/de/zalando/aruha/nakadi/config/SecurityConfiguration.java
tmuehl: Ah, true. I wasn't thinking of different teams within the same company.

PR #126, opened 2016-02-22
Aruha 71 support partitioning

tmuehl:
Implements ARUHA-73 and ARUHA-74


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/126" target="_blank">ReviewNinja

rcillo:
I was thinking about the implications of defining an ordering key. It means that all events must have those keys otherwise it's going to throw an error. Wouldn't it be useful if we just allowed ordering keys in required fields declared in the schema?

src/main/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+                    .map(okf -> traversableJsonEvent.get(okf).toString().hashCode())
+                    .mapToInt(hc -> hc)
+                    .sum();
+
+            return String.valueOf(abs(hashValue) % numberOfPartitions);
+
+        } catch (RuntimeException e) {
+            final Throwable cause = e.getCause();
+            if (cause != null && cause instanceof InvalidOrderingKeyFieldsException) {
+                throw (InvalidOrderingKeyFieldsException) cause;
```
v-stepanov: O_o
I don't really get what magic happens here.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -50,39 +53,55 @@ public EventPublishingController(final TopicRepository topicRepository,
...
+        if (!eventType.getOrderingKeyFields().isEmpty()) {
+            partitionId = orderingKeyFieldsPartitioningStrategy.calculatePartition(eventType, eventAsJson, 8);
+        } else {
+            // Will be replaced later:
+            partitionId = "1";
+        }
+        return partitionId;
+    }
+
+    private JSONObject validateSchema(final String event, final EventType eventType) throws EventValidationException {
```
v-stepanov: for me validateSchema method that returns JSONObject seems so be a little bit weird :)
In fact it does two things: 1) converts to JSONObject 2) Validates event against the schema. So I think splitting into two methods will make sense here (IMHO)

src/main/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategy.java
v-stepanov: oh, nevermind. Now I got that.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -50,39 +53,55 @@ public EventPublishingController(final TopicRepository topicRepository,
...
     }
 
-    private Optional<ValidationError> validateSchema(final String event, final EventType eventType) {
+    private String applyPartitioningStrategy(final EventType eventType, final JSONObject eventAsJson) throws InvalidOrderingKeyFieldsException {
+        String partitionId;
+        if (!eventType.getOrderingKeyFields().isEmpty()) {
+            partitionId = orderingKeyFieldsPartitioningStrategy.calculatePartition(eventType, eventAsJson, 8);
+        } else {
+            // Will be replaced later:
+            partitionId = "1";
```
v-stepanov: for me it looks like a fallback strategy, I'm not really sure if we planned to have such functionality.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -50,39 +53,55 @@ public EventPublishingController(final TopicRepository topicRepository,
...
             LOG.error("error posting to partition", e);
             return create(e.asProblem(), nativeWebRequest);
         }
     }
 
-    private Optional<ValidationError> validateSchema(final String event, final EventType eventType) {
+    private String applyPartitioningStrategy(final EventType eventType, final JSONObject eventAsJson) throws InvalidOrderingKeyFieldsException {
+        String partitionId;
+        if (!eventType.getOrderingKeyFields().isEmpty()) {
+            partitionId = orderingKeyFieldsPartitioningStrategy.calculatePartition(eventType, eventAsJson, 8);
```
v-stepanov: In general the number of partitions can be easily get from KafkaProducer with partitionsFor method (https://kafka.apache.org/090/javadoc/index.html?org/apache/kafka/clients/producer/KafkaProducer.html)

Or another option can be to grab it from application.yml: https://github.com/zalando/nakadi/blob/nakadi-jvm/src/main/resources/application.yml#L23

src/main/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategy.java
tmuehl: Yes, it's a "magic" to workout the god old problem that `Function` doesn't allow to throw an Exception. Any better idea?

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
tmuehl: I want to avoid to parse the Json multiple times with no reason. The problem is that parsing the Json itself is already part of the validation, i.e. it can fail.

If you see validation as an _aspect_ of the Json parsing, it fits.

An alternative is to parse the Json before `validateSchema`, hand over a JSONObject to the method, and catch the `JSONException` in the Controller method. I'd be also fine with this, but for the moment I preferred the approach above, since it keeps the Controller method free from all other ExceptionHandling but NakadiExceptions. Moreover, if you look at the `@Valid` annotation for Controller method parameters it works more or less the same.

tmuehl:
@rcillo Orderings keys don't need to be defined. The strategy is only applied if at least one ordering key is present. Just leave them blank for a schema, if you want to.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
tmuehl: We talked about this when we planned the partitioning and agreed that we want to have these two strategies at this level:

* `OrderingKeyFieldsStrategy`
* `HardcodedPartitionStrategy`

And since we want to avoid the complexity of configuration the strategy for now, we agreed to decide based on whether the `OrderingKeyFields` property is set or not.

As soon as it is configurable this hardcoded partition "strategy" will be removed.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
tmuehl: I wanted to avoid the costly operation of asking Kafka for the number of partitions every time. Hence, I asked whether we expect the number of partitions for a topic to change. But a hardcoded `8` is not good, I agree. I will ask Kafka and cache the number of partitions per event type.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
rcillo: @tmuehl idea of splitting into to methods would look like this?

```java
JSONObject eventAsJson = parseEvent(event) // can be a private method that throws a domain exception
validateSchema(eventAsJson, eventType) // throws validation exception
```

Wdyt? Json would be parsed only once and validate does only schema validation.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
rcillo: I think it's time for us to discuss this caching in a broader context. I'm also interested in not doing expensive operations during event submission, in the context of schema validation.

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    @Test
+    public void partitionsAreEquallyDistributed() {
+        fillPartitionsWithEvents(simpleEventType, partitions, 10000);
+
+        final double[] eventDistribution = partitions.stream().map(p -> p.size()).mapToDouble(value -> value * 1.0).toArray();
+        final double variance = calculateVarianceOfUniformDistribution(eventDistribution);
+
+        assertThat(variance, lessThan(1.5));
```
rcillo: Why 1.5?

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
v-stepanov: Ok, I see. I probably forgot about that. Sorry :(

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
v-stepanov: @rcillo +1 
this is exactly what I thought of :)

src/main/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategy.java
v-stepanov: No, no, it's totally fine. 
I just thought that here you throw the same exception which you caught :)
But then I noticed that you throw the cause.

src/main/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+import org.json.JSONObject;
+
+import java.util.List;
+
+import static java.lang.Math.abs;
+
+public class OrderingKeyFieldsPartitioningStrategy implements PartitioningStrategy {
+
+    @Override
+    public String calculatePartition(final EventType eventType, final JSONObject event, final int numberOfPartitions) throws InvalidOrderingKeyFieldsException {
```
v-stepanov: As for me - the implementation is too kafka-specific (assuming that partition is a number from 0 to partitionsNum - 1).
I think we should abstract from that with a help of some new method of TopicRepository like getPartitionByIndex.

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
tmuehl: Haha, very good question. It's just because... I ran `testVariance` with different values to get a feeling for what is acceptable.
We had a discussion about a decent strategy to test the event distribution over the partitions, yesterday, too, without a good conclusion.

Do you have any idea what would be an acceptable variance for a discrete uniform distribution like the distribution of events over partitions?

src/main/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategy.java
tmuehl: Yes, I was thinking about exactly this. Ideally you should pass over the available partition names to the strategy as well. I then decided that for now it only makes it more complicated. For sure, this doesn't work for messaging brokers that don't provide simple increasing numbers for their partitions, but this is a step ahead.

With having the PartitionCache implemented (see discussion about the number of partitions handed over to this method in the Controller), we can use the cache, to cache the partition names. I will refactor in a second step.

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
```diff
@@ -0,0 +1,187 @@
...
+        final JSONObject jsonObject = new JSONObject();
+        jsonObject.put("sku", sku);
+        jsonObject.put("categoryId", categoryId);
+        jsonObject.put("name", name);
+        jsonObject.put("color", color);
+        jsonObject.put("price", price);
+        return jsonObject;
+    }
+
+    public static <T> Consumer<T> exceptionSafe(ConsumerWithException<T> consumer) {
```
v-stepanov: so you managed to cheat the system :-D

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
rcillo: It looks like an approachable problem from a statistical point of view. But tackling this problem would add not too much value to the project and since I'm no expert I'll just take it.

If we knew the sample variance distribution, we could choose a safe interval where the build would fail with controlled estimated frequency.

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
```diff
@@ -0,0 +1,187 @@
...
+
+        final int numberOfSamples = 10000;
+
+        for (int run = 0; run < 100; run++) {
+            final double[] dist = new double[8];
+            for (int i = 0; i < numberOfSamples; i++) {
+                dist[random.nextInt(dist.length)]++;
+            }
+            final double variance = calculateVarianceOfUniformDistribution(dist);
+            System.out.println(Arrays.toString(dist) + " = " + variance);
```
rcillo: :scissors:

rcillo: I see its ignored.

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
tmuehl: :D

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
```diff
@@ -0,0 +1,187 @@
...
+        final double x_sum = stream(samples).sum();
+        final double x_pow2_sum = stream(samples).map(d -> pow(d, 2)).sum();
+
+        final double n = x_sum;
+
+        final double expectedValue = (n / samples.length);
+
+
+        final double variance = (1.0 / n) * (x_pow2_sum - (1.0 / n) * pow(x_sum, 2));
+        return abs(variance - expectedValue);
```
rcillo: What does the difference between the variance and the expected value mean?

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
tmuehl: `1.5` is already very tolerant. Run the `testVariance` test locally and compare the values. :)
I was thinking of a retry for this test but seemed to much effort compared to the value for me.

PR #127, opened 2016-02-23
aruha-63 improve validation error messages

rcillo:
When an event has multiple validation errors then we concatenate all of
then in a single message.

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
+++ b/src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
@@ -175,13 +175,11 @@ class JSONSchemaValidator implements EventValidator {
     private void collectErrorMessages(ValidationException e, StringBuilder builder) {
         builder.append(e.getMessage());
 
-        if (!e.getCausingExceptions().isEmpty()) {
-            e.getCausingExceptions().stream()
-                    .forEach(causingException -> {
-                        builder.append("\n");
-                        collectErrorMessages(causingException, builder);
-                    });
-        }
+        e.getCausingExceptions().stream()
+                .forEach(causingException -> {
+                    builder.append("\n");
+                    collectErrorMessages(causingException, builder);
+                });
     }
 }
 
--- a/src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
+++ b/src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
@@ -166,16 +166,21 @@ class JSONSchemaValidator implements EventValidator {
             return Optional.empty();
         } catch (final ValidationException e) {
             StringBuilder builder = new StringBuilder();
-            builder.append(e.getMessage());
-
-            if (!e.getCausingExceptions().isEmpty()) {
-                builder.append("\n");
-                builder.append(e.getCausingExceptions().stream()
-                        .map(ValidationException::getMessage)
-                        .collect(Collectors.joining("\n")));
-            }
+            collectErrorMessages(e, builder);
 
             return Optional.of(new ValidationError(builder.toString()));
+        }
+    }
+
+    private void collectErrorMessages(ValidationException e, StringBuilder builder) {
+        builder.append(e.getMessage());
+
+        if (!e.getCausingExceptions().isEmpty()) {
+            e.getCausingExceptions().stream()
+                    .forEach(causingException -> {
+                        builder.append("\n");
+                        collectErrorMessages(causingException, builder);
+                    });
         }
     }
 }
--- a/src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
+++ b/src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
@@ -1,26 +1,22 @@
 package de.zalando.aruha.nakadi.validation;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.jsonorg.JsonOrgModule;
+import de.zalando.aruha.nakadi.domain.EventType;
+import de.zalando.aruha.nakadi.domain.ValidationStrategyConfiguration;
+import org.everit.json.schema.Schema;
+import org.everit.json.schema.ValidationException;
+import org.everit.json.schema.loader.SchemaLoader;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/127" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -1,26 +1,22 @@
...
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.datatype.jsonorg.JsonOrgModule;
-
-import de.zalando.aruha.nakadi.domain.EventType;
-import de.zalando.aruha.nakadi.domain.ValidationStrategyConfiguration;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
```
v-stepanov: I always love this code formatters wars in imports section :-D

rcillo: yes, =/

src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -169,7 +165,17 @@ public JSONSchemaValidator(final JSONObject effectiveSchema) {
 
             return Optional.empty();
         } catch (final ValidationException e) {
-            return Optional.of(new ValidationError(e.getMessage()));
+            StringBuilder builder = new StringBuilder();
+            builder.append(e.getMessage());
+
+            if (!e.getCausingExceptions().isEmpty()) {
```
rcillo: I'm wondering if I should have a recursive implementation for errors on deeply nested structures. I'll try and tell.

src/test/java/de/zalando/aruha/nakadi/validation/JSONSchemaValidationTest.java
```diff
@@ -76,15 +76,19 @@ public void schemaValidationShouldRespectEventTypeDefinition() {
...
+                "{\"foo\": \"bar\", \"bar\": {\"foo\": \"baz\"}, \"extra\": \"i should be no problem\"}");
+        final JSONObject invalidEventMissingMultipleFields = new JSONObject(
+                "{\"bar\": {\"foobar\": \"baz\"}, \"extra\": \"i should be no problem\", \"name\": \"12345\"}");
 
         assertThat(validator.validate(validEvent), isAbsent());
-        assertThat(validator.validate(invalidEventMissingFoo).get().getMessage(), equalTo("#: required key [foo] not found"));
+        assertThat(validator.validate(invalidEventMissingBar).get().getMessage(), equalTo("#: required key [bar] not found"));
         assertThat(validator.validate(invalidEventMissingNameField).get().getMessage(), equalTo("name is required"));
+        assertThat(validator.validate(invalidEventMissingMultipleFields).get().getMessage(),
+                equalTo("#: 2 schema violations found\n#/bar: required key [foo] not found\n#: required key [foo] not found"));
```
rcillo: Nested errors are displayed like this. First it displays the path, like "/bar/foo/baz/whatever" than the error message. So there's no need for recursively looking into causing exceptions, from what I got so far.

rcillo:
Actually, recursive error message was needed https://github.com/zalando/nakadi/commit/2188ef760cd7be97376c215c3cc4007079c8a330

It's a special case when you have multiple errors on a nested object.

Do you think that concatenating with "\n" is enough? In order to make it readable, someone would be required to properly display the "\n" as breaklines, at least.

src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -169,7 +165,22 @@ public JSONSchemaValidator(final JSONObject effectiveSchema) {
...
+        }
+    }
+
+    private void collectErrorMessages(ValidationException e, StringBuilder builder) {
+        builder.append(e.getMessage());
+
+        if (!e.getCausingExceptions().isEmpty()) {
+            e.getCausingExceptions().stream()
+                    .forEach(causingException -> {
+                        builder.append("\n");
```
bocytko: mutating lambdas?

src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -169,7 +165,22 @@ public JSONSchemaValidator(final JSONObject effectiveSchema) {
...
+            collectErrorMessages(e, builder);
+
+            return Optional.of(new ValidationError(builder.toString()));
+        }
+    }
+
+    private void collectErrorMessages(ValidationException e, StringBuilder builder) {
+        builder.append(e.getMessage());
+
+        if (!e.getCausingExceptions().isEmpty()) {
```
bocytko: This is not needed, right? You can just iterate over an empty collection, which is a noop.

src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -169,7 +165,22 @@ public JSONSchemaValidator(final JSONObject effectiveSchema) {
...
+            e.getCausingExceptions().stream()
+                    .forEach(causingException -> {
+                        builder.append("\n");
+                        collectErrorMessages(causingException, builder);
```
tmuehl: Challenge: Do you manage to do it without recursion? :wink:

src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
rcillo: I guess that's the purpose of `forEach`, since it returns `void`, right?

src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
rcillo: Good catch :+1: done in https://github.com/zalando/nakadi/commit/f39d94918c1371a3f943688a29445aaf94cdcbe2

src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
rcillo: I can only imagine that using an additional data structure, like a stack. Do you think it's possible to do that even without such a data structure?

tmuehl: Don't know. The solution I have in my head is also with a stack, yes. :)

PR #129, opened 2016-02-23
Aruha 71 support partitioning

tmuehl:
Implements ARUHA-73 and ARUHA-74


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/129" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.List;
+
+import static java.lang.Math.abs;
+
+public class OrderingKeyFieldsPartitioningStrategy implements PartitioningStrategy {
+
+    @Override
+    public String calculatePartition(final EventType eventType, final JSONObject event, final String[] partitions) throws InvalidOrderingKeyFieldsException {
+        final List<String> orderingKeyFields = eventType.getOrderingKeyFields();
+        if (orderingKeyFields.isEmpty()) {
```
rcillo: It seems that this method would return 0 in the case of empty ordering key fields. I'm not sure. Maybe we could just not check and return 0 silently?

src/main/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategy.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    private static class JsonPath {
+        private final JSONObject jsonObject;
+
+        private JsonPath(final JSONObject jsonObject) {
+            this.jsonObject = jsonObject;
+        }
+
+        public Object get(final String path) {
+            final String[] fields = StringUtils.split(path, '.');
```
rcillo: I think we should split composed keys in the format `"key_1.1"."key_1.2"."key_1.3"` e.g. considering that fields might contain `.`

rcillo: dot-notation is so much cleaner. Maybe we should state in the API that json should not contain dots in the field names.

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
```diff
@@ -0,0 +1,270 @@
...
+        fillPartitionsWithRandomEvents(simpleEventType, partitions, 1000);
+
+        checkThatEventsWithSameKeysAreInSamePartition(partitions);
+    }
+
+    @Test
+    @Ignore
+    public void partitionsAreEvenlyDistributed_usingRandomEvents() {
+        // This is a probabilistic test.
+        // The probability that it fails is approx. 0.577%
```
rcillo: I thought we would replace this test by a deterministic one. Wdyt?

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
```diff
@@ -0,0 +1,270 @@
...
+        final double variance = calculateVarianceOfUniformDistribution(eventDistribution);
+
+        assertThat(variance, lessThan(1.5));
+    }
+
+    @Test
+    public void partitionsAreEvenlyDistributed() throws IOException {
+        loadEventSamples();
+
+        assertThat("Event sample set A is not evenly distributed with strategy", varianceForEvents(EVENT_SAMPLES_A), lessThan(1.5));
```
rcillo: :+1: determinism in tests

src/main/java/de/zalando/aruha/nakadi/repository/InMemoryTopicRepository.java
```diff
@@ -80,13 +80,18 @@ private MockPartition getPartitionStorage(final String topicId, final String par
...
         if (mockTopic == null) {
-            throw new InternalNakadiException("No such topic");
+            throw new InternalNakadiException("No such topic '" + topicId + "'");
         }
 
         return mockTopic.partitions.values().stream().map(p -> new TopicPartition(topicId, p.id)).collect(toList());
     }
 
     @Override
+    public String[] listPartitionNames(final String topicId) throws NakadiException {
```
v-stepanov: why array?

tmuehl:

src/main/java/de/zalando/aruha/nakadi/partitioning/InvalidOrderingKeyFieldsException.java
```diff
@@ -0,0 +1,17 @@
+package de.zalando.aruha.nakadi.partitioning;
```
rcillo: Why is this exception in a different package?

tmuehl: That wasn't by intention. Usually I like to structure my packages by business logic not technically, that's probably why this class ended up here. But I see, we have a technical structure in the project, I'll move it. Thanks for pointing out!

src/test/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategyTest.java
tmuehl: see below :)

src/main/java/de/zalando/aruha/nakadi/partitioning/OrderingKeyFieldsPartitioningStrategy.java
tmuehl: I don't see it as a responsibility of this strategy to fall back to another strategy if it cannot be applied. If someone tries to apply this strategy although it's not applicable, I see it as an error. Don't you think so?

rcillo: Your point makes sense. But I think it's a matter of defining the meaning of this strategy in the API.

rcillo:
It looks pretty solid for me. I just wasn't sure why you left some tests with `@Ignore` annotation.

Fix build and :shipit:

tmuehl:
@rcillo Thanks! I added a "ignore reason" to the ignored tests.

src/main/java/de/zalando/aruha/nakadi/util/JsonPathAccess.java
```diff
@@ -0,0 +1,90 @@
...
+
+    private static class JsonPathTokenizer {
+        private final char[] path;
+        private int pos = 0;
+
+        private JsonPathTokenizer(final String path) {
+            this.path = path.toCharArray();
+        }
+
+        public synchronized String nextToken() {
```
CyberDem0n: Does it really needs synchronization? If not, then tokenBuilder object could be created only once.
But on every iteration it needs to be reset by calling `setLength(0)`

src/main/java/de/zalando/aruha/nakadi/util/JsonPathAccess.java
```diff
@@ -0,0 +1,90 @@
...
+            boolean inSingleQuote = false;
+            boolean escaped = false;
+
+            while (pos < path.length) {
+                final char c = path[pos];
+
+                pos++;
+
+                if (escaped) {
+                    tokenBuilder.append(c);
```
CyberDem0n: Not every character can be `unescaped`, for example `\\`, `\"`, '\n', '\r' can be, but  `\s` can't.

src/main/java/de/zalando/aruha/nakadi/util/JsonPathAccess.java
```diff
@@ -0,0 +1,90 @@
...
+                    escaped = false;
+                } else {
+
+                    if (c == '\\') {
+                        escaped = true;
+                    } else if (c == '\'') {
```
CyberDem0n: Why single quote?
Shouldn't the double quote be used?
Actually I'm ok with single quote, but people could have different opinion...
May be we can support both cases?

tmuehl: Yes, supporting both cases would be the best. Single quotes make it easier to use in json property values (which are already double-quoted strings). But since it is not needed (the api guidelines doesn't allow dots in fieldnames anyway), I would avoid putting more effort into it.

src/main/java/de/zalando/aruha/nakadi/util/JsonPathAccess.java
tmuehl: Ok, valid point. However, I would not put more effort into it, because it's simply not needed right now.

PR #132, opened 2016-02-24
Fix travis build

v-stepanov:
I temporarily disabled codacy as it fails the build. Seems that repo.jpm4j.org doesn't work.
Will turn it back when it's good.
@jkliff @tmuehl @rcillo @CyberDem0n please review


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/132" target="_blank">ReviewNinja

PR #133, opened 2016-02-26
Aruha 104 - Event Type in memory cache

rcillo:
Caching event types in memory could significantly improve the
performance of event publishing, since we need to load event types every
time to check for the schema.

We decided to use an eventual consistent cache synchronization, which
means that it might take some time to propagate changes to all nakadi
instances. This solution was chosen over fully synchronized one due to
CAP (performance) and it fits well the most usual use cases.

In order to test this feature, you can just boot the application normally, 
post some new events using curl 

`curl -v -X POST -d @event-type.json http -H "Content-type: application/json" http://localhost:8080/event-types`

Then a simple GET in order to cache the event type.

`curl -X GET http://localhost:8080/event-types/event-name`

Then, there's no need to launch another instance of the app, you can
just log in the storages docker and simulated updates directly there

```
$ docker exec -ti local-storages bash
# cd /opt/kafka_2.11-0.9.0.0/bin
# sh zookeeper-shell.sh localhost:2181
# set /nakadi/event_types/event-name something
```

This should invalidate the cache, since there a watcher on every instance listening to this znode.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/133" target="_blank">ReviewNinja

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private final EventTypeRepository dbRepo = mock(EventTypeRepository.class);
+    private final CuratorFramework client;
+
+    public EventTypeCacheTest() throws Exception {
+        ZooKeeperHolder zkHolder = new ZooKeeperHolder("127.0.0.1:2181", "", "", 0);
+        zkHolder.init();
+        String connecString = zkHolder.get().getZookeeperClient().getCurrentConnectionString();
+        RetryPolicy retryPolicy = new ExponentialBackoffRetry(1000, 3);
+        CuratorFramework cf = CuratorFrameworkFactory.newClient(connecString, retryPolicy);
```
rcillo: Maybe I should just build this connection string directly.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCachedRepository.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.List;
+
+public class EventTypeCachedRepository implements EventTypeRepository {
+
+    private final EventTypeRepository repository;
+
+    private final EventTypeCache cache;
+
+    public EventTypeCachedRepository(EventTypeRepository repository,
```
rcillo: This a "decorated" version of the repository, in the sense that it just adds some functionality on top of the DB repo without changing the interface.

I also isolated all caching behaviour in the cache class, not to mess up with things here.

I thought this approach would be a less intrusive way of adding the caching. But feel free to make drastic suggestions of design change.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private LoadingCache<String,EventType> setupInMemoryCache(EventTypeRepository dbRepo) {
+        CacheLoader<String, EventType> loader = new CacheLoader<String, EventType>() {
+            public EventType load(String key) throws NoSuchEventTypeException {
+                return dbRepo.findByName(key);
+            }
+        };
+
+        return CacheBuilder.newBuilder().maximumSize(100000).build(loader);
```
rcillo: This is arbitrarily chosen number, just to make sure memory would not increase wildly. But I haven't checked for heap size if it would fit.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }
+
+    private void addCacheChangeListener(final LoadingCache<String, EventType> cache, PathChildrenCache cacheSync) {
+        PathChildrenCacheListener listener = new PathChildrenCacheListener() {
+            @Override
+            public void childEvent(CuratorFramework client, PathChildrenCacheEvent event) throws Exception {
+                switch ( event.getType() ) {
+                    case CHILD_UPDATED: {
+                        invalidateCacheKey(event);
```
rcillo: When something changes, just invalidate the cache and waits for the nest request to reload it, if necessary.

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
```diff
@@ -0,0 +1,169 @@
...
+        cf.start();
+        this.client = cf;
+    }
+
+    @Before
+    public void setUp() throws Exception {
+        if (client.checkExists().forPath("/nakadi/event_types/event-name") != null) {
```
rcillo: I will extract this repeated code.

rcillo:
Some local performance testing, to get a feeling of how much it improves performance

The speedup is around 1.6x. Not that impressive.

## Not cached

$ ab -p event.json -T application/json -c 10 -n 10000 http://localhost:8080/event-types/NAME_PLACEHOLDER/events
This is ApacheBench, Version 2.3 <$Revision: 1604373 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests


Server Software:        Apache-Coyote/1.1
Server Hostname:        localhost
Server Port:            8080

Document Path:          /event-types/NAME_PLACEHOLDER/events
Document Length:        0 bytes

Concurrency Level:      10
Time taken for tests:   33.225 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      3450000 bytes
Total body sent:        1890000
HTML transferred:       0 bytes
Requests per second:    300.98 [#/sec] (mean)
Time per request:       33.225 [ms] (mean)
Time per request:       3.322 [ms] (mean, across all concurrent requests)
Transfer rate:          101.40 [Kbytes/sec] received
                        55.55 kb/s sent
                        156.96 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    2   1.8      1      19
Processing:     7   31  20.6     28     566
Waiting:        6   30  20.6     27     560
Total:          8   33  20.5     30     566

Percentage of the requests served within a certain time (ms)
  50%     30
  66%     35
  75%     38
  80%     40
  90%     46
  95%     53
  98%     62
  99%     70
 100%    566 (longest request)


## Cached

ab -p event.json -T application/json -c 10 -n 10000 http://localhost:8080/event-types/NAME_PLACEHOLDER/events
This is ApacheBench, Version 2.3 <$Revision: 1604373 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests


Server Software:        Apache-Coyote/1.1
Server Hostname:        localhost
Server Port:            8080

Document Path:          /event-types/NAME_PLACEHOLDER/events
Document Length:        0 bytes

Concurrency Level:      10
Time taken for tests:   20.255 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      3450000 bytes
Total body sent:        1890000
HTML transferred:       0 bytes
Requests per second:    493.72 [#/sec] (mean)
Time per request:       20.255 [ms] (mean)
Time per request:       2.025 [ms] (mean, across all concurrent requests)
Transfer rate:          166.34 [Kbytes/sec] received
                        91.13 kb/s sent
                        257.46 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    2   2.5      1     133
Processing:     4   17   7.2     16     107
Waiting:        4   16   7.6     15     106
Total:          7   20   6.7     18     145

Percentage of the requests served within a certain time (ms)
  50%     18
  66%     20
  75%     22
  80%     23
  90%     28
  95%     32
  98%     36
  99%     41
 100%    145 (longest request)

tmuehl:
@rcillo Thanks for the benchmark!
True, 1.6x is not such impressive. Have you switched off Authentication?
However, the impact in production system might be bigger since the database is not running on the same host and we get rid of the network latency.

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
tmuehl: What do you mean?

tmuehl:
@rcillo Haven't checked everywhere, but almost all method parameters can be final.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCachedRepository.java
tmuehl: `CachingEventTypeRepository` or `CachedEventTypeRepository` since it implements `EventTypeRepository`?

rcillo: Do you think `EventTypeCachedRepository` sounds incorrect for an English speaker?

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
rcillo: What the lines above are doing is kind of dumb, because I first create a client using `ZooKeeperHolder` just to get the connection string in order to create another client.

So I thought it would be enough to hardcode the connection string. I'll try it out and comment the commit link :+1:

rcillo: @tmuehl https://github.com/zalando/nakadi/commit/b464726602cd8cc61ba5747890468e5f3877c478

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
rcillo: Maybe it should be made configured with an app property?

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCachedRepository.java
tmuehl: It's more because of class naming convention.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
v-stepanov: I think that would be better if it is configurable

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
```diff
@@ -0,0 +1,154 @@
...
+        EventTypeCache etc = new EventTypeCache(dbRepo, client);
+
+        EventType et = buildEventType();
+
+        Mockito
+                .doReturn(et)
+                .when(dbRepo)
+                .findByName(et.getName());
+
+        etc.get(et.getName());
```
v-stepanov: I think here you can also check the cache actually returned correct event type (first time and second time)

src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
```diff
@@ -26,8 +32,19 @@
...
+
     @Bean
-    public EventTypeRepository eventTypeRepository() {
-        return new EventTypeDbRepository(jdbcTemplate, jsonConfig.jacksonObjectMapper());
+    public EventTypeRepository eventTypeRepository() throws Exception {
+        final EventTypeRepository dbRepo = new EventTypeDbRepository(jdbcTemplate, jsonConfig.jacksonObjectMapper());
+
+        final String connectString = zookeeperConfig.zooKeeperHolder().get().getZookeeperClient().getCurrentConnectionString();
+        final RetryPolicy retryPolicy = new ExponentialBackoffRetry(1000, 3);
+        final CuratorFramework client = CuratorFrameworkFactory.newClient(connectString, retryPolicy);
```
v-stepanov: why do you create a seconds CuratorFramework here? 
We already have one in https://github.com/zalando/nakadi/blob/nakadi-jvm/src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZooKeeperHolder.java#L29

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
```diff
@@ -0,0 +1,140 @@
...
+
+        this.zkClient = zkClient;
+        this.cache = setupInMemoryCache(dbRepo);
+        this.cacheSync = setupCacheSync(zkClient);
+    }
+
+    public void updated(final String name) {
+        try {
+            final String path = getZNodePath(name);
+            zkClient.setData().forPath(path, new byte[0]);
```
v-stepanov: will it actually trigger the watchers if you set the same value as it had before?

src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
rcillo: The `AdminUtils` close the connection right after creating a topic, for example. I needed the connection to remain open in order to receive the notification.

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
rcillo: :+1: done https://github.com/zalando/nakadi/commit/990243b6d348d62cd23c2a568c7f7da52a5bdf7c

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
rcillo: Yes, I tried it. It always triggers independently of the value.

rcillo:
Codacy is complaining about an instance variable that could be made local. But what codacy is missing is that this way the object would be garbage collected. We need to have a reference to that object otherwise the cache would run out of sync after sometime.

CyberDem0n:
Regarding performance tests:
Sure, apache benchmark is nice test utility, but unfortunately it supports only HTTP/1.0. That mean nakadi is closing connection after processing each request. Establishing of new connection could be relatively expensive. We should try to run the same performance tests with Siege (https://www.joedog.org/siege-home/), which supports HTTP/1.1

src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
CyberDem0n: @rcillo, `AdminUtils` is using not `Curator` but `ZkUtils` to work with Zookeeper.

Currently we are using the same Exhibitor/ZooKeeper for Kafka and for Nakadi, that means we should not create new CuratorFramework but use already existing one. It is not doing much besides periodically fetching list of Kafka brokers from ZooKeeper.

rcillo: I'll investigate further, but using the already existing one did not work. The application wasn't receiving watcher triggers.

v-stepanov: Actually not. In `createTopic` method the curator is not used. We use it only to get the connection string.

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+        executeWithRetry(() -> {
+                    try {
+                        etc.get(et.getName());
+                        verify(dbRepo, times(2)).findByName(et.getName());
+                    } catch (NoSuchEventTypeException e) {
+                        fail();
+                    } catch (ExecutionException e) {
+                        fail();
+                    } catch (Exception e) {
```
v-stepanov: these three can be collapsed into one catch

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
```diff
@@ -0,0 +1,159 @@
...
+import static junit.framework.Assert.assertNull;
+import static junit.framework.Assert.fail;
+import static org.echocat.jomon.runtime.concurrent.Retryer.executeWithRetry;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+
+public class EventTypeCacheTest {
```
v-stepanov: Where are all the `final`s in this test? :)

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/UserJourneyAT.java
```diff
@@ -77,15 +79,20 @@ public void userJourneyM1() {
...
+                    jsonRequestSpec()
+                            .when()
+                            .get("/event-types/" + TEST_EVENT_TYPE)
+                            .then()
+                            .statusCode(OK.value())
+                            .and()
+                            .body("owning_application", equalTo("my-app"))
+                            .body("category", equalTo("new-data"));
+                },
+                new RetryForSpecifiedTimeStrategy<Void>(5000).withExceptionsThatForceRetry(AssertionError.class)
```
v-stepanov: `@SuppressWarnings("unchecked")` ?

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
```diff
@@ -0,0 +1,159 @@
...
+                        verify(dbRepo, times(2)).findByName(et.getName());
+                    } catch (NoSuchEventTypeException e) {
+                        fail();
+                    } catch (ExecutionException e) {
+                        fail();
+                    } catch (Exception e) {
+                        fail();
+                    }
+                },
+                new RetryForSpecifiedTimeStrategy<Void>(5000).withExceptionsThatForceRetry(AssertionError.class)
```
v-stepanov: `@SuppressWarnings("unchecked")` ?

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/UserJourneyAT.java
v-stepanov: `@SuppressWarnings("unchecked")` ?

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
```diff
@@ -0,0 +1,128 @@
...
+        final String path = getZNodePath(eventType.getName());
+        zkClient
+                .create()
+                .creatingParentsIfNeeded()
+                .withMode(CreateMode.PERSISTENT)
+                .forPath(path, new byte[0]);
+    }
+
+    public void removed(final String name) throws Exception {
+        final String path = ZKPaths.makePath(ZKNODE_PATH, name);
```
CyberDem0n: Why not to use `getZNodePath` method here?

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
```diff
@@ -0,0 +1,128 @@
...
+        zkClient.delete().forPath(path);
+    }
+
+    private void initParentCacheZNode(final CuratorFramework zkClient) throws Exception {
```
CyberDem0n: This method is not really needed, since you are using `creatingParentsIfNeeded()` everywhere.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCachedRepository.java
```diff
@@ -0,0 +1,93 @@
...
+            return cache.get(name);
+        } catch (ExecutionException e) {
+            LOG.error("Failed to load event type from cache '" + name + "'", e);
+            throw new InternalNakadiException("Failed to load event type", e);
+        }
+    }
+
+    @Override
+    public void update(final EventType eventType) throws InternalNakadiException, NoSuchEventTypeException {
+        EventType original = this.repository.findByName(eventType.getName());
```
CyberDem0n: final

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCachedRepository.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    @Override
+    public List<EventType> list() {
+        return this.repository.list();
+    }
+
+    @Override
+    public void removeEventType(final String name) throws InternalNakadiException, NoSuchEventTypeException {
+        EventType original = this.repository.findByName(name);
```
CyberDem0n: final

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/UserJourneyAT.java
rcillo: :+1:  done https://github.com/zalando/nakadi/commit/bbdda25f5d50a73aeb65d06f87bd20f8c29738da

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
rcillo: :+1:  done https://github.com/zalando/nakadi/commit/bbdda25f5d50a73aeb65d06f87bd20f8c29738da

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
rcillo: :+1:  done https://github.com/zalando/nakadi/commit/bbdda25f5d50a73aeb65d06f87bd20f8c29738da

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/UserJourneyAT.java
rcillo: https://github.com/zalando/nakadi/commit/ee4e934465bbd25c5ecff6e4e20c02bb03a79258 also

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/EventTypeCacheTest.java
rcillo: done https://github.com/zalando/nakadi/commit/bbdda25f5d50a73aeb65d06f87bd20f8c29738da

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
rcillo: Done :+1: https://github.com/zalando/nakadi/commit/33c7901ceb9749b3937638708d5b54b26661fb0f

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
rcillo: I shouldn't be using `creatingParentsIfNeeded()` everywhere, since the node `/nakadi/event_types` should be created only on application bootstrap. This node should be created before the initialisation of the remote cache and never afterwards.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCachedRepository.java
rcillo: done

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCachedRepository.java
rcillo: done :+1:

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
CyberDem0n: In most cases it's good to use `creatingParentsIfNeeded()` because it costs nothing.

I.e., it tries to create your node and if it gets NoNodeExistsException it will create parent node and retry.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCachedRepository.java
```diff
@@ -0,0 +1,93 @@
...
+import de.zalando.aruha.nakadi.exceptions.NoSuchEventTypeException;
+import de.zalando.aruha.nakadi.repository.DuplicatedEventTypeNameException;
+import de.zalando.aruha.nakadi.repository.EventTypeRepository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+
+public class EventTypeCachedRepository implements EventTypeRepository {
```
tmuehl: `CachedEventTypeRepository` or `CachingEventTypeRepository` ?

rcillo: Done https://github.com/zalando/nakadi/commit/d4b0c3961d49f156433cafa96ee9b9d1ee7ffac3

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
rcillo: :+1: restored the `creatingParentsIfNeeded()` call.

src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
rcillo: You were right. Done here https://github.com/zalando/nakadi/commit/c3aa6f2c4079e436bf2834caa96f4bf737e27ccc

v-stepanov:
:shipit:

PR #140, opened 2016-02-29
Fixes 138 Allow dots in event-type names

tmuehl:
Disables suffix detection by Spring, e.g. ".htm", ".xml" or ".json". Hence, content-type negotiation via file type suffix isn't possible anymore. On the other hand, it's now possible to have `PathVariable`s containing a dot.

### DIFF
--- a/src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/UserJourneyAT.java
+++ b/src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/UserJourneyAT.java
@@ -9,6 +9,8 @@ import org.junit.Before;
 import org.junit.Test;
 
 import java.io.IOException;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
 
 import static com.jayway.restassured.http.ContentType.JSON;
 import static de.zalando.aruha.nakadi.utils.TestUtils.randomString;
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/140" target="_blank">ReviewNinja

v-stepanov:
Do we want to have a test for that? :)
I can be a good practice to write a test for the bug fixed.

rcillo:
@v-stepanov @tmuehl maybe changing the user journey test would be a good idea. Making it more "realistic".

tmuehl:
@v-stepanov Yes, you are right, I provide one.

jkliff:
+1 for the test; user journey seems suitable

tmuehl:
According to your suggestions, I changed the `UserJourneyAT` to use an event-type name with a dot in it.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/UserJourneyAT.java
```diff
@@ -19,7 +19,7 @@
 
 public class UserJourneyAT extends RealEnvironmentAT {
 
-    private static final String TEST_EVENT_TYPE = randomString();
+    private static final String TEST_EVENT_TYPE = "order.ORDER_CANCELLED";
```
v-stepanov: such a name will limit the possibility to run this UserJourney test on a real system (because it can conflict with real name)
Maybe `randomString() + "." + randomString()` ?

tmuehl: Makes sense, I change it

bocytko:
:+1:

PR #142, opened 2016-02-29
aruha-104 validator cache

rcillo:
Caching the event type validator spares some CPU since its assembly is
CPU intensive.

Once event type and its validator is cached, we achieved a speedup of
2.2x, pushing time per request as low as 14ms, locally.

## After Validator caching

$ ab -p event.json -T application/json -c 10 -n 10000 http://localhost:8080/event-types/NAME_PLACEHOLDER/events
This is ApacheBench, Version 2.3 <$Revision: 1604373 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests

Server Software:        Apache-Coyote/1.1
Server Hostname:        localhost
Server Port:            8080

Document Path:          /event-types/NAME_PLACEHOLDER/events
Document Length:        0 bytes

Concurrency Level:      10
Time taken for tests:   14.445 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      3450000 bytes
Total body sent:        1890000
HTML transferred:       0 bytes
Requests per second:    692.29 [#/sec] (mean)
Time per request:       14.445 [ms] (mean)
Time per request:       1.444 [ms] (mean, across all concurrent requests)
Transfer rate:          233.24 [Kbytes/sec] received
                        127.78 kb/s sent
                        361.02 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    3   2.4      3      14
Processing:     3   10   2.7     10      30
Waiting:        3    9   2.7      9      29
Total:          4   14   3.8     13      32

Percentage of the requests served within a certain time (ms)
  50%     13
  66%     15
  75%     16
  80%     17
  90%     19
  95%     21
  98%     22
  99%     24
 100%     32 (longest request)

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
+++ b/src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
@@ -6,6 +6,8 @@ import de.zalando.aruha.nakadi.repository.db.EventTypeCachedRepository;
 import de.zalando.aruha.nakadi.repository.db.EventTypeDbRepository;
 import de.zalando.aruha.nakadi.repository.kafka.KafkaConfig;
 import de.zalando.aruha.nakadi.repository.zookeeper.ZookeeperConfig;
+import de.zalando.aruha.nakadi.validation.EventBodyMustRespectSchema;
+import de.zalando.aruha.nakadi.validation.ValidationStrategy;
 import org.apache.curator.RetryPolicy;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.CuratorFrameworkFactory;
@@ -36,15 +38,25 @@ public class RepositoriesConfig {
     private ZookeeperConfig zookeeperConfig;
 
     @Bean
-    public EventTypeRepository eventTypeRepository() throws Exception {
-        final EventTypeRepository dbRepo = new EventTypeDbRepository(jdbcTemplate, jsonConfig.jacksonObjectMapper());
-
+    public EventTypeCache eventTypeCache() throws Exception {
         final String connectString = zookeeperConfig.zooKeeperHolder().get().getZookeeperClient().getCurrentConnectionString();
         final RetryPolicy retryPolicy = new ExponentialBackoffRetry(1000, 3);
         final CuratorFramework client = CuratorFrameworkFactory.newClient(connectString, retryPolicy);
 
         client.start();
 
-        return new EventTypeCachedRepository(dbRepo, new EventTypeCache(dbRepo, client));
+        EventBodyMustRespectSchema strategy = new EventBodyMustRespectSchema();
+        ValidationStrategy.register(EventBodyMustRespectSchema.NAME, strategy);
+
+        return new EventTypeCache(dbRepo(), client);
+    }
+
+    @Bean
+    public EventTypeRepository eventTypeRepository() throws Exception {
+        return new EventTypeCachedRepository(dbRepo(), eventTypeCache());
+    }
+
+    private EventTypeRepository dbRepo() {
+        return new EventTypeDbRepository(jdbcTemplate, jsonConfig.jacksonObjectMapper());
     }
 }
--- a/src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
+++ b/src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
@@ -45,7 +45,7 @@ public class RepositoriesConfig {
 
         client.start();
 
-        EventBodyMustRespectSchema strategy = new EventBodyMustRespectSchema();
+        final EventBodyMustRespectSchema strategy = new EventBodyMustRespectSchema();
         ValidationStrategy.register(EventBodyMustRespectSchema.NAME, strategy);
 
         return new EventTypeCache(dbRepo(), client);
--- a/src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
+++ b/src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
@@ -4,8 +4,12 @@ import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
 import de.zalando.aruha.nakadi.domain.EventType;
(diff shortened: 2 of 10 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/142" target="_blank">ReviewNinja

v-stepanov:
there are few tests failed

src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
```diff
@@ -36,15 +38,25 @@
...
+
+        return new EventTypeCache(dbRepo(), client);
+    }
+
+    @Bean
+    public EventTypeRepository eventTypeRepository() throws Exception {
+        return new EventTypeCachedRepository(dbRepo(), eventTypeCache());
+    }
+
+    private EventTypeRepository dbRepo() {
```
v-stepanov: If you don't put @Bean annotation for this method - it will create db repository every time this method is called.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
```diff
@@ -76,7 +86,7 @@ private void initParentCacheZNode(final CuratorFramework zkClient) throws Except
         }
     }
 
-    private void addCacheChangeListener(final LoadingCache<String, EventType> cache, final PathChildrenCache cacheSync) {
+    private void addCacheChangeListener(final LoadingCache<String, EventType> eventTypeCache, LoadingCache<String, EventTypeValidator> validatorCache, final PathChildrenCache cacheSync) {
```
v-stepanov: `With great power comes great responsibility`. If we don't use Jalopy - we should care about code formatting ourselves. Please try to avoid such long lines.

src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
```diff
@@ -36,15 +38,25 @@
...
-
+    public EventTypeCache eventTypeCache() throws Exception {
         final String connectString = zookeeperConfig.zooKeeperHolder().get().getZookeeperClient().getCurrentConnectionString();
         final RetryPolicy retryPolicy = new ExponentialBackoffRetry(1000, 3);
         final CuratorFramework client = CuratorFrameworkFactory.newClient(connectString, retryPolicy);
 
         client.start();
 
-        return new EventTypeCachedRepository(dbRepo, new EventTypeCache(dbRepo, client));
+        EventBodyMustRespectSchema strategy = new EventBodyMustRespectSchema();
```
v-stepanov: final ?

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
CyberDem0n: missing final for validatorCache

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
```diff
@@ -107,12 +119,26 @@ private PathChildrenCache setupCacheSync(final CuratorFramework zkClient) throws
...
         cacheSync.start();
 
-        addCacheChangeListener(cache, cacheSync);
+        addCacheChangeListener(eventTypeCache, validatorCache, cacheSync);
 
         return cacheSync;
     }
 
-    private LoadingCache<String,EventType> setupInMemoryCache(final EventTypeRepository dbRepo) {
+    private LoadingCache<String, EventTypeValidator> setupInMemoryValidatorCache(LoadingCache<String, EventType> eventTypeCache) {
```
CyberDem0n: final

src/main/java/de/zalando/aruha/nakadi/config/RepositoriesConfig.java
tmuehl: I would also suggest giving it a meaningful name like `eventTypeRepository`.
However, you then have two `EventTypeRepository` bean which will lead to trouble when using `@Autowired`. Hence, it would be good having The `DatabaseEventTypeRepository` not as a spring bean. That leads me to the question why `EventTypeCachedRepository` needs a `EventTypeCache` and both beans need a 'EventTypeRepository'.

tmuehl: Maybe you can merge `EventTypeCachedRepository` and `EventTypeCache` into one class, that transparently caches. The `EventTypeCachedRepository` doesn't contain much logic at all.

rcillo: @tmuehl I'm trying to solve this intertwined dependencies issues. To explain why it's the way it is:

1. `EventTypeCachedRepository` needs an `EventTypeCache` because it needs to notify the cache when something changes in the repository. E.g. when someone updates an event type, we need to notify zookeeper that it changed. Talking to zookeeper is responsibility of `EventTypeCache`.

2. `EventTypeCache` needs an `EventTypeDbRepository` because it loads `EventType`s from database on cache miss.

3. `EventTypeCachedRepository` needs one `EventTypeDbRepository` because it needs to persist changes in the Db.

rcillo: @tmuehl the problem with merging is that `EventTypeCachedRepository` would also be responsible for handling validators, which seems weird.

src/main/java/de/zalando/aruha/nakadi/repository/db/EventTypeCache.java
rcillo: This is definitely the result of some automatic refactoring. Gonna fix it.

rcillo:
I'v addressed the comments and fixed the tests. The only discussion that needs a final word is this https://github.com/zalando/nakadi/pull/142#discussion_r54579410 but I don't have much to say about it.

PR #147, opened 2016-03-02
This unchecked cast causes a warning

tmuehl:
...that breaks the build if the compiler is configured to be restrictive.

Any better suggestion to get rid of that warning?


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/147" target="_blank">ReviewNinja

rcillo:
Do you think it would be a good idea to make our build more restrictive, so we would have feedback from travis on our PRs?

tmuehl:
@rcillo yes, I think so

v-stepanov:
:shipit:

PR #149, opened 2016-03-02
Invalid Enum constant error message

### LINKED ISSUE #146: Ugly error message when using invalid enums in json body
**As a** Nakadi user
**I want to** get a useful error response if I specify invalid enum constants in my json body
**Such that** I can debug my request easily

For example, if someone posts an event-type json using invalid enum constants, she ends up with an error response like this:

```
{
  "detail": "Could not read document: No enum constant de.zalando.aruha.nakadi.domain.EventTypeSchema.Type.JSON_SCHEMX (through reference chain: de.zalando.aruha.nakadi.domain.EventType[\"schema\"]->de.zalando.aruha.nakadi.domain.EventTypeSchema[\"type\"]); nested exception is com.fasterxml.jackson.databind.JsonMappingException: No enum constant de.zalando.aruha.nakadi.domain.EventTypeSchema.Type.JSON_SCHEMX (through reference chain: de.zalando.aruha.nakadi.domain.EventType[\"schema\"]->de.zalando.aruha.nakadi.domain.EventTypeSchema[\"type\"])",
  "status": 400,
  "title": "Bad Request",
  "type": "http://httpstatus.es/400"
}
```

Todo:
- There should be no Java Classes and internals exposed
- It would be nice if the error message contains the valid possible values for the enum

tmuehl:

Fixes #146


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/149" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/controller/ExceptionHandling.java
```diff
@@ -36,5 +38,16 @@ private String generateErrorTraceId() {
         return "ETI" + RandomStringUtils.randomAlphanumeric(24);
     }
 
-
+    @Override
+    @ExceptionHandler
+    public ResponseEntity<Problem> handleMessageNotReadableException(final HttpMessageNotReadableException exception, final NativeWebRequest request) {
```
rcillo: It's not very clear why this code exists. It handles `JsonMappingException` in a very specific way. Maybe we should add a test for that or just document this code, explaining the motivation.

tmuehl: You are right, it looks very arbitrary. I'll add some documentation.

PR #155, opened 2016-03-04
ARUHA-126 working on API for event-type deletion

v-stepanov:
@jkliff @rcillo @CyberDem0n @tmuehl 
Please take a look. As a starting point of discussion of API for event-type deletion.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/155" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -351,6 +353,54 @@ paths:
...
+    delete:
+      tags:
+        - schema-registry-api
+      security:
+        - oauth2: [nakadi.admin]
+      description: |
+        Deletes an `EventType` identified by its name. The underlying Kafka topic and all events of this `EventType`
+        will also be removed.
+        It's important to note that topic deletion in Kafka can take some time, so before it is completelly removed
+        it will not be possible to create an `EventType` with the same name as removed one had (you will get 409
```
rcillo: Are you sure it's gonna return a 409? Because the 409 is a result of a duplicated key in the database. If you already removed the entry from the database, it might have trouble to create the topic but it's going to return a 500.

v-stepanov: @rcillo 
No, I mean that will make it return 409 in that case :)

rcillo: Ah, ok. Maybe we need to improve the controller code a little bit in order to handle specific failures in kafka topic creation. It's doable. :+1:

PR #160, opened 2016-03-07
Aruha 120 per eventtype metrics

tmuehl:
Introduced the following metrics per event type:
* Timer for successful publishing (including counter)
* Counter for the number of events failed to publish
* Counter for the current number of consumers

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
@@ -41,6 +41,8 @@ import static org.zalando.problem.spring.web.advice.Responses.create;
 public class EventPublishingController {
 
     private static final Logger LOG = LoggerFactory.getLogger(EventPublishingController.class);
+    public static final String SUCCESS_METRIC_NAME = "success";
+    public static final String FAILED_METRIC_NAME = "failed";
 
     private final TopicRepository topicRepository;
     private final EventTypeRepository eventTypeRepository;
@@ -67,7 +69,7 @@ public class EventPublishingController {
             final EventType eventType = eventTypeRepository.findByName(eventTypeName);
 
             try {
-                final Timer successfullyPublishedTimer = metricRegistry.timer(metricNameFor(eventTypeName, "success"));
+                final Timer successfullyPublishedTimer = metricRegistry.timer(metricNameFor(eventTypeName, SUCCESS_METRIC_NAME));
                 final Timer.Context successfullyPublishedTimerContext = successfullyPublishedTimer.time();
 
                 final JSONObject eventAsJson = parseJson(event);
@@ -80,7 +82,7 @@ public class EventPublishingController {
 
                 return status(HttpStatus.CREATED).build();
             } catch (Exception e) {
-                metricRegistry.counter(metricNameFor(eventTypeName, "failed")).inc();
+                metricRegistry.counter(metricNameFor(eventTypeName, FAILED_METRIC_NAME)).inc();
                 throw e;
             }
 
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
@@ -1,5 +1,7 @@
 package de.zalando.aruha.nakadi.controller;
 
+import com.codahale.metrics.MetricRegistry;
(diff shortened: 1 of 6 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/160" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -61,14 +68,23 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName, @Reque
         try {
             final EventType eventType = eventTypeRepository.findByName(eventTypeName);
 
-            final JSONObject eventAsJson = parseJson(event);
+            try {
+                final Timer successfullyPublishedTimer = metricRegistry.timer(metricNameFor(eventTypeName, SUCCESS_METRIC_NAME));
```
v-stepanov: I'm not sure about this place. Will it create object with every call? Can we create it only one time?

tmuehl: It will create a timer once and the MetricRegistry will return the same timer the next time. And I am not sure, as well.

We are measuring per event type. That's why the event type has to be taken into account and we cannot have one timer in the class. We would need a timer per event type. One possibility is to cache the timers. I was considering this but came to the conclusion that this is premature optimisation. Let's first try this out and improve it if necessary.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -71,7 +75,13 @@ public StreamingResponseBody streamEvents(
             final NativeWebRequest request, final HttpServletResponse response) throws IOException {
 
         return outputStream -> {
+
+            Counter consumerCounter = null;
+
             try {
+                consumerCounter = metricRegistry.counter(metricNameFor(eventTypeName, CONSUMERS_COUNT_METRIC_NAME));
+                consumerCounter.inc();
```
v-stepanov: I'm not sure if it should be increased here, because at this point the stream is not really started yet. There are still some checks to be performed. But that is of course a point to argue what is considered to be a started stream.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -61,14 +68,23 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName, @Reque
         try {
             final EventType eventType = eventTypeRepository.findByName(eventTypeName);
```
rcillo: Why isn't this operation measured, as well?

tmuehl: I don't want to mess up the stats with metrics for event types that doesn't exist. But you are right that is an important call regarding performance and should be measured.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
tmuehl: I don't care. If it fails because of checks it will anyway remove that client immediately again. At this point it's a "potential" additional consumer.

src/test/java/de/zalando/aruha/nakadi/controller/EventStreamControllerTest.java
```diff
@@ -277,6 +287,57 @@ public void whenExceptionIsThrownThenInternalServerError() throws NakadiExceptio
...
+        // "connect" to the server
+        final StreamingResponseBody responseBody = controller.streamEvents(TEST_EVENT_TYPE, 0, 0, 0, 0, 0, null,
+                requestMock, responseMock);
+
+
+        final LinkedList<Thread> clients = new LinkedList<>();
+        final Counter counter = metricRegistry.counter(metricNameFor(TEST_EVENT_TYPE, EventStreamController.CONSUMERS_COUNT_METRIC_NAME));
+
+        // create clients...
+        for (int i=0; i<3; i++) {
```
rcillo: I have nothings against the number 3, but I feel that only one client would be enough to test it. I'm not suggesting that you should change, but just would like to share my thoughts on this subject.

tmuehl: Why one if you can have multiple? And I really want to test parallelism.

PR #161, opened 2016-03-07
Aruha 108 category aware validation

rcillo:
This PR adds the idea of an effective schema. An effective schema is an schema that is aware of categories.

If the event type is of business category, then we should expect it to have "metadata" in the root of the every event.

If from data category, then we should expect to find `data_op`, `data_type` and `data` at the root. The later should contain the original `schema` as a value.

Do you think we should be strict about what fields are allowed at the root object of a data event?

If so, I could add an `additionalProperties: false` to it.

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
+++ b/src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
@@ -31,7 +31,7 @@ public class EventBodyMustRespectSchema extends ValidationStrategy {
     @Override
     public EventValidator materialize(final EventType eventType, final ValidationStrategyConfiguration vsc) {
 
-        JSONObject effectiveSchema = EventValidation.effectiveSchema(eventType);
+        final JSONObject effectiveSchema = EventValidation.effectiveSchema(eventType);
 
         final JSONSchemaValidator defaultSchemaValidator = new JSONSchemaValidator(effectiveSchema);
 
--- a/src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
+++ b/src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
@@ -31,8 +31,10 @@ public class EventBodyMustRespectSchema extends ValidationStrategy {
     @Override
     public EventValidator materialize(final EventType eventType, final ValidationStrategyConfiguration vsc) {
 
-        final JSONSchemaValidator defaultSchemaValidator = new JSONSchemaValidator(
-                new JSONObject(eventType.getSchema().getSchema()));
+        JSONObject effectiveSchema = EventValidation.effectiveSchema(eventType);
+
+        final JSONSchemaValidator defaultSchemaValidator = new JSONSchemaValidator(effectiveSchema);
+
         if (vsc.getAdditionalConfiguration() == null) {
             return defaultSchemaValidator;
         }
@@ -151,10 +153,8 @@ class JSONSchemaValidator implements EventValidator {
     private static final Logger LOG = LoggerFactory.getLogger(JSONSchemaValidator.class);
 
     private final Schema schema;
-    private final JSONObject effectiveSchema;
 
     public JSONSchemaValidator(final JSONObject effectiveSchema) {
-        this.effectiveSchema = effectiveSchema;
         schema = SchemaLoader.load(effectiveSchema);
     }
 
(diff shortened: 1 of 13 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/161" target="_blank">ReviewNinja

tmuehl:
@rcillo Yes, I think, we should be strict. There is no reason why somebody should add random properties that are not in the schema.

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -1,39 +1,75 @@
...
+        return schema;
     }
 
-    public static EventTypeValidator lookup(final EventType eventType) {
-        return eventTypeValidators.get(eventType.getName());
+    private static void addToRequired(final JSONObject schema, final String[] toBeRequired) {
+        final Set<String> required = new HashSet<>(Arrays.asList(toBeRequired));
+
+        for(int i = 0; i < schema.getJSONArray("required").length(); i++) {
+            required.add(schema.getJSONArray("required").getString(i));
```
tmuehl: You are getting the JSONArray two times in each iteration. You can extract it as a variable.

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -1,39 +1,75 @@
...
+        }
+
+        schema.put("required", required);
```
tmuehl: Can't you add the toBeRequired fields directly to the json instead of creating a new array, adding everything and overriding the existing array?

tmuehl: Ah, got it. To ensure no entry is duplicated.

rcillo: I did it this way to avoid duplicating entries. For example, if the user of this method asks to add "metadata" but "metatada" was already there, then it's better to it this way.

src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
```diff
@@ -31,8 +31,10 @@
     @Override
     public EventValidator materialize(final EventType eventType, final ValidationStrategyConfiguration vsc) {
 
-        final JSONSchemaValidator defaultSchemaValidator = new JSONSchemaValidator(
-                new JSONObject(eventType.getSchema().getSchema()));
+        JSONObject effectiveSchema = EventValidation.effectiveSchema(eventType);
```
tmuehl: Maybe, we should check, in the case of a business event, that the schema provided by the user doesn't contain a metadata field.

rcillo: I agree with you that there should be a check. But I would argue that at this point, there is no useful feedback we could provide to the user.

I think the check should be done on the registry of the event type. Since we did not had this sort of checking before, this would require us to write a "clean up" routine in the database, as well. So I thought about not doing anything in this task. Wdyt @v-stepanov @CyberDem0n @jkliff ?

rcillo:
One thing worth mentioning is that I chose not to persist the effective schema. Instead, it will be generated on demand. Since we have caching, this should not add any significant overhead.

From my experience it's preferable to generate values on demand, if possible. Persisting the effective schema could be lead to many issues. Let's suppose there's a bug in the effective schema generation. If we stored it in the database, we would be required to implement an update routine to fix all existing values (extra maintenance effort).

Another issue is the distance between generation and usage of the effective schema. It makes things harder to understand since it would be generated during persistence, which is far from the usage context, event publishing.

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
rcillo: :+1: done https://github.com/zalando/nakadi/commit/c0946c933b4c5c611657d53d918c896527f22cfc

src/main/java/de/zalando/aruha/nakadi/validation/EventBodyMustRespectSchema.java
rcillo: Done https://github.com/zalando/nakadi/commit/900258233dd340ef8bed66c1d7e35847c3ceaca3

v-stepanov: agree with Tobias

v-stepanov:
looks good to me :shipit:

PR #162, opened 2016-03-07
Aruha 126 event type deletion

v-stepanov:
@rcillo @tmuehl @jkliff @CyberDem0n please review


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/162" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -351,6 +353,54 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+    delete:
+      tags:
+        - schema-registry-api
+      security:
+        - oauth2: [nakadi.admin]
+      description: |
+        Deletes an `EventType` identified by its name. The underlying Kafka topic and all events of this `EventType`
```
tmuehl: I would rather speak of the "underlying" infrastructure instead of explicitly talking about Kafka.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -65,6 +66,12 @@ public EventTypeController(EventTypeRepository eventTypeRepository, TopicReposit
         }
 
         try {
+            // can happen if EventType already exists or if topic deletion
+            // was not yet processed after EventType deletion
+            if (topicRepository.topicExists(eventType.getName())) {
```
rcillo: Another approach would be to change the code for `topicRepository.createTopic` as follows:

```java
@Override
    public void createTopic(final String topic, final int partitionsNum, final int replicaFactor,
                            final long retentionMs, final long rotationMs) throws TopicCreationException {
        ZkUtils zkUtils = null;
        try {
            final String connectionString = zkFactory.get().getZookeeperClient().getCurrentConnectionString();
            zkUtils = ZkUtils.apply(connectionString, settings.getZkSessionTimeoutMs(),
                    settings.getZkConnectionTimeoutMs(), false);

            final Properties topicConfig = new Properties();
            topicConfig.setProperty("retention.ms", Long.toString(retentionMs));
            topicConfig.setProperty("segment.ms", Long.toString(rotationMs));

            AdminUtils.createTopic(zkUtils, topic, partitionsNum, replicaFactor, topicConfig);
        } catch (TopicExistsException e) {
            throw new DuplicatedEventTypeNameException("EventType with name " + topic +
                        " already exists (or wasn't completely removed yet)");
        } catch (Exception e) {
            throw new TopicCreationException("Unable to create topic", e);
        }
        finally {
            if (zkUtils != null) {
                zkUtils.close();
            }
        }
    }
```


Then you would not be required to change the controller code. This can be handy if we are using the repository outside of the controller context.

api/nakadi-event-bus-api.yaml
v-stepanov: But we anyway expose that we use kafka in out swagger file:
`The single currently supported messaging infrastructure is Kafka`
Don't see much sense to hide that.

rcillo: I think that "underlying infrastructure" may sound too abstract for users. I guess they are pretty much aware that it's Kafka.

rcillo:
:shipit: if you don't think the changes suggested are worth it https://github.com/zalando/nakadi/pull/162#discussion_r55334683

v-stepanov:
@rcillo That was a good point, I currently working on implementing that

api/nakadi-event-bus-api.yaml
tmuehl: I my eyes, they shouldn't. We may change this at any point in time. It might help people who are familiar with Kafka to understand some of the Nakadi principles, but in general Nakadi is not about Kafka.

The thing they should be aware of in terms of deletion is that they delete the topic including all events permanently. That physically a Kafka topic is deleted is implementation details.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
v-stepanov: good point. Implemented it with this commit https://github.com/zalando/nakadi/commit/7cb7a5bf3693e1bc9db7ba4666c400996da2274c

PR #164, opened 2016-03-08
ARUHA-113 initial work on batching api

jkliff:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/164" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -416,28 +416,35 @@ paths:
       security:
         - oauth2: ['nakadi_event_stream.write']
       description: |
-        Publishes one `Event` of this `EventType`. (**Note** *Format and response semantics of publishing of
-        batches of Events will be added in the short term.*)
+        Publishes a batch `Event`s of this `EventType`. All items must be of the same EventType.
```
valgog: a batch **of** Events

valgog: I think it is not only the *same* envettype but also the event type specified by the `{name}`

api/nakadi-event-bus-api.yaml
```diff
@@ -416,28 +416,35 @@ paths:
...
+        in the first case of failure. If the offending validation rule
+        provides information about the violation it will be included in the `BatchItemResponse`.
         If the `EventType` defines schema validation it will be performed at this moment.
 
         1. Once the validation succeeded, the content of the Event is updated according to the enrichment
         rules in the order the rules are defined in the `EventType`.
         No preexisting value might be changed (even if added by an enrichment rule). Violations on this will
-        force the immediate **rejection** of the Event with a 409 Conflict status. The invalid overwrite
-        attempt will be included in the response's Problem object.
+        force the immediate **rejection** of the Event. The invalid overwrite
```
valgog: only the Event that is affected? What about the next events in the "stripe"?

api/nakadi-event-bus-api.yaml
```diff
@@ -1270,3 +1281,43 @@ definitions:
...
+  BatchItemResponse:
+    description: |
+      A status corresponding to one individual Event's publishing attempt.
+    properties:
+      eid:
+        type: string
+        format: uuid
+        description: |
+          eid of the corresponding item.
+      publishing_status:
```
v-stepanov: I think the outcome of our meeting had a little bit different structure.

pushed:
 - yes
 - no

status:
 - none
 - checked
 - partitioned
 - enriched
 - published
 - failed (or maybe even several status codes for failed, showing on which step it failed)

api/nakadi-event-bus-api.yaml
```diff
@@ -1270,3 +1281,43 @@ definitions:
...
+        type: string
+        enum:
+          - FAILED
+          - SUBMITTED
+          - ABORTED
+        description: |
+          Indicator of the submission of the Event. SUBMITTED indicates successful submission, including commit on he underlying broker. FAILED indicates the message submission was not possible and can be resubmitted if so desired.
+
+          ABORTED indicates that no further attempt at publishing the affected item was attempted because of a failure on another item.
```
valgog: Do we need to include information about the source of the problem in case of the `ABORTED`?

api/nakadi-event-bus-api.yaml
```diff
@@ -1270,3 +1281,43 @@ definitions:
...
+      step:
+        type: string
+        enum:
+          - NONE
```
valgog: `NONE`? what is the meaning of `NONE` in this context? Is it `PUBLISHED`? `COMMITTED`?

api/nakadi-event-bus-api.yaml
jkliff: - about yes/no: failed/submitted is the same; and there we had agreed on not being boolean. 
- about the step: is the same, but without failed. failed is no step, but an outcome. the names used are just more in line with what the api already specifies. the semantic is the same.

valgog: This could be just `status`? or can there be other stati?

api/nakadi-event-bus-api.yaml
jkliff: - the status is actually changed only for that event (if event i is invalid, this does not make even i+1 invalid).
- the impact on the streaming is explained in a paragraph below.

api/nakadi-event-bus-api.yaml
v-stepanov: - failed/submitted is also fine for me, but you also have here ABORTED value. I think the main reason why we wanted a separate 'publish' field - is to be able to have a simple answer for a question "Was event published or not?"
-  ok, so for failed event we will also show the last step it passed, right?

api/nakadi-event-bus-api.yaml
jkliff: none means none of the others. would in practice mean that not even the validation (first step) was yet attempted for this event.

api/nakadi-event-bus-api.yaml
jkliff: i believe would be great to do so, but i would leave details to the textual description (ABORTED, "Event publishing aborted due to validation failure on event foo-bar-1bc-123" )

api/nakadi-event-bus-api.yaml
```diff
@@ -1260,3 +1272,51 @@ definitions:
         description: Documentation for the enrichment.
     required:
       - name
+
+
+  BatchItemResponse:
+    description: |
+      A status corresponding to one individual Event's publishing attempt.
+    properties:
+      eid:
```
rcillo: Since this is user generated data, it might be absent. Isn't it a problem?

api/nakadi-event-bus-api.yaml
```diff
@@ -1260,3 +1272,51 @@ definitions:
...
+        type: string
+        enum:
+          - FAILED
+          - SUBMITTED
+          - ABORTED
+        description: |
+          Indicator of the submission of the Event within a Batch.
+          - SUBMITTED indicates successful submission, including commit on he underlying broker.
+          - FAILED indicates the message submission was not possible and can be resubmitted if so desired.
+          - ABORTED indicates that no further attempt at publishing the affected item was attempted because
```
rcillo: I understand the definition, but what do you think of "indicates that this item has not been processed due to a failure on a previous item in the batch".

api/nakadi-event-bus-api.yaml
```diff
@@ -1260,3 +1272,51 @@ definitions:
...
+      step:
+        type: string
+        enum:
+          - NONE
+          - VALIDATION
+          - ENRICHMENT
+          - PARTITIONING
+          - PUBLISHING
+        description: |
+          Indicator of the last step in the pulbishing process this Event reached.
```
rcillo: Inclusive or exclusive? For example, if it's inclusive, it means that an item with step VALIDATION had a validation issue. If it's exclusive, then it wasn't even validated.

From what I understand, this is an "inclusive" semantics. Wdyt about making it even more obvious "this Event reached, inclusively"?

rcillo:
I have one last question, which is implementation specific. What's the maximum batch size, in bytes, that we are going to accept?

rcillo:
:shipit:

PR #168, opened 2016-03-09
Aruha 114 metadata validation

rcillo:
In this PR I implemented the EventMetadata validation using json-schema capabilities.

I could not use `date-time` formats since they are not implemented by the json-schema validator library of choice. Instead, I just implemented a simple regex, which is obviously not compliant with the RFC but provides enough safety to make sure values will respect some date formatting.

Another approach would be to use some Java code to check the format.

I also deleted some tests written previously since they were checking for an intermediate structure and not the actual validation functionality.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/168" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -51,10 +49,33 @@ private static JSONObject wrapSchemaInData(final JSONObject schema) {
...
-        schema.getJSONObject("properties").put("metadata", new JSONObject("{\"type\": \"object\"}"));
+        final JSONObject metadata = new JSONObject();
+        final JSONObject metadataProperties = new JSONObject();
+
+        final JSONObject uuid = new JSONObject("{\"type\": \"string\", \"pattern\": \"^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$\"}");
+        final JSONObject eventTypeString = new JSONObject("{\"type\": \"string\", \"enum\": [\"" + eventType.getName() + "\"]}");
+        final JSONObject string = new JSONObject("{\"type\": \"string\"}");
+
+        // We know this is not rfc3339. We think this is enough by now
+        final String rfc3339 = "^[0-9]{4}-?[0-9]{2}-?[0-9]{2}T?[0-9]{2}:?[0-9]{2}:?[0-9]{2}(.[0-9]+)?(Z|[+-][0-9]{2}:?[0-9]{2})$";
```
v-stepanov: I think it would be really great if in our API for `occurred_at` field we specify an example so that our users will know which date-format they can specify.

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -51,10 +49,33 @@ private static JSONObject wrapSchemaInData(final JSONObject schema) {
...
+        final JSONObject dateTime = new JSONObject("{\"type\": \"string\", \"pattern\": \"" + rfc3339 + "\"}");
+
+        metadataProperties.put("eid", uuid);
+        metadataProperties.put("event_type", eventTypeString);
+        metadataProperties.put("occurred_at", dateTime);
+        metadataProperties.put("parent_eid", uuid);
+        metadataProperties.put("root_eid", uuid);
+        metadataProperties.put("flow_id", string);
```
v-stepanov: I'm missing 'received_at' field here

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -51,10 +49,33 @@ private static JSONObject wrapSchemaInData(final JSONObject schema) {
...
+        final JSONObject string = new JSONObject("{\"type\": \"string\"}");
+
+        // We know this is not rfc3339. We think this is enough by now
+        final String rfc3339 = "^[0-9]{4}-?[0-9]{2}-?[0-9]{2}T?[0-9]{2}:?[0-9]{2}:?[0-9]{2}(.[0-9]+)?(Z|[+-][0-9]{2}:?[0-9]{2})$";
+        final JSONObject dateTime = new JSONObject("{\"type\": \"string\", \"pattern\": \"" + rfc3339 + "\"}");
+
+        metadataProperties.put("eid", uuid);
+        metadataProperties.put("event_type", eventTypeString);
+        metadataProperties.put("occurred_at", dateTime);
+        metadataProperties.put("parent_eid", uuid);
```
v-stepanov: in the API it is called `parent_eids` and it is an array of uuid's

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -51,10 +49,33 @@ private static JSONObject wrapSchemaInData(final JSONObject schema) {
...
+        metadataProperties.put("root_eid", uuid);
```
v-stepanov: we don't have this one in our API

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
jkliff: I don't think there's any valid format where the - is optional. (javascript's own or iso 8601)

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
rcillo: This is why https://github.com/zalando/nakadi/commit/bb462e54b4e9f03faf0ab6912887dd7e6c5d892b

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
rcillo: Actually I would avoid discussing details of the each ISO or RFC. If we want to make dashes required, lets do it.

rcillo: done https://github.com/zalando/nakadi/commit/58fcdfd12f82b9cafc014af50801aebdb6332aaa

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
rcillo: done https://github.com/zalando/nakadi/commit/eb7caed572d30eb3aba8f0fb595cb3ee6e4a264e

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
rcillo: You're right. It was based on an outdated version of the api. Removed in https://github.com/zalando/nakadi/commit/c4fe0aa7ffb3139d9abcfb387b9f83e3ebe52c66

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
v-stepanov: You create `arrayOfUUIDs` but you don't use it.

rcillo: https://github.com/zalando/nakadi/commit/f8db717b34b16aa985c2ee48254814a549684a27

v-stepanov:
:shipit:

rcillo:
I added on more commit https://github.com/zalando/nakadi/commit/aa5c282b6f3b4e7adfb46fff353bdb0954366042 to make it easier to maintain in the future. Parsing json strings seems awkward.

v-stepanov:
:shipit:

PR #172, opened 2016-03-11
aruha-114 improve api by providing date-time examples and check for rfc3339

rcillo:
As suggested during validation, I'm adding some datetime examples.

Currently, `date-time` is not fully implemented by the library of choice, since it's specified as optional in the json-schema http://json-schema.org/latest/json-schema-validation.html#anchor106

But, if implemented, it should follow rfc3339 as stated in http://json-schema.org/latest/json-schema-validation.html#anchor108

We currently just validate using a regex which accepts a subset of rfc3339. So, this way, if we start to accept the complete rfc in the future, this should be backward compatible.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/172" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/validation/MetadataValidator.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Optional;
+
+public class MetadataValidator implements EventValidator {
+    private static final String DATETIME_FORMAT_STRING = "yyyy-MM-dd'T'HH:mm:ssXXX";
+    private static final String DATETIME_FORMAT_STRING_SECFRAC = "yyyy-MM-dd'T'HH:mm:ss.SSSXXX";
```
v-stepanov: I'm not really sure, these two format-strings - do they support all possible cases of the rfc?

rcillo: I couldn't think about a counter example.

v-stepanov:
:shipit:

api/nakadi-event-bus-api.yaml
```diff
@@ -888,6 +888,7 @@ definitions:
               Timestamp of creation of the Event generated by the producer.
             type: string
             format: date-time
+            example: '1996-12-19T16:39:57-08:00'
```
valgog: I would use the +01:00 for the time zone 😀

Apropos, does API guild allow non-UTF times?

rcillo: I'm not sure about the API guild, I'll ask then, but swagger supports rfc 3339.

rcillo: From the guidelines:

"Timestamps are passed in UTC-related format via APIs and should be stored in UTC (i.e. without any offset information). Localization based on UTC should be done by the services that provide user interfaces, if required."

It's not clear for me that we should not accept anything that's not UTC. I would interpret that we should accept it and then manipulate it in order to store an UTC version of the value.

Wdyt?

rcillo: https://docs.google.com/document/d/1Btz_xm3cpwG_N76qe82klQLFGjRW1e0ZqS60RTaw4Jw/edit#heading=h.ucw2lhbf6v6b

valgog: I would be very careful with time parsing. There are so many corner cases, that is is better to work with special library like JODA

Have loom here:https://rafaelsantosbra.wordpress.com/tag/rfc-3339/

rcillo: You are right. Date and time formatting is no simple task. Thanks for the reference. :+1: 

I checked JODA time and they use a slightly different format http://joda-time.sourceforge.net/apidocs/org/joda/time/format/ISODateTimeFormat.html#dateTime()  and http://joda-time.sourceforge.net/apidocs/org/joda/time/format/ISODateTimeFormat.html#dateTimeNoMillis()

The main difference is that they would format UTC as Z instead of +00:00.

But lets use Joda, so we do not have to be aware of the format :+1:

rcillo: Unfortunately, it looks like that Joda implementation is more lenient than ours. I tested with the string `"1996-03-15T16:39:57+0100"` and it was accepted by their implementation.

But the RFC  states that the `:` is mandatory. So I'll stay with our implementation, by now. We should be comply with RFC 3339 and not ISO 8601, which is not intended "for internet".

PR #175, opened 2016-03-15
Feature/aruha 118

jkliff:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/175" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -1210,6 +1219,22 @@ definitions:
...
       - schema
 
+  EventTypeStatistics:
+    type: object
+    description: |
+      Operational statistics for an EventType. This data is generated by Nakadi based on the runtime and
+      might be used to guide changes in internal parameters.
+
+    properties:
+      expected_write_rate:
```
rcillo: Maybe we should be more specific about the extension of the value. Is it "events/second" or "events/milisecond"? The same applies to the other fields.

rcillo: Except for the non dimensional ones.

bocytko: @rcillo how about two fields: `time` (long) and `unit` (like Java's TimeUnit). This should cover all cases.

rcillo: Looks like a good idea. It would be easier to parse too.

jkliff: This is not finished.
I don't see the need to offer such flexibility at this time, but obviously further descriptions about this are important and will be added.

api/nakadi-event-bus-api.yaml
```diff
@@ -1228,12 +1229,28 @@ definitions:
...
+          Write rate for events of this EventType. This rate encompasses all producers of this EventType for
+          a Nakadi cluster.
+
+          Measured in kb/minutes.
+
+      message_size:
         type: integer
-      expected_read_parallelism:
+        description: |
+          Average message size for each Event of this EventType. Includes in the count the whole serialized
```
rcillo: Wdyt about "Average size of the payload in bytes"?

jkliff: in your terminology, what exactly does the "payload" includes?

api/nakadi-event-bus-api.yaml
```diff
@@ -1228,12 +1229,28 @@ definitions:
...
+      read_parallelism:
         type: integer
-      expected_write_parallelism:
+        description: |
+          Amount of parallel consumers to this EventType.
+
+      write_parallelism:
         type: integer
+        description: |
+          Amount of parallel writes to this EventType.
```
rcillo: writes or writers?

api/nakadi-event-bus-api.yaml
rcillo: Forget about it, I just checked that one could understand this as been only the contents of the "data" property.

jkliff: exactly this ambiguity I wanted to avoid. 
to be clear, my intention here is to allow the same value to be produced by internal measurements of nakadi -- blind to the content and interested really on the size of what would be actually written to disk -- and the client. 
this means in practice all of the data *after* enrichment of metadata. 

off course, for setting the default statistics (by the client) can require some extra hops for him to get the value right, but we probably can't measure on our side any differently (nor it would be useful) and we can provide later some tooling for the client to get this numnber right.

rcillo: As I said previously, forget about it.

rcillo:
I have no further comments :+1: 

:shipit:

PR #178, opened 2016-03-17
Aruha 72 support definition of partitioning strategy

v-stepanov:
@rcillo @jkliff @tmuehl @CyberDem0n Please review


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/178" target="_blank">ReviewNinja

src/test/java/de/zalando/aruha/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -140,6 +145,45 @@ public void whenPostWithNoSchemaSchemaThenReturn422() throws Exception {
...
+
+        final Problem expectedProblem = invalidProblem("partition_resolution_strategy.name", "may not be null");
+
+        postEventType(eventType)
+                .andExpect(status().isUnprocessableEntity())
+                .andExpect(content().string(matchesProblem(expectedProblem)));
+    }
+
+    @Test
+    public void whenPostWithKnownPartitioningStrategyThenReturn201() throws Exception {
```
rcillo: I was wondering if this test isn't already covered by the existing 201 test scenario. Maybe we should just add the patition strategy to the default event type instantiated by `buildDefaultEventType`. Wdyt?

v-stepanov: I would not like to do that, because we agreed that partitioning strategy should not be a required field. So I think default event-type should contain only the minimum amount of fields needed.

rcillo: :+1:

v-stepanov: So In general this test just checks that event-type creation with partitioning strategy defined is performed correctly.

rcillo:
Are you going to change the default number of partitions to 1 in this PR?

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -67,12 +73,13 @@ public EventTypeController(final EventTypeRepository eventTypeRepository, final
...
+            validatePartitioningStrategy(eventType);
             eventTypeRepository.saveEventType(eventType);
             topicRepository.createTopic(eventType.getName());
             return status(HttpStatus.CREATED).build();
-        } catch (final InvalidEventTypeException e) {
-            return create(e.asProblem(), nativeWebRequest);
-        } catch (final DuplicatedEventTypeNameException e) {
+        } catch (final InvalidEventTypeException | NoSuchPartitioningStrategyException |
+                DuplicatedEventTypeNameException e) {
+            LOG.warn(e.getMessage(), e);
```
rcillo: What's the semantics of a warning?

src/main/java/de/zalando/aruha/nakadi/domain/PartitionResolutionStrategy.java
```diff
@@ -0,0 +1,32 @@
...
+                                       @JsonProperty("doc") @Nullable final String doc) {
+        this.name = name;
+        this.doc = doc;
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    @Nullable
```
rcillo: Do you have to repeat the annotation used in the definition of the property here as well?

src/main/java/de/zalando/aruha/nakadi/exceptions/NoSuchPartitioningStrategyException.java
```diff
@@ -0,0 +1,9 @@
+package de.zalando.aruha.nakadi.exceptions;
+
+public class NoSuchPartitioningStrategyException extends UnprocessableEntityException {
+
+    public NoSuchPartitioningStrategyException(String message) {
```
rcillo: final :grin:

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
v-stepanov: What do you mean?

v-stepanov:
@rcillo yes! thanks for reminding. Now I set default partitions numbers to 1.

rcillo:
Maybe we could cache the result of `resolvePartition`?

Since it does a call to `listPartitionNames` which is expensive.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
rcillo: I'm confused about when to use warning and when to use debug, for example. Because this seems to something for debug as I understand it. So I would like to know your opinion.

src/main/java/de/zalando/aruha/nakadi/domain/PartitionResolutionStrategy.java
v-stepanov: I think it will be useful for people that will use the method to see that it can return null

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
v-stepanov: I think we never discussed that in our team.
But I see your point. It should not really be a warning as it is just usual work of an application. So Debug or Info will fit better here. Will fix that.

src/main/java/de/zalando/aruha/nakadi/exceptions/NoSuchPartitioningStrategyException.java
v-stepanov: done!

v-stepanov:
@rcillo 
About caching. I think caching the result of `resolvePartition` is not that interesting as it takes event as an input parameter (so to cache the result of specific event is not really interesting).
But caching the result of `listPartitionNames ` can make sense. Will check that now.

rcillo:
@v-stepanov sure. You're right. It makes much more sense to cache reusable info only.

v-stepanov:
After discussion with @rcillo we agreed that we will not implement the caching now because it's already kind of cached by Kafka client. So we will come to this point if it really will be a performance issue.

rcillo:
@v-stepanov one last question I have is if do we need to do something with already existing event types and events.

Since we changed the default number of partitions and we'll be acting as if it's true by default, maybe it could introduce some issues or change the expected behaviour of existing event type publishing.

v-stepanov:
@rcillo 
In general there's really an issue with event-types which have already 8 partitions and don't have partitioning strategy defined. 
As we planned to have "random" as a default one this can be a problem (it will work well only for event-types with one partition)
Possible solution can be to have a specific case (if partitioning strategy is not present and there are more than one partition - push to 0 as it was before). Or we can just Inform our users that they should re-create event-types (and I prefer this one, as we will not have to deal with legacy in this case)

rcillo:
I agree that warning users would be enough. But then we should give then some time to make adjustments which could slow down the release process. But anyway, I think that adding code to handle a legacy scenario is undesirable.

:shipit:

PR #179, opened 2016-03-17
aruha-113 add multi-status code 207

rcillo:
In order to more precisely communicate partial success scenarios, we are
introducing 207 Multi-status code as a possible http response code.

In case all events have been published successfully, then we just return
200 with empty body, otherwise we provide detailed status information in
the form of an array of BatchItemResponse's.

We considered using 422 as a response answer but it just doesn't feel correct
since we could have a mix of successes and failures in the batch.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -468,7 +468,9 @@ paths:
           required: true
       responses:
         '200':
-          description: Ok
+          description: All events in the batch have been successfully published.
+        '207':
+          description: At least one event has failed to be submitted.
           schema:
             type: array
             items:
@@ -487,10 +489,6 @@ paths:
               $ref: '#/definitions/Problem'
         '405':
           description: Not allowed.
-          schema:
-              $ref: '#/definitions/Problem'
-        '409':
-          description: Conflict.
           schema:
               $ref: '#/definitions/Problem'
         '500':
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -467,7 +467,7 @@ paths:
         '200':
           description: All events in the batch have been successfully published.
         '207':
-          description: At least one event has failed to be submitted.
+          description: At least one event has failed to be submitted. The batch might be partially submitted.
           schema:
             type: array
             items:
@@ -488,6 +488,12 @@ paths:
           description: Not allowed.
           schema:
               $ref: '#/definitions/Problem'
+        '422':
+          description: At least one event failed to be validated, enriched or partitioned. None were submitted.
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/BatchItemResponse'
         '500':
           description: Server error
           schema:
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -435,9 +435,6 @@ paths:
         1. The incoming Event's relative ordering is evaluated according to the rule on the
         `EventType`. Failure to evaluate the rule will **reject** the Event.
 
-        The response in successful operation is an array of same length as the input, including a
-        `BatchItemResponse` corresponding to each item's status.
-
         Given the batched nature of this operation, any violation on validation or failures on enrichment or
         partitioning will cause the whole batch to be rejected, i.e. none of its elements are pushed to the
         underlying broker.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/179" target="_blank">ReviewNinja

v-stepanov:
I still think that 422 is a an appropriate answer when not a single event was published. So that we can have
- 200 when everything successful
- 207 when some are published and some are failed
- 422 when no event is published successfully

@rcillo @jkliff what do you guys think?

jkliff:
@v-stepanov I don't think so, but maybe I am not seeing why you think so. Do you mind explaining the reasoning behind it?

rcillo:
@v-stepanov I think it makes sense if all of the failures are due to client error. The same way, if all of the events fail due to kafka being down, I think i could be 503, for example.

This is an important discussion. I think that the answer to @v-stepanov question will have as consequences the removal of other statuses, like 503, for example.

v-stepanov:
@jkliff @rcillo 
Well, for me everything that starts with 2xx is considered to be a "successful", and if everything failed - there's not much success in that. However in the documentation of 207 it says that you can even use in in failure situations. (https://httpstatuses.com/207)
`The response MAY be used in success, partial success and also in failure situations.`
So in general you are right
.
But I also think that consulting with API guild can also be useful in this case.

rcillo:
@v-stepanov I just realised that it can be useful to reserve 422 for complete failure. For example, let's suppose there's an error on the desserialization of the RequestBody. This error is not "localized" to a single event and should result in a complete failure with 422.

So if we state that 422 should come together with a list of BatchItemResponse's then we have no way to express errors that are not specific to a single event.

tfrauenstein:
If think it depends on the failure type. I would 1st check all client side failures, like producing of event that does not match its type scheme. If there is any client failure event in the batch I would react "atomic" with client error status code (e.g. 422) plus return problem object array providing the details and no state change. If it is during server side state change processing phase, I would return either 200 for complete success or 207 for partial state change and success complemented with return problem object array providing the details.

jkliff:
Except for failures directly related to the underlying broker (i.e. after validating, etc), all issues are treated atomically (no state change). 
Any case of success and failure will always produce item-by-item responses. 
I agree with @v-stepanov and @tfrauenstein. In being a 4xx, 422 indicates that the opreation was not successful as a whole and this is a simple indication to the client that it wasn't correct. 
200 works as a simple indicator that the whole was successful and 207 serves any situation where some items only were not successful (server side failure in @tfrauenstein terms).

This fits directly with @v-stepanov's proposal. 

I support.

dehora:
Just so I understand, what would the return codes be in these cases posting a batch containing two events, A and B - 

- A passes the EventValidationStrategy, and B fails

- Both A and B fail the EventValidationStrategy

- A can be sent onwards to its partition, and B can't

- Both A and B can't be sent to their partitions

rcillo:
I added a new commit with 422 scenario https://github.com/zalando/nakadi/pull/179/commits/37c1efcd62ce76ea867e691ad6473dbde86b74d5

Answering @dehora questions:

> A passes the EventValidationStrategy, and B fails

Response status 422 and response body `[ BatchItemResponse A { step: VALIDATION, publishing_status: ABORTED }, BatchItemResponse B { step: VALIDATION, publishing_status: FAILED, detail: "the specifics of the validation error" } ]`

> Both A and B fail the EventValidationStrategy

Only A will fail, since it comes first in the batch.

Response status 422 and response body `[ BatchItemResponse A { step: VALIDATION, publishing_status: FAILED, detail: "the specifics of the validation error" }, BatchItemResponse B { step: NONE, publishing_status: ABORTED } ]`

> A can be sent onwards to its partition, and B can't

Response status 207 and response body `[ BatchItemResponse A { step: NONE, publishing_status: SUBMITTED }, BatchItemResponse B { step: PUBLISHING, publishing_status: FAILED, detail: "the specifics of the publishing error" } ]`

> Both A and B can't be sent to their partitions

Response status 207 and response body `[ BatchItemResponse A { step: PUBLISHING, publishing_status: FAILED, detail: "the specifics of the publishing error" }, BatchItemResponse B { step: PUBLISHING, publishing_status: FAILED, detail: "the specifics of the publishing error" } ]`

dehora:
@rcillo :+1: thanks

rcillo:
Thanks to @tfrauenstein and @dehora for the feedback. I'll merge this PR.

PR #180, opened 2016-03-18
Aruha 131 different partitioning strategies

v-stepanov:
@rcillo @jkliff @tmuehl @CyberDem0n 
please review


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/180" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -920,6 +920,12 @@ definitions:
               from the header into the metadata by Nakadi to avoid clients having to explicitly copy this.
             type: string
             example: 'JAh6xH4OQhCJ9PutIV_RYw'
+          partition:
+            description: |
+              This field indicates the partition to which the event should be published to. Used only for partitioning
```
jkliff: I'd replace 'Used only for ...' with 'Required when the partitioning strategy used is ...'

api/nakadi-event-bus-api.yaml
```diff
@@ -1246,7 +1258,15 @@ definitions:
     properties:
       name:
         type: string
-        description: Name of the strategy.
+        enum: ['hash', 'user_defined', 'random']
+        description: |
+          Name of the strategy. Possible values are:
+          - 'hash' will use the event field(s) defined in 'partitioning_key_fields' property of `EventType` as a source
+            for a hash function that will caclulate the partition where the event will be put.
```
rcillo: Maybe make it even more clear that "all events with the same value in this field will go to the same partition, and consequently be ordered".

src/main/java/de/zalando/aruha/nakadi/partitioning/RandomPartitioningStrategy.java
```diff
@@ -0,0 +1,32 @@
...
+import static de.zalando.aruha.nakadi.exceptions.ExceptionWrapper.wrapFunction;
+import static java.lang.Math.abs;
+
+public class RandomPartitioningStrategy implements PartitioningStrategy {
+
+    private static final Random random = new Random();
+
+    @Override
+    public String calculatePartition(final EventType eventType, final JSONObject event, final List<String> partitions) {
+        if (partitions.size() == 1) {
```
rcillo: It seems that this case is already covered by the code in the `else` body. Maybe we could just remove the `if`?

src/test/java/de/zalando/aruha/nakadi/partitioning/RandomPartitioningStrategyTest.java
```diff
@@ -0,0 +1,37 @@
...
+        final List<String> partitions = ImmutableList.of("a", "b", "c", "d", "e", "f", "g", "h");
+        final Set<String> resolvedPartitions = newHashSet();
+
+        final int numberOfRuns = 1000;
+        for (int run = 0; run < numberOfRuns; run++) {
+            final String resolvedPartition = strategy.calculatePartition(null, null, partitions);
+            resolvedPartitions.add(resolvedPartition);
+        }
+
+        assertThat(resolvedPartitions, equalTo(newHashSet(partitions)));
```
rcillo: The odd are very low, but it's not necessarily true. I think we don't need to test this since we rely on a battle tested java library for random generation.

I would ratter inject a mock and test that the code interacts with java library in a specific way (if possible, because it maybe is final).

Otherwise I would be ok with not testing this scenario at all.

rcillo: Or make this test deterministic by using the same seed. E.g, make sure that the result you get is the same every time.

src/main/java/de/zalando/aruha/nakadi/partitioning/RandomPartitioningStrategy.java
jkliff: +1

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -154,11 +157,24 @@ public EventTypeController(final EventTypeRepository eventTypeRepository, final
...
     }
 
-    private void validatePartitioningStrategy(final EventType eventType) throws NoSuchPartitioningStrategyException {
+    private void validatePartitioningStrategy(final EventType eventType) throws NoSuchPartitioningStrategyException,
+            InvalidEventTypeException {
         final PartitionResolutionStrategy partitioningStrategy = eventType.getPartitionResolutionStrategy();
-        if (partitioningStrategy != null && !partitionResolver.strategyExists(partitioningStrategy.getName())) {
-            throw new NoSuchPartitioningStrategyException("partitioning strategy does not exist: " +
-                    partitioningStrategy.getName());
+        if (partitioningStrategy != null ) {
```
jkliff: what if it is null? 
is the ever valid?

src/test/java/de/zalando/aruha/nakadi/partitioning/RandomPartitioningStrategyTest.java
v-stepanov: I calculated a little bit, this test can fail with probability
`0.000000000000000000000000000000000000000000000000000000000814972603061156`
Which is pretty low I believe :)

src/main/java/de/zalando/aruha/nakadi/partitioning/RandomPartitioningStrategy.java
v-stepanov: That is true, the only reason why I made this branch is to save some performance here (as the random will still be calculated even that only one value is possible)
If most of event-types will have only one partition - this will be called quite often.

src/test/java/de/zalando/aruha/nakadi/partitioning/RandomPartitioningStrategyTest.java
jkliff: I'd say: one shoudl test that: 
a) the data lands in the partition returned by the strategy. this is irrespective of the strategy or how it is implemented. a mock would do the trick and afaik this is covered.
b) for implementations with our logic, test the logic (user defined, hashed) deterministically. for the random case, i suggest something slightly different than what @rcillo suggests about the seed: treat the whole random generation as something you'd mock; say, have a randomNumberProvider that in reality is just Random.nextInt, but under test is mocked to be something you expect it to be, and than the partition computation is deterministic.

api/nakadi-event-bus-api.yaml
v-stepanov: Yes, makes sense. Done.

api/nakadi-event-bus-api.yaml
v-stepanov: Yes, good point. I added that.

api/nakadi-event-bus-api.yaml
```diff
@@ -920,6 +920,12 @@ definitions:
               from the header into the metadata by Nakadi to avoid clients having to explicitly copy this.
             type: string
             example: 'JAh6xH4OQhCJ9PutIV_RYw'
+          partition:
```
jkliff: I have some further editorial suggestions about the api that I would edit in your branch directly if you don't mind.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
v-stepanov: Yes, null is allowed here. Because we agreed that partitioning strategy should be not required and 
1) We will have 'random' strategy as a default together with:
2) We will have 1 partition for EventType by default (which is already there)

api/nakadi-event-bus-api.yaml
v-stepanov: sure, feel free to do that.

src/test/java/de/zalando/aruha/nakadi/partitioning/RandomPartitioningStrategyTest.java
v-stepanov: @rcilo @jkliff changed test, it now will only test that the random generator is used correctly; https://github.com/zalando/nakadi/pull/180/commits/ba718add035aba867081911b931a1f8c784619d5

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
v-stepanov: @jkliff changed it now as we agreed with you

api/nakadi-event-bus-api.yaml
rcillo: @jkliff are you going to make any further changes to this PR? Otherwise I would merge it in order to avoid conflicts, since I'm also changing the event publishing controller.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -154,12 +157,23 @@ public EventTypeController(final EventTypeRepository eventTypeRepository, final
...
             throw new NoSuchPartitioningStrategyException("partitioning strategy does not exist: " +
                     partitioningStrategy.getName());
         }
+        else if (HASH_STRATEGY.equals(partitioningStrategy.getName()) &&
+                eventType.getPartitioningKeyFields().isEmpty()) {
+            throw new InvalidEventTypeException("partitioning_key_fields field should be set for " +
+                    "partition resolution strategy 'hash'");
+        }
+        else if (USER_DEFINED_STRATEGY.equals(partitioningStrategy.getName()) &&
+                UNDEFINED.equals(eventType.getCategory())) {
```
tmuehl: Why is it forbidden for `undefined` category? We could treat the `metadata` field as a reserved one (even for `undefined`) and add it silently to the effective schema. Surely, if this is the case, event type registration should fail if the schema contains a `metadata` field.

rcillo: I also think it's reasonable. I can't remember why we decided to limit this functionality to Business and DataChange events. That would be easy to also provide it to Undefined using effect schema, as you just said.

api/nakadi-event-bus-api.yaml
```diff
@@ -1246,7 +1254,16 @@ definitions:
     properties:
       name:
         type: string
-        description: Name of the strategy.
+        enum: ['hash', 'user_defined', 'random']
```
tmuehl: I'd get rid of the description of the strategies in the API and the field as an enum. Otherwise it has to be kept consistent to the documentation of all existing strategies. Furthermore, the registered strategies are generally meant to be dynamic.

tmuehl: Users can ask the api to get a list of available strategies and their documentation.

api/nakadi-event-bus-api.yaml
```diff
@@ -1167,8 +1169,13 @@ definitions:
...
+          Determines how the assignment of the event to a Partition should be handled. If not set - the 'random'
+          partition resolution strategy will be used, which will put the events to a random partition and that is not
+          a problem if your `EventType` type has only one partition (which is default). If your `EventType` will have
+          more than one partition you should think about setting some other strategy than 'random' if you care about
+          ordering of your events.
+          See more detailed description of available partition resolution strategies in `PartitionResolutionStrategy`
+          definition
         type: object
         allOf:
           - $ref: '#/definitions/PartitionResolutionStrategy'
```
tmuehl: Why not simply a string which is the name of the strategy?

tmuehl:
We sometimes call it `PartitioningStrategy` and `PartitionResolutionStrategy`. I know that was already introduced in a PR before. Let's be consistent here. I'd personally would prefer `PartitioningStrategy` at least in the code.

Furthermore, we have a class `PartitionResolutionStrategy` and an interface `PartitioningStrategy` which have nothing to do with each other. I think it's a bit confusion - at least I had to think to understand what represents what. I'd suggest to rename `PartitionResolutionStrategy` to something that emphasises that this represents the registered object, e.g. `PartitioningStrategyDescription` and the `PartitioningStrategy` exposes another method

```
PartitioningStrategyDescription describe();
```

src/main/java/de/zalando/aruha/nakadi/partitioning/UserDefinedPartitioningStrategy.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.List;
+
+public class UserDefinedPartitioningStrategy implements PartitioningStrategy {
+
+    @Override
+    public String calculatePartition(final EventType eventType, final JSONObject event, final List<String> partitions)
+            throws PartitioningException {
+        try {
+            final String partition = event.getJSONObject("metadata").getString("partition");
```
tmuehl: I'd null check `event.getJSONObject("metadata")`. I know, you know that the partitioning strategy is applied after an event is validated against it's schema and that the schema currently requires the metadata field to be present. However, the interface of the function doesn't guarantee all that and I wouldn't rely on this.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -154,12 +157,23 @@ public EventTypeController(final EventTypeRepository eventTypeRepository, final
...
+    private void validatePartitioningStrategy(final EventType eventType) throws NoSuchPartitioningStrategyException,
+            InvalidEventTypeException {
         final PartitionResolutionStrategy partitioningStrategy = eventType.getPartitionResolutionStrategy();
-        if (partitioningStrategy != null && !partitionResolver.strategyExists(partitioningStrategy.getName())) {
+        if (!partitionResolver.strategyExists(partitioningStrategy.getName())) {
             throw new NoSuchPartitioningStrategyException("partitioning strategy does not exist: " +
                     partitioningStrategy.getName());
         }
+        else if (HASH_STRATEGY.equals(partitioningStrategy.getName()) &&
+                eventType.getPartitioningKeyFields().isEmpty()) {
```
tmuehl: Maybe we also check that the specified `PartitioningKeyFields`are valid and required fields.

src/main/java/de/zalando/aruha/nakadi/service/Registry.java
```diff
@@ -0,0 +1,32 @@
+package de.zalando.aruha.nakadi.service;
+
+import com.google.common.collect.ImmutableList;
+import de.zalando.aruha.nakadi.domain.PartitionResolutionStrategy;
+import de.zalando.aruha.nakadi.partitioning.PartitioningStrategy;
+
+import java.util.List;
+
+public class Registry {
```
tmuehl: `Registry` sounds so generic and can be basically everything. :) Maybe `StrategiesRegistry` or a dedicated `PartitioningStrategiesRegistry`?

api/nakadi-event-bus-api.yaml
rcillo: I would ratter keep the documentation in the API yml file only. Since it's more convenient for the users to understand Nakadi capabilities without been forced to run some GET in the command line.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
v-stepanov: I don't really have a strong preference here. I just remember that we decided to forbid that for undefined strategy as it doesn't have a `metadata` as a standard field. But in general I see no problem to allow that for undefined category

src/main/java/de/zalando/aruha/nakadi/service/Registry.java
v-stepanov: ![]
(https://s-media-cache-ak0.pinimg.com/736x/85/34/76/8534760ab5e299800888507c1b69a406.jpg)

v-stepanov: But in general the idea was that this registry will also contain stuff like ValidationStrategies and Enrichment strategies. So it really can be `StrategiesRegistry`. I will rename.

v-stepanov:
@tmuehl 
Checked your comments. Good points :+1: 
Will try to resolve some issues when will have some time.

src/main/java/de/zalando/aruha/nakadi/service/Registry.java
tmuehl: :smile:

src/main/java/de/zalando/aruha/nakadi/partitioning/UserDefinedPartitioningStrategy.java
rcillo: Would you throw an exception in case it happens? Because if so I would say the behaviour would be the same, since `getJSONObject` also throws an exception in case of null.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
rcillo: What would be a valid `PartitioningKeyFields`? Would it be a field stated in the schema?

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
rcillo: For the sake of simplicity, we decided to ship it without support for undefined. Let's do it later, if necessary.

rcillo:
I fixed merge conflicts introduced by the my last changes to nakadi-jvm.

I'll do some small adjustments to this PR so we can close a new release, ok?

1. rename registry https://github.com/zalando/nakadi/pull/180#discussion_r56994344
2. rename partition strategy description https://github.com/zalando/nakadi/pull/180#issuecomment-199832041

Do you think it will do it?

api/nakadi-event-bus-api.yaml
```diff
@@ -823,6 +822,21 @@ paths:
 
         If the EventType creation is to have a specific partitioning (besides the default), one can
         consult over this method the available possibilities.
+
+        Nakadi offers currently, out of the box, the following strategies:
+
+        - `random`: Resulution of the target partition happens randomly (events are evenly distributed on the
```
rcillo: /Resulution/Resolution/

api/nakadi-event-bus-api.yaml
```diff
@@ -823,6 +822,21 @@ paths:
...
+          topic's partitions).
+
+        - `user_defined`: Target partition is defined by the client. As long as the indicated partition
+          exists, Event assignment will respect this value. Correctnes of the relative ordering of events is
```
rcillo: /Correctnes/Correctness/

api/nakadi-event-bus-api.yaml
```diff
@@ -823,6 +822,21 @@ paths:
...
+          under the responsibility of the Producer.
+          Requires that the client provides the target partition on `metadata.paritition` (See
```
rcillo: /paritition/partition/

api/nakadi-event-bus-api.yaml
```diff
@@ -1204,9 +1216,9 @@ definitions:
         items:
           type: string
         description: |
-          Indicator of the fields used for guaranteeing the ordering of Events of this type (used only by 'hash'
-          PartitionResolutionStrategy). Regired if 'partition_resolution_strategy' is set to 'hash'. In other cases it
-          is ignored.
+          Required when 'partition_resolution_strategy' is set to 'hash'. Must be absent otherwise.
+          Indicates the fields used for evaluatiion the partition of Events of this type.
```
rcillo: /evaluatiion/evaluation/

api/nakadi-event-bus-api.yaml
```diff
@@ -365,11 +365,11 @@ paths:
...
         - oauth2: [nakadi.admin]
       description: |
-        Deletes an `EventType` identified by its name. The underlying Kafka topic and all events of this `EventType`
-        will also be removed.
-        It's important to note that topic deletion in Kafka can take some time, so before it is completelly removed
-        it will not be possible to create an `EventType` with the same name as removed one had (you will get 409
-        response code from creation endpoint)
+        Deletes an `EventType` identified by its name. The underlying Kafka topic and all events of this
+        `EventType` will also be removed.
+        **Note**: Kafka's topic deletion happens asynchronously w.r.t. this DELETE call; therefore creation
```
rcillo: what's `w.r.t.`?

tmuehl: _with respect to_

rcillo: Thanks :smile: Maybe we should replace with `with respect to` so that people like me have an easier time understanding this doc.

PR #182, opened 2016-03-22
ARUHA-131 smaller suggestions to the api documentation

tmuehl:
@rcillo @v-stepanov @jkliff @CyberDem0n Please review

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1205,9 +1205,9 @@ definitions:
           type: string
         description: |
           Indicator of the fields used for guaranteeing the ordering of Events of this type (used only by 'hash'
-          PartitionResolutionStrategy).
-          If set MUST be a valid required field as defined in the schema.
-          It MUST be set if 'partition_resolution_strategy' is set to 'hash'. In other cases it is ignored.
+          PartitionResolutionStrategy). Regired if 'partition_resolution_strategy' is set to 'hash'. In other cases it
+          is ignored.
+          If set it MUST be a valid required field as defined in the schema.
 
     required:
       - name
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/182" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -1260,7 +1261,7 @@ definitions:
           - 'hash' will use the event field(s) defined in 'partitioning_key_fields' property of `EventType` as a source
             for a hash function that will caclulate the partition where the event will be put. All events with the same
             value in this field(s) will go to the same partition, and consequently be ordered.
-          - 'user_defined' will use 'metadata'.'partition' property of incoming event to know the partition where to put
+          - 'user_defined' will use `metadata.partition` property of incoming event to know the partition where to put
```
rcillo: Why using back ticks  for this field and not for the one above `partitioning_key_fields`?

api/nakadi-event-bus-api.yaml
```diff
@@ -1203,10 +1204,10 @@ definitions:
         items:
           type: string
         description: |
-          Indicator of the field used for guaranteeing the ordering of Events of this type (used only by 'hash'
+          Indicator of the fields used for guaranteeing the ordering of Events of this type (used only by 'hash'
           PartitionResolutionStrategy).
           If set MUST be a valid required field as defined in the schema.
-          It MUST be set if 'partition_resolution_strategy' is set to 'hash'.
+          It MUST be set if 'partition_resolution_strategy' is set to 'hash'. In other cases it is ignored.
```
rcillo: You could also use the same wording as before "In case of another partitioning strategy it is ignored.". Its very clear that way.

api/nakadi-event-bus-api.yaml
tmuehl: Yes, should be for the above `partitioning_key_fields` also, have missed that. :)

api/nakadi-event-bus-api.yaml
tmuehl: Yes, I was thinking about that, didn't wanted to alter the original too much. But since you think the same, I change it to make it consistent.

rcillo:
Nice improvements :shipit:

tmuehl:
@rcillo Thanks :)

PR #183, opened 2016-03-22
ARUHA-153 Remove Authorization header for localhost calls

tmuehl:
@rcillo @v-stepanov @jkliff @CyberDem0n @sezera Please review

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,6 @@ To stop the running Nakadi again:
 
 ```sh
 curl --request POST \
-     --header "Authorization: Bearer $TOKEN" \
      --header "Content-Type:application/json" \
      http://localhost:8080/event-types -d '{
   "name": "order.ORDER_RECEIVED",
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/183" target="_blank">ReviewNinja

rcillo:
Maybe we should not try to validate the token when authentication is disabled. Then these instructions would be valid for all environments. Wdyt?

tmuehl:
@rcillo I think, it's ok to validate it if it's provided. But I don't have a strong opinion about this.
Nevertheless it should be removed from the Readme.

Maybe it's worth to add an extra line in Readme telling about the Authorization. But the Readme on Github is not a Readme how to use the Zalando Services.

rcillo:
@tmuehl ok. Since it's targeted in localhost, then it seems reasonable to remove it.

:shipit:

PR #186, opened 2016-03-23
Aruha 112 event bulk submission

rcillo:
This PR introduces a backward incompatible change to the current API.

Once merged, the old event submission endpoint, which accepts a single event, is going to require an array of events.

That mean that in the case of single event submission, clients should wrap the event with "[]".

You should find further details related to design decisions on each commit message. It might also be easier to review.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/186" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -167,6 +178,42 @@ public void postEvent(final String topicId, final String partitionId, final Stri
...
+            } catch (InterruptException | SerializationException | BufferExhaustedException e) {
+                LOG.error("Error posting event to kafka", e);
+                item.setPublishingStatus(EventPublishingStatus.FAILED);
+                throw new EventPublishingException("Error publishing message to kafka", e);
+            }
+        }
+
+        // this is not the most precise way of waiting for the send to complete, since it will flush all
+        // other concurrent partitions in the accumulator.
+        kafkaProducer.flush();
```
v-stepanov: This can bring certain problems. As we have a single kafkaProducer instance for the whole application this flush will block not only till all events of this batch are published but till all events that are sent to this kafkaProducer are published. So in a situation when we process 20 different batches at the same time the behaviour here can be quite strange. 

Have you considered a possibility to use another send method that returns future?
`final Future<RecordMetadata> future = kafkaProducer.send(record);`
And then wait till al futures are finished.

(or if you want to stick to callback-implementation - you can then create manual synchronisation with some synchronised counter or something like that)

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -167,6 +178,42 @@ public void postEvent(final String topicId, final String partitionId, final Stri
...
+    }
+
+    private Callback kafkaSendCallback(final BatchItem item) {
+        return new Callback() {
```
v-stepanov: can be replaced with hipster lambda

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: Your're right. I changed it here https://github.com/zalando/nakadi/pull/186/commits/8f4adc36f670555c5b1019b7285f34510dc144f4 :+1:

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: :sunglasses: https://github.com/zalando/nakadi/pull/186/commits/d7496d02979b8ea0f542b837d963e93c79e6ea65

rcillo:
I added one more commit to restore an accidentally removed unit test for metrics collection https://github.com/zalando/nakadi/pull/186/commits/33b4dd5433775cdcc50d1b20b8725760febb7bd7

src/main/java/de/zalando/aruha/nakadi/repository/TopicRepository.java
```diff
@@ -35,9 +37,11 @@ void createTopic(String topic, int partitionsNum, int replicaFactor, long retent
 
     void postEvent(String topicId, String partitionId, String payload) throws NakadiException;
```
rcillo: This is no longer used. I'll remove it.

src/main/java/de/zalando/aruha/nakadi/service/EventPublisher.java
```diff
@@ -0,0 +1,140 @@
...
+
+    public EventPublisher(final TopicRepository topicRepository,
+                         final EventTypeCache cache,
+                         final PartitionResolver partitionResolver) {
+        this.topicRepository = topicRepository;
+        this.cache = cache;
+        this.partitionResolver = partitionResolver;
+    }
+
+    public EventPublishResult publish(JSONArray events, String eventTypeName) throws NoSuchEventTypeException,
```
rcillo: One could ask why I mixed a result object with exceptions here. For example:

1. if there is a validation partitioning or publishing errors, it returns a result object
2. if the event type cannot be found, it throws an exception

That's because in the former scenarios we want not only to inform about an error but also to provide granular information about the status of every item. So we would be required to create new exceptions types to hold the "result" information.

It seemed to be better to make a distinction between exceptions (caused by unexpected things) and a result object (expected).

This distinction sometimes blurry so if you fill we should decide between exceptions only or result only, please, I would like to read your reasoning.

rcillo:
I'll do some experimentation locally with different batch configurations (trigger values) and see which performs best. But there's no guarantee that it will also work for other environments where networking configuration is different.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -69,57 +50,33 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName, @Reque
...
-            return doWithMetrics(eventTypeName, startingTime, () -> {
-                final JSONObject eventAsJson = parseJson(event);
-                validateSchema(eventAsJson, eventType);
-                String partitionId = partitionResolver.resolvePartition(eventType, eventAsJson);
-
-                topicRepository.postEvent(eventTypeName, partitionId, event);
-
-                return status(HttpStatus.CREATED).build();
+            return doWithMetrics(eventTypeName, System.nanoTime(), () -> {
+                final JSONArray events = new JSONArray(event);
```
tmuehl: Maybe rename die parameter to `events` or `eventsBatch` or `eventsAsString`? :)

src/main/java/de/zalando/aruha/nakadi/service/EventPublisher.java
```diff
@@ -0,0 +1,141 @@
...
+
+public class EventPublisher {
+    private static final Logger LOG = LoggerFactory.getLogger(EventPublisher.class);
+
+    private final TopicRepository topicRepository;
+    private final EventTypeCache cache;
+    private final PartitionResolver partitionResolver;
+
+    public EventPublisher(final TopicRepository topicRepository,
+                         final EventTypeCache cache,
```
tmuehl: Rename to `eventTypeCache`? From reading a variable just called `cache` in the code, I don't know what is actually cached.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
rcillo: :+1: done here https://github.com/zalando/nakadi/pull/186/commits/cb41c64f8507f6c28f8df5ae7146b5f0810db9f4

src/main/java/de/zalando/aruha/nakadi/service/EventPublisher.java
rcillo: :+1: https://github.com/zalando/nakadi/pull/186/commits/cb41c64f8507f6c28f8df5ae7146b5f0810db9f4

tmuehl:
Looks good :+1:

PR #187, opened 2016-03-23
Aruha 154 hotfix datetime validation

### LINKED ISSUE #185: Date time validations not compliant with rfc 3339
The following date is rejected "2016-03-23T12:52:06.8Z" by nakadi but it should be accepted, since 1 digit milliseconds is specified by the rfc https://www.ietf.org/rfc/rfc3339.txt as valid.

rcillo:
The current validation rejects DateTimes with variable number of
millisecond digits (it's accepts only zero or 3 digits milliseconds strings).

It's better to use Joda validation since it handles such scenarios,
even though it accepts timezones without ":" which should be
invalid by the RFC 3339.

But it's better to accept such small variation than reject valid dates.

Fixes #185

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -154,7 +154,7 @@ dependencies {
     compile 'org.apache.curator:curator-recipes:2.9.1'
 
     // json
-    compile 'org.everit.json:org.everit.json.schema:1.1.0'
+    compile 'org.everit.json:org.everit.json.schema:1.2.0'
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-json-org:2.4.0'
 
     // scala dependencies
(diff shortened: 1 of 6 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/187" target="_blank">ReviewNinja

rcillo:
If we want to avoid accepting timezones without ":" to be 100% compliant with the RFC 3339, then I would be required to implement some code using regular expressions, like https://github.com/restlet/restlet-framework-java/blob/master/modules/org.restlet/src/org/restlet/engine/util/InternetDateFormat.java

If you think it's worth it, I would like to read your reasoning.

PR #191, opened 2016-03-29
ARUHA-156 Collect metrics for event batch sizes

tmuehl:
@rcillo @jkliff Please review


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/191" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -50,8 +53,9 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName, @Reque
         LOG.trace("Received event {} for event type {}", event, eventTypeName);
 
         try {
-            return doWithMetrics(eventTypeName, System.nanoTime(), () -> {
+            return doWithTimerMetric(eventTypeName, System.nanoTime(), () -> {
                 final JSONArray eventsAsJsonObjects = new JSONArray(event);
+                batchSizeHistogram.update(eventsAsJsonObjects.length());
```
rcillo: Do you think the size in bytes would be an interesting metric as well?

rcillo:
Maybe we should complement metrics tests in the controller with a new assertion regarding this new measurement.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
tmuehl: Yes, I think so. I added it to the ticket ARUHA-156 and gonna implement it in this PR.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -75,17 +85,16 @@ private ResponseEntity response(final EventPublishResult result) {
         }
     }
 
-    private ResponseEntity doWithMetrics(final String eventTypeName, final long startingNanos,
-                                         final EventProcessingTask task) throws NakadiException {
+    private ResponseEntity doWithTimerMetric(final EventTypeMetrics eventTypeMetrics, final long startingNanos,
```
jkliff: why pass a start time and calculate the delta yourselve if, doWithTimerMetric, receives the metric and from it could extract a timer (c  = timer.time() then c.stop()) ?

rcillo: I think the motivation for that is to be able to pass an initial time that is not right after the invocation of this method.

Maybe it would make sense to change it now. Seems a good idea to control time inside of this method.

tmuehl: The initial reason was that in the moment the timer is started I don't know if the process is going to succeed or fail. I want to report it differently, especially because we use the timer also as a counter. 

But after the refactoring it is not necessary to pass in the start time from outside. The `System.nanoTime()` can be called inside the method. I change it.

tmuehl:
@rcillo Regarding metric tests: done.

src/main/java/de/zalando/aruha/nakadi/metrics/EventTypeMetricRegistry.java
```diff
@@ -0,0 +1,36 @@
...
+            synchronized (metricsPerEventType) {
+                eventTypeMetrics = getOrAddMetricsFor(eventTypeName);
+            }
+        }
+        return eventTypeMetrics;
+    }
+
+    private EventTypeMetrics getOrAddMetricsFor(final String eventTypeName) {
+        EventTypeMetrics eventTypeMetrics = metricsPerEventType.get(eventTypeName);
+        if (eventTypeMetrics == null) {
```
rcillo: Mult-threading is hard. One way of not having to handle it ourselves would be to use a guava cache here. Wdyt?

antban: Why not to use ConcurrentHashMap?

rcillo: `ConcurrentHashMap` would not solve the problem of duplicated generation of `EventTypeMetrics` would it?

rcillo: You're right. There's a `computeIfAbsent`.

jkliff: this look very similar to the singleton doubly locked (anti) pattern. 
isn't it simpler to preinitialize these together with the initial loading of each event type (and accordingly the creation?)

tmuehl: @antban @rcillo Oh yes, `ConcurrentHashMap`, I wasn't aware of that `computeIfAbsent`, thanks!

@jkliff I think it's not an anti-pattern under every circumstances and depends on some compiler behaviours. I thought they were not present in the current Java versions anymore, but maybe it's only safe for volatile field. Anyway, I will use the `ConcurrentHashMap`.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -67,6 +71,12 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName, @Reque
         }
     }
 
+    private void reportMetrics(final EventTypeMetrics eventTypeMetrics, final String eventsAsString, final JSONArray eventsAsJsonObjects) {
+        final int numberOfEventsInBatch = eventsAsJsonObjects.length();
+        eventTypeMetrics.getEventsPerBatchHistogram().update(numberOfEventsInBatch);
+        eventTypeMetrics.getAverageEventSizeInBytesHistogram().update(eventsAsString.length() / numberOfEventsInBatch);
```
v-stepanov: can `numberOfEventsInBatch` be `0` ?

tmuehl: Ah, good point :+1:

v-stepanov:
:shipit:

PR #194, opened 2016-03-30
aruha-114 fix ambiguity when requiring eid

### LINKED ISSUE #192: is eid in metadata now required or not?
in https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#L879 it says regarding the metadata/eid "Clients are allowed to generate this", which sounds to me like its not mandatory to create this by yourself as a producer.

in https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#L922 though it says that eid is required.

which one is the correct interpretation?

rcillo:
It was not clear enought if eid is a required metadata field or not.

The correct interpretation is the later: it is required.

Our usage of Kafka provides the safest persistence guarantee available,
which has "at least once" semantics.

> acks=all This means the leader will wait for the full set of in-sync
replicas to acknowledge the record. This guarantees that the record will
not be lost as long as at least one in-sync replica remains alive. This
is the strongest available guarantee.

On the other hand it means that we could reply with error when it would
be successfully published. That would happen in the very rare event of
timeout during replicas acknowledgement. See page 81, for example
http://www.slideshare.net/miguno/apache-kafka-08-basic-training-verisign

That said, it's really important to provide a way for consumers to
handle duplicates. That is one of the reasons why it's important to
provide an UUID.

Fixes #192

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -876,8 +876,8 @@ definitions:
             description: |
               Identifier of this Event.
 
-              Clients are allowed to generate this and this SHOULD be guaranteed to be unique from the
-              perspective of the producer. Consumers MIGHT use this value to assert uniqueness of reception
+              Clients MUST generate this and this SHOULD be guaranteed to be unique from the perspective of the
+              producer. Consumers MIGHT use this value to assert uniqueness of reception
               of the Event.
             type: string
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/194" target="_blank">ReviewNinja

rcillo:
Fixed in #195

PR #199, opened 2016-03-31
ARUHA-158 Limit event type names to 255 characters

### LINKED ISSUE #184: Registering an event type with name bigger than 255 chars returns 500
Kafka does not accept topic names with more than 255 chars. Since this is not validated, it fails to create such topic and returns a 500.

It would be better to validate name length informing users with a 422 and a proper detail message.

tmuehl:
Fixes bug ARUHA-158
Fixes #184 
Kafka limits topic names to 255 characters. So we do with event type names.

@rcillo @jkliff @v-stepanov Please review


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/199" target="_blank">ReviewNinja

rcillo:
Nice :+1: 

:shipit:

PR #201, opened 2016-03-31
ARUHA-157 Update Kafka Broker List for KafkaProducer

### LINKED ISSUE #188: Nakadi producers does not recover from kafka broker shutdowns
Besides periodically refreshing the list of available brokers https://github.com/zalando/nakadi/blob/nakadi-jvm/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java#L90-L98 Nakadi does not refresh it's producer instances with up-to-date properties (list of brokers). 

Since producers are created during initialisation phase only https://github.com/zalando/nakadi/blob/nakadi-jvm/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaFactory.java#L16, they will not use recently created brokers. It may cause inconsistent behaviour when publishing events.

Nakadi should recover from such scenarios by discovering recently created brokers and updating producers with such information.

tmuehl:
When the Kafka brokers change, create a new producer to take new brokers into account and remove old ones.

Fixes #188

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaFactory.java
+++ b/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaFactory.java
@@ -10,7 +10,7 @@ import java.util.Properties;
 
 class KafkaFactory implements KafkaPropertiesListener {
     private final KafkaLocationManager kafkaLocationManager;
-    private KafkaProducer<String, String> kafkaProducer;
+    private volatile KafkaProducer<String, String> kafkaProducer;
 
     public KafkaFactory(final KafkaLocationManager kafkaLocationManager) {
         this.kafkaLocationManager = kafkaLocationManager;
--- a/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaFactory.java
+++ b/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaFactory.java
@@ -18,7 +18,7 @@ class KafkaFactory implements KafkaPropertiesListener {
         kafkaLocationManager.registerPropertiesListener(this);
     }
 
-    public Producer<String, String> createProducer() {
+    public Producer<String, String> getProducer() {
         return kafkaProducer;
     }
 
--- a/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaFactory.java
+++ b/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaFactory.java
@@ -6,14 +6,16 @@ import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.Producer;
 
 import java.util.List;
+import java.util.Properties;
 
-class KafkaFactory {
+class KafkaFactory implements KafkaPropertiesListener {
     private final KafkaLocationManager kafkaLocationManager;
-    private final KafkaProducer<String, String> kafkaProducer;
+    private KafkaProducer<String, String> kafkaProducer;
 
     public KafkaFactory(final KafkaLocationManager kafkaLocationManager) {
         this.kafkaLocationManager = kafkaLocationManager;
         kafkaProducer = new KafkaProducer<>(kafkaLocationManager.getKafkaProperties());
+        kafkaLocationManager.registerPropertiesListener(this);
     }
 
     public Producer<String, String> createProducer() {
@@ -29,4 +31,8 @@ class KafkaFactory {
         return new NakadiKafkaConsumer(getConsumer(), topic, kafkaCursors, pollTimeout);
     }
 
+    @Override
+    public void updateProperties(final Properties kafkaProperties) {
+        kafkaProducer = new KafkaProducer<>(kafkaProperties);
+    }
 }
--- a/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
+++ b/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
@@ -6,10 +6,8 @@ import org.json.JSONException;
 import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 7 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/201" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -91,13 +97,23 @@ private void init() {
     @Scheduled(fixedDelay = 30000)
```
jkliff: I know this is not a change in this ticket but: 
1. shouldn't this fixedDelay be parameterized?
2. is it every (reasonable) possible that kafkaProperties is null? do we gain anything by handling this silently?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -91,13 +97,23 @@ private void init() {
...
     private void updateBrokers() {
         if (kafkaProperties != null) {
+            final Properties oldProperties = getKafkaProperties();
+
             final List<Broker> brokers = fetchBrokers();
             if (!brokers.isEmpty()) {
```
jkliff: here too: is it ever valid that there are no brokers? should we even try to recover from this?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -91,13 +97,23 @@ private void init() {
...
             }
         }
     }
 
+    private void notifyPropertiesListeners() {
+        kafkaPropertiesListeners.stream().forEach(listener -> listener.updateProperties(getKafkaProperties()));
+    }
+
     public Properties getKafkaProperties() {
         return (Properties) kafkaProperties.clone();
```
jkliff: out of curiosity: why clone? is it ever modified by the receiver?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -29,4 +31,8 @@ public NakadiKafkaConsumer createNakadiConsumer(final String topic, final List<K
         return new NakadiKafkaConsumer(getConsumer(), topic, kafkaCursors, pollTimeout);
     }
 
+    @Override
+    public void updateProperties(final Properties kafkaProperties) {
+        kafkaProducer = new KafkaProducer<>(kafkaProperties);
```
jkliff: a few issues that might or might not relevant:
1. what i can't follow from this is: how this actually changes the kafkaProducer in the repository? there createProducer gave a reference to the previous kafkaProducer and there it was not changed by this update until afurther createConsumer, which does not happen.
2. allright that @Scheduled is thread safe, but kafkaProducer here is not really safe for access, right? the updateProperties too can theoretically be called by other means that will mess things up. wouldn't it make sense to lock updateProperties to write wrt to reads on createProducer?
3. does it make sense to offer the possibility to explicitly trigger this reloading? this might come in handy...

rcillo: From my understanding:

1. I think you're right. The reference to the old producer would not be update in the `KafkaTopicRepository`. Then, even if we do have a new producer here, it would not be used. I think we need to change the repository to always ask for this "producer singleton". Or we need a "setter" in the repository in order to update the reference to the new producer.

2 and 3. This seems to be too much hypothetical. But since it's as simple as locking, I would not be against it. That could prevent future developers from expecting this method to provide insufficient thread safety.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
tmuehl: I don't know. That's what is was, I haven't touched it but was asking myself the same.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
tmuehl: Same here, haven't touched that code. :) And I agree with you, if there is no single broker that's a bigger issue not solved by just keeping the old brokers.

Maybe the idea was that if the list of brokers is empty the reason is that the reading from Zookeeper was faulty. So rather stay with the current list. But I don't know.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaLocationManager.java
tmuehl: I guess to be immutable. You may not want to trust your caller. :)

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaFactory.java
tmuehl: 1. The name of the method `createProducer()` is misleading here. It should be called whenever a kafka producer is needed and not be cached or kept by a caller (same as for the consumer). I gonna change the name of the method. But you hit a very good point: I wanted to check if every caller does it right and totally forgot.
2. I agree with @rcillo, I think it's very hypothetical. Changes to the properties should only happen within this class and thus it's the responsibility of the one who does it to make it right and eventually call `notifyPropertiesListeners()`. I would prefer to not overcomplicated it here. And even if somebody else would call `updateProperties()` it shouldn't be an issue since assigning a new kafka producer should be atomic as well a returning the field. However, I evaluate making the `kafkaProducer` field volatile as you suggested in another discussion.
3. Oh yes, I will add this!

rcillo:
What's the current status of this PR? @tmuehl do you know if was it a false positive? If it's not necessary to update the producer, then why is it necessary to update brokers list periodically?

rcillo:
We no longer experienced this issue after a fix in buku.

PR #202, opened 2016-03-31
aruha-72 remove doc from the partition strategy

rcillo:
Providing a way for users to submit to define a doc for fixed number
strategies adds no value, since the documentation should never change
nor be defined by users.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/202" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/service/StrategiesRegistry.java
```diff
@@ -9,20 +9,13 @@
 public class StrategiesRegistry {
 
     public static final PartitionStrategyDescriptor HASH_PARTITION_STRATEGY =
-            new PartitionStrategyDescriptor("hash", "This strategy will use the event " +
```
jkliff: oh, I apologize! i didn't know it was already being exposed with the current implementation... 
my suggestion was more on the pragmatic side. 

do you really think we should remove this after all?

rcillo: I think so. This is duplicating the description from the API. It increases the cost of maintenance and provides nothing in return, since this is fixed. I think it would be better to keep documentation in a single place, the API yml.

tmuehl: We had this discussion since I once proposed to remove the description from the API. I would strongly vote for not removing the description from the code (and rather remove it from the API or keep it in both places).

I see the strategies not as part of the api but something that theoretically could be configured and is a open registry. A description for a strategy hence should be come a long with the strategy and not with Nakadi's API. That's my 2c.

What I meant, when we discussed this offline was that I would change the field in the API from

```
partition_strategy:
        allOf:
          - $ref: '#/definitions/EventPartitionStrategy'
```

to

```
partition_strategy:
        type: string
```

referring to the name of a strategy.

rcillo: I would agree if the registry was dynamic. 

Removing it from the API documentation would require users interested in using our API to get a token, learn about our environments or even deploy a locally, assemble the proper curl with headers and authentication just to read an excerpt of immutable text.

I would not discard the possibility of this changing in the future, but I would prefer adding complexity when it's required and not before hand.

tmuehl: Generally it's thought to be dynamic, configurable and extendable, like the enrichment and validation. But I agree with you that it's not a good approach that the user has to get an access token to be able to read some documentation. Still, for me it's not part of the api. Maybe it is some other external documentation and would fit into a document like the Readme. We can also use the wiki here on github. I still have the feeling that a strategy (no matter if it's partition, validation or enrichment) should bring along it's explanation. As an intermediate solution, I'm fine with it.

Anyway, if we remove the `doc` field, only the `name` remains. I'd remove that field as well and just have a string with a name referring to a strategy (as mentioned above).

v-stepanov: +1 for removing the class (and removing `EventPartitionStrategy` from swagger)
As it has now only `name` field - it doesn't make much sense.

tmuehl: Another issue, when keeping the documentation in the API is where to keep it?

* `GET /registry/partition-strategies`
* Definition of `eventtype.partition_strategy`
* Definition of `EventPartitionStrategy`

I really think, it is not the right place to keep it.

rcillo: Let's discuss this live tomorrow, ok?

rcillo:
I'll close this PR until we have a clearer definition.

PR #204, opened 2016-04-04
ARUHA-160 Added closing of kafka consumer when stream is closed

v-stepanov:
@tmuehl @rcillo @jkliff @CyberDem0n @antban 
just a small bugfix (kafka consumer was not closed properly), please review

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
@@ -77,6 +77,7 @@ public class EventStreamController {
         return outputStream -> {
 
             Counter consumerCounter = null;
+            EventConsumer eventConsumer = null;
 
             try {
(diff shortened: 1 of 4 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/204" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -163,6 +163,9 @@ public StreamingResponseBody streamEvents(
                 if (consumerCounter != null) {
                     consumerCounter.dec();
                 }
+                if (eventConsumer != null) {
```
tmuehl: +1 for putting the curly brace to the same line as the conditional. :)

v-stepanov: Hey, I always do that! :)
You probably mean that I put `else` or `catch` on the next line. Like
```
if () {
    ....
}
else {
    ....
}
```

PR #206, opened 2016-04-05
aruha-112 don't display null detail

rcillo:
Display `detail: null` could confuse users. Display empty string
instead.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/206" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/domain/BatchItemResponse.java
```diff
@@ -3,7 +3,7 @@
 public class BatchItemResponse {
     private EventPublishingStatus publishingStatus = EventPublishingStatus.ABORTED;
     private EventPublishingStep step = EventPublishingStep.NONE;
-    private String detail;
+    private String detail = "";
```
v-stepanov: But you wanted to have a static constant. What made you change your mind? :)

rcillo: I believed in this answer http://stackoverflow.com/a/1171707 it seems that Java "interns" string literals so to reuse them. I didn't experimented but found plenty examples all over the internet.

PR #207, opened 2016-04-05
Aruha 134

v-stepanov:
- removed nakadi.read scope
- adjusted scope name to what we actually have
- removed trailing slash automapping to avoid security problems
@jkliff @CyberDem0n @rcillo @tmuehl @antban please review

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -105,15 +105,13 @@ securityDefinitions:
     flow: implicit
     authorizationUrl: 'https://auth.example.com/oauth2/tokeninfo'
     scopes:
-      nakadi.read: |
-        Generic read scope for operations that do not require more specific restrictions.
-      nakadi.admin: |
-        Grants access for administrative operations in Nakadi.
-      nakadi_event_type.write: |
+      nakadi.config.write: |
+        Grants access for changing Nakadi configuration.
+      nakadi.event_type.write: |
         Grants access for applications to define and update EventTypes.
-      nakadi_event_stream.write: |
+      nakadi.event_stream.write: |
         Grants access for applications to submit Events.
-      nakadi_event_stream.read: |
+      nakadi.event_stream.read: |
         Grants access for consuming Event streams.
 
 paths:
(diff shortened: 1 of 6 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/207" target="_blank">ReviewNinja

tmuehl:
@v-stepanov In which way do the trailing slashes raise security issues?

v-stepanov:
@tmuehl 
Before my fix it was possible to avoid scope check by just adding "/" at the end of URL. As Spring by default maps that to URL without slash, and Spring security only checks the URL without slash at the end.

tmuehl:
@v-stepanov Have you tried to put the more specific rule (e.g. `"/event-types/*/events"`) at the top and use `**` instead at the end of one rule, e.g.

```
.antMatchers(GET, "/event-types/*").access(hasScope(uidScope))
```

v-stepanov:
@tmuehl I considered this possibility - to extend security masks with ** to cover more. But I think that can lead to different overlapping scenarios and to bugs as a result. I would prefer to have a more specific mask for each security role that covers exactly only endpoint itself.

tmuehl:
@v-stepanov I think it would all work fine, if you put the more specific endpoints to the top. :)

v-stepanov:
@tmuehl Yes, but that already makes you think about it and gives an opportunity to make a bug :)
Why do you prefer that? To keep the possibility to have `/` at the end?

tmuehl:
@v-stepanov Yes, I like to have the possibility to be able to have a trailing slash. In the past I had cases when I was wondering what's going wrong when calling an endpoint and it turned out it was the missing / present trailing slash. I like to not need to think about if omitting the slash or not, if it is considered to be the same by the server. It can quickly lead to small but nerve-wracking bugs on the client side. But that's maybe my personally feelings, so feel free to ignore if you don't agree. :)

Particularly in cases where I expect the api to return a list of resources, i.e. where the resource is a bucket, it feels more natural for me to treat it like a directory, e.g. `/event-types/`.

Furthermore, what you suggest ist blacklisting, while restricting the entire endpoint `/event-types/**` and only set different access control settings for specific endpoints is whitelisting, what I prefer.

v-stepanov:
@tmuehl 
I discussed that with api-guild, they also think that it's better to allow trailing slash as optional to be more flexible for the user who uses API. So I changed that back to optional "/" at the end, and I changed scopes masking to cover more with ** so that we don't have that security leak.

PR #208, opened 2016-04-05
aruha-112 set detail in case of publishing exceptions

rcillo:
When an exception happens on publishing, update response status and
detail message.

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
+++ b/src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
@@ -35,20 +35,8 @@ public class BatchItem {
         this.response.setStep(step);
     }
 
-    // avoid race condition with kafka producer callback
-    synchronized public void setPublishingStatus(final EventPublishingStatus publishingStatus) {
+    synchronized public void updateStatusAndDetail(final EventPublishingStatus publishingStatus, final String detail) {
         this.response.setPublishingStatus(publishingStatus);
-    }
-
-    public void setDetail(String detail) {
         this.response.setDetail(detail);
-    }
-
-    // avoid race condition with kafka producer callback
-    synchronized public void safeFail(final String reason) {
-        if (this.getResponse().getPublishingStatus() != SUBMITTED) {
-            this.setDetail(reason);
-            this.response.setPublishingStatus(FAILED);
-        }
     }
 }
--- a/src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
+++ b/src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
@@ -2,6 +2,9 @@ package de.zalando.aruha.nakadi.domain;
 
 import org.json.JSONObject;
 
+import static de.zalando.aruha.nakadi.domain.EventPublishingStatus.FAILED;
+import static de.zalando.aruha.nakadi.domain.EventPublishingStatus.SUBMITTED;
+
 public class BatchItem {
     private final BatchItemResponse response;
     private final JSONObject event;
@@ -32,11 +35,20 @@ public class BatchItem {
         this.response.setStep(step);
     }
 
-    public void setPublishingStatus(final EventPublishingStatus publishingStatus) {
+    // avoid race condition with kafka producer callback
+    synchronized public void setPublishingStatus(final EventPublishingStatus publishingStatus) {
         this.response.setPublishingStatus(publishingStatus);
     }
 
     public void setDetail(String detail) {
         this.response.setDetail(detail);
+    }
+
+    // avoid race condition with kafka producer callback
+    synchronized public void safeFail(final String reason) {
+        if (this.getResponse().getPublishingStatus() != SUBMITTED) {
+            this.setDetail(reason);
+            this.response.setPublishingStatus(FAILED);
+        }
     }
 }
--- a/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -217,7 +217,7 @@ public class KafkaTopicRepository implements TopicRepository {
             if (exception == null) {
                 item.updateStatusAndDetail(EventPublishingStatus.SUBMITTED, "");
             } else {
-                LOG.error("Failed to publish event " + item.getEvent().toString(), exception);
+                LOG.error("Failed to publish event", exception);
                 item.updateStatusAndDetail(EventPublishingStatus.FAILED, "internal error");
             }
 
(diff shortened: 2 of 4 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/208" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -197,9 +197,17 @@ public void syncPostBatch(final String topicId, final List<BatchItem> batch) thr
             final boolean isSuccessful = done.await(settings.getKafkaSendTimeoutMs(), TimeUnit.MILLISECONDS);
 
             if (!isSuccessful) {
+                for (final BatchItem item : batch) {
+                    item.safeFail("timed out");
```
v-stepanov: Why do you set it for each item?
The posible scenario can be that all events except one were submitted and only one timed out and in that case `isSuccessful` will be false.

v-stepanov: ah ok. Now I see that safeFail will only set new status if event was not submitted.
Everything should be fine then.

rcillo: `isSuccessful` will be false in case the submit operation exceeded the time specified by the config.

But when it happens there could be some successfully published items already. So I try to fail every item.

There could be race conditions between this update and the `callback` updates, so I used the `synchronized` keyword.

src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepositoryTest.java
```diff
@@ -184,6 +192,52 @@ public void canPostAnEvent() throws Exception {
...
+        } catch (EventPublishingException e) {
+            assertThat(item.getResponse().getPublishingStatus(), equalTo(EventPublishingStatus.FAILED));
+            assertThat(item.getResponse().getDetail(), equalTo("timed out"));
+        }
+    }
+
+    @Test
+    @SuppressWarnings("unchecked")
+    public void whenPostEventOverflowsBufferThenUpdateItemStatus() throws Exception {
+        BatchItem item = new BatchItem(new JSONObject());
```
v-stepanov: final

src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepositoryTest.java
```diff
@@ -184,6 +192,52 @@ public void canPostAnEvent() throws Exception {
...
+        item.setPartition("1");
+        List<BatchItem> batch = new ArrayList<>();
```
v-stepanov: final

src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepositoryTest.java
```diff
@@ -184,6 +192,52 @@ public void canPostAnEvent() throws Exception {
     }
 
     @Test
+    @SuppressWarnings("unchecked")
+    public void whenPostEventTimesOutThenUpdateItemStatus() throws Exception {
+        BatchItem item = new BatchItem(new JSONObject());
```
v-stepanov: final

src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepositoryTest.java
```diff
@@ -184,6 +192,52 @@ public void canPostAnEvent() throws Exception {
...
+        item.setPartition("1");
+        List<BatchItem> batch = new ArrayList<>();
```
v-stepanov: final

v-stepanov:
@rcillo I remember we were discussing the possibility of renaming `safeFail` method. What is your current opinion about that?

rcillo:
@v-stepanov I'm thinking if isn't there a clearer solution for this issue.

There are too many things going on here:

1. there's a race condition update between kafka producer IO thread (invoking the callback) and the request thread (timeout)
2. we should always prioritise the information provided by the callback, if available. E.g. we should not override it, since it's the most up-to-date information we have about the publishing status. That's why I check for the status before updating.

From these two statements I think that the desirable scenario would be to always have this field updated by the callbacks. I'm researching if isn't there anyway of doing this.

I think that in case of "caller timeout" (thread waited for more than `getKafkaSendTimeoutMs`) we should throw a 500.

src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
```diff
@@ -32,11 +35,20 @@ public void setStep(final EventPublishingStep step) {
         this.response.setStep(step);
     }
 
-    public void setPublishingStatus(final EventPublishingStatus publishingStatus) {
+    // avoid race condition with kafka producer callback
+    synchronized public void setPublishingStatus(final EventPublishingStatus publishingStatus) {
```
tmuehl: I think there is no value in making this method `synchronized`. `setPublishingStatus(publishingStatus)` is already "almost" atomic. And if not it would be the responsibility of that method to synchronize.

What I meant in our discussion is to remove the setters for `publishingStatus` and `detail` and create one new method which is synchronized:

```
synchronized public void setPublishingStatusAndDetails(final EventPublishingStatus publishingStatus, final String cetail) {
...
}
```

That way it's avoided that the publishing status and the detail message are out of inconsistent.

rcillo: Yes, it would be simpler that way. Just did it in the last commit https://github.com/zalando/nakadi/pull/208/commits/95472cfec3e8bbf78b1a7d400be4276920cbd06f

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -197,20 +197,28 @@ public void syncPostBatch(final String topicId, final List<BatchItem> batch) thr
...
+        }
+    }
+
     private Callback kafkaSendCallback(final BatchItem item, final CountDownLatch done) {
         return (metadata, exception) -> {
             if (exception == null) {
-                item.setPublishingStatus(EventPublishingStatus.SUBMITTED);
+                item.updateStatusAndDetail(EventPublishingStatus.SUBMITTED, "");
             } else {
                 LOG.error("Failed to publish event " + item.getEvent().toString(), exception);
```
rcillo: Do you think it's ok to log the event? I was concerned about security.

v-stepanov: I think we should not log event here, but not because of security resons but because it's not really important what is the content of message here. I believe a failure with publishing to kafka doesn't depend on message content, because kafka doesn't care what is inside a message.

rcillo: :+1: sure. Removed it.

rcillo:
Closing in favor of #209

PR #210, opened 2016-04-08
Aruha 166 remove doc from strategies

v-stepanov:
@jkliff @rcillo @tmuehl @CyberDem0n @antban please review


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/210" target="_blank">ReviewNinja

v-stepanov:
Script to run on DB:
```sql
UPDATE zn_data.event_type
SET et_event_type_object = regexp_replace(et_event_type_object :: TEXT, '"partition_strategy"\s*:\s*{.*"name"\s*:\s*"hash"[^}]*}', '"partition_strategy":"hash"') :: JSONB
WHERE et_event_type_object :: TEXT ~ '"partition_strategy"\s*:\s*{.*"name"\s*:\s*"hash"[^}]*}';

UPDATE zn_data.event_type
SET et_event_type_object = regexp_replace(et_event_type_object :: TEXT, '"partition_strategy"\s*:\s*{.*"name"\s*:\s*"random"[^}]*}', '"partition_strategy":"random"') :: JSONB
WHERE et_event_type_object :: TEXT ~ '"partition_strategy"\s*:\s*{.*"name"\s*:\s*"random"[^}]*}';

UPDATE zn_data.event_type
SET et_event_type_object = regexp_replace(et_event_type_object :: TEXT, '"partition_strategy"\s*:\s*{.*"name"\s*:\s*"user_defined"[^}]*}', '"partition_strategy":"user_defined"') :: JSONB
WHERE et_event_type_object :: TEXT ~ '"partition_strategy"\s*:\s*{.*"name"\s*:\s*"user_defined"[^}]*}';
```

api/nakadi-event-bus-api.yaml
```diff
@@ -835,11 +831,11 @@ paths:
           on those fields end in the same partition.
       responses:
         '200':
-          description: Returns a list of all `EventPartitionStrategy` known to Nakadi
+          description: Returns a list of all partitioning strategies known to Nakadi
           schema:
             type: array
             items:
-              $ref: '#/definitions/EventPartitionStrategy'
+              type: string
```
tmuehl: Like!

src/main/java/de/zalando/aruha/nakadi/domain/EventType.java
```diff
@@ -32,8 +32,7 @@
     @JsonIgnore
     private final List<ValidationStrategyConfiguration> validationStrategies = Lists.newArrayList();
 
-    @Valid
-    private PartitionStrategyDescriptor partitionStrategy = RANDOM_PARTITION_STRATEGY;
+    private String partitionStrategy = PartitionStrategy.RANDOM_STRATEGY;
```
rcillo: I'm not sure, but if you make it an `Enum` instead of a `String` wouldn't it be validated by default?

I mean, if the user provides a value that is not in the `enum` then it would fail. I have to check, but if yes, then it would remove the need to validate manually.

rcillo: I have no preferences here. It's correct as is :+1:

rcillo:
:shipit:

src/main/java/de/zalando/aruha/nakadi/domain/EventType.java
v-stepanov: I also don't really have a preference here. I just thought that String will allow us to be more flexible in the future.

PR #211, opened 2016-04-08
Introduce version controller that provides access to /scm-source.json file

antban:
Generation of scm source is included to docker image build task.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/211" target="_blank">ReviewNinja

build.gradle
```diff
@@ -272,8 +272,12 @@ task startStoragesInDocker() {
     outputs.upToDateWhen { false }
 }
 
+task generateScmSourceJson(type: Exec) {
+    commandLine "scm-source", "-f", "scm-source.json"
```
v-stepanov: Build fails because it can't find 'scm-source' command. What this command is supposed to be?
https://travis-ci.org/zalando/nakadi/jobs/121664971

v-stepanov:
@antban 
build still fails https://travis-ci.org/zalando/nakadi/jobs/121683115

build.gradle
rcillo: I think we should only generate this `scm-source.json` when building images that we push to pierone. Those generated by travis just to run tests do not require this.

rcillo: So maybe we should add this to our private repo.

src/main/java/de/zalando/aruha/nakadi/controller/VersionController.java
```diff
@@ -0,0 +1,98 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping(value = "/version", produces = MediaType.APPLICATION_JSON)
+public class VersionController {
+
+    private static final String SCM_SOURCE_FILE = "/scm-source.json";
+
+    private static class ScmSource {
+        @JsonProperty("author")
```
rcillo: I think that there's not need for `@JsonProperty`. This should be correctly inferred by the mapper.

src/main/java/de/zalando/aruha/nakadi/controller/VersionController.java
```diff
@@ -0,0 +1,98 @@
...
+        @JsonProperty("status")
+        private String status;
+        @JsonProperty("url")
+        private String url;
+
+        public String getAuthor() {
+            return author;
+        }
+
+        public void setAuthor(String author) {
```
rcillo: I know this might seem silly, but I love adding `final` in every whenever possible.

v-stepanov: Yes, I think in all Java teams here it's an unofficial rule that everything that can be immutable/final should be immutable/final.

build.gradle
antban: I think that we need to build this file each time we create docker image:
 - scm-source is a public tool, and a public way to describe version (no need to move to private repo)
 - scm-source is a way to describe what's in box (so it's better to include version every time we build image, as it's a part of docker image).

src/main/java/de/zalando/aruha/nakadi/controller/VersionController.java
antban: ok

src/main/java/de/zalando/aruha/nakadi/controller/VersionController.java
antban: ok

build.gradle
```diff
@@ -272,8 +272,37 @@ task startStoragesInDocker() {
...
+    exec {
+        commandLine "git", "status", "--porcelain"
+        standardOutput = status
+    }
+    status = status.toString().trim().replaceAll("[\n\r]", "\\\\n")
+    if (status) {
+        rev = "$rev (locally modified)"
+    }
+    def scmSource = new FileOutputStream("scm-source.json")
+    scmSource.write("{\"url\": \"git:${url}\", \"revision\":\"${rev}\", \"author\": \"${System.getenv()['USER']}\", \"status\": \"${status}\"}".getBytes())
```
rcillo: Should the user set `USER` environment variable in order for this script to be successful? Or is it automatically setup somehow?

antban: On Mac and Linux it works.

rcillo: Sure, thanks :+1:

rcillo:
Looks good :+1: 

Do you think it would be good to add a test for the controller? Just to check that it reads from the correct file, parses correctly and provide the right information.

antban:
@rcillo Thinking about it

antban:
@rcillo Actually I can't do test without building docker image. And after that I need actually to run some kind of integration test, but I haven't seen such tests... Please fix me if I'am wrong

rcillo:
@antban I agree that generating the `scm-source.json` would be complicated, but maybe it would have some value to test the controller code. I thought about having a hardcoded `scm-source.json` file as a test resource, just like we do for events, for example `complex-event.json`.

v-stepanov:
@antban Then I think it's a good case for making an acceptance test. Our acceptance tests run against application built and running in docker container.
You can find some here: https://github.com/zalando/nakadi/tree/nakadi-jvm/src/acceptance-test

antban:
@v-stepanov Cool, I'll create it

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/VersionControllerAT.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.Test;
+import org.springframework.util.StringUtils;
+
+public class VersionControllerAT extends BaseAT {
+    private static final String VERSION_RESOURCE = "/version";
+    private static final ObjectMapper mapper = (new JsonConfig()).jacksonObjectMapper();
+
+    @Test(timeout = 10000)
+    public void scmSourceShouldContainCorrectData() {
+        given().when().get(VERSION_RESOURCE).then().statusCode(HttpStatus.SC_OK).body(is(new Matcher<Object>() {
```
rcillo: I didn't expected to have a custom matcher for these assertions. What it does it to check if the json structure matches the expected one, right? Maybe we could achieve it with some already existing json matcher.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/VersionControllerAT.java
```diff
@@ -0,0 +1,74 @@
...
+                }
+            }
+
+            @Override
+            public void describeMismatch(Object item, Description mismatchDescription) {
+                mismatchDescription.appendText("Error: " + error + " for " + item);
+            }
+
+            @Override
+            public void _dont_implement_Matcher___instead_extend_BaseMatcher_() {
```
v-stepanov: I think this method speaks for itself :) You should extend BaseMatcher, not implement Matcher.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/VersionControllerAT.java
v-stepanov: I agree with Ricardo at this point. We already have JsonPath library included in our project (https://github.com/jayway/JsonPath). I think you can use it here.

v-stepanov: And maybe you even don't need any specific Json matcher, but the standard one offered by restassured. Like:
`given().when()......then().assertThat().body("revision", notNullValue()).and() .....`

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/VersionControllerAT.java
antban: Removed that crap

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/VersionControllerAT.java
antban: Yep, my fault, fixed

rcillo:
Looks good for me. In my opinion there's only one merge conflict holding us from merging this.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/VersionControllerAT.java
```diff
@@ -0,0 +1,25 @@
...
+import de.zalando.aruha.nakadi.config.JsonConfig;
+import org.apache.http.HttpStatus;
+import static org.hamcrest.CoreMatchers.allOf;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import org.junit.Test;
+
+public class VersionControllerAT extends BaseAT {
+    private static final String VERSION_RESOURCE = "/version";
+    private static final ObjectMapper mapper = (new JsonConfig()).jacksonObjectMapper();
```
v-stepanov: not used, can be removed

antban: yep, fixed

src/main/java/de/zalando/aruha/nakadi/config/SecurityConfiguration.java
```diff
@@ -58,11 +58,13 @@ public void configure(final HttpSecurity http) throws Exception {
                     .antMatchers(POST, "/event-types/**").access(hasScope(eventTypeWriteScope))
                     .antMatchers(PUT, "/event-types/**").access(hasScope(eventTypeWriteScope))
                     .antMatchers(GET, "/health/**").permitAll()
+                    .antMatchers(GET, "/version/**").permitAll()
```
v-stepanov: Sorry, forgot to mention that when first looking at pull request. As you added an endpoint and defined security rules for it - please extend authentication tests to cover that.
https://github.com/zalando/nakadi/blob/nakadi-jvm/src/test/java/de/zalando/aruha/nakadi/security/BasicModeAuthenticationTest.java
https://github.com/zalando/nakadi/blob/nakadi-jvm/src/test/java/de/zalando/aruha/nakadi/security/FullModeAuthenticationTest.java
https://github.com/zalando/nakadi/blob/nakadi-jvm/src/test/java/de/zalando/aruha/nakadi/security/OffModeAuthenticationTest.java

antban: Added tests for /version url, refactored tests for authentication

src/test/java/de/zalando/aruha/nakadi/security/OffModeAuthenticationTest.java
```diff
@@ -1,46 +1,31 @@
...
+                .forEach(this::checkHasCorrectResponseStatus);
+    }
+
+    public void checkHasCorrectResponseStatus(Endpoint endpoint) {
+        try {
+            mockMvc.perform(endpoint.toRequestBuilder()).andExpect(STATUS_NOT_401_OR_403);
+        } catch (Exception e) {
+            throw new AssertionError("Exception when calling endpoint: " + endpoint, e);
+        } catch (AssertionError e) {
+            throw e;
```
v-stepanov: If you just throw exception here - then this block doesn't really make sense, because you just throw what you caught. What can add value here is to create and throw AssertionError that contains the endpoint (like few lines above), because if mockMvc check fails - it does not provide the information about the endpoint.

src/test/java/de/zalando/aruha/nakadi/security/AuthenticationTest.java
```diff
@@ -165,9 +173,19 @@ public void setup() {
...
+        endpointsWithNoScope.forEach(this::checkOkResponse);
+    }
+
+    private void checkOkResponse(Endpoint endpoint) {
+        try {
+            mockMvc.perform(endpoint.toRequestBuilder()).andExpect(status().isOk());
+        } catch (Exception e) {
+            throw new AssertionError("Exception occurred while calling endpoint: " + endpoint, e);
+        } catch (AssertionError e) {
+            throw e;
```
v-stepanov: I left my comment about the similar stuff in OffModeAuthenticationTest

v-stepanov: Acceptance tests fail in travis https://travis-ci.org/zalando/nakadi/jobs/122999756
It timed out at the place where it waits for nakadi-container. Maybe be it is somehow connected with changes you did in Makefile or gradle. On my PC the build and tests work fine for your branch. I think you can first try to relaunch the build on travis, maybe that was some travis problem.

v-stepanov:
Acceptance tests fail in travis https://travis-ci.org/zalando/nakadi/jobs/122999756
It timed out at the place where it waits for nakadi-container. Maybe be it is somehow connected with changes you did in Makefile or gradle. On my PC the build and tests work fine for your branch. I think you can first try to relaunch the build on travis, maybe that was some travis problem.

v-stepanov:
:shipit:

PR #212, opened 2016-04-08
Aruha 68 and 69: Enrichment

rcillo:
Applies enrichment to events. It mutates the original event object in
order not to create to many objects and keep things as fast as possible.

We only have one enrichment strategy at the moment, but it's done in a
way that would be possible to add new strategies or even change the
registry implementation to a dynamic one.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/212" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/enrichment/Enrichment.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Set;
+
+public class Enrichment {
+    private final EnrichmentsRegistry registry;
+
+    public Enrichment(final EnrichmentsRegistry registry) {
+        this.registry = registry;
+    }
+
+    public void validate(final EventType eventType) throws InvalidEventTypeException {
```
v-stepanov: there is no check that the enrichment strategy specified is known to Nakadi

src/main/java/de/zalando/aruha/nakadi/enrichment/EnrichmentsRegistry.java
```diff
@@ -0,0 +1,20 @@
...
+import com.google.common.collect.ImmutableMap;
+import de.zalando.aruha.nakadi.domain.EnrichmentStrategyDescriptor;
+import de.zalando.aruha.nakadi.domain.PartitionStrategyDescriptor;
+
+import java.util.List;
+import java.util.Map;
+
+public class EnrichmentsRegistry {
+    private static final Map<EnrichmentStrategyDescriptor, EnrichmentStrategy> ENRICHMENT_STRATEGIES =
+        ImmutableMap.<EnrichmentStrategyDescriptor, EnrichmentStrategy>builder()
```
v-stepanov: Why don't you use ImmutableMap.of() here? It would save you few lines on code

src/main/java/de/zalando/aruha/nakadi/service/EventPublisher.java
```diff
@@ -65,6 +74,20 @@ public EventPublishResult publish(final JSONArray events, final String eventType
         }
     }
 
+    private void enrich(final List<BatchItem> batch, final EventType eventType) throws EnrichmentException {
+        for (BatchItem item : batch) {
+            try {
+                item.setStep(EventPublishingStep.ENRICHING);
+                enrichment.enrich(item.getEvent(), eventType);
```
v-stepanov: result of enrichment is ignored here (the return value)

rcillo: Actually there's no need to return the result since this method just updates the object. Do you think it would be better to change the interface not to return anything?

v-stepanov: Yes, I think there's no good reason to return anything as the method mutates the parameter you pass.

src/main/java/de/zalando/aruha/nakadi/enrichment/Enrichment.java
```diff
@@ -0,0 +1,49 @@
...
+        if (eventType.getCategory() == EventCategory.UNDEFINED && !eventType.getEnrichmentStrategies().isEmpty()) {
+            throw new InvalidEventTypeException("must not define enrichment strategy for undefined event type");
+        }
+
+        final Set<EnrichmentStrategyDescriptor> uniqueStrategies = new HashSet<EnrichmentStrategyDescriptor>();
```
v-stepanov: 1) Diamond <> can be used.
2) HashSet constructor can take collection as a parameter, so these two lines can be replaced with one.

src/main/java/de/zalando/aruha/nakadi/enrichment/Enrichment.java
rcillo: Since I'm using an `enum` for `EnrichmentStrategyDescriptor`, if the user provides an invalid value, then he will see a 400 answer.

```
{"type":"http://httpstatus.es/400","title":"Bad Request","status":400,"detail":"Illegal enum value: 'metadata_strategy'. Possible values: [metadata_enrichment] (through reference chain: de.zalando.aruha.nakadi.domain.EventType[\"enrichment_strategies\"]->java.util.ArrayList[0])"}
```

I know it's ugly to expose internals, but that's how spring does with it's validations.

v-stepanov: Ok, I see, so it's really validated.
About the exposing externals - that's the trade-off of having enum and automatic spring validation :)

src/main/java/de/zalando/aruha/nakadi/service/EventPublisher.java
rcillo: Done :+1: https://github.com/zalando/nakadi/pull/212/commits/efa3a156d610e2bba86dd2654f7e6a52083963f9

src/main/java/de/zalando/aruha/nakadi/enrichment/EnrichmentsRegistry.java
rcillo: Done :+1: https://github.com/zalando/nakadi/pull/212/commits/efa3a156d610e2bba86dd2654f7e6a52083963f9

src/main/java/de/zalando/aruha/nakadi/enrichment/Enrichment.java
rcillo: Done :+1: https://github.com/zalando/nakadi/pull/212/commits/efa3a156d610e2bba86dd2654f7e6a52083963f9

src/main/java/de/zalando/aruha/nakadi/enrichment/EnrichmentStrategy.java
```diff
@@ -0,0 +1,9 @@
+package de.zalando.aruha.nakadi.enrichment;
+
+import de.zalando.aruha.nakadi.domain.EventType;
+import de.zalando.aruha.nakadi.exceptions.EnrichmentException;
+import org.json.JSONObject;
+
+public interface EnrichmentStrategy {
+    JSONObject enrich(JSONObject event, EventType eventType) throws EnrichmentException;
```
v-stepanov: Here as well, as the original event will be enriched I don't see any value to also return it. When I look at such method it makes me think that original event will not be modified (as it has a return value).

src/main/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategy.java
```diff
@@ -0,0 +1,39 @@
...
+            setFlowId(metadata);
+        } catch (final JSONException e) {
+            throw new EnrichmentException("enrichment error", e);
+        }
+
+        return event;
+    }
+
+    private void setFlowId(final JSONObject metadata) {
+        metadata.put("flow_id", FlowIdUtils.peek());
```
v-stepanov: I didn't really look deep into it, but how do we ensure that we peek a right flowId here? At the end it's a static stack, right? Or is it somehow separated for each thread?

src/test/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategyTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Test
+    public void setReceivedAtWithSystemTimeInUTC() throws Exception {
+        final EventType eventType = buildDefaultEventType();
+        final JSONObject event = buildBusinessEvent();
+
+        assertThat(event.getJSONObject("metadata").optString("received_at"), isEmptyString());
+
+        try {
+            org.joda.time.DateTimeUtils.setCurrentMillisFixed(0);
```
v-stepanov: Why use a full-name import here? I don't see any conflicts.

src/main/java/de/zalando/aruha/nakadi/enrichment/EnrichmentStrategy.java
rcillo: Sure. Done https://github.com/zalando/nakadi/pull/212/commits/d4383c503da333b9573c8fd0e739cc1f504304d9

src/main/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategy.java
rcillo: That's is good question. I did some research and discovered that it's implemented on top of log4j NDC. Reading NDC doc it states that it's isolated on a thread level. Since the thread that is executing this line is the request's thread, we are guaranteed to get the correct flow id.

src/test/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategyTest.java
rcillo: Done https://github.com/zalando/nakadi/pull/212/commits/5efc443e400fdde9ea1fef96124e48a89da9a454

v-stepanov:
:shipit:

rcillo:
Could you please take another look at this PR. I just need to merge `nakadi-jvm` in order to fix conflicts.

src/main/java/de/zalando/aruha/nakadi/config/NakadiConfig.java
```diff
@@ -75,6 +77,9 @@ public EventStreamFactory eventStreamFactory() {
     }
 
     @Bean
+    public Enrichment enrichment() { return new Enrichment(new EnrichmentsRegistry()); }
```
v-stepanov: I know we don't have coding guidelines, but I believe our unofficial approach is to split this to three lines.

rcillo: Looking at the rest of this file it doesn't seem to follow any pattern.

PR #213, opened 2016-04-12
fix for test that was failing sometimes

v-stepanov:
Dmitriy found out that this test can fail. I believe it fails if it runs as the very first one after the kafka topic was created. This is a fix.

### DIFF
--- a/src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/PartitionsControllerAT.java
+++ b/src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/PartitionsControllerAT.java
@@ -1,6 +1,7 @@
 package de.zalando.aruha.nakadi.webservice;
 
 import com.jayway.restassured.response.Response;
+import de.zalando.aruha.nakadi.domain.Cursor;
 import de.zalando.aruha.nakadi.repository.kafka.KafkaTestHelper;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/213" target="_blank">ReviewNinja

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/PartitionsControllerAT.java
```diff
@@ -141,11 +142,16 @@ public void whenGetPartitionAndWriteMessageThenOffsetInPartitionIsIncreased() th
...
-        final long offsetBefore = Long.parseLong(partitionInfoBefore.get("newest_available_offset"));
-        final long offsetAfter = Long.parseLong(partitionInfoAfter.get("newest_available_offset"));
+        final long offsetBefore = getNewestOffsetAsLong(partitionInfoBefore);
+        final long offsetAfter = getNewestOffsetAsLong(partitionInfoAfter);
         assertThat(offsetAfter, is(offsetBefore + delta));
     }
 
+    private Long getNewestOffsetAsLong(final Map<String, String> partitionInfo) {
+        final String offset = partitionInfo.get("newest_available_offset");
+        return Cursor.BEFORE_OLDEST_OFFSET.equals(offset) ? -1 : Long.parseLong(offset);
```
rcillo: Since I had some hard time to understand this, maybe we should just add some comment in the lines of "Handles the case of empty partition, when the current offset would be "BEGIN".

antban: And actually I think that we need to have two tests - one for testing BEGIN and one for testing real values like 0 or 1

v-stepanov: @rcillo I see your point, but to be consistent - then we should add the same comment to all other places in code where BEFORE_OLDEST_OFFSET is used.

@antban We agreed not to have too much acceptance tests, but try to cover more with unit tests. And in unit test all cases regarding offsets are covered (it's in https://github.com/zalando/nakadi/blob/nakadi-jvm/src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepositoryTest.java)

rcillo: @v-stepanov ok. I remember the pain of deciding for `BEGIN`. The effort of educating everyone is lower than commenting everywhere.

PR #215, opened 2016-04-13
fix_success_metric Fix metric names and usage for unsuccessful requests

antban:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/215" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -47,18 +42,27 @@ public EventPublishingController(final EventPublisher publisher, final EventType
...
         LOG.trace("Received event {} for event type {}", eventsAsString, eventTypeName);
+        final EventTypeMetrics eventTypeMetrics = eventTypeMetricRegistry.metricsFor(eventTypeName);
 
         try {
-            final EventTypeMetrics eventTypeMetrics = eventTypeMetricRegistry.metricsFor(eventTypeName);
+            final ResponseEntity response = postEventInternal(eventTypeName, eventsAsString, nativeWebRequest, eventTypeMetrics);
+            eventTypeMetrics.incrementResponseCount(response.getStatusCode().value());
+            return response;
+        } catch (RuntimeException ex) {
+            eventTypeMetrics.incrementResponseCount(-1); // yep, dunno response code here, related to event type.
```
rcillo: This should throw a 500, doesn't it?

antban: On this stage I can not say anything about response code, because it will be mapped later.
On the other side - InternalNakadiException creates 500 status code, and I need to separate it from unknown status code.

rcillo: Maybe we should check. I think that every uncaught exception is handled as 500.

v-stepanov: Yes, I think Spring turns every uncaught exception into 500. I think it would be correct to consider that here we have 500 response code.

v-stepanov:
👍

PR #217, opened 2016-04-18
Calculate partition count using statistics object

antban:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/217" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/domain/EventTypeStatistics.java
```diff
@@ -0,0 +1,73 @@
+package de.zalando.aruha.nakadi.domain;
+
+import java.util.Objects;
+import javax.validation.constraints.Min;
+import javax.validation.constraints.NotNull;
+
+public class EventTypeStatistics {
+    @NotNull
+    @Min(value = 0, message = "expected_write_rate can't be less then zero")
+    private Integer expectedWriteRate; // messages per minute
```
v-stepanov: A need of comment here shows that the field name is not informative enough. I suggest to transform comment into field name.
In my opinion the same can be applied to the next field (it can be something like `messageSizeInBytes`)

api/nakadi-event-bus-api.yaml
```diff
@@ -1250,7 +1250,7 @@ definitions:
           Write rate for events of this EventType. This rate encompasses all producers of this EventType for
           a Nakadi cluster.
 
-          Measured in kb/minutes.
```
rcillo: Should we state that we expect a mean event size of 10kb or 1kb?

antban: there is another parameter in EventTypeStatistics - message_size

src/main/java/de/zalando/aruha/nakadi/domain/EventTypeStatistics.java
antban: ok, I'll change it

v-stepanov:
@antban Just a general comment. You miss `final` in a lot of places :)
Can you please review your code and add that where needed.

src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaPartitionsCalculatorTest.java
```diff
@@ -0,0 +1,128 @@
...
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
+import static org.hamcrest.Matchers.lessThanOrEqualTo;
+import static org.hamcrest.Matchers.notNullValue;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class KafkaPartitionsCalculatorTest {
+
+    private static final String TEST_DATA = "[" +
```
v-stepanov: I would suggest to move this to a separate json-file. (I think we already have json-files for some of the tests)
Json inside Java string looks really ugly :)

src/main/java/de/zalando/aruha/nakadi/domain/EventTypeStatistics.java
rcillo: @v-stepanov I think this name comes from the API. If we conclude that API names are not clear, then I would suggest to change it as well.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -68,6 +68,7 @@ public EventTypeController(final EventTypeRepository eventTypeRepository,
     }
 
     @RequestMapping(method = RequestMethod.POST)
+    @SuppressWarnings("unused")
```
v-stepanov: Which warning do you try to suppress with this?

src/main/java/de/zalando/aruha/nakadi/domain/EventTypeStatistics.java
v-stepanov: @rcillo 
Sure, that will simplify life of our users.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaConfig.java
```diff
@@ -31,8 +34,15 @@ public KafkaFactory kafkaFactory() {
     }
 
     @Bean
-    public KafkaTopicRepository kafkaTopicRepository() {
-        return new KafkaTopicRepository(zookeeperConfig.zooKeeperHolder(), kafkaFactory(), kafkaRepositorySettings());
+    public KafkaPartitionsCalculator kafkaPartitionsCalculator(
+            @Value("${nakadi.kafka.instanceType}") final String instainceType,
```
v-stepanov: `instainceType` typo

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
antban: ok, I'll change

src/main/java/de/zalando/aruha/nakadi/domain/EventTypeStatistics.java
```diff
@@ -0,0 +1,73 @@
...
+    @NotNull
+    @Min(value = 1, message = "message_size can't be less then 1")
+    private Integer messageSize; // bytes
+    @NotNull
+    @Min(value = 1, message = "problem with read_parallelism: at least one reader expected")
```
rcillo: Is it necessary to state the field name where there was a problem? I think this is already present by default in the final json problem message. In other places in the code we don't have this, for example `EventType.name` message is `format not allowed`.

src/main/java/de/zalando/aruha/nakadi/domain/EventType.java
```diff
@@ -44,6 +42,9 @@
     @NotNull
     private EventTypeSchema schema;
 
+    @Valid
+    private EventTypeStatistics defaultStatistic;
```
v-stepanov: I think we forgot to tell you:
We have an agreement that if something can be null it should be markes as `@Nullable`
(so everything is none-null by default, and we avoid having @Nonnull everywhere)

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaPartitionsCalculator.java
```diff
@@ -0,0 +1,150 @@
...
+            return getBestPartitionsCount(ceil.getValue(), mbsPerSecond);
+        } else if (ceil == null) {
+            return getBestPartitionsCount(floor.getValue(), mbsPerSecond);
+        } else {
+            final int floorResult = getBestPartitionsCount(floor.getValue(), mbsPerSecond);
+            if (Objects.equals(floor.getKey(), ceil.getKey())) {
+                return floorResult;
+            }
+            final int ceilResult = getBestPartitionsCount(ceil.getValue(), mbsPerSecond);
+            return floorResult + (ceilResult - floorResult) * (messageSize - floor.getKey()) / (ceil.getKey() - floor.getKey());
```
rcillo: Have you considered other rounding strategies? Do you think it would be worth checking?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -162,8 +162,7 @@ else if (Cursor.BEFORE_OLDEST_OFFSET.equals(pInfo.getNewestAvailableOffset())) {
                         .orElse(false));
     }
 
-    @Override
-    public void postEvent(final String topicId, final String partitionId, final String payload) throws ServiceUnavailableException {
+    void postEvent(final String topicId, final String partitionId, final String payload) throws ServiceUnavailableException {
```
rcillo: I forgot to remove this method. I know it's out of the scope of this task, but if you want to remove it, feel free.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -359,4 +357,22 @@ private void doWithZkUtils(final ZkUtilsAction action) throws Exception {
...
+
+    private Integer calculateKafkaPartitionCount(final EventTypeStatistics stat) {
+        if (null == stat) {
+            return settings.getDefaultTopicPartitionCount();
+        }
+        final int maxPartitionsDueParallelism = Math.max(stat.getReadParallelism(), stat.getWriteParallelism());
+        if (maxPartitionsDueParallelism >= settings.getMaxTopicPartitionCount()) {
+            return settings.getMaxTopicPartitionCount();
+        }
+        return Math.min(8, Math.max(
```
rcillo: Aren't lines 365-368 redundant?

```java
private Integer calculateKafkaPartitionCount(final EventTypeStatistics stat) {
        if (null == stat) {
             return settings.getDefaultTopicPartitionCount();
        }
        final int maxPartitionsDueParallelism = Math.max(stat.getReadParallelism(), stat.getWriteParallelism());
        return Math.min(settings.getMaxTopicPartitionCount(), Math.max(
                                  maxPartitionsDueParallelism,
                                  calculatePartitionsAccordingLoad(stat.getExpectedWriteRate(), stat.getMessageSize())));
```

src/main/resources/partitions_statistics.json
```diff
@@ -0,0 +1,31 @@
+[
+  {
+    "name": "t2.large",
```
rcillo: I know that this experiment result has been around for a while, but I think we are not gonna use this sorte of instance outside of our sandbox environment, probably. We are currently experimenting with the following instances types:

* M3
        - m3.large
        - m3.xlarge
        - m3.2xlarge
* C4
       - c4.large
       - c4.xlarge
       - c4.2xlarge
       - c4.4xlarge

Should I reconsider the instance types in the experiment or should we create a new ticket in order  to reevaluate the present statistics?

src/main/java/de/zalando/aruha/nakadi/repository/TopicRepository.java
```diff
@@ -2,30 +2,27 @@
...
  */
 public interface TopicRepository {
 
     List<Topic> listTopics() throws NakadiException;
 
-    void createTopic(String topic) throws TopicCreationException, DuplicatedEventTypeNameException;
-
-    void createTopic(String topic, int partitionsNum, int replicaFactor, long retentionMs, long rotationMs)
-            throws TopicCreationException, DuplicatedEventTypeNameException;
+    void createTopic(String topic, EventTypeStatistics defaultStatistics) throws TopicCreationException, DuplicatedEventTypeNameException;
```
rcillo: I would be ok with changing the API to receive the whole `EventType` object instead of adding a separate parameter. Wdyt?

src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaPartitionsCalculatorTest.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test
+    public void testLoadNullForUnknownName() throws IOException {
+        try {
+            KafkaPartitionsCalculator.load(objectMapper, "null", getTestStream());
+            Assert.fail("Unknown calculator must throw IllegalArgumentException only");
+        } catch (IllegalArgumentException ignored) {
+        } catch (RuntimeException ex) {
+            Assert.fail("Unknown calculator must throw IllegalArgumentException only, but found: " + ex.getClass().getName());
```
rcillo: Maybe we could use `@Test(expected = IllegalArgumentException.class)` ?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaPartitionsCalculator.java
antban: I think that this rounding is enough. The best we can do is to get better statistics.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
antban: ok

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
antban: nope, they are not redundant. I just do not want to do calculations if they are not needed. In case you provided `calculatePartitionsAccordingLoad` will be called even if it won't work at all. But I can change it anyway.
The problem here is that I forgot to replace `8` with `getMaxTopicPartitionCount` call

src/main/resources/partitions_statistics.json
antban: In current version it was the only one statistics I have. I decided to put all the knowledge it into code, so we can review it and select correct values just replacing configuration string. When we will get new statistics - we will just put it here.

src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaPartitionsCalculatorTest.java
antban: ok

src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaPartitionsCalculatorTest.java
antban: yep, it's the best way.

src/main/java/de/zalando/aruha/nakadi/repository/TopicRepository.java
antban: I do not want to be in situation, where I can't understand which data exactly used for this call.

antban:
@v-stepanov @rcillo Guys, do you have any other comments?

rcillo:
@antban looks good for me 👍 

:shipit:

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -189,6 +189,18 @@ private void validateUpdate(final String name, final EventType eventType) throws
 
         validateName(name, eventType);
         validateSchemaChange(eventType, existingEventType);
+        eventType.setDefaultStatistic(
+                validateStatisticsUpdate(existingEventType.getDefaultStatistic(), eventType.getDefaultStatistic()));
+    }
+
+    private EventTypeStatistics validateStatisticsUpdate(final EventTypeStatistics existing, final EventTypeStatistics newStatistics) throws InvalidEventTypeException {
```
v-stepanov: what exactly this method does?

antban: This method is used on updating event type, and if no default_statistics provided, tries to leave previous one. If default_statistics is provided - checks that it's the same as it was on creation.

PR #220, opened 2016-04-20
ARUHA-172: Made java opts configurable through senza file and dockerfile

v-stepanov:
@rcillo @antban please review
That will give the possibility to set Xmx/Xms (and other jvm params) when deploying nakadi

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,5 +9,5 @@ ADD scm-source.json /scm-source.json
 EXPOSE 8080
 
 # run the server when a container based on this image is being run
-ENTRYPOINT java -jar -Djava.security.egd=file:/dev/./urandom nakadi.jar
+ENTRYPOINT java $JAVA_OPTS -jar -Djava.security.egd=file:/dev/./urandom nakadi.jar
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,5 +9,5 @@ ADD scm-source.json /scm-source.json
 EXPOSE 8080
 
 # run the server when a container based on this image is being run
-ENTRYPOINT java $JAVA_OPTS -jar -Djava.security.egd=file:/dev/./urandom nakadi.jar
+ENTRYPOINT java $JAVA_OPTS -Djava.security.egd=file:/dev/./urandom -jar nakadi.jar
 
--- a/senza.yaml
+++ b/senza.yaml
@@ -54,6 +54,9 @@ SenzaInfo:
     - ReplicaFactor:
         Description: "Replica factor for event broker"
         Default: "3"
+    - JavaOpts:
+        Description: "Java opts to be passed to jvm"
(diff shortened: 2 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/220" target="_blank">ReviewNinja

Dockerfile
```diff
@@ -9,5 +9,5 @@ ADD scm-source.json /scm-source.json
 EXPOSE 8080
 
 # run the server when a container based on this image is being run
-ENTRYPOINT java -jar -Djava.security.egd=file:/dev/./urandom nakadi.jar
+ENTRYPOINT java $JAVA_OPTS -jar -Djava.security.egd=file:/dev/./urandom nakadi.jar
```
rcillo: Shouldn't it be possible to do through environment variables?

antban: Actually -Djava.security.egd=file:/dev/./urandom is one of that java options. Strange to put it on different sides of -jar

v-stepanov: @rcillo not that I'm aware of. How is it possible to do that?

@antban Yes, it's one of these options, but I don't want to make it configurable I think it should always be there. I can move it to the left from -jar if you prefer :)

v-stepanov: changed params order

rcillo: @v-stepanov from my limited knowledge of JVM it's just a matter of exporting an environment variable, like `JAVA_OPTS='-Djava.awt.headless=true'` and new java processes would use it. Makes sense?

Then one could pass these variables as parameters to the container.

v-stepanov: @rcillo Not exactly. Jvm doesn't check for JAVA_OPTS environment variable. It is usually done by scripts that run jvm (in tomcat, ant etc.)
At least this is what I've read on stackoverflow and my local test confirmed that.

rcillo: It looks like that Oracle JVM uses another env variable https://docs.oracle.com/javase/7/docs/webnotes/tsg/TSG-VM/html/envvars.html#gbmsy

v-stepanov: But still there's no guarantee that it will supported in other JVMs. I would feel more safe with passing it directly when running java. This is an approach a lot of java apps use.

rcillo: Sure 👍 I have nothing against the proposed approach. Since it's still possible to use the environment variable, I think it's not an issue to add this one here.

antban: @v-stepanov My point was only about the order
@rcillo We have two different calls for getting different variables:
```
        System.getenv("name") // gets system dependent environment variable
        System.getProperty("name") // gets system property passed using -D
```
Probably tool, that you provided link to tries to get variables from different places.
For our purposes: we need -X and other switches, that can't be passed using environment vairables.

PR #227, opened 2016-04-29
aruha-175 Improved error message for invalid offset streaming

### LINKED ISSUE #158: Reading events: Content-Type missing on HTTP 412 precondition failed
An invalid query (partition with `newest_available_offset = BEGIN`) on `[...]/events` leads to HTTP 412. The response does not contain a `Content-Type` header, however the values are matching the problem type (`application/problem+json`).

```
$ http --stream http://localhost:8080/event-types/order.ORDER_CANCELLED/events X-Nakadi-Cursors:"[{\"partition\": \"1\", \"offset\": \"0\"}]"
HTTP/1.1 412 Precondition Failed
Cache-Control: no-cache, no-store, max-age=0, must-revalidate
Content-Length: 111
Date: Sat, 05 Mar 2016 23:28:19 GMT
Expires: 0
Pragma: no-cache
Server: Apache-Coyote/1.1
X-Content-Type-Options: nosniff
X-Flow-Id: 1PoiUMR1oA34KtxvbJirPHhY
X-Frame-Options: DENY
X-XSS-Protection: 1; mode=block

{"type":"http://httpstatus.es/412","title":"Precondition Failed","status":412,"detail":"cursors are not valid"}
```

rcillo:
Some users were confused when the provided offsets were resulting in
error. So instead of providing a boolean answer we now display more
detailed reasons for:

1. out of bounds offset
2. invalid offset value (nonnumerical nor "BEGIN")
3. empty partition
4. inexistent partition.

Also fixes #158

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
@@ -180,7 +180,7 @@ public class EventStreamController {
             case EMPTY_PARTITION:
                 return "partition " + cursor.getPartition() + " is empty";
             case UNAVAILABLE:
-                return "offset " + cursor.getOffset() + " for partition " + cursor.getPartition() + " unavailable";
+                return "offset " + cursor.getOffset() + " for partition " + cursor.getPartition() + " is unavailable";
             default:
                 return "invalid offset " + cursor.getOffset() + " for partition " + cursor.getPartition();
         }
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
@@ -189,6 +189,7 @@ public class EventStreamController {
     private void writeProblemResponse(final HttpServletResponse response, final OutputStream outputStream,
                                       final Response.StatusType statusCode, final String message) throws IOException {
         response.setStatus(statusCode.getStatusCode());
+        response.setContentType("application/json+problem");
         jsonMapper.writer().writeValue(outputStream, Problem.valueOf(statusCode, message));
     }
 }
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
@@ -189,7 +189,7 @@ public class EventStreamController {
     private void writeProblemResponse(final HttpServletResponse response, final OutputStream outputStream,
                                       final Response.StatusType statusCode, final String message) throws IOException {
         response.setStatus(statusCode.getStatusCode());
-        response.setContentType("application/json+problem");
+        response.setContentType("application/problem+json");
         jsonMapper.writer().writeValue(outputStream, Problem.valueOf(statusCode, message));
     }
 }
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
@@ -6,6 +6,8 @@ import com.codahale.metrics.annotation.Timed;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import de.zalando.aruha.nakadi.domain.Cursor;
+import de.zalando.aruha.nakadi.domain.CursorError;
+import de.zalando.aruha.nakadi.exceptions.InvalidCursorException;
 import de.zalando.aruha.nakadi.exceptions.NakadiException;
 import de.zalando.aruha.nakadi.repository.EventConsumer;
 import de.zalando.aruha.nakadi.repository.TopicRepository;
@@ -33,7 +35,6 @@ import java.io.OutputStream;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.stream.Collectors;
 
 import static de.zalando.aruha.nakadi.metrics.MetricUtils.metricNameFor;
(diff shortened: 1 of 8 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/227" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -120,9 +121,12 @@ public StreamingResponseBody streamEvents(
                 }
 
                 // check that offsets are not out of bounds and partitions exist
-                if (cursors != null && !topicRepository.areCursorsValid(topic, cursors)) {
-                    writeProblemResponse(response, outputStream, PRECONDITION_FAILED, "cursors are not valid");
-                    return;
+                if (cursors != null) {
+                    final Optional<String> errorMessage = topicRepository.validateCursors(topic, cursors);
```
antban: For me it's incorrect layering.
TopicRepository is something that can throw exception or provide some error code.
Controller instead is something that can create user message.

rcillo: @antban if I understood, your suggestion would be to create 4 types of exceptions, capture then here and map then to different error messages?

antban: Well, actually no. My suggestion is to remove `areCursorsValid` method from topic repository interface, create Exception type `CursorInvalidException` with property reason, and validate cursor while creating consumers, throwing exceptions sometimes.

antban: And in EventStreamController take care of creating correct error message according to catched CursorInvalidException.

rcillo: @antban thank you for the clarification 👍 sounds pretty much better 😄

rcillo: @antban I did some changes. I did not refactored other code that uses cursos. I just changed the validation aspect of it. Tried to move things around in order to get a better design as by your suggestions.

src/test/java/de/zalando/aruha/nakadi/controller/EventStreamControllerTest.java
```diff
@@ -214,12 +216,12 @@ public void whenNoCursorsThenLatestOffsetsAreUsed() throws NakadiException, IOEx
...
     @Test
-    public void whenNormalCaseThenParametersArePassedToConfigAndStreamStarted() throws IOException, NakadiException {
+    public void whenNormalCaseThenParametersArePassedToConfigAndStreamStarted() throws Exception {
         final EventConsumer eventConsumerMock = mock(EventConsumer.class);
         when(topicRepositoryMock.topicExists(eq(TEST_EVENT_TYPE))).thenReturn(true);
-        when(topicRepositoryMock.areCursorsValid(eq(TEST_EVENT_TYPE), eq(ImmutableList.of(new Cursor("0", "0")))))
-                .thenReturn(true);
-        when(topicRepositoryMock.createEventConsumer(eq(TEST_EVENT_TYPE), eq(ImmutableMap.of("0", "0"))))
+//        when(topicRepositoryMock.validateCursors(eq(TEST_EVENT_TYPE), eq(ImmutableList.of(new Cursor("0", "0")))))
+//                .thenReturn(Optional.empty());
```
antban: I think that we need to remove comments

antban:
de.zalando.aruha.nakadi.webservice.EventStreamReadingAT > whenInvalidCursorsThenPreconditionFailed FAILED

rcillo:
@antban @v-stepanov could you please review it again. Tests are passing 👍

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -173,11 +173,23 @@ public StreamingResponseBody streamEvents(
...
     }
 
+    private String invalidCursorMessage(final CursorError error, final Cursor cursor) {
+        switch (error) {
+            case PARTITION_NOT_FOUND:
+                return "non existing partition " + cursor.getPartition();
+            case EMPTY_PARTITION:
+                return "partition " + cursor.getPartition() + " is empty";
+            case UNAVAILABLE:
+                return "offset " + cursor.getOffset() + " for partition " + cursor.getPartition() + " unavailable";
```
v-stepanov: **is** unavailable ?

rcillo: From my understanding it would change the syntactical function of it from adnominal to adverbial, but would have the same semantical meaning. But I'm always not so sure which version would sound better for a native speaker or even if it's correct as is.

rcillo: Maybe I should move it to closer to the noun, like "unavailable offset for partition".

v-stepanov: I think I don't really get your point. Probably it's just too complicated for me :)
My point is that `unavailable` is an adjective and for me `offset 5482 for partition 2 unavailable` sounds the same as for example `partition 3 empty` or `rose red`

v-stepanov: "unavailable offset X for partition Y" sounds also good for me

v-stepanov:
:shipit:

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
rcillo: Sure. But I did some research and it seems that "is unavailable" could be better since it emphasises the transience of the state. It means that it's unavailable now, but could had been available in the past or even might be available in the future. So I changed it 👍

PR #228, opened 2016-04-29
aruha-136 Expose effective schema

rcillo:
When publishing events whose `EventType` category is either "business"
or "data", additional fields are required other than those described in
the schema.

This endpoint exposes this effective schema, e.g. the schema used to
validate events in fact.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -394,7 +394,7 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
-  /event-types/{name}/effective_schema:
+  /event-types/{name}/effective-schema:
     get:
       tags:
         - schema-registry-api
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -141,6 +141,8 @@ paths:
     get:
       tags:
         - schema-registry-api
+      security:
+        - oauth2: ['uid']
       description: Returns a list of all registered `EventType`s
       parameters:
         - name: X-Flow-Id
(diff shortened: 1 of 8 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/228" target="_blank">ReviewNinja

src/test/java/de/zalando/aruha/nakadi/utils/TestUtils.java
```diff
@@ -89,6 +89,10 @@ public static String resourceAsString(final String resourceName, final Class cla
         return IOUtils.toString(clazz.getResourceAsStream(resourceName));
     }
 
+    public static String getStringFromFile(final String resourceName) throws IOException {
```
v-stepanov: I remember we do it in several places in code already, so your method will be a god use there.
Can you please search for `Resources.toString` and replace what you found with this new `getStringFromFile` method you created.

api/nakadi-event-bus-api.yaml
```diff
@@ -392,6 +394,46 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+  /event-types/{name}/effective_schema:
```
v-stepanov: 1) For me this endpoint looks not completely RESTful. Were there any alternatives for endpoint URL?
2) Why do you use underscore here? I think for other places we use dashes (e.g. `event-types`)

rcillo: The alternative would be to add a new field to the event-type. We would have to remember updating it every time something changes and also do some data migration in case we change the structure of the effective schema.

Regarding second point, it's just that it's my first contribution to the api. I'll change it.

@jkliff do you have any opinions on this one?

v-stepanov: Yeah I don't really like the alternative with having another field for event type :(
But maybe we can ask API guild about that?

api/nakadi-event-bus-api.yaml
```diff
@@ -392,6 +394,46 @@ paths:
...
+        - name: X-Flow-Id
+          in: header
+          description: |
+            The flow id of the request, which is written into the logs and passed to called services. Helpful
+            for operational troubleshooting and log analysis.
+          type: string
+          format: flow-id
+      responses:
+        '200':
+          description: Ok
```
antban: Is there any other description of schema object?

antban: I mean that there is not schema provided and no description of returned object type.

rcillo: The description would be the same as above. I don't know what else to say here.

rcillo: Wdyt of:

"A JSON with the schema used for validating events".

antban: Exactly!

rcillo: Done 👍

rcillo:
This will be addressed in another opportunity. We have other priorities now.

PR #231, opened 2016-05-02
aruha-181 Set eid in the response when present

rcillo:
In order to make it easier for producers to correlate a response with an
event we are setting the eid as a property in the batch item response.

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
+++ b/src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
@@ -15,7 +15,7 @@ public class BatchItem {
 
         Optional.
                 ofNullable(event.optJSONObject("metadata"))
-                .map(e -> e.optString("eid"))
+                .map(e -> e.optString("eid", null))
                 .ifPresent(this.response::setEid);
     }
 
--- a/src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
+++ b/src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
@@ -14,7 +14,7 @@ public class BatchItem {
         this.event = event;
 
         Optional.
-                of(event.optJSONObject("metadata"))
+                ofNullable(event.optJSONObject("metadata"))
                 .map(e -> e.optString("eid"))
                 .ifPresent(this.response::setEid);
     }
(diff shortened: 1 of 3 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/231" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
```diff
@@ -13,6 +12,11 @@
     public BatchItem(final JSONObject event) {
         this.response = new BatchItemResponse();
         this.event = event;
+
+        Optional.
+                of(event.optJSONObject("metadata"))
```
v-stepanov: That will throw `NullPointerException` if `event.optJSONObject("metadata")` is null.
Better use `Optional.ofNullable`.

rcillo: Fixed 👍  https://github.com/zalando/nakadi/pull/231/commits/1b2aaead761a63afbe8edc92051fbf3f06eaa304

You were faster than CI to spot that out.

src/main/java/de/zalando/aruha/nakadi/domain/BatchItem.java
```diff
@@ -13,6 +12,11 @@
...
+                .map(e -> e.optString("eid"))
```
v-stepanov: `e.optString("eid")` will return empty string is "eid" field is not found. As result the next statement (ifPresent(....)) will always be executed.
I think if you change it like `.map(e -> e.optString("eid", null))` it will work correct

rcillo: Then I think we need to improve the api definition. It was not clear for me if we should reply with empty string of null. For example, details is empty but never null.

v-stepanov: But it will not be null. `map` method will treat `null` as empty and code inside `ifPresent` will not be executed. And `eid` will stay with it's default value (empty string) :)

rcillo: Done https://github.com/zalando/nakadi/pull/231/commits/6c756289edb9203bff349266134042178e9f3ba9 👍

v-stepanov:
:shipit:

PR #234, opened 2016-05-10
aruha-175 cursors partition and offset must not be null

rcillo:
As reported during validation, when passing null partition or offsets, 500 was returned instead of 412.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/234" target="_blank">ReviewNinja

v-stepanov:
looks good to me :shipit:

PR #235, opened 2016-05-10
aruha-151 fix definitions on data change events

rcillo:
Since we wrapping the original json schema inside a wrapper schema,
references to definitions in the root of the provided JSON were broken.

This commit fixes this issue by moving definitions to the root of the
wrapper json.

If there are other definitions at other levels of the provided json
schema, that would stil be an unsolved issue.

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
+++ b/src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
@@ -45,6 +45,8 @@ public class EventValidation {
 
         addMetadata(wrapper, eventType);
 
+        moveDefinitionsToRoot(wrapper, schema);
+
         final JSONObject properties = wrapper.getJSONObject("properties");
 
         properties.put("data_type", new JSONObject().put("type", "string"));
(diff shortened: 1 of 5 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/235" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/validation/EventValidation.java
```diff
@@ -59,6 +61,14 @@ private static JSONObject wrapSchemaInData(final JSONObject schema, final EventT
         return wrapper;
     }
 
+    private static void moveDefinitionsToRoot(final JSONObject wrapper, final JSONObject schema) {
```
v-stepanov: I think we should have a test for the functionality that is performed in this method.

rcillo: I didn't do so because I wanted to avoid coupling tests to the implementation. I added a test case that shows that "definitions" should work without worrying about what goes on in the internals of validation.

I could improve test case for the effective schema task once it's merged. Wdyt?

v-stepanov: Ah ok. It's actually tested by the test you added. Then good.

v-stepanov:
:shipit:

PR #236, opened 2016-05-11
ARUHA-117 Fix documentation regarding event_type_statistics, fix overflow error.

antban:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/236" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -1202,16 +1202,12 @@ definitions:
 
       default_statistics:
         type: object
+        allOf:
```
v-stepanov: why not just:
```
default_statistics:
  $ref: '#/definitions/EventTypeStatistics'
```
?

antban: There are 2 reasons.
 - While using different ui generators description field on `EventTypeStatistics` will be used itself, and I do not want it (because someday there will be measured_statistics).
 - Default statistics can hold some additional information.

v-stepanov: ok

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -350,7 +351,8 @@ private void doWithZkUtils(final ZkUtilsAction action) throws Exception {
         }
     }
 
-    private Integer calculateKafkaPartitionCount(final EventTypeStatistics stat) {
+    @VisibleForTesting
```
v-stepanov: I'm such a big hater of this annotation :)
But we can discuss if we want that or not in Nakadi in one of our meetings.
For now it's ok for me.

src/test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepositoryTest.java
```diff
@@ -256,6 +257,17 @@ public void canGetPartition() throws NakadiException {
...
 
     @Test
+    public void testIntegerOverflowOnStatisticsCalculation() throws NakadiException {
+        when(settings.getMaxTopicPartitionCount()).thenReturn(1000);
+        final EventTypeStatistics statistics = new EventTypeStatistics();
+        statistics.setReadParallelism(1);
+        statistics.setWriteParallelism(1);
+        statistics.setMessagesPerMinute(1000000000);
+        statistics.setMessageSize(1000000000);
+        assertThat(kafkaTopicRepository.calculateKafkaPartitionCount(statistics), equalTo(6));
```
v-stepanov: I expected that for 16666666 GB/s it will give the max number of partitions :)

antban: According to collected data - It's not so. The speed is not increasing, so it will select the fastest partition count

PR #237, opened 2016-05-11
The description and example belong to the item and not to the array

tmuehl:
@v-stepanov @rcillo @jkliff @antban

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -901,10 +901,10 @@ definitions:
             items:
               type: string
               format: uuid
-            description: |
-              Event identifier of the Event that caused the generation of this Event.
-              Set by the producer.
-            example: '105a76d8-db49-4144-ace7-e683e8f4ba46'
+              description: |
+                Event identifier of the Event that caused the generation of this Event.
+                Set by the producer.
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/237" target="_blank">ReviewNinja

rcillo:
👍 thanks for the contribution.

api/nakadi-event-bus-api.yaml
```diff
@@ -901,10 +901,10 @@ definitions:
...
               type: string
               format: uuid
-            description: |
-              Event identifier of the Event that caused the generation of this Event.
-              Set by the producer.
-            example: '105a76d8-db49-4144-ace7-e683e8f4ba46'
+              description: |
+                Event identifier of the Event that caused the generation of this Event.
+                Set by the producer.
+              example: '105a76d8-db49-4144-ace7-e683e8f4ba46'
```
v-stepanov: I know that it wasn't affected by you, but shouldn't the example be an array?

rcillo: I don't think so because the example is for the item.

v-stepanov: Ahhh ok. You actually moved it to the right, so it's now under "items". Good.

PR #238, opened 2016-05-11
aruha-184 fix schema doc

rcillo:
The format of the serialized string should follow the specified
"schema.type", as stated in the EventTypeSchema object definition.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1222,8 +1222,11 @@ definitions:
     properties:
       type:
         type: string
-        example: 'json-schema'
-        description: The type of schema definition (avro, json-schema, etc).
+        enum:
+          - JSON_SCHEMA
+        description: |
+          The type of schema definition. Currently only JSON_SCHEMA v4 is supported, but in the future there
+          could be others.
       schema:
         type: string
         description: |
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/238" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -1177,8 +1177,7 @@ definitions:
         allOf:
           - $ref: '#/definitions/EventTypeSchema'
         description: |
-          The schema for this EventType. This is expected to be a json schema in yaml format (other formats
-          might be added in the future).
```
v-stepanov: If you remove this then I think it will make sense to write in 'EventTypeSchema' definition that only currently supported schema type is json. What do you think?

rcillo: Yes, we have it there. We have the `type` field whose type is `array` of enums. The only valid value in the enum is `json_schema`.

v-stepanov: Can you please point me this place, because for me it is a string https://github.com/zalando/nakadi/blob/aruha-184-fix-schema-doc/api/nakadi-event-bus-api.yaml#L1223

rcillo: @v-stepanov my mistake. There were so many small tasks going on that I confused myself with a local branch. https://github.com/zalando/nakadi/pull/238/commits/c126b06d857b099773ef395641bafb1c740befbd

v-stepanov:
:shipit:

PR #239, opened 2016-05-11
aruha-185 fix enrichment api documentation

rcillo:
Trying to create an event type without any specified enrichment strategy
result in:

```
Failed to create event type end2end_monitor.test_invoked, code: 422,
message: {u'status': 422, u'type': u'http://httpstatus.es/422',
u'detail': u'must define metadata enrichment strategy', u'title':
u'Unprocessable Entity'}
```

But according to specs default enrichment_strategies is `metadata-enrichment`.

So in this PR we made it clear that enrichment strategy is required, when it
should specify it and why.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1157,12 +1157,16 @@ definitions:
           upon reception (and after validation) of an Event and is only possible on fields that are not
           defined on the incoming Event.
 
+          For event types in categories 'business' or 'data' it's mandatory to use metadata_enrichment strategy. For
+          'undefined' event types it's not possible to use this strategy, since metadata field is not required.
+
           See documentation for the write operation for details on behaviour in case of unsuccessful
           enrichment.
         type: array
         items:
           type: string
-        default: ['metadata-enrichment']
+          enum:
+            - metadata_enrichment
 
       partition_strategy:
         description: |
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/239" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -1217,6 +1221,7 @@ definitions:
       - name
       - category
       - owning_application
+      - enrichment_strategies
```
v-stepanov: is it possible to create undefined event-type without enrichment_strategies field

rcillo: It is. Actually, this field is required in the case of business and data change events. I don't know how to get rid of these exceptions. This category concept is really hard to deal with.

v-stepanov: Then I believe it should not be in a list of required fields.
I think it's ok - we can just say that for certain categories we have additional event-type validation.

v-stepanov: But in general I agree. These categories make me think of:

rcillo: Then you think I should say in the description that it might be required in special cases?

v-stepanov: Yes, I think you can say that it is required for categories business and data.

PR #240, opened 2016-05-11
Aruha 190 hila api - create subscription

v-stepanov:
@rcillo @antban @jkliff 
API for subscription creation endpoint


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/240" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1131,35 @@ definitions:
...
           $ref: '#/definitions/Event'
 
+  Subscription:
+    description: Subscription is a high level consumption unit.
+    properties:
+      application_id:
+        type: string
+        example: 'price-service'
+        description: The id of application that will consume from the subscription.
+      event_types:
```
jkliff: I would leave this at this time as singular. We can offer (and think about the details) of providing multi topic streaming at a later stage.

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1131,35 @@ definitions:
         items:
           $ref: '#/definitions/Event'
 
+  Subscription:
```
jkliff: Maybe "StreamSubscription"?

api/nakadi-event-bus-api.yaml
```diff
@@ -743,6 +743,54 @@ paths:
...
+          in: body
+          description: Subscription to create
+          schema:
+            $ref: '#/definitions/Subscription'
+          required: true
+      responses:
+        '200':
+          description: Subscription was successfuly created
+          schema:
+            type: string
```
jkliff: format: uuid
example: ...

api/nakadi-event-bus-api.yaml
```diff
@@ -743,6 +743,54 @@ paths:
...
+            description: id of subscription that was created
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '403':
+          description: Client is not authorized to perform this operation
+          schema:
+            $ref: '#/definitions/Problem'
+        '404':
```
jkliff: really? 404 on this??

api/nakadi-event-bus-api.yaml
```diff
@@ -743,6 +743,54 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+  /subscriptions:
+    post:
```
jkliff: there should be a GET on the collection and one for the stream itself, right?

api/nakadi-event-bus-api.yaml
v-stepanov: I don't think it will be possible to switch at later stage because it will be backwards incompatible change in the API - which is forbidden http://zalando.github.io/restful-api-guidelines/compatibility/Compatibility.html#must-dont-break-backward-compatibility
In the past we had such a change for event-pushing endpoint (from single event to array of events) and it caused a lot of pain to our users, and we were not even live.
That's why I think it will makes sense to make it an array from the very beginning because we will need to support that anyway very soon.

api/nakadi-event-bus-api.yaml
v-stepanov: As we will not have any other types of subscriptions I don't see need to make the name more specific.

api/nakadi-event-bus-api.yaml
v-stepanov: good point. Fixed.

api/nakadi-event-bus-api.yaml
v-stepanov: Yes, of course, we will have it. 
But I just wanted to describe the subscription creation endpoint - what we already discussed and agreed on.

api/nakadi-event-bus-api.yaml
v-stepanov: Why not? Because it's a POST?

api/nakadi-event-bus-api.yaml
```diff
@@ -743,6 +743,56 @@ paths:
...
+        return the same subscription id.
+      parameters:
+        - name: subscription
+          in: body
+          description: Subscription to create
+          schema:
+            $ref: '#/definitions/Subscription'
+          required: true
+      responses:
+        '200':
```
rcillo: We are using 201 in event type creation. Shouldn't we use the same here?

v-stepanov: My problem here is that we don't obviously create a subscription, in a case when one already exists we just return an id. So I'm not sure if we can always return 201.
Maybe we can return 201 if it is really created and 200 if we just return id?

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1133,35 @@ definitions:
...
+        type: array
+        items:
+          type: string
+        description: |
+          EventTypes to subscribe to.
+          The order is not important. Subscriptions that differ only be the order of EventTypes will be considered
+          the same and will have the same id.
+          * Currently only subscription to a single EventType is supported. Subscriptions with more than one EventType
+          in event_types property will be rejected.
+      use_case:
```
rcillo: <begin Talmudic discussion>I would argue that `group_id` is a more agnostic name for this property. The name `use_case` feel like a description field for me. I would understand that I should write some meaningful text here, but in reality it's more about a "short" identifier.<end>

v-stepanov: I don't have a strong opinion here, but I think that having a name like "something_id" will make our users think that they actually need to get this id from somewhere. And having it without "id" may provoke entering just a text description. So each approach has its pros and cons.

api/nakadi-event-bus-api.yaml
rcillo: I really think POST is the wrong verb for this endpoint. Since it should indempotent, we should use PUT to create this resource.

api/nakadi-event-bus-api.yaml
```diff
@@ -743,6 +743,62 @@ paths:
...
             $ref: '#/definitions/Problem'
 
+  /subscriptions:
+    post:
+      tags:
+        - managed-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        This endpoint creates a Subscription for EventType(s). The subscription is needed to be able to consume
```
rcillo: I think it would ok to have it in plural without parenthesis. In a programatic point of view it makes sense to refer to a collection of 1 as a collection.

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1139,101 @@ definitions:
         items:
           $ref: '#/definitions/Event'
 
+  Subscription:
+    description: |
+      Subscription is a high level consumption unit. Creating a subscription user subscribes to event-types and
```
rcillo: I feel the need to stress the key properties of a subscription. Wdyt of `Subscriptions allows applications to easily scale the number of clients by managing consumed event offsets and distributing load between instances.`

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1139,101 @@ definitions:
...
+      later can read this subscription with many streams.
+      The key properties that identify subscription are 'owning_application', 'event_types' and 'use_case'.
+      It's not possible to have two different subscriptions with these properties being the same.
+      'id' and 'created_at' properties should not be specified when creating subscription as these properties are
```
rcillo: I think this information could be in the description for each property.

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1139,101 @@ definitions:
...
+      generated by Nakadi.
+    properties:
+      id:
+        type: string
+        description: |
+          Id of subscription that was created. Is generated by Nakadi, should not be specified when creating
+          subscription.
+        format: uuid
```
rcillo: Do we want to have it as uuid? I don't see need to make the format explicit. Just a `string` is enough for clients to use it.

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1139,101 @@ definitions:
...
+          - BEGIN
+          - END
+        description: |
+          Position to start reading events from. BEGIN - read from the oldest available event. END - read from the most
+          recent offset. Will be applied when the first client for this subscription starts consuming.
+        default: 'END'
+      autocommit:
+        type: boolean
+        description: |
+          Specifies if the autocommit mode should be used for this subscription. When operating in autocommit mode
```
rcillo: It seems that this first sentence could be removed without harming clarity. I mean ✂️  `Specifies if the autocommit mode should be used for this subscription.` for the sake of brevity.

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1139,101 @@ definitions:
...
+          Nakadi considers all events that were sent from Nakadi to be committed. It's very easy to loose events when
+          running in this mode, use autocommit mode only if you know what you are doing.
```
rcillo: 😅

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1139,101 @@ definitions:
...
+        default: false
+      allowed_uncommitted_event_count:
```
rcillo: wdyt of `max_uncommitted_event_count`?

api/nakadi-event-bus-api.yaml
v-stepanov: I think John wanted to have uuid for this. 
@jkliff what is your idea about that?

api/nakadi-event-bus-api.yaml
v-stepanov: done

api/nakadi-event-bus-api.yaml
v-stepanov: I like your description, I added that.

api/nakadi-event-bus-api.yaml
v-stepanov: +1
Done

api/nakadi-event-bus-api.yaml
v-stepanov: I would prefer to have it as it is. For me the description seems to be more "smooth" with the first sentence included.

api/nakadi-event-bus-api.yaml
v-stepanov: I like it. Changed.

rcillo:
I'm just not a 100% convinced that having `commit_timeout`, `commit_period`, `max_uncommitted_event_count` is in some way better than having `max_uncommitted_event_count` and `timeout`.

In my understanding, we are controlling 2 different things: throughput and liveness. And for that reason, only 2 parameters should suffice.

In my understanding `commit_timeout` and `commit_period` mix both ideas into one making it somehow a mess in my mind.

rcillo:
But anyway, I think that it also does not harm to have all 3. So, for me, it's a :shipit: 

I think it's clearly defined 👍

v-stepanov:
@rcillo When writing the description for these three properties I also thought that it can make sense to have only two of them.
My main question here is: what can be the use-case to pause the stream after some amount of time (if no commit came). For now I don't really know if someone needs that. 
In my mind I prefer a little bit more having two params. But I can live with both solutions.
Let's see what others think. @antban @jkliff ?

api/nakadi-event-bus-api.yaml
rcillo: As discussed afterwards, using PUT is not an option here.

api/nakadi-event-bus-api.yaml
rcillo: I remember that @antban had a point about idempotency. Returning two different status could be against the mathematical definition of the term. But anyways, it still makes sense to me.

antban: What I meant about idempotency - any action performed on second time must have same result as it was done once. 
so it's really better to return only one response code. If one need to check subscription existence - than it's better just to make GET.

api/nakadi-event-bus-api.yaml
antban: Actually it's illegal argument (so imho - BAD REQUEST)

api/nakadi-event-bus-api.yaml
antban: I vote for single word naming :)

api/nakadi-event-bus-api.yaml
antban: I agree with @rcillo that we don't need to provide as less as possible information here. We can use uuid, and can not. It's not client's business. Two things we need to provide - type and maximum length.

api/nakadi-event-bus-api.yaml
antban: @rcillo  I think that `use_case` is actually a better name than `group_id`, because it have some meaning. `group_id` don't have it.

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1139,99 @@ definitions:
...
+          and event_types.
+        default: 'none'
+      read_from:
+        type: string
+        enum:
+          - BEGIN
+          - END
+        description: |
+          Position to start reading events from. BEGIN - read from the oldest available event. END - read from the most
+          recent offset. Will be applied when the first client for this subscription starts consuming.
```
antban: @v-stepanov @rcillo @jkliff 
I don't agree with ''Will be applied when the first client for this subscription starts consuming"
Consider the following:
client starts consuming (server sends 1000 messages), but client did't answered anything to it (did not commit offsets). Than the moment of "consumption start" is not defined, but we already need to handle these offsets somehow.
I think that offsets must be calculated on time of creation of subscription.

api/nakadi-event-bus-api.yaml
v-stepanov: My only point here for not having the same response code is:
Let's imagine situation: one client creates a subscription with parameters A, he gets 201 and in response he gets a subscription with parameters A.
Now another client also tried to create subscription with the same owning_application, event_types and use_case but with parameters B. He will get 201 and in response he gets a subscription with parameters A. I think for client that will be quite unexpected, so we can at least differ this situation with different response code.

What do you think?

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1139,99 @@ definitions:
...
+        default: 30
+      commit_timeout:
+        type: integer
+        description: |
+          The amount of seconds Nakadi waits for client to commit offset after max_uncommitted_event_count or
+          commit_period is reached. After the timeout expires the connection to the client will be closed by Nakadi.
+          Specifying 0 for this property will make it unlimited.
+          It is possible to override this property for a single stream when starting a stream for the subscription.
+          The property is ignored when autocommit is true.
+        default: 300
```
antban: I think that these 3 parameters are related to streams by itself, probably it's not the time for it right now. (Yes I think that it's cool that we can set defaults for subscription consumers on subscription creation, but imho it's not the right time for that). The only thing that is really important here is internal subscription parameters (like rebalancing timeouts, rebalancing strategies, etc.)

api/nakadi-event-bus-api.yaml
v-stepanov: I see your point but there are trade offs for each solution.
If we fetch offsets when subscription is created then it also has some issues:

Example 1, user creates subscription and defines that he wants to read from BEGIN. Nakadi fetches the earliest available offset (let's say 43809) and defines it as the one to start streaming from. 10 hours after creation of subscription user creates first stream and starts reading events. Offset 43809 is not available any more so the stream is started from offset 44102. So the BEGIN will anyway be the earliest available offsets when first client started.

Example 2, user creates subscription and defines that he wants to read from END. Nakadi fetches the earliest available offset (let's say 57098) and defines it as the one to start streaming from. When first client connects and starts reading that is not a real end any more, there were few hundreds of events pushed between subscription creation and start of stream.
Another possibility here is that first stream was created one week after subscription creation and at that time event END offset is expired. So the stream will be started from the earliest offset that is currently available

That's why I think that defining BEGIN and END relatively to subscription creation will be quite confusing for the user.

api/nakadi-event-bus-api.yaml
v-stepanov: I think that most of the users would prefer to have these stream parameters the same of all clients, so I think it's ok to have that from the beginning in subscription.
The parameters for rebalancing we will certainly have here, I will add them after we discuss rebalancing on our meeting.

api/nakadi-event-bus-api.yaml
antban: Regarding 1: The same situation could be reached in normal cases
Regarding 2: It's really the case showing that it's bad to use END:
 - 2 clients start using nakadi. One is died, not commiting anything. second is continuing to commit. First is restarted and starts reading. From which point? END - Then it's strange, because part of events are lost. The position from which streaming was started? - strage, because there were no commits, and it's like nothing was read.

v-stepanov: I will remove this phrase "Will be applied when the first client for this subscription starts consuming." for now to postpone this discussion for later.

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1235,60 @@ definitions:
...
+      clients by managing consumed event offsets and distributing load between instances.
+      The key properties that identify subscription are 'owning_application', 'event_types' and 'use_case'.
+      It's not possible to have two different subscriptions with these properties being the same.
+    properties:
+      id:
+        type: string
+        description: |
+          Id of subscription that was created. Is generated by Nakadi, should not be specified when creating
+          subscription.
+      created_at:
```
antban: Long time passed.. but this is still in PR, so I'll add it. I think that we need to split Subscription into 2 classes:
 - Subscripiton
 - StoredSubscription

StoredSubscription will hold all the values from Subscription and additionaly it will hold id and created_at field. From my point of view it will simplify development.

v-stepanov: I had this in the beginning, but then switched to a single object. There are not so much properties we generate in Nakadi, so the difference will be not so big.
Also if we have two objects, then some day we will have to have two object for EventType as well (as EventType will have creationTime which is generated by nakadi)

rcillo: I think the best way to understand trade-offs is to implement both. But since that would cost twice, I'm ok with the having only Subscription.

api/nakadi-event-bus-api.yaml
jkliff: i guess I though this was under /subscriptions, not /subscriptions/{...}/cursors. 
ignore my comment.

api/nakadi-event-bus-api.yaml
jkliff: For the record after offline discussion: will be consumer_group.

api/nakadi-event-bus-api.yaml
whiskeysierra: > Returning two different status could be against the mathematical definition of the term.

Idempotency in REST refers to the state change, not the response. Meaning, two idempotent requests should produce the same state on the server. It's totally valid to return e.g. 409 Conflict if a subscription already exists. The actual fact *the subscription is now present* is true, in both cases.

api/nakadi-event-bus-api.yaml
whiskeysierra: You can declare properties as `readOnly: true`, that should cover your use case.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-13

api/nakadi-event-bus-api.yaml
whiskeysierra: Maybe I'm missing a fundamental part of the concept here, but how exactly do subscriptions work? I was expecting to specify a URI in here which then receives events, being pushed by Nakadi. Am I wrong?

antban: @whiskeysierra Probably you are trying to find complete specification and it's not ready still.
the WIP version is in https://github.com/zalando/nakadi/tree/feature/ARUHA-190-subscription-api

PR #243, opened 2016-05-13
ARUHA-190: added endpoints to commit/expose offsets

v-stepanov:
@antban @rcillo @jkliff 
Please take a look. The endpoints for committing/exposing offsets of subscription.
We didn't really discuss it yet, so this is more like a PR to start a discussion about that.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -799,6 +799,103 @@ paths:
             $ref: '#/definitions/Problem'
 
 
+  /subscriptions/{subscription_id}/offsets:
+    get:
+      tags:
+        - managed-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: Exposes the currently committed offsets of a subscription.
+      parameters:
+        - name: subscription_id
+          in: path
+          type: string
+          description: Id of subscription
+          required: true
+      responses:
+        '200':
+          description: Ok
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Cursor'
+        '204':
+          description: |
+            No Content. If the reading from the subscription was not started yet - there is nothing to expose.
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '403':
+          description: Client is not authorized to perform this operation
+          schema:
+            $ref: '#/definitions/Problem'
+        '404':
+          description: Subscription not found
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Server error
+          schema:
+            $ref: '#/definitions/Problem'
+        '503':
+          description: Service (temporarily) unavailable
+          schema:
+            $ref: '#/definitions/Problem'
+    put:
+      tags:
+        - managed-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        Endpoint for committing offsets of the subscription. The offsets are per partition of EventType. One can use
+        this endpoint to commit the offsets for any of the partitions that are part of this subscription.
+        With every commit the offsets can only increase so it is not possible to commit offset that is lower than
+        already committed offset for the appropriate partition; if such request is made - it will not have any effect
+        on the committed offsets.
+      parameters:
+        - name: subscription_id
+          in: path
+          type: string
+          description: Id of subscription
+          required: true
+        - name: cursors
+          in: body
+          description: Cursors with offsets to commit.
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Cursor'
+          required: true
+      responses:
+        '200':
+          description: Offsets were committed or ignored
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '403':
+          description: Client is not authorized to perform this operation
+          schema:
+            $ref: '#/definitions/Problem'
+        '404':
+          description: Subscription not found
+          schema:
+            $ref: '#/definitions/Problem'
+        '422':
+          description: Unprocessable Entity
+          schema:
+            $ref: '#/definitions/Problem'
+        '500':
+          description: Server error
+          schema:
+            $ref: '#/definitions/Problem'
+        '503':
+          description: Service (temporarily) unavailable
+          schema:
+            $ref: '#/definitions/Problem'
+
+
   '/registry/validation-strategies':
     get:
       tags:
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -819,14 +819,6 @@ paths:
             type: array
             items:
               $ref: '#/definitions/Cursor'
-        '204':
-          description: |
-            No Content. If the reading from the subscription was not started yet - there is nothing to expose.
-            Will return an empty list of cursors to be consistent with successful scenario.
-          schema:
-            type: array
-            items:
-              $ref: '#/definitions/Cursor'
         '401':
           description: Client is not authenticated
           schema:
@@ -875,6 +867,8 @@ paths:
       responses:
         '200':
           description: Offsets were committed
+        '204':
+          description: No Content. The cursors which were tried to be committed are older than already committed ones.
         '401':
           description: Client is not authenticated
           schema:
@@ -885,10 +879,6 @@ paths:
             $ref: '#/definitions/Problem'
         '404':
           description: Subscription not found
-          schema:
-            $ref: '#/definitions/Problem'
-        '409':
-          description: Conflict. The cursors which were tried to be committed are older than already committed ones.
           schema:
             $ref: '#/definitions/Problem'
         '422':
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -799,7 +799,7 @@ paths:
             $ref: '#/definitions/Problem'
 
 
-  /subscriptions/{subscription_id}/offsets:
+  /subscriptions/{subscription_id}/cursors:
     get:
       tags:
         - managed-api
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/243" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -799,6 +799,103 @@ paths:
             $ref: '#/definitions/Problem'
 
 
+  /subscriptions/{subscription_id}/offsets:
```
rcillo: Did you considered "cursors"?

api/nakadi-event-bus-api.yaml
```diff
@@ -799,6 +799,103 @@ paths:
...
+      responses:
+        '200':
+          description: Ok
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Cursor'
+        '204':
+          description: |
+            No Content. If the reading from the subscription was not started yet - there is nothing to expose.
```
rcillo: I would argue that we should expose an empty list. I say that because then it is more homogeneous. If you program a client oriented to lists, handling both lists with or without something should fit in the same case, making it simpler.

api/nakadi-event-bus-api.yaml
```diff
@@ -799,6 +799,103 @@ paths:
...
+            $ref: '#/definitions/Problem'
+    put:
+      tags:
+        - managed-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        Endpoint for committing offsets of the subscription. The offsets are per partition of EventType. One can use
+        this endpoint to commit the offsets for any of the partitions that are part of this subscription.
+        With every commit the offsets can only increase so it is not possible to commit offset that is lower than
```
rcillo: Can I change my mind about it? I was considering an alternative to this ignoring strategy. If we versioned each `Cursor` then we could let the client use optimistic versioning. For example:

1. client commits first time.
2. server tags commit cursor with version 1 and reply to client in the response body this VERSION in the cursor (or as companion structure).
3. client commits a new cursors with VERSION 1.
4. server updates cursors with VERSION 2.

This way, if the client commits with a different version, we could reply with an error and deny to commit.

The advantage of this approach is that clients could rewind and we do not loose the safety of handling concurrent commits.

api/nakadi-event-bus-api.yaml
```diff
@@ -799,6 +799,103 @@ paths:
...
+          in: body
+          description: Cursors with offsets to commit.
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Cursor'
+          required: true
+      responses:
+        '200':
+          description: Offsets were committed or ignored
```
rcillo: It could be misleading to reply with 200 when they are ignored. I think we should only reply with 200 in case of accepting the commit. I think 409 conflict would fit nicely. Even better if we adopt the optimistic versioning.

api/nakadi-event-bus-api.yaml
v-stepanov: yes, this is what I had in the beginning :)
I just thought that for user it will be more clear what 'offsets' mean.
But in general 'cursors' will be more consistent, as we really mean cursors (offsets per partition)
So I will change that to 'cursors'

api/nakadi-event-bus-api.yaml
v-stepanov: I think that will make things very complicated. As I understand you want to have a separate version for each cursor (cursor = partition + offset), so the client will have to track the version for each partition he streams from (because there can be different version for different partition)

And I don't really see how it will help to rewind. Can you please provide some more details on that?

api/nakadi-event-bus-api.yaml
v-stepanov: I could not really find a suitable error code for that.
I just didn't really consider that to be an error situation, so I didn't consider 4xx responses.
But in general I'm fine with returning 409 in a case when offsets are ignored.

api/nakadi-event-bus-api.yaml
v-stepanov: 👍 
Will change that to empty list.

api/nakadi-event-bus-api.yaml
v-stepanov: fixed

api/nakadi-event-bus-api.yaml
v-stepanov: fixed

api/nakadi-event-bus-api.yaml
v-stepanov: added 409 in a case of ignore. I think that is better than having 200 for both situations.

api/nakadi-event-bus-api.yaml
rcillo:
```
client A              Nakadi              client B

    +                    +                   +
    |POST initial Cursor |                   |
    +-------------------->                   |
    |OK Cursor V1        |                   |
    <--------------------+                   |
    |                    |GET Cursor V1      |
    |                    +------------------->
    |                    |                   |
    |PUT Cursor V1 update|                   |
    +-------------------->                   |
    |OK Cursor V2        |                   |
    <--------------------+                   |
    |                    |PUT Cursor V1 update
    |                    <-------------------+
    |                    |Conflict           |
    |                    +------------------->
    |                    |                   |
    |                    |                   |
    |                    |                   |
    +                    +                   +
```

This way, the first update to reach with a certain version "wins". Subsequent updates should be targeted to the most recent version.

But you are right about rewinding. We should have additional parameters for that, so let's ignore this by now.

api/nakadi-event-bus-api.yaml
rcillo: Sorry if I was not clear enough. I think we should stick to 200 only. Empty list *is* content. So 204 would not be our best option here.

api/nakadi-event-bus-api.yaml
v-stepanov: I don't think that will be a case for us, as the clients consume from different partitions and they don't intersect in any way. Their commits are independent.

rcillo: Ok. Maybe I'm considering other aspects, as for example having multiple consumers for the same partition. Lets just keep in mind that we have other ways of controlling concurrent commits.

api/nakadi-event-bus-api.yaml
antban: I agree with @rcillo 204 - is not a valuable answer

api/nakadi-event-bus-api.yaml
```diff
@@ -799,6 +799,112 @@ paths:
...
+          type: string
+          description: Id of subscription
+          required: true
+      responses:
+        '200':
+          description: Ok
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Cursor'
```
antban: The result that is retrieved here is not usable at all. We will have cursors like that 
```
[{'partition': '0', offset: 120}, {'partition': '1', offset: 100}]
```
It doesn't point to anywhere. My point is that we need to think about it and to include event type to cursor (or extended cursor) right now.

api/nakadi-event-bus-api.yaml
```diff
@@ -799,6 +799,112 @@ paths:
...
+        - name: cursors
+          in: body
+          description: Cursors with offsets to commit.
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Cursor'
+          required: true
+      responses:
+        '200':
```
antban: Maybe 204 is better?

api/nakadi-event-bus-api.yaml
```diff
@@ -799,6 +799,112 @@ paths:
...
+        '403':
+          description: Client is not authorized to perform this operation
+          schema:
+            $ref: '#/definitions/Problem'
+        '404':
+          description: Subscription not found
+          schema:
+            $ref: '#/definitions/Problem'
+        '409':
+          description: Conflict. The cursors which were tried to be committed are older than already committed ones.
```
antban: We need to discuss it. I think that it's ok to commit previous cursors (or commit the same several times). If client tries to commit async, than it's normal situation when commits are not in exact order.
Consider example with rebalancing and late commits from previous owner. The new one won't ever be able to commit it's offsets.

api/nakadi-event-bus-api.yaml
v-stepanov: Good point. With current situation user will have to do additional request to know the event-type of this subscription, but it is not really a show stopper.
In general I see no problem to add event-type to cursor, because it is backwards compatible change. But I think the right moment to do that is when we switch to multi-event-type consumption (for low and high levels)

api/nakadi-event-bus-api.yaml
v-stepanov: For me it doesn't matter that much. Why you want to have 204 here?

api/nakadi-event-bus-api.yaml
v-stepanov: In general I agree with you. Probably it is better that committing older offsets should not be an error.

api/nakadi-event-bus-api.yaml
antban: Because there is no content inside ^)

api/nakadi-event-bus-api.yaml
rcillo: But then, how can the client be informed that his commit updated server state or not?

api/nakadi-event-bus-api.yaml
rcillo: Wdyt about using HATEOAS and linking to the event type?

rcillo:
I think this is good enough, even though we could not agree on a 100% of the details 👍

rcillo:
:shipit:

PR #245, opened 2016-05-20
ARUHA-193 Do not use ELB connection caching on streaming responses

antban:
And use tcp elb listener, instead of http elb (http elb does not inform about client connection close)

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
@@ -14,6 +14,7 @@ import de.zalando.aruha.nakadi.repository.TopicRepository;
 import de.zalando.aruha.nakadi.service.EventStream;
 import de.zalando.aruha.nakadi.service.EventStreamConfig;
 import de.zalando.aruha.nakadi.service.EventStreamFactory;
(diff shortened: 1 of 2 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/245" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -142,6 +143,7 @@ public StreamingResponseBody streamEvents(
                         streamLimit, batchTimeout, streamTimeout, streamKeepAliveLimit);
 
                 response.setStatus(HttpStatus.OK.value());
+                response.addHeader(HttpHeaders.CONNECTION, "close");
```
v-stepanov: What effect will it have?

antban: For now it will have no effect. But imho it's better to say that this connection must be closed instead of reusing it. We do not know how ELB is implemented, so maybe it will try to reuse connection (some day), but we expect that streaming will be ended only when connection is closed. That's why I added it.

v-stepanov:
:shipit:

PR #248, opened 2016-05-20
ARUHA-190: Started definition of HiLA endpoint for creating a stream

v-stepanov:
WIP

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -798,6 +798,141 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+  /subscriptions/{subscription_id}/events:
+    get:
+      tags:
+        - stream-api
+        - managed-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        ---
+      parameters:
+        parameters:
+          - name: subscription_id
+            in: path
+            type: string
+            description: Id of subscription
+            required: true
+          - name: commit_timeout
+            in: query
+            description: |
+              The amount of time allowed between commits. If no commit is performed within this amount of time - the
+              connection will be closed from Nakadi side. Specified in seconds.
+            type: integer
+            format: int32
+            required: false
+            default: 60
+          - name: max_uncommitted_event_count
+            in: query
+            description: |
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/248" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -798,6 +798,141 @@ paths:
...
+              The amount of time allowed between commits. If no commit is performed within this amount of time - the
+              connection will be closed from Nakadi side. Specified in seconds.
+            type: integer
+            format: int32
+            required: false
+            default: 60
+          - name: max_uncommitted_event_count
+            in: query
+            description: |
+              The amount of uncommitted events Nakadi will stream before pausing the stream. When in paused state and
```
rcillo: I was thinking in a way to remove the duplication of "stream" from the sentence "will stream before pausing the stream". Do you think that just removing "the stream" from the end would be harmless? Like "The amount of uncommitted events Nakadi will stream before pausing."

api/nakadi-event-bus-api.yaml
```diff
@@ -798,6 +798,141 @@ paths:
...
+              commit comes - the stream will continue.
```
rcillo: I think this could be interpreted in a slightly inaccurate way. It's not that stream will resume after a commit. But it would resume if the commit offset decreased the number of uncommitted events.

api/nakadi-event-bus-api.yaml
```diff
@@ -798,6 +798,141 @@ paths:
...
+
+              * Stream initialization will fail if `stream_limit` is lower than `batch_limit`.
+            type: integer
+            format: int32
+            required: false
+            default: 0
+          - name: batch_flush_timeout
+            in: query
+            description: |
+              Maximum time in seconds to wait for the flushing of each chunk (per partition).
```
rcillo: Do you think we should make explicit some constraints? For example, it should be lower than `commit_timeout`. Otherwise we could create a stream that is bricked, as users won't receive events before `batch_flush_timeout` so they won't have a proper `cursor` to commit (remembering that clients only commit cursors as an opaque structure).

PR #250, opened 2016-05-23
ARUHA-190: added basic API for getting single/all subscriptions

v-stepanov:
@rcillo @antban @jkliff the endpoints for getting one/all subscription. Please review.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -797,7 +797,74 @@ paths:
           description: Service (temporarily) unavailable
           schema:
             $ref: '#/definitions/Problem'
+    get:
+      tags:
+        - managed-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: Lists all subscriptions that exist in a system.
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Subscription'
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '403':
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/250" target="_blank">ReviewNinja

api/nakadi-event-bus-api.yaml
```diff
@@ -797,7 +797,74 @@ paths:
           description: Service (temporarily) unavailable
           schema:
             $ref: '#/definitions/Problem'
+    get:
+      tags:
+        - managed-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: Lists all subscriptions that exist in a system.
```
v-stepanov: I remember we wanted to have it available only for "admin" scope.
The only problem I have with that - if someone will loose the id of subscription he created, it will not be possible for him to find it in any way. So maybe it will make sense to make this list-all endpoint to be available to everyone.

api/nakadi-event-bus-api.yaml
```diff
@@ -797,7 +797,74 @@ paths:
           description: Service (temporarily) unavailable
           schema:
             $ref: '#/definitions/Problem'
+    get:
```
v-stepanov: Do we want to have some query parameters for the first version for filtering of subscriptions during list-all operation? (Like "owning_application" or "event_types")

rcillo: I don't think so. We do not have these filters for event type listing endpoint, as well. At some point we will be required to add pagination to all of the listing endpoints.

PR #254, opened 2016-05-25
Aruha 202 subscription persistence

v-stepanov:
Subscription persistence 
@rcillo @antban @jkliff please review

### DIFF
--- a/database/nakadi/10_data/04_tables/02_subscription.sql
+++ b/database/nakadi/10_data/04_tables/02_subscription.sql
@@ -1,7 +1,10 @@
 CREATE TABLE IF NOT EXISTS zn_data.subscription (
-  s_id varchar(255) NOT NULL PRIMARY KEY CHECK (s_id <> ''),
+  s_id varchar(36) NOT NULL PRIMARY KEY CHECK (s_id <> ''),
   s_subscription_object jsonb NOT NULL,
-  CHECK (s_subscription_object->>'id' = s_id)
+  CHECK (s_subscription_object->>'id' = s_id),
+  CHECK (s_subscription_object->>'owning_application' IS NOT NULL),
+  CHECK (s_subscription_object->>'event_types' IS NOT NULL),
+  CHECK (s_subscription_object->>'use_case' IS NOT NULL)
 );
 
 CREATE INDEX ON zn_data.subscription USING gin (s_subscription_object);
--- a/database/nakadi/10_data/04_tables/02_subscription.sql
+++ b/database/nakadi/10_data/04_tables/02_subscription.sql
@@ -4,11 +4,11 @@ CREATE TABLE IF NOT EXISTS zn_data.subscription (
   CHECK (s_subscription_object->>'id' = s_id),
   CHECK (s_subscription_object->>'owning_application' IS NOT NULL AND s_subscription_object->>'owning_application' <> ''),
   CHECK (s_subscription_object->>'event_types' IS NOT NULL AND s_subscription_object->>'event_types' <> '[]'),
-  CHECK (s_subscription_object->>'use_case' IS NOT NULL AND s_subscription_object->>'use_case' <> '')
+  CHECK (s_subscription_object->>'consumer_group' IS NOT NULL AND s_subscription_object->>'consumer_group' <> '')
 );
 
 CREATE INDEX ON zn_data.subscription USING gin (s_subscription_object);
 
 CREATE UNIQUE INDEX ON zn_data.subscription ((s_subscription_object->>'owning_application'),
                                              (s_subscription_object->>'event_types'),
-                                             (s_subscription_object->>'use_case'));
+                                             (s_subscription_object->>'consumer_group'));
--- a/database/nakadi/10_data/04_tables/02_subscription.sql
+++ b/database/nakadi/10_data/04_tables/02_subscription.sql
@@ -2,9 +2,9 @@ CREATE TABLE IF NOT EXISTS zn_data.subscription (
   s_id varchar(36) NOT NULL PRIMARY KEY CHECK (s_id <> ''),
   s_subscription_object jsonb NOT NULL,
   CHECK (s_subscription_object->>'id' = s_id),
-  CHECK (s_subscription_object->>'owning_application' IS NOT NULL),
-  CHECK (s_subscription_object->>'event_types' IS NOT NULL),
-  CHECK (s_subscription_object->>'use_case' IS NOT NULL)
+  CHECK (s_subscription_object->>'owning_application' IS NOT NULL AND s_subscription_object->>'owning_application' <> ''),
+  CHECK (s_subscription_object->>'event_types' IS NOT NULL AND s_subscription_object->>'event_types' <> '[]'),
+  CHECK (s_subscription_object->>'use_case' IS NOT NULL AND s_subscription_object->>'use_case' <> '')
 );
 
 CREATE INDEX ON zn_data.subscription USING gin (s_subscription_object);
--- a/database/nakadi/10_data/04_tables/02_subscription.sql
+++ b/database/nakadi/10_data/04_tables/02_subscription.sql
@@ -1,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS zn_data.subscription (
+  s_id varchar(255) NOT NULL PRIMARY KEY CHECK (s_id <> ''),
+  s_subscription_object jsonb NOT NULL,
+  CHECK (s_subscription_object->>'id' = s_id)
+);
+
+CREATE INDEX ON zn_data.subscription USING gin (s_subscription_object);
+
+CREATE UNIQUE INDEX ON zn_data.subscription ((s_subscription_object->>'owning_application'),
+                                             (s_subscription_object->>'event_types'),
+                                             (s_subscription_object->>'use_case'));
--- a/src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepositoryTest.java
+++ b/src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepositoryTest.java
@@ -1,12 +1,18 @@
 package de.zalando.aruha.nakadi.repository.db;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
+import de.zalando.aruha.nakadi.config.JsonConfig;
 import de.zalando.aruha.nakadi.domain.Subscription;
 import de.zalando.aruha.nakadi.exceptions.DuplicatedSubscriptionException;
 import org.junit.Before;
 import org.junit.Test;
 
+import java.util.Map;
+
 import static java.util.UUID.randomUUID;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.core.IsEqual.equalTo;
 
 public class SubscriptionDbRepositoryTest extends AbstractDbRepositoryTest {
 
@@ -20,6 +26,24 @@ public class SubscriptionDbRepositoryTest extends AbstractDbRepositoryTest {
     public void setUp() {
         super.setUp();
         repository = new SubscriptionDbRepository(template, mapper);
+    }
+
+    @Test
+    public void whenCreateSubscriptionThenOk() throws Exception {
+
+        final Subscription subscription = createSubscription();
+        repository.saveSubscription(subscription);
+
+        final int rows = template.queryForObject("SELECT count(*) FROM zn_data.subscription", Integer.class);
+        assertThat("Number of rows should be 1", rows, equalTo(1));
+
+        final Map<String, Object> result =
+                template.queryForMap("SELECT s_id, s_subscription_object FROM zn_data.subscription");
+        assertThat("Id is persisted", result.get("s_id"), equalTo(subscription.getId()));
+
+        final ObjectMapper mapper = (new JsonConfig()).jacksonObjectMapper();
+        final Subscription saved = mapper.readValue(result.get("s_subscription_object").toString(), Subscription.class);
+        assertThat("Saved subscription equal to original one", saved, equalTo(subscription));
     }
 
     @Test(expected = DuplicatedSubscriptionException.class)
--- a/src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepositoryTest.java
+++ b/src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepositoryTest.java
@@ -1,14 +1,18 @@
 package de.zalando.aruha.nakadi.repository.db;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
 import de.zalando.aruha.nakadi.config.JsonConfig;
 import de.zalando.aruha.nakadi.domain.Subscription;
 import de.zalando.aruha.nakadi.exceptions.DuplicatedSubscriptionException;
+import de.zalando.aruha.nakadi.exceptions.InternalNakadiException;
+import de.zalando.aruha.nakadi.exceptions.NoSuchEventTypeException;
 import org.junit.Before;
 import org.junit.Test;
 
 import java.util.Map;
+import java.util.Set;
 
 import static java.util.UUID.randomUUID;
 import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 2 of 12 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/254" target="_blank">ReviewNinja

database/nakadi/10_data/04_tables/02_subscription.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS zn_data.subscription (
+  s_id varchar(255) NOT NULL PRIMARY KEY CHECK (s_id <> ''),
```
valgog: Are subscription IDs not UUID? 

If you are going paranoid about the text is, it would make sence to also trim the value in the check.

database/nakadi/10_data/04_tables/02_subscription.sql
```diff
@@ -0,0 +1,11 @@
...
+  s_subscription_object jsonb NOT NULL,
```
valgog: This is really one crazy mongo approach 😂

database/nakadi/10_data/04_tables/02_subscription.sql
```diff
@@ -0,0 +1,11 @@
...
+  CHECK (s_subscription_object->>'id' = s_id)
+);
+
+CREATE INDEX ON zn_data.subscription USING gin (s_subscription_object);
+
+CREATE UNIQUE INDEX ON zn_data.subscription ((s_subscription_object->>'owning_application'),
```
valgog: I think you also will need some checks on the not null not empty for the fields used in that index.

src/main/java/de/zalando/aruha/nakadi/domain/Subscription.java
```diff
@@ -0,0 +1,88 @@
...
+    @NotNull
+    private String owningApplication;
+
+    @NotNull
+    @Size(min = 1, max = 1)
+    private Set<String> eventTypes;
+
+    private String useCase = "none";
+
+    private DateTime createdAt = new DateTime(DateTimeZone.UTC);
```
rcillo: I was thinking if should it be done here or in the persistence layer. Managing this in the persistence layer seems to make sense because this piece of information is related to persistence. I could even imagine a generic class handling these updates for every database update for every table.

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepositoryTest.java
```diff
@@ -0,0 +1,113 @@
...
+        insertSubscriptionToDB(subscription);
+
+        // get subscription by key properties and compare to original
+        final Subscription gotSubscription = repository.getSubscription("myapp", ImmutableSet.of("second-et", "my-et"),
+                "mycase");
+        assertThat("We found the needed subscription", gotSubscription, equalTo(subscription));
+    }
+
+    private void insertSubscriptionToDB(final Subscription subscription) throws JsonProcessingException {
+        template.update("INSERT INTO zn_data.subscription (s_id, s_subscription_object) VALUES (?, ?::jsonb)",
```
valgog: Why not to use subscription creation API to do that?

src/main/java/de/zalando/aruha/nakadi/exceptions/DuplicatedSubscriptionException.java
```diff
@@ -0,0 +1,15 @@
+package de.zalando.aruha.nakadi.exceptions;
+
+import javax.ws.rs.core.Response;
+
+public class DuplicatedSubscriptionException extends NakadiException {
```
rcillo: What do you think about having a generic "DuplicatedRecordException" ?

src/main/java/de/zalando/aruha/nakadi/exceptions/NoSuchSubscriptionException.java
```diff
@@ -0,0 +1,18 @@
+package de.zalando.aruha.nakadi.exceptions;
+
+import javax.ws.rs.core.Response;
+
+public class NoSuchSubscriptionException extends NakadiException {
```
rcillo: The same here suggestion here "NotFoundRecordException".

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -0,0 +1,82 @@
...
+
+        final String sql = "SELECT s_subscription_object FROM zn_data.subscription " +
+                "WHERE s_subscription_object->>'owning_application' = ? " +
+                "AND replace(s_subscription_object->>'event_types', ' ', '') = ? " +
+                "AND s_subscription_object->>'use_case' = ? ";
+        try {
+            final String eventTypesJson = jsonMapper.writer().writeValueAsString(newTreeSet(eventTypes));
+            return jdbcTemplate.queryForObject(sql, new Object[]{owningApplication, eventTypesJson, useCase},
+                    rowMapper);
+        } catch (JsonProcessingException e) {
```
rcillo: missing `final` =P

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -0,0 +1,82 @@
...
+            throw new InternalNakadiException("Serialization problem during getting event type", e);
+        } catch (EmptyResultDataAccessException e) {
+            throw new NoSuchSubscriptionException("Subscription does not exist.", e);
```
rcillo: Do we use dots at the end of messages?

src/main/java/de/zalando/aruha/nakadi/config/JsonConfig.java
```diff
@@ -37,6 +39,8 @@ public ObjectMapper jacksonObjectMapper() {
         objectMapper.registerModule(enumModule());
         objectMapper.registerModule(new Jdk8Module());
         objectMapper.registerModule(new ProblemModule());
+        objectMapper.registerModule(new JodaModule());
+        objectMapper.configure(WRITE_DATES_AS_TIMESTAMPS , false);
```
rcillo: Do you think that we need any additional tests for this? Or is it already covered by current tests?

database/nakadi/10_data/04_tables/02_subscription.sql
antban: Why not to use s_id SERIAL?

database/nakadi/10_data/04_tables/02_subscription.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS zn_data.subscription (
+  s_id varchar(255) NOT NULL PRIMARY KEY CHECK (s_id <> ''),
+  s_subscription_object jsonb NOT NULL,
+  CHECK (s_subscription_object->>'id' = s_id)
```
antban: Why do we need to include s_id into Subscription object?

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    public Subscription getSubscription(final String owningApplication, final Set<String> eventTypes,
+                                        final String useCase)
+            throws NoSuchSubscriptionException, InternalNakadiException {
+
+        final String sql = "SELECT s_subscription_object FROM zn_data.subscription " +
+                "WHERE s_subscription_object->>'owning_application' = ? " +
+                "AND replace(s_subscription_object->>'event_types', ' ', '') = ? " +
```
antban: Why this replace is needed?

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -0,0 +1,82 @@
...
+
+public class SubscriptionDbRepository extends AbstractDbRepository {
+
+    private final SubscriptionMapper rowMapper = new SubscriptionMapper();
+
+    public SubscriptionDbRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper objectMapper) {
+        super(jdbcTemplate, objectMapper);
+    }
+
+    public void saveSubscription(final Subscription subscription) throws InternalNakadiException,
```
antban: What is the plan for usage this method?
From my point of view it must be like that:
```
id = saveSubscription(subscriptionData)
```

src/main/java/de/zalando/aruha/nakadi/domain/Subscription.java
antban: And I can imagine situation where user adds createdAt field into subscription object while storing it.

database/nakadi/10_data/04_tables/02_subscription.sql
valgog: SERIAL is an integer. We also could use integers for the subscription IDs of cause, but then you have a sequence generator in the database.

database/nakadi/10_data/04_tables/02_subscription.sql
valgog: I think @v-stepanov really wants to use only the subscription IDs, so in this case it would be a UUID (what I can see from the code)

database/nakadi/10_data/04_tables/02_subscription.sql
v-stepanov: In API this is String. And I thought that we will generate it in Nakadi (e.g. UUID)
The max length I just forgot to adjust, will change it to 36.

database/nakadi/10_data/04_tables/02_subscription.sql
v-stepanov: Don't see any reasons to make it different than event-types table :)

database/nakadi/10_data/04_tables/02_subscription.sql
v-stepanov: For me the question is "do we really need to have s_id as a separate table column?" as we anyway store the whole object to json. Probably it makes sense for indexing purposes. And I don't see a good reason not to store it in json.
And also that will make this functionality consistent with event-types table (where we store the name as a separate table column and in json object)

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepositoryTest.java
v-stepanov: It's a bootstrap for test. It should not use any of the project's code (because the project code is actually what we test).

database/nakadi/10_data/04_tables/02_subscription.sql
v-stepanov: changed length to 36

database/nakadi/10_data/04_tables/02_subscription.sql
v-stepanov: good point, I now added some checks for those fields.

src/main/java/de/zalando/aruha/nakadi/domain/Subscription.java
v-stepanov: @rcillo As it is a field inside json it's kinda difficult to populate it with some default value on DB level :) So I decided to go with easiest path.

@antban I think we even defined in our API that this field should not be set when creating subscriptions. If we let user set it - it will not be the creation time then.

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: I thought we don't have it for exceptions :) But ok, I will add that.

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: Because Postgres returns it like "['et1', 'et2']" with a space after coma 😜

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: My idea was that subscription id is generated already when subscription is passed to this method. And after that a) subscription is saved or b) DuplicatedSubscriptionException is thrown if subscription with the same parameters already exists

src/main/java/de/zalando/aruha/nakadi/domain/Subscription.java
antban: Added comment to api. https://github.com/zalando/nakadi/pull/240/files#r64699868

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
antban: @v-stepanov Vhtang! Ktulhu is coming.

On this stage there are 2 possible ways to get problems!
 - Duplicate primary key
 - Unique constraint for app, event_types and use_case

And this method is trying to mix several different actions into one.
According to use cases that we discussed a lot, there are basically 2 methods related to 'work with stored'/'storing' subscription:
 - `Create or get subscription` based on (app, use_case, event_types)
 - `get subscription` by nakadi generated id (won't discuss the format)
These methods may be directly mapped to dao layer (as you see - they are even called like dao methods)
According to this PR, we actually need to add another method - get subscription by app_id, et and so on. 
I wold prefer another method (if we really need this nakadi id generator):

```
storedSubscription = saveSubscription(subscription, idGenerator); // idGenerator = Supplier<String>
```

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
antban: @v-stepanov 
As far as I understand indexes, if they are exists for, could not be used?

And another one thing - tree set, that was used for storing `eventTypesJson` sorts strings.
I failed to find the same for creation of subscription. Then simply reordering eventTypes will introduce new subscription.

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
antban: And again, I don't like the idea of storing in subscription itself, because of:
 - at least 2 additional checks
 - id is not a part of subscription that was initialy created by user (and that's the difference from EventType)
 - not logical to store it twice

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepositoryTest.java
valgog: In my experience this is not really a nice way to do bootstrap. Because this way you are binding yourself to the table and object structure in the tests, and this is not what you really want. This increases maintenance costs without any real need for that.

database/nakadi/10_data/04_tables/02_subscription.sql
rcillo: We need a key-value store. By the time I created the event-type table, I did provided an extra column for the `name` because I did not know that postgres offers some consistency checks in json objects properties. But I see no harm in using the same approach. We could in the future remove both columns from both tables and have it implemented in a consistent way across the project.

database/nakadi/10_data/04_tables/02_subscription.sql
rcillo: @v-stepanov where are the checks? I though @valgog was talking about database constraints.

rcillo:
Being pragmatic, I think that there is too much discussion for such a simple feature. Besides having comments here, I would say that it's your call @v-stepanov to make adjustments of merging. It's already good enough for me. I think that there will be other more relevant discussions in future implementations.

database/nakadi/10_data/04_tables/02_subscription.sql
v-stepanov: @rcillo I added the checks to 02_subscription.sql
https://github.com/zalando/nakadi/blob/ARUHA-202-subscription-persistence/database/nakadi/10_data/04_tables/02_subscription.sql#L5
Or do you mean something else?

src/main/java/de/zalando/aruha/nakadi/exceptions/DuplicatedSubscriptionException.java
v-stepanov: I don't think that we will have too many entities in Nakadi to have a generic Exception. For now I'm fine with separate exceptions.

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: fixed, removed dots

src/main/java/de/zalando/aruha/nakadi/config/JsonConfig.java
v-stepanov: Now, we don't test it anywhere.
Do you think it makes sense to implement it as a separate test for objectMapper ?

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: The event-types are sorted when you set event-types for a subscription:
https://github.com/zalando/nakadi/blob/ARUHA-202-subscription-persistence/src/main/java/de/zalando/aruha/nakadi/domain/Subscription.java#L51

src/main/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: I completely understand your idea here but for me it is just a matter of implementation. It can be done by this way or by that way. I don't really see one or other implementation as a game changer.

If we look on storing id twice from a different angle we can say say we store the whole Subscription object into json-field and id is part of subscription object. So in general that is logical from other point if view.

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/db/SubscriptionDbRepositoryTest.java
v-stepanov: I'm pretty fine with binding to table and its structure in test - because this is actually what I test: that DAO method modifies/reads the table in a correct way.

If some class "foo" has methods "readBlah" and "writeBlah" - I think it's wrong to use its "readBlah" method to test its "writeBlah" method and vise versa. That will break test isolation. And if "foo" class is broken your test can still be green because you use class "foo" to test itself.

I think it is possible to test like that only for really high-level tests like API-tests which are more like User Journeys.

v-stepanov:
@rcillo 
I agree. We discuss a lot of things which are just different possible implementations. I think it doesn't make much sense to change implementation if the change doesn't bring indisputable benefits.

But I"m more than fine to read all the comments with possible suggestions and alternatives. Discussions in PRs are really very valuable for me.

rcillo:
:shipit:

src/main/java/de/zalando/aruha/nakadi/config/JsonConfig.java
rcillo: I think that if we are able to save it and then read it back into an object it's enough. Just test that this serialize-desserialize works 👍

v-stepanov: this we do test :)

PR #255, opened 2016-05-25
ARUHA-201 Add meter for events per second publishing

antban:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/255" target="_blank">ReviewNinja

rcillo:
Adding some UT for that would be valuable 👍

v-stepanov:
:shipit:

Issue #224, opened 2016-04-27
Link to Swagger API is broken


MALPI:
The link mentioned in the README.md which should referr to the swagger API is broken and leads to a png. 

http://online.swagger.io/validator?url=https://raw.githubusercontent.com/zalando/nakadi/nakadi-jvm/api/nakadi-event-bus-api.yaml

LappleApple:
+1, same for me

dehora:
I had a look and it seems the swagger.io validator returns a request with no content type 

```
bdehora$ curl -v https://online.swagger.io/validator/\?url\=https://raw.githubusercontent.com/zalando/nakadi/nakadi-jvm/api/nakadi-event-bus-api.yaml
*   Trying 23.22.16.221...
* Connected to online.swagger.io (23.22.16.221) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384
* Server certificate: *.swagger.io
* Server certificate: DigiCert SHA2 Secure Server CA
* Server certificate: DigiCert Global Root CA
> GET /validator/?url=https://raw.githubusercontent.com/zalando/nakadi/nakadi-jvm/api/nakadi-event-bus-api.yaml HTTP/1.1
> Host: online.swagger.io
> User-Agent: curl/7.43.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Date: Tue, 31 May 2016 17:46:11 GMT
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Methods: GET, POST, DELETE, PUT
< Access-Control-Allow-Headers: Content-Type
< Cache-Control: no-cache
< Connection: close
< Server: Jetty(9.2.9.v20150224)
```

I'm fairly sure this is screwing up github's [image proxy](https://github.com/atmos/camo). The answer here might be to remove the badge.

PR #265, opened 2016-06-01
Remove badge link to swagger validator.

dehora:
The validator service at online.swagger.io doesn't
return a content type, which seems to be confusing
github's image proxy. Removing it for now as it
looks broken.

For #224


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/265" target="_blank">ReviewNinja

PR #269, opened 2016-06-08
ARUHA-215 Subscription streaming

antban:
- ***NOT*** tested
 - initial implementation
 - covers all the use cases.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/269" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/config/NakadiConfig.java
```diff
@@ -11,15 +11,16 @@
 import de.zalando.aruha.nakadi.controller.EventPublishingController;
 import de.zalando.aruha.nakadi.controller.EventStreamController;
 import de.zalando.aruha.nakadi.controller.PartitionsController;
+import de.zalando.aruha.nakadi.controller.VersionController;
 import de.zalando.aruha.nakadi.enrichment.Enrichment;
 import de.zalando.aruha.nakadi.enrichment.EnrichmentsRegistry;
-import de.zalando.aruha.nakadi.controller.VersionController;
```
v-stepanov: you kill kittens again

src/main/java/de/zalando/aruha/nakadi/config/SecurityConfiguration.java
```diff
@@ -53,6 +53,7 @@ public void configure(final HttpSecurity http) throws Exception {
             http.authorizeRequests()
                     .antMatchers(GET, "/event-types/*/partitions/**").access(hasScope(eventStreamReadScope))
                     .antMatchers(GET, "/event-types/*/events/**").access(hasScope(eventStreamReadScope))
+                    .antMatchers(GET, "/subscriptions/*/events/**").access(hasScope(eventStreamReadScope))
```
v-stepanov: Please don't forget to extend security tests.

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
```diff
@@ -0,0 +1,160 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StreamingContext implements SubscriptionStreamer {
+    public final StreamParameters parameters;
+    public final Session session;
+    public final ZkSubscriptionClient zkClient;
+    public final KafkaClient kafkaClient;
+    public final SubscriptionOutput out;
+    public final long kafkaPollTimeout;
```
v-stepanov: why have you decided to have these fields public?

src/main/java/de/zalando/aruha/nakadi/config/SecurityConfiguration.java
v-stepanov: ah, ok. You already did that :)

src/main/java/de/zalando/aruha/nakadi/domain/Cursor.java
```diff
@@ -8,6 +8,7 @@
 public class Cursor {
 
     public static final String BEFORE_OLDEST_OFFSET = "BEGIN";
+    public static final String AFTER_NEWEST_OFFSET = "END";
```
rcillo: Since pointers are exclusive, e.g. if you have offsets [1, 2, 3, 4], and your cursors points to 2, then it should stream from 3 and 4.

That said, I would argue that the name of this constant should be `NEWEST_OFFSET`.

v-stepanov: +1
As we have exclusive offsets reading, it's not the one after newest, but the newest itself

src/main/java/de/zalando/aruha/nakadi/exceptions/NoStreamingSlotsAvailable.java
```diff
@@ -0,0 +1,14 @@
...
+import javax.ws.rs.core.Response;
+
+public class NoStreamingSlotsAvailable extends NakadiException {
+    public NoStreamingSlotsAvailable(final int totalSlots) {
+        super("No free slots for streaming available. Total slots: " + totalSlots);
+    }
+
+    @Override
+    protected Response.StatusType getStatus() {
+        return Response.Status.BAD_REQUEST;
```
rcillo: Thinking from the point of view of a developer implementing a client for Nakadi subscriptions, how is it possible for me to detect that this scenario is happening? Should I programatically match over the error message? Or is it enough to match on the status code? Is there any other scenarios that could generate a 400 other than this one?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -230,6 +231,30 @@ private Callback kafkaSendCallback(final BatchItem item, final CountDownLatch do
...
+                    .partitionsFor(topicId)
+                    .stream()
+                    .map(p -> new org.apache.kafka.common.TopicPartition(topicId, p.partition()))
+                    .toArray(org.apache.kafka.common.TopicPartition[]::new);
+            if (position.equals(Cursor.BEFORE_OLDEST_OFFSET)) {
+                consumer.seekToBeginning(kafkaTPs);
+            } else if (position.equals(Cursor.AFTER_NEWEST_OFFSET)) {
+                consumer.seekToEnd(kafkaTPs);
+            } else {
+                throw new ServiceUnavailableException("Bad offset specification " + position + " for topic " + topicId);
```
rcillo: Why service unavailable?

src/main/java/de/zalando/aruha/nakadi/domain/Cursor.java
antban: Agree with NEWEST, and again, I think that we need to decide which values to use: https://github.com/zalando/nakadi/pull/270#discussion-diff-66392347R20

src/main/java/de/zalando/aruha/nakadi/service/subscription/ExactWeightRebalancer.java
```diff
@@ -0,0 +1,113 @@
...
+        final int totalWeight = IntStream.of(weigths).sum();
+        final int fixed = itemCount / totalWeight;
+        final int[] result = IntStream.of(weigths).map(w -> fixed * w).toArray();
+        if (fixed == 0) {
+            Arrays.fill(result, 1);
+        }
+
+        int left = itemCount - IntStream.of(result).sum();
+
+        // Yes, it's bad way, I know. But failed to figure out other simple way for exact case.
```
rcillo: Actually, a small change here could simplify things a little, in my opinion.

```
static int[] splitByWeight(final int itemCount, final int[] weigths) {
        final int totalWeight = IntStream.of(weigths).sum();
        final int fixed = (int) Math.ceil((float) itemCount / (float) totalWeight);
        int left = itemCount;
        int[] result = new int[weigths.length]

        for (int i = 0; i < weigths.length && left > 0; ++i) {
             final int v = Math.min(left, weigths[i] * fixed);
             result[i] = v;
             left -= v;
         }
         return result;
}
```

src/main/java/de/zalando/aruha/nakadi/service/subscription/ExactWeightRebalancer.java
```diff
@@ -0,0 +1,113 @@
...
+        // sorted client ids.
+        final List<String> activeClientIds = activeClients.keySet().stream().sorted().collect(Collectors.toList());
+        // the main part of rebalance - calculate count for each partition.
+        final int[] partitionsPerClient = splitByWeight(
+                currentPartitions.length,
+                activeClientIds.stream().mapToInt(c -> activeClients.get(c).weight).toArray());
+
+        // Stage 1. Select partitions that are not assigned to any EXISTING client.
+        final Set<Partition> unassignedPartitions = Stream.of(currentPartitions)
+                .filter(p -> Partition.State.UNASSIGNED.equals(p.getState()) || !activeClients.containsKey(p.getSessionOrNextSession()) || !activeClients.containsKey(p.getSession()))
```
rcillo: I had a hard time to understand all 3 conditions. Maybe could we have some intermediary steps better describing each condition?

src/main/java/de/zalando/aruha/nakadi/service/subscription/ExactWeightRebalancer.java
```diff
@@ -0,0 +1,113 @@
...
+        final Map<String, Session> activeClients = Stream.of(sessions).collect(Collectors.toMap(c -> c.id, c -> c));
+        // sorted client ids.
+        final List<String> activeClientIds = activeClients.keySet().stream().sorted().collect(Collectors.toList());
+        // the main part of rebalance - calculate count for each partition.
+        final int[] partitionsPerClient = splitByWeight(
+                currentPartitions.length,
+                activeClientIds.stream().mapToInt(c -> activeClients.get(c).weight).toArray());
+
+        // Stage 1. Select partitions that are not assigned to any EXISTING client.
+        final Set<Partition> unassignedPartitions = Stream.of(currentPartitions)
```
rcillo: From what I understood, these are not "unassigned partitions" but ratter "partitions that should be reassigned" which is slightly different. For example, we might have a partition that is assigned but the session is gone, therefore it should be reassigned. Technically it's still assigned.

src/main/java/de/zalando/aruha/nakadi/service/subscription/ExactWeightRebalancer.java
```diff
@@ -0,0 +1,113 @@
...
+import java.util.Set;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+class ExactWeightRebalancer implements BiFunction<Session[], Partition[], Partition[]> {
+    @Override
+    public Partition[] apply(final Session[] sessions, final Partition[] currentPartitions) {
+        final Map<String, Session> activeClients = Stream.of(sessions).collect(Collectors.toMap(c -> c.id, c -> c));
```
rcillo: We should stick to one name. Either session or client.

src/main/java/de/zalando/aruha/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -0,0 +1,125 @@
...
+            out.flush();
+        }
+    }
+
+    @RequestMapping(value = "/subscriptions/{subscription_id}/events", method = RequestMethod.GET)
+    public StreamingResponseBody streamEvents(
+            @PathVariable("subscription_id") final String subscriptionId,
+            @RequestParam(value = "window_size", required = false, defaultValue = "100") final int windowSize,
+            @RequestParam(value = "commit_timeout", required = false, defaultValue = "30") final int commitTimeout,
+            @RequestParam(value = "batch_limit", required = false, defaultValue = "1") final int batchLimit,
```
v-stepanov: Why do you use primitive type for some of params and Integer for the others?

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamParameters.java
```diff
@@ -0,0 +1,66 @@
+package de.zalando.aruha.nakadi.service.subscription;
+
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+
+public class StreamParameters {
```
v-stepanov: Just wanted to point out that there is class https://github.com/zalando/nakadi/blob/nakadi-jvm/src/main/java/de/zalando/aruha/nakadi/service/EventStreamConfig.java
And it has a lot of common properties with your StreamParameters class
Maybe it makes sense to extract a common parent for them.

src/main/java/de/zalando/aruha/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -0,0 +1,125 @@
...
+                streamer = subscriptionStreamerFactory.build(
+                        subscriptionId,
+                        StreamParameters.of(batchLimit, streamLimit, batchTimeout, streamTimeout, streamKeepAliveLimit, windowSize, commitTimeout),
+                        output);
+                streamer.stream();
+            } catch (final NoSuchSubscriptionException e) {
+                output.onException(e);
+            } catch (final InterruptedException ex) {
+                LOG.warn("Interrupted while streaming with " + streamer, ex);
+                Thread.currentThread().interrupt();
```
v-stepanov: Why you do interrupt here?

src/main/java/de/zalando/aruha/nakadi/service/subscription/model/Session.java
```diff
@@ -0,0 +1,42 @@
+package de.zalando.aruha.nakadi.service.subscription.model;
+
+import java.util.UUID;
+
+public class Session {
+    public final String id;
+    public final int weight;
```
v-stepanov: what is weight?

src/main/java/de/zalando/aruha/nakadi/service/subscription/model/Session.java
```diff
@@ -0,0 +1,42 @@
+package de.zalando.aruha.nakadi.service.subscription.model;
+
+import java.util.UUID;
+
+public class Session {
+    public final String id;
```
v-stepanov: why have you decided to use public fields?

src/main/java/de/zalando/aruha/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -0,0 +1,125 @@
...
+            SubscriptionStreamer streamer = null;
+            final SubscriptionOutputImpl output = new SubscriptionOutputImpl(response, outputStream);
+            try {
+                streamer = subscriptionStreamerFactory.build(
+                        subscriptionId,
+                        StreamParameters.of(batchLimit, streamLimit, batchTimeout, streamTimeout, streamKeepAliveLimit, windowSize, commitTimeout),
+                        output);
+                streamer.stream();
+            } catch (final NoSuchSubscriptionException e) {
+                output.onException(e);
```
v-stepanov: If the subscription doesn't exist user will get BAD_REQUEST, I think more suitable is NOT_FOUND (I think in API it's also 404)

src/main/java/de/zalando/aruha/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+
+        @Override
+        public void onException(final Exception ex) {
+            LOG.warn("Exception occurred while streaming", ex);
+            if (!headersSent) {
+                headersSent = true;
+                try {
+                    if (ex instanceof NakadiException) {
```
v-stepanov: I'm not sure that it will be correct to return only 400 for Nakadi Exception and 503 for everything else. 
I think that in reality there are more options.

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamParameters.java
antban: It's mixed with different data, but I'll try to refactor.

src/main/java/de/zalando/aruha/nakadi/controller/SubscriptionStreamController.java
antban: Because catching of InterruptedException clears Thread's interrupted flag. If I won't do that, the guy that is stopping this thread will fail.

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
```diff
@@ -0,0 +1,356 @@
...
+        this.lock = new InterProcessSemaphoreMutex(curatorFramework, "/nakadi/locks/subscription_" + subscriptionId);
+        this.log = LoggerFactory.getLogger("zk." + subscriptionId);
+    }
+
+    public void setExceptionListener(final Consumer<Exception> exceptionListener) {
+        this.exceptionListener = exceptionListener;
+    }
+
+    @Override
+    public void lock(final Runnable function) {
```
v-stepanov: in fact it does:
1) lock
2) function execution
3) release
I think the name doesn't completely reflects what happens.
Maybe "runLocked" ?

src/main/java/de/zalando/aruha/nakadi/service/subscription/model/Session.java
antban: It's the estimation of client's ability to take load. For the first implementation it's always 1. But it could change in time.

src/main/java/de/zalando/aruha/nakadi/service/subscription/model/Session.java
antban: I use public fields if:
 - the field is final
 - the actual value that field is holding is not changed in time.

Reasons: 
 - I don't see any reasons to create getters that will just return field value.

v-stepanov:
@antban 
General comment. 
I'm not fan of the approach you use for nulls. I think that will make our code null-unsafe. 
You have a lot of methods where you catch the exception at the end and return null (like here for example https://github.com/zalando/nakadi/pull/269/files#diff-89991a50509920ddef27210b439ee7f4R207). I think this is a bad approach, which will lead to NPEs at the end.
The alternatives can be:
1) Throw exception instead of returning null.
2) Return Optional

src/main/java/de/zalando/aruha/nakadi/service/subscription/model/Session.java
v-stepanov: The reason is to follow Encapsulation principle, not to expose the inner structure of your class. 

>> I don't see any reasons to create getters that will just return field value.
In 99% cases getter just returns field value.

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
```diff
@@ -0,0 +1,356 @@
...
+        final ZKSubscription init() {
+            set();
+            return this;
+        }
+
+        protected void setInternal() throws Exception {
+            curatorFramework.getChildren().usingWatcher(this).forPath(getSubscriptionPath("/clients"));
+        }
+
+        private boolean set() {
```
v-stepanov: 1) You don't seem to use a return value of this method.
2) The name is a bit too general. Can it be something like "setWatch" ?

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
```diff
@@ -0,0 +1,356 @@
...
+            int counter = 0;
+            for (final Map.Entry<Partition.PartitionKey, Long> p : partitionToOffset.entrySet()) {
+                final String partitionPath = getPartitionPath(p.getKey());
+                curatorFramework.create().creatingParentsIfNeeded().withMode(CreateMode.PERSISTENT).forPath(
+                        partitionPath,
+                        serializeNode(null, null, Partition.State.UNASSIGNED));
+                curatorFramework.create().withMode(CreateMode.PERSISTENT).forPath(
+                        partitionPath + "/offset",
+                        (String.valueOf(p.getValue())).getBytes(CHARSET));
+                if (++counter % 100 == 0) {
```
v-stepanov: This is just crazy :-D
I thought that `++i` is used only on tech interviews to confuse candidates

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
```diff
@@ -0,0 +1,160 @@
...
+        this.currentState = new StreamCreatedState(this);
+        this.kafkaClient = kafkaClient;
+        this.kafkaPollTimeout = kafkaPollTimeout;
+    }
+
+    @Override
+    public void stream() throws InterruptedException {
+        // Because all the work is processed inside one thread, there is no need in
+        // additional lock.
+        currentState = new StreamCreatedState(this);
```
rcillo: I think that there's not need to do it again since it was already done in the constructor.

src/main/java/de/zalando/aruha/nakadi/service/subscription/SubscriptionStreamerFactory.java
```diff
@@ -0,0 +1,69 @@
...
+        this.subscriptionDbRepository = subscriptionDbRepository;
+        this.topicRepository = topicRepository;
+    }
+
+    public SubscriptionStreamer build(
+            final String subscriptionId,
+            final StreamParameters streamParameters,
+            final SubscriptionOutput output) throws NoSuchSubscriptionException {
+
+        // Make it here to trigger subscription not found exception earlier.
```
rcillo: Please, clarify this comment.

src/main/java/de/zalando/aruha/nakadi/service/subscription/ExactWeightRebalancer.java
antban: ok

src/main/java/de/zalando/aruha/nakadi/service/subscription/ExactWeightRebalancer.java
antban: Well, no. It won't work: itemCount = 10, weights=[3,3,3]
result = [4, 4, 2], I want [4, 3, 3]

antban: Even worse: result will be [6, 4, 0]

src/main/java/de/zalando/aruha/nakadi/exceptions/NoStreamingSlotsAvailable.java
antban: There are 3 candidates for that:
- 429 TOO MANY REQUESTS (which is not the part of official rfc2616)
- 409 Conflict (part of rfc, but too broad)
- 418 I'm a teapot (I didn't used this code anywhere, and want to try, and it's not a part of rfc)

429 Makes more sense, but I believe that it's not supported in different clients (like python one). So probably the best way is to use 409.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
antban: Agree, on this level of processing I can't throw it. I'll wait for subscription creation endpoint and will 
use enumeration that is there and this will fix all problems.

src/main/java/de/zalando/aruha/nakadi/service/subscription/ExactWeightRebalancer.java
antban: I moved method to partition, now it's completely clear what it does

src/main/java/de/zalando/aruha/nakadi/service/subscription/ExactWeightRebalancer.java
antban: yep, didn't cleared here.

src/main/java/de/zalando/aruha/nakadi/controller/SubscriptionStreamController.java
antban: I thought, that values, that are not specified will be nulls, and I don't want to check them on `0`.
So everything that has default value is primitive and everything that doesn't have it and could be omitted - object.

src/main/java/de/zalando/aruha/nakadi/service/subscription/model/Session.java
antban: I don't think that in this case encapsulation is really would be encapsulation. It will be just a set of getters, that will add additional calls.
I'll fix it in another way, so it will be simpler to use and will save us time instead of staring holly war.

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
antban: ok

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
antban: ok, my fault

src/main/java/de/zalando/aruha/nakadi/service/subscription/SubscriptionStreamerFactory.java
antban: Load subscription here to trigger `subscription not found` exception earlier

antban: I'll just remove the comment.

src/main/java/de/zalando/aruha/nakadi/service/subscription/ExactWeightRebalancer.java
rcillo: @antban you're right. I could change it to behave as expected but then it would no longer be so simple. I missed some details there. So, just let it as is 👍 

```java
static int[] splitByWeight(final int itemCount, final int[] weigths) {
        final int totalWeight = IntStream.of(weigths).sum();
        int left = itemCount;
        int[] result = new int[weigths.length]

        for (int i = 0; i < weigths.length; ++i) {
             final float fixed = (float) itemCount / (float) totalWeight;
             final int v = (int) Math.ceil(Math.min(left, weigths[i] * fixed));
             result[i] = v;
             left -= v;
             itemCount -= v;
             totalWeight -= weigths[i];
         }
         return result;
}
```

src/main/java/de/zalando/aruha/nakadi/exceptions/NoStreamingSlotsAvailable.java
rcillo: I see. I even took some time to look for other options but came to a similar conclusion as you. Maybe using status code is not enough to programatically handle this scenario in an elegant way. We probably should have additional fields in the problem object returned with some application code for the error. I say so because matching over a error message that is dynamically generated would be bad for client implementation.

Something like `error_code: no_slots_available`.

When I implemented the event validation, I missed the opportunity to add more structure to the error message and today I see that it imposes some limitations for clients to handle that programatically.

The error return values are not so well defined in the API as well.

src/main/java/de/zalando/aruha/nakadi/service/subscription/SubscriptionStreamerFactory.java
rcillo: Ok. Got it 👍

src/main/java/de/zalando/aruha/nakadi/controller/SubscriptionStreamController.java
antban: But actually it's so.
If Exception occured somewhere in ZK framework - we can not say that it's something else but 503.
If we made a bug - the same situation.

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
antban: Refactored and removed it. After removing exceptionListeners it was required to refresh these listeners.

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
antban: This counter is needed to print out that something is happening while creating really big subscription. If it's not needed - I can remove it.

antban: And regarding ++ there id c++ task: 
what will be printed: 
```
    int i = 0;
    int j = ++i + i++;
    std::cout << i << "," << j << std::endl;
```

v-stepanov: @antban 
2,2
?
:-D

v-stepanov: P.S. I don't have any problem with counter itself. I just think that using ++ operator in the same line with comparation is not the best practice. But it's only my opinion, I don't force you to change that.

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
```diff
@@ -0,0 +1,170 @@
...
+            } catch (final RuntimeException ex) {
+                LOG.error("Failed to process task " + task + ", code carefully!", ex);
+                switchState(new CleanupState(ex));
+            }
+        }
+    }
+
+    public void switchState(final State newState) {
+        this.addTask(() -> {
+            if (currentState != null) {
```
rcillo: I think this branch should never be executed, right?

Because, when processing `newState = null` we already set the `currentState = null` which would cause the `while` loop from line 90 to exit.

Is this correct?

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -0,0 +1,366 @@
...
+        if (null != getParameters().streamTimeoutMillis) {
+            scheduleTask(this::shutdownGracefully, getParameters().streamTimeoutMillis, TimeUnit.MILLISECONDS);
+        }
+
+        this.lastCommitMillis = System.currentTimeMillis();
+        scheduleTask(this::checkCommitTimeout, getParameters().commitTimeoutMillis, TimeUnit.MILLISECONDS);
+    }
+
+    private void checkCommitTimeout() {
+        if (!isCurrent()) {
```
rcillo: Why is it required to check if we are in the current state? When does it happen that while running this function we are in a different state than the expected?

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    @Override
+    public ZKSubscription subscribeForOffsetChanges(final Partition.PartitionKey key, final Runnable commitListener) {
+        log.info("subscribeForOffsetChanges");
+        return ChangeListener.forData(curatorFramework, getPartitionPath(key) + "/offset", commitListener);
+    }
+
+    @Override
+    public long getOffset(final Partition.PartitionKey key) {
```
v-stepanov: Why you treat offset as long here?
As I remember the initial idea was that we treat offset as String to have the theoretical possibility to support Kinesis in the future. So that we convert the String offset to long only in kafka-specific code (e.g. KafkaTopicRepository class)

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
```diff
@@ -0,0 +1,263 @@
...
+            log.debug("Lock taken " + function.hashCode());
+            try {
+                function.run();
+            } finally {
+                log.info("Releasing runLocked for " + function.hashCode());
+                try {
+                    lock.release();
+                } catch (final Exception e) {
+                    log.error("Failed to release runLocked", e);
+                    throw e;
```
v-stepanov: If you get to `finally` block with an exception - you will loose it as you throw a new one.

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
```diff
@@ -0,0 +1,263 @@
...
+    @Override
+    public ZKSubscription subscribeForOffsetChanges(final Partition.PartitionKey key, final Runnable commitListener) {
+        log.info("subscribeForOffsetChanges");
+        return ChangeListener.forData(curatorFramework, getPartitionPath(key) + "/offset", commitListener);
+    }
+
+    @Override
+    public long getOffset(final Partition.PartitionKey key) {
+        try {
+            return Long.parseLong(new String(curatorFramework.getData().forPath(getPartitionPath(key)), CHARSET));
```
v-stepanov: I'm not sure that the path is correct. Shouldn't it be with "/offset" at the end?

antban: Yes, you are right, I changed it in unit tests branch

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
antban: Because all the code that is written in StreamingState and PartitionData uses this functionality now (it uses offsets to calculate sent data size).
I decided not to make current solution overcomplicated.

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/StreamingState.java
antban: This task is scheduled once and than it's just rescheduled. If Exception occurred during offsets refresh or in some other case - the state will probably switch to some other state. For example - there are 2 tasks in queue: `p1` and `checkCommitTimeout`. `checkCommitTimeout` is deferred and will be executed after 1 second. p1 generates exception, which causes change in tasks queue, now its `switchState` and `checkCommitTimeout`. `switchState` will switch state once executed, after that there will be only one task - `checkCommitTimeout`. But it's strange to execute it, because no one cares about commit timeouts anymore.
There are basically 3 approaches for that:
 - Remember current state while scheduling task (and this can be easily done)
 - Make StreamingContext act as a proxy for states (complex, but it's standard approach for state machines)
 - Ignore processing in each case where it's needed. 

I like the first approach, but it was simpler to debug and understand what is going on with the last one.

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
antban: Yes, you are correct, but actually exit check could be done by different means (not just checking null state). In theory - state machine is in StoppedState at the end.
Actually method `stream()` is external one and is placed here for simplification (I have thoughts about removing this method to some other place, to have at most N threads (<10) for subscription streaming, that will process many subscriptions (>10) at a time). That's why I decided to make `switchState` in fully correct way (with null check).

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
v-stepanov: I examined more code and I see that you use Kafka-specific stuff in a lot of places. 
I'm afraid that will completely ruin all the previous effort of keeping all kafka stuff on kafka layer.

src/main/java/de/zalando/aruha/nakadi/service/subscription/KafkaClient.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
+
+public class KafkaClient {
+    private final Subscription subscription;
+    private final KafkaTopicRepository topicRepository;
+
+    KafkaClient(final Subscription subscription, final KafkaTopicRepository topicRepository) {
+        this.subscription = subscription;
```
v-stepanov: Why have you decided to have this class stateful? (as well as CuratorZkSubscriptionClient)
It's not possible to have it as Spring bean as it has to be created for each subscription, and it's difficult to write tests for classes that use this class (additional factory is needed to allow mocking this class in test code)

src/main/java/de/zalando/aruha/nakadi/service/subscription/KafkaClient.java
```diff
@@ -0,0 +1,38 @@
...
+import de.zalando.aruha.nakadi.domain.Cursor;
+import de.zalando.aruha.nakadi.domain.Subscription;
+import de.zalando.aruha.nakadi.exceptions.NakadiException;
+import de.zalando.aruha.nakadi.repository.kafka.KafkaTopicRepository;
+import de.zalando.aruha.nakadi.service.subscription.model.Partition;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
+
+public class KafkaClient {
```
v-stepanov: I'm not sure that KafkaClient is a good name for this class, it's a bit confusing.

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
```diff
@@ -0,0 +1,170 @@
...
+                LOG.info("Switching state from " + currentState.getClass().getSimpleName());
+                currentState.onExit();
+            } else {
+                LOG.info("Connection died, no new state switches available");
+                return;
+            }
+            currentState = newState;
+            if (null != currentState) {
+                LOG.info("Switching state to " + currentState.getClass().getSimpleName());
```
v-stepanov: I think for all these messages DEBUG level is more suitable

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
```diff
@@ -0,0 +1,170 @@
...
+        this.addTask(() -> {
+            if (currentState != null) {
+                LOG.info("Switching state from " + currentState.getClass().getSimpleName());
+                currentState.onExit();
+            } else {
+                LOG.info("Connection died, no new state switches available");
+                return;
+            }
+            currentState = newState;
+            if (null != currentState) {
```
v-stepanov: I'm a big hater of `null` state. I think it is better to create some constant INITIAL_STATE of type State and assign is to currentState when StreamingContext class is created.

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
```diff
@@ -0,0 +1,170 @@
...
+        return out;
+    }
+
+    public long getKafkaPollTimeout() {
+        return kafkaPollTimeout;
+    }
+
+    @Override
+    public void stream() throws InterruptedException {
+        currentState = new State() {
```
v-stepanov: What is happening here? 😕

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/StartingState.java
```diff
@@ -0,0 +1,53 @@
...
+     */
+    private void createSubscriptionLocked() {
+        // check that subscription initialized in zk.
+        if (getZk().createSubscription()) {
+            // if not - create subscription node etc.
+            getZk().fillEmptySubscription(getKafka().getSubscriptionOffsets());
+        } else {
+            final Session[] sessions = getZk().listSessions();
+            final Partition[] partitions = getZk().listPartitions();
+            if (sessions == null || partitions == null) {
```
v-stepanov: I think you have some unnecessary null checks (not only here). I don't see any cases where that can be null.
Moreover in Nakadi code we have a following null-policy:
If some method can return null - it should be marked as @Nullable. Everything else that is not marked @Nullable is considered nonnull by default and does not require null checks.

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
```diff
@@ -0,0 +1,269 @@
...
+                }
+            }
+            return partitions.toArray(new Partition[partitions.size()]);
+        } catch (final Exception e) {
+            throw new SubscriptionWrappedException(e);
+        }
+    }
+
+    @Override
+    public Session[] listSessions() {
```
v-stepanov: Why have you decided to work with arrays not with lists?
Moreover at the end of this method you actually have list which you have to convert to array.

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
```diff
@@ -0,0 +1,170 @@
...
+            @Override
+            public void onEnter() {
+            }
+        };
+
+        // Add first task - switch to starting state.
+        switchState(new StartingState());
+
+        while (null != currentState) {
+            final Runnable task = taskQueue.poll(1, TimeUnit.MINUTES);
```
v-stepanov: Just wondering why the timeout is 1 minute. It waits when the queue will have some tasks from the states, right?

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
```diff
@@ -0,0 +1,170 @@
...
+                LOG.info("Switching state to " + currentState.getClass().getSimpleName());
+                currentState.setContext(this);
+                currentState.onEnter();
+            } else {
+                LOG.info("No next state found, dying");
+            }
+        });
+    }
+
+    public void registerSession() {
```
rcillo: It's misleading to have a method called register session that also triggers rebalance.

src/main/java/de/zalando/aruha/nakadi/service/subscription/model/Partition.java
```diff
@@ -0,0 +1,124 @@
...
+        return new Partition(key, session, nextSession, state);
+    }
+
+    public Partition moveToSessionId(final String sessionId, final Collection<String> existingSessionIds) {
+        switch (state) {
+            case UNASSIGNED:
+                return toState(State.ASSIGNED, sessionId, null);
+            case ASSIGNED:
+                if (sessionId.equals(this.session) || !existingSessionIds.contains(this.session)) {
+                    return toState(State.ASSIGNED, sessionId, null);
```
rcillo: Why would it move from `ASSIGNED` to `ASSIGNED` and not change session? Just to nullify the `nextSession`? When would this happen?

rcillo: I mean, why not return the same object?

In the case of the second condition (after `||`) I do understand that the current session no longer exists so it should be moved to reassigning.

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
```diff
@@ -0,0 +1,170 @@
...
+            }
+        }
+    }
+
+    public boolean isInState(final State state) {
+        return currentState == state;
+    }
+
+    public void addTask(final Runnable task) {
+        taskQueue.offer(task);
```
v-stepanov: How do you decide which things to execute within this task queue and which things to execute without queue?

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/ClosingState.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    private void reactOnTopologyChange() {
+        final Map<Partition.PartitionKey, Partition> partitions = new HashMap<>();
+        getZk().runLocked(() -> Stream.of(getZk().listPartitions())
+                .filter(p -> getSessionId().equals(p.getSession()))
+                .forEach(p -> partitions.put(p.getKey(), p)));
+        final Set<Partition.PartitionKey> freeRightNow = new HashSet<>();
+        final Set<Partition.PartitionKey> addListeners = new HashSet<>();
+        for (final Partition p : partitions.values()) {
```
rcillo: I think one thing that could improve readability would be to split it in many scenarios. Even if that means iterating multiple times over the collection. Something like:

1. freeCommitedPartitions
2. freeNotAssignedPartitions
3. ...

and so on. Then write some comments for each method explaining the necessity for it. For someone reading this code for the first time, it handles so many scenarios that it's almost impossible for me to tell why or if is there any missing scenario to be handled.

antban: There is a problem with this approach. If I'll do like that than I probably can skip one of scenarios (or forget about it), just because they are in different places, and I'm not sure if all the scenarios was covered.
Instead of it - I make a decision about every item that belongs to this session. In this case I will react on every change. After making decision I perform actions. And yes, there must be more comments.

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamParameters.java
antban: Decided to use old and new ones.

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/StreamingState.java
antban: Remember current state while scheduling task (and this can be easily done) - decision.

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/StartingState.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * 1. Checks, that subscription node is present in zk. If not - creates it.
+     * <p>
+     * 2. Registers session.
+     * <p>
+     * 3. Switches to streaming state.
+     */
+    private void createSubscriptionLocked() {
```
rcillo: This method creates the subscription, registers the session and switches the states. But it's name tells us that it's responsibility is to create the subscription only.

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -0,0 +1,366 @@
...
+            return;
+        }
+        if (null != topologyChangeSubscription) {
+            topologyChangeSubscription.refresh();
+        }
+        reactNoTopologyChange();
+    }
+
+    private void reactNoTopologyChange() {
```
rcillo: This name is more about the context of execution and not what it does. I mean, by reading `reactNoTopologyChange` I know that it's doing something when a topology change happens but the change set is empty. But to understand what it does I have to read the code.

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -0,0 +1,366 @@
...
+        getZk().runLocked(() -> {
+            final Partition[] assignedPartitions = Stream.of(getZk().listPartitions())
+                    .filter(p -> getSessionId().equals(p.getSession()))
+                    .toArray(Partition[]::new);
+            addTask(() -> refreshTopologyUnlocked(assignedPartitions));
+        });
+    }
+
+    private void refreshTopologyUnlocked(final Partition[] assignedPartitions) {
```
rcillo: This name does not make it clear what this code is about. From my understanding this is about switching streaming partitions. Be it starting newly assigned partitions or releasing no longer assigned ones.

What about "reassignPartitionsAndUpdateConsumers"?

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -0,0 +1,366 @@
...
+                    Collectors.toMap(
+                            k -> k,
+                            k -> new TopicPartition(k.topic, Integer.valueOf(k.partition))));
+            // Ignore order
+            kafkaConsumer.assign(new ArrayList<>(kafkaKeys.values()));
+            // Check if offsets are available in kafka
+            kafkaConsumer.seekToBeginning(kafkaKeys.values().toArray(new TopicPartition[kafkaKeys.size()]));
+            kafkaKeys.forEach((key, kafka) -> offsets.get(key).ensureDataAvailable(kafkaConsumer.position(kafka)));
+            //
+            kafkaKeys.forEach((k, v) -> kafkaConsumer.seek(v, offsets.get(k).getSentOffset()));
```
v-stepanov: There is a bug here that will result in double sending of message. 
"offsets" are the offsets that are committed (at least in the scenario when you start stream for the partition) and when you call seek() with these offsets that will deliver you the last committed event again. 
(User committed offset 4391, and when you do seek(x, 4391) - you will then receive event with offset 4391 again from kafkaConsumer)

PR #270, opened 2016-06-08
Aruha 230 subscription creation endpoint

v-stepanov:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/270" target="_blank">ReviewNinja

v-stepanov:
please don't merge it, I forgot to add the feature toggle :)

src/main/java/de/zalando/aruha/nakadi/controller/SubscriptionController.java
```diff
@@ -0,0 +1,97 @@
...
+            if (!noneExistingEventTypes.isEmpty()) {
+                final String errorMessage = "Failed to create subscription, event type(s) not found: '" +
+                        StringUtils.join(noneExistingEventTypes, "','") + "'";
+                LOG.debug(errorMessage);
+                return create(NOT_FOUND, errorMessage, nativeWebRequest);
+            }
+
+            // generate subscription id and try to create subscription in DB
+            subscription.setId(UUID.randomUUID().toString());
+            subscriptionRepository.saveSubscription(subscription);
```
antban: If client will provide created_at field, than incorrect data will be written. I think, that we need two objects - one to store data and the second one to return it. And the second one is a child of the first one.

src/main/java/de/zalando/aruha/nakadi/domain/Subscription.java
```diff
@@ -10,9 +10,15 @@
...
 
 import static com.google.common.collect.Sets.newTreeSet;
+import static de.zalando.aruha.nakadi.domain.Subscription.InitialPosition.END;
 
 public class Subscription {
 
+    public enum InitialPosition {
+        BEGIN,
+        END
+    }
```
antban: According to current implementation, cursor can have offset named 'BEGIN' (with capital letters)
And this one will be in lowercase, because it's enumeration. I think that we need to make it the same way in both cases.

src/main/java/de/zalando/aruha/nakadi/controller/SubscriptionController.java
rcillo: I think that the persistence layer should manage this field.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/SubscriptionAT.java
```diff
@@ -0,0 +1,83 @@
...
+
+        // when we try to create that subscription again - we should get status 200
+        // and the subscription that already exists should be returned
+        response = given()
+                .body(subscription)
+                .contentType(JSON)
+                .post(ENDPOINT);
+
+        // assert status code
+        response
```
rcillo: It seems to be a good idea to also check for the subscription in the body again.

src/main/java/de/zalando/aruha/nakadi/domain/Subscription.java
```diff
@@ -23,10 +29,15 @@
     @Size(min = 1, max = 1)
     private Set<String> eventTypes;
 
+    @NotNull
     private String consumerGroup = "none";
 
+    @NotNull
```
rcillo: This would force this field to be populated by clients. I think it should be managed by the persistence layer.

src/test/java/de/zalando/aruha/nakadi/controller/SubscriptionControllerTest.java
```diff
@@ -0,0 +1,164 @@
...
+        existingSubscription.setId("123");
+        existingSubscription.setStartFrom(Subscription.InitialPosition.BEGIN);
+        existingSubscription.setCreatedAt(new DateTime(DateTimeZone.UTC));
+        when(subscriptionRepository.getSubscription(eq("app"), eq(ImmutableSet.of("myET")), any()))
+                .thenReturn(existingSubscription);
+
+        postSubscription(subscription)
+                .andExpect(status().isOk())
+                .andExpect(content().contentTypeCompatibleWith(APPLICATION_JSON))
+                .andExpect(content().string(sameJSONAs(jsonHelper.asJsonString(existingSubscription))));
```
rcillo: I see that you tested it here. So I think there's no need to test it again in the AT as I commented.

rcillo:
What about an endpoint for only getting or even for listing subscriptions? Wdyt? Even if it would be implemented in another PR.

v-stepanov:
@rcillo 
>What about an endpoint for only getting or even for listing subscriptions? Wdyt? Even if it would be implemented in another PR.

I think we will definitely have it. But subscription creation endpoint is more urgent, so we started with it.

rcillo:
:shipit:

v-stepanov:
@antban @rcillo 
Hey Guys. I made a bunch of changes according to your feedback.
- Split subscription object into two classes SubscriptionBase and Subscription
- Id and createdAt are now generated in repository class
- BEGIN and END are now text constants in uppercase
I hope that will make you happy :)

PR #271, opened 2016-06-09
Fix field name of EventType

tmuehl:
In the API it's called "default_statistics" but it was implemented as "default_statistic" (without an "s")

@v-stepanov @rcillo @antban @jkliff

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
@@ -80,7 +80,7 @@ public class EventTypeController {
             enrichment.validate(eventType);
             partitionResolver.validate(eventType);
             eventTypeRepository.saveEventType(eventType);
-            topicRepository.createTopic(eventType.getName(), eventType.getDefaultStatistic());
+            topicRepository.createTopic(eventType.getName(), eventType.getDefaultStatistics());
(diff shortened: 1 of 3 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/271" target="_blank">ReviewNinja

v-stepanov:
@tmuehl 
Thanks a lot for finding that.
It will require fixing that in all our DBs 😞

v-stepanov:
@tmuehl 
You actually increases test coverage by 5.70% with this commit 😆

tmuehl:
Oh ok, that's probably why it wasn't exposed in the event types when calling `/event-types`. :(

Haha, yes, I was already wondering, don't know what I did. 😄

rcillo:
For data migration

`UPDATE zn_data.event_type SET et_event_type_object = replace(et_event_type_object::text, 'default_statistic', 'default_statistics')::jsonb WHERE et_event_type_object::text LIKE '%default_statistic%';`

v-stepanov:
@rcillo regarding migration script, what if event-type is named "order_52_default_statistic"?
:trollface:

rcillo:
@v-stepanov I thought about that. There are none, but what's your suggestion?

v-stepanov:
@rcillo No, I'm just trolling. It should work fine.

PR #275, opened 2016-06-10
Resolve #166: Support for event json schema as object

### LINKED ISSUE #166: Support event type json-schema definition as object
It is inconvenient to provide event schema definition as a string.
Suggestion is to support schema_object field of type object along side/instead of schema string field

Also there is yaml mentioned, but json expected
https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#L1182

AlexanderYastrebov:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/275" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/config/JsonConfig.java
```diff
@@ -33,14 +39,15 @@
     @Bean
     @Primary
     public ObjectMapper jacksonObjectMapper() {
-        final ObjectMapper objectMapper = new ObjectMapper()
-                .setPropertyNamingStrategy(CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES);
+        final ObjectMapper objectMapper = new ObjectMapper();
 
+        objectMapper.setPropertyNamingStrategy(CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES);
         objectMapper.registerModule(enumModule());
+        objectMapper.registerModule(eventTypeSchemaModule());
```
jkliff: This is a very convenient idea if the followin holds:
- API spec is changed to support schema as object in the case type is json_schema, still be a blob of text otherwise. 
- migration script exists and is applied to all existing event types.

ePaul: Also, for compatibility a schema string should also be allowed.

AlexanderYastrebov: String is allowed. More over object value is a convenience, string value is default

AlexanderYastrebov: That means that type is still string and you do not need any migrations at all

AlexanderYastrebov: Also listing the schema will return string. This could be next improvement.

AlexanderYastrebov:
Other option can be introduction of JSON_SCHEMA_OBJECT so schema field will be object on creation and on listing

src/main/java/de/zalando/aruha/nakadi/config/JsonConfig.java
jkliff: In my view it's either or. And I would support having it as a json object (there's no point in keeping a string except for backwards compatibility).
In any of the cases, clarification on API is needed.

rcillo: I think this features follows the *robustness principle* where we are flexible on the input side, as far as it's not ambiguous, but we are strict in the output format. Since this is even backward compatible, I would say that it's completely fine to merge 👍

api/nakadi-event-bus-api.yaml
```diff
@@ -1250,6 +1250,8 @@ definitions:
 
           To have a generic, undefined schema it is possible to define the schema as `"schema":
           "{\"additionalProperties\": true}"`.
+
+          If type of schema is JSON_SCHEMA then schema can be passed as json object for convenience.
```
dehora: Object level schema would be awesome. I'd avoid overloading the type of the `schema` field as that's going to complicate readers, even though it can be done by stringiness checking of the data when it's received. Defining a `json_schema_object` enum option and a new `schema_object` field would provide a clean form of extension - if the enum is `json_schema_object` look in the `schema_object` field, no jackson modules needed - it's less magical :) Also for the json schema case (and avro if that gets supported), it'll be easy for the server to populate either field from the other. 

```yaml
  EventTypeSchema:
    properties:
      type:
        type: string
        enum:
          - json_schema
          - json_schema_object
      schema:
        type: string
      schema_object:
        type: object
    required:
      - type
```

Fwiw, today's string schema field is the most generic  option as it can support non-JSON schema (eg swagger yaml or protobufs), so it's probably worth keeping around with schema_object as a special consideration for the current common case (json formatted schema definitions)

AlexanderYastrebov: It seems a better option for EventTypeSchema to have mandatory field type and then use it as a `discriminator`, so other fields are defined on their own definitions `json_schema`  and `json_schema_object`.
The conversion between schema types in general not always possible, so if it was defined like a json_schema then it should stay a string, if it was json_schema_object then it should stay object to avoid confusion.

rcillo:
Currently, schema versioning is pretty immature and basic. We do not have versioning, we do not have decoupled schema validation from event publishing. E.g. schema related features should evolve quite a lot in the near future. That would probably create lots of opportunity to treat schema as a resource on it's own. One side effect of this decoupling would be to allow for content negotiation, which is the ideal way of providing such a feature in my opinion.

I'm against adding more complexity in order to provide such convenience at this moment. I think this should be discussed in a broader context of evolving schema registry functionality and not just as an isolated addition.

rcillo:
I'm closing this PR since we agreed that this field should not have it's type overloaded.

PR #281, opened 2016-06-14
Aruha 238 offsets commit

v-stepanov:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/281" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/controller/CursorsController.java
```diff
@@ -0,0 +1,75 @@
...
+        this.cursorsCommitService = cursorsCommitService;
+        this.featureToggleService = featureToggleService;
+    }
+
+    @RequestMapping(value = "/subscriptions/{subscriptionId}/cursors", method = RequestMethod.PUT)
+    public ResponseEntity<?> commitCursors(@PathVariable("subscriptionId") final String subscriptionId,
+                                           @RequestBody final List<Cursor> cursors,
+                                           final NativeWebRequest request) {
+
+        if (!featureToggleService.isFeatureEnabled("high_level_api")) {
```
antban: It's time to create constant

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -332,18 +332,22 @@ public EventConsumer createEventConsumer(final String topic, final List<Cursor>
         return kafkaFactory.createNakadiConsumer(topic, kafkaCursors, settings.getKafkaPollTimeoutMs());
     }
 
+    public int compareOffsets(final String firstOffset, final String secondOffset) {
+        try {
+            final long first = toKafkaOffset(firstOffset);
+            final long second = toKafkaOffset(secondOffset);
+            return Long.compare(first, second);
+        } catch (final NumberFormatException e) {
+            throw new IllegalArgumentException("Incorrect offset format, should be long", e);
```
antban: Need to include values that actually caused exception

src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZookeeperUtils.java
```diff
@@ -0,0 +1,28 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(ZookeeperUtils.class);
+
+    private ZookeeperUtils() {
+    }
+
+    public static <V> V runLocked(final Callable<V> callable, final InterProcessLock lock) throws Exception {
+        try {
+            lock.acquire();
+            return callable.call();
```
antban: `lock.acure();` must be outside of try section

src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZookeeperUtils.java
```diff
@@ -0,0 +1,28 @@
...
+        } finally {
+            try {
+                lock.release();
+            } catch (final Exception e) {
+                LOG.warn("Error occurred when releasing ZK lock", e);
```
antban: I don't like the idea of eating exception. In this case it means that there was a failure while releasing lock, so probably it will be better to throw this exception instead of returning valid result.

src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZookeeperUtils.java
v-stepanov: Why?

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
```diff
@@ -0,0 +1,50 @@
...
+                if (topicRepository.compareOffsets(cursor.getOffset(), currentOffset) > 0) {
+                    zkHolder.get().setData().forPath(offsetPath, cursor.getOffset().getBytes(CHARSET));
+                    return true;
+                } else {
+                    return false;
+                }
+            }, zkLockFactory.createLock(offsetPath));
+        } catch (Exception e) {
+            throw new ServiceUnavailableException("Error communicating with zookeeper", e);
+        }
```
antban: exception on compareOffsets is eaten here. User won't be able to see it

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
```diff
@@ -0,0 +1,50 @@
...
+    public CursorsCommitService(final ZooKeeperHolder zkHolder, final TopicRepository topicRepository,
+                                final ZooKeeperLockFactory zkLockFactory) {
+        this.zkHolder = zkHolder;
+        this.topicRepository = topicRepository;
+        this.zkLockFactory = zkLockFactory;
+    }
+
+    public boolean commitCursor(final String subscriptionId, final String eventType, final Cursor cursor)
+            throws ServiceUnavailableException, NoSuchSubscriptionException {
+
```
antban: Isn't it better to perform cursor validation inside `commitCursor` method call?

src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZookeeperUtils.java
v-stepanov: I don't think that it is better to throw that exception because:
1) If we came to `finally` block from "success"-branch then with throwing exception we will loose the result of callable.call()
2) If we came to `finally` block from "exception"-branch then with throwing exception we will loose the original exception that was thrown by callable.call()
3) I don't see how we can recover from exception in lock.release, so I don't see much sense in throwing that.

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
v-stepanov: I actually also thought about that possibility but then I changed my mind.
`commitCursor` is designed to commit only one cursor. Validating single cursor is not effective because to validate a cursor you need to get a list of partitions from kafka. So if you need to validate 10 cursors you will need to have 10 calls to kafka.
That's why I decided to validate all cursors together before the call of `commitCursors`

src/main/java/de/zalando/aruha/nakadi/controller/CursorsController.java
v-stepanov: Sounds reasonable 👍 
Created constant

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
v-stepanov: After discussion with @antban it was decided that it makes sense to make commitCursors method that gets a batch of cursors and it will do validation and will call commitCursor for each cursor.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/SubscriptionAT.java
```diff
@@ -71,6 +77,53 @@ public void testSubscriptionCreation() throws IOException {
...
+                .statusCode(HttpStatus.SC_NO_CONTENT);
+
+        // check that committed offset in Zookeeper is not changed
+        committedOffset = getCommittedOffsetFromZk(eventType, subscription, "0", curator);
+        assertThat(committedOffset, equalTo("25"));
+    }
+
+    private String getCommittedOffsetFromZk(final EventType eventType, final Subscription subscription,
+                                            final String partition, final CuratorFramework curator) throws Exception {
+        final String path = format("/nakadi/subscriptions/{0}/topics/{1}/{2}/offset", subscription.getId(),
```
rcillo: What do you think of a path like "/nakadi/subscriptions/{0}/topics/{1}/partitions/{2}/offset"?

v-stepanov: Well, I actually use what @antban designed in his PR. But in general I agree with him, as topics will not hold anything else than partitions it makes sense to skip "partitions" pathNode to save some space.
But if there is theoretical possibility that topic node will hold something else besides partitions, it will make sense to have that "partitions" node to avoid problems in the future.
@antban what is your position on that point?

src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZookeeperUtils.java
antban: Because in other case you will try to release lock that was not accured

v-stepanov: If the exception is thrown I don't have guarantee that it was not acquired. So stay on the safe side it makes sense for me to try to release it.

src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZookeeperUtils.java
antban: Regarding third point - that's the idea - we can not recover, so we need to inform caller that something went wrong.

src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZookeeperUtils.java
antban: Well, in general there is an approach for that: if method returned with exception, than it's up to this method to rollback all the data back. So in this case it's like that.
Even more - if acqure() throwed Exception, than it's guaranteed that release() will throw exception :)

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
```diff
@@ -0,0 +1,120 @@
...
+    public boolean commitCursors(final String subscriptionId, final List<Cursor> cursors)
+            throws InvalidCursorException, NoSuchSubscriptionException, ServiceUnavailableException {
+
+        final Subscription subscription = subscriptionRepository.getSubscription(subscriptionId);
+        final String eventType = subscription.getEventTypes().iterator().next();
+
+        createSubscriptionInZkIfNeeded(subscription);
+        topicRepository.validateCommitCursors(eventType, cursors);
+
+        boolean allCommitted = true;
```
rcillo: Maybe you could be a little bit more hipster and use some stream method like `allMatch`, for example?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: Shouldn't offsets be validated before calling this method?

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/SubscriptionAT.java
antban: When I was writing this I thought about 2 possible solutions: 
 - have `/nakadi/subscriptions/{0}/partitions/{1}_{2}/offset` key in zk
 - have more structured approach like it's implemented now (or Ricardo pointed).
But honestly - I don't think that having 'partitions' key will help us in any way. Instead, it will force us to create more zk data and to think about it.

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -238,6 +240,7 @@ private Callback kafkaSendCallback(final BatchItem item, final CountDownLatch do
                     .stream()
                     .map(p -> new org.apache.kafka.common.TopicPartition(topicId, p.partition()))
                     .toArray(org.apache.kafka.common.TopicPartition[]::new);
+            consumer.assign(Arrays.asList(kafkaTPs));
```
rcillo: Why is this necessary?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -370,6 +377,30 @@ private void validateCursors(final String topic, final List<Cursor> cursors) thr
         }
     }
 
+    public void validateCommitCursors(final String topic, final List<Cursor> cursors) throws InvalidCursorException {
```
rcillo: Why did you split this validation into two methods `validateCommitCursors` and `validateCursors`?

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
v-stepanov: method throws checked exceptions :(
no hipsters allowed :(

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
v-stepanov: not eaten any more

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
```diff
@@ -0,0 +1,120 @@
...
+            return runLocked(() -> {
+                final String currentOffset = new String(zkHolder.get().getData().forPath(offsetPath), CHARSET);
+                if (topicRepository.compareOffsets(cursor.getOffset(), currentOffset) > 0) {
+                    zkHolder.get().setData().forPath(offsetPath, cursor.getOffset().getBytes(CHARSET));
+                    return true;
+                } else {
+                    return false;
+                }
+            }, zkLockFactory.createLock(offsetPath));
+        } catch (final IllegalArgumentException e) {
```
rcillo: Why not validating then at the beginning of the method?

src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZookeeperUtils.java
v-stepanov: The main question for me here is if the caller of this method is more interested in actual function's result/exception or in lock exception.
And I thought that actual function's result/exception is more interesting for him.
Wdyt?

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
antban: Because it's not possible to seek without assigning partitions

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
v-stepanov: Because they are validated in a different way. 
Commit cursors don't have to be validated against available offsets

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
v-stepanov: cursors already get here validated in `commitOffsets` method (https://github.com/zalando/nakadi/blob/ARUHA-238-offsets-commit/src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java#L59), so theoretically this IllegalArgumentException will never happen (but if it will happen - it will be handled in a correct way)

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
v-stepanov: @rcillo they are validated before this method is called. But maybe someone will call this method without prior validation, so we need to provide this basic NumberFormatException handling

src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZookeeperUtils.java
v-stepanov: Ok, you win :)
I moved lock.acquire() out of try block

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/SubscriptionAT.java
rcillo: Ok. I think that even kafka uses some smart naming pattern for log files, so I would not oppose to it. Looks like the only nested resource to topics are partitions. I hope nothing else will show up any time soon.

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
rcillo: I think that if it's not supposed to happen, then if it happens it's a bug and we better throw an ugly exception that will cause a 500 ratter than 422 (is this the case?).

src/main/java/de/zalando/aruha/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: I'm not sure of it. Feels like defensive programming.

rcillo:
I'm pretty satisfied with the current status of this implementation. For me it's a :shipit: 

I'll just wait for @antban 's approve since the has some open comments.

src/main/java/de/zalando/aruha/nakadi/service/CursorsCommitService.java
v-stepanov: I also thought of that. The only conditions when it can happen - is if the code is changed. And if the code is changed - then probably 422 will be more appropriate than 500

v-stepanov:
@antban are there still some points that you want to be fixed/changed? Or are you ready to give a star? :)

antban:
:shipit:

Issue #272, opened 2016-06-09
data_key_fields is defined in the API but not implemented.


tmuehl:
Hence, when trying to post en event-type with `data_key_fields` property set, it's rejected with an 400 Response.

v-stepanov:
Yeah, we are really missing functionality for that. I'm not sure yet if we will implement that or remove from the API for now.

PR #285, opened 2016-06-14
ARUHA-240: Remove data_key_fields from definition.

dehora:
The `data_key_fields` isn't implemented in the code and setting it
stops event type creation. This removes the property from the API
definition for the moment; apart from breaking event type creation,
it's not a certainty the property will be used (there's a TBD in
the documentation).

For #272.


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/285" target="_blank">ReviewNinja

PR #286, opened 2016-06-14
ARUHA-240: lower case some enums in the definition.

dehora:
Follow up to 00448ed to lower case the metadata_enrichment and
json_schema enum values and align the code with the API
definition.

After this the only uppercase enums are the 4 letters used for the
`data_op` value (these match the implementation, which checks they
are uppercase when sent).


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/286" target="_blank">ReviewNinja

Issue #287, opened 2016-06-15
Consider removing validation_strategies from the API definition


dehora:
The `validation_strategies` [property](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#L1153-L1166) is described in the API for event types, along with a [registry url](https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#L750-L773), but is not in the code.

I'm wondering if it should be removed for the time being from the API definition and reintroduced when it's available. In particular sending it with a new event type stops the event type from being created. The yaml descriptions could be updated to indicate that schema validation happens for business and data types by default.

----

The sever will reject an event type post that contains the field with a 400. 

```
{
    "detail": "Unrecognized field \"validation_strategies\" (class de.zalando.aruha.nakadi.domain.EventType), not marked as ignorable (8 known properties: \"schema\", \"default_statistic\", \"partition_strategy\", \"name\", \"owning_application\", \"category\", \"enrichment_strategies\", \"partition_key_fields\"])\n at [Source: java.io.PushbackInputStream@28ac13f6; line: 6, column: 29] (through reference chain: de.zalando.aruha.nakadi.domain.EventType[\"validation_strategies\"])",
    "status": 400,
    "title": "Bad Request",
    "type": "http://httpstatus.es/400"
}
```

It doesn't appear as default data. 

```
$ curl -v http://localhost:8080/event-types | python -mjson.tool
> GET /event-types HTTP/1.1
> 
< HTTP/1.1 200 OK

[
    {
        "category": "business",
        "default_statistic": null,
        "enrichment_strategies": [
            "metadata_enrichment"
        ],
        "name": "business_order_received_1",
        "owning_application": "acme-order-service",
        "partition_key_fields": [],
        "partition_strategy": "random",
        "schema": {
            "schema": "{ \"properties\": { \"order_number\": { \"type\": \"string\" } } }",
            "type": "json_schema"
        }
    }
]
```

The `/registry/validation-strategies` resource is missing as well:

```sh
$ curl -v http://localhost:8080/registry/validation-strategies | python -mjson.tool 

> GET /registry/validation-strategies HTTP/1.1
> 
< HTTP/1.1 404 Not Found

{
    "error": "Not Found",
    "message": "No message available",
    "path": "/registry/validation-strategies",
    "status": 404,
    "timestamp": "2016-06-15T11:41:16.072+0000"
}
```

[...]

PR #288, opened 2016-06-15
ARUHA-242: Remove validation_strategies from the API yaml.

dehora:
This removes the `validation_strategies` property from event
type, and the corresponding registry path to see validations
available in the installation. These features are not
implemented yet by the server, and they can be added back
when they're available.

For #287


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/288" target="_blank">ReviewNinja

Issue #222, opened 2016-04-26
Nakadi API: `additionalProperties: true/false` is not valid Swagger/OpenAPI syntax.


ePaul:
There are some slight differences between Swagger's (now OpenAPI's) model schema and JSON schema. One of them is that the `additionalProperties` key [allows only object values](https://github.com/swagger-api/swagger-codegen/issues/1318#issuecomment-146302879) (which mean schemas), not boolean `true` (which would be equivalent to an empty object) nor `false` (which is the default if the key is omitted).

This makes the Nakadi API in the current form unusable with Swagger Codegen (these are simply ignored).

dehora:
@ePaul what's the best way to express open for extensions? I would have gone with

```yaml
"additionalProperties": {}
```

but there's a comment [here](https://github.com/swagger-api/swagger-codegen/issues/1318#issuecomment-146841035) that suggests that isn't accepted either, so I'm not sure.

dehora:
@ePaul @rcillo there's a pr here to help with swagger tooling: https://github.com/zalando/nakadi/issues/290. The codegen result and appearance in the ui are linked from there.

PR #290, opened 2016-06-15
ARUHA-247: Optimise API definition for swagger tooling.

dehora:
ARUHA-247: Optimise API definition for swagger tooling.

This changes the API definition to help swagger ui and codegen tools. Because 
there are structural changes to the object definitions, the API version is 
bumped to 0.6.0. The API JSON format is *not* being changed, just how the 
objects represent the data:

 - BusinessEvent and DataChangeEvent extend Event to make them available to
   swagger codegen's api client as subclasses. This specific subclassing
   technique for allOf allows codegen to subclass from Event but not do weird
   stuff like remove fields in the subclass (it's fiddly, but works).

 - Change how BusinessEvent and DataChangeEvent inline Metadata to make it
   visible to swagger ui, and avoid redundant Java classes. Avoid "Inline"
   anonymous objects in the ui.

 - Change how EventTypeStatistics and EventTypeSchema are inlined to EventType
   to make them visible to swagger ui. Avoid "Inline" anonymous objects in the
   ui.

 - Remove DataChangeEventQualifier and make its fields part of DataChangeEvent.
   DataChangeEventQualifier isn't used anywhere else.

 - Remove Metrics and /metrics as they don't do anything at the moment.

 - Remove non-interesting response codes from path definitions. Make sure all
   resources document happy-path responses and 401, otherwise only document
   notable responses.

 - Remove additionalProperties true/false declarations. Swagger doesn't accept 
   them, and they don't make or break the API (we can assume open for extension
   in the true cases, and closing for extension for the business and data 
   categories isn't strictly needed.

For #222

The result of codegen is visible here: https://github.com/dehora/nakadi-swagger-gen 

Note: try the [w=1 view of the diff](https://github.com/zalando/nakadi/pull/290/files?w=1) to review the changes (a chunk of the changes to Metadata are indendation)

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -90,7 +90,7 @@ info:
     parameters (number of partitions, retention times, etc), as well as their modification.
 
 
-  version: '0.5.1'
+  version: '0.6.0'
   contact:
     name: Team Aruha @ Zalando
     email: team-aruha+nakadi-maintainers@zalando.de
@@ -116,28 +116,6 @@ securityDefinitions:
         Grants access for consuming Event streams.
 
 paths:
-  /metrics:
-    get:
-      tags:
-        - monitoring
-      summary: Get monitoring metrics
-      responses:
-        '200':
-          description: Ok
-          schema:
-            $ref: '#/definitions/Metrics'
-        '401':
-          description: Client is not authenticated
-          schema:
-            $ref: '#/definitions/Problem'
-        '500':
-          description: Server error
-          schema:
-            $ref: '#/definitions/Problem'
-        '503':
-          description: Service (temporarily) unavailable
-          schema:
-            $ref: '#/definitions/Problem'
   /event-types:
     get:
       tags:
@@ -162,15 +140,6 @@ paths:
           description: Client is not authenticated
           schema:
             $ref: '#/definitions/Problem'
-        '500':
-          description: Server error
-          schema:
-            $ref: '#/definitions/Problem'
-        '503':
-          description: Service (temporarily) unavailable
-          schema:
-            $ref: '#/definitions/Problem'
-
     post:
       tags:
         - schema-registry-api
@@ -234,15 +203,6 @@ paths:
           description: Unprocessable Entity
           schema:
             $ref: '#/definitions/Problem'
-        '500':
-          description: Server error
-          schema:
-            $ref: '#/definitions/Problem'
-        '503':
-          description: Service (temporarily) unavailable
-          schema:
-            $ref: '#/definitions/Problem'
-
 
   /event-types/{name}:
     get:
@@ -270,18 +230,6 @@ paths:
             $ref: '#/definitions/EventType'
         '401':
           description: Client is not authenticated
-          schema:
-            $ref: '#/definitions/Problem'
-        '404':
-          description: EventType not found
-          schema:
-            $ref: '#/definitions/Problem'
-        '500':
-          description: Server error
-          schema:
-            $ref: '#/definitions/Problem'
-        '503':
-          description: Service (temporarily) unavailable
           schema:
             $ref: '#/definitions/Problem'
 
@@ -324,23 +272,10 @@ paths:
           description: Client is not authenticated
           schema:
             $ref: '#/definitions/Problem'
-        '404':
-          description: EventType not found.
-          schema:
-              $ref: '#/definitions/Problem'
         '422':
           description: Unprocessable Entity
           schema:
             $ref: '#/definitions/Problem'
-        '500':
-          description: Server error
-          schema:
-            $ref: '#/definitions/Problem'
-        '503':
-          description: Service (temporarily) unavailable
-          schema:
-            $ref: '#/definitions/Problem'
-
     delete:
       tags:
         - schema-registry-api
@@ -370,22 +305,6 @@ paths:
           description: EventType is successfuly removed
         '401':
           description: Client is not authenticated
-          schema:
-            $ref: '#/definitions/Problem'
-        '403':
-          description: Client is not authorized to perform this operation
-          schema:
-            $ref: '#/definitions/Problem'
-        '404':
-          description: EventType not found.
-          schema:
-              $ref: '#/definitions/Problem'
-        '500':
-          description: Server error
-          schema:
-            $ref: '#/definitions/Problem'
-        '503':
-          description: Service (temporarily) unavailable
           schema:
             $ref: '#/definitions/Problem'
 
@@ -456,22 +375,10 @@ paths:
             type: array
             items:
               $ref: '#/definitions/BatchItemResponse'
-        '400':
-          description: Bad request
-          schema:
-            $ref: '#/definitions/Problem'
         '401':
           description: Client is not authenticated
           schema:
             $ref: '#/definitions/Problem'
-        '404':
-          description: EventType not found.
-          schema:
-              $ref: '#/definitions/Problem'
-        '405':
-          description: Not allowed.
-          schema:
-              $ref: '#/definitions/Problem'
         '422':
           description: |
             At least one event failed to be validated, enriched or partitioned. None were submitted.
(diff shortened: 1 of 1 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/290" target="_blank">ReviewNinja

dehora:
What it looks like in a local swagger ui (note that the only "Inline" objects are arrays):

nakadiswagger

dehora:
This looks like an improvement, but there's still an issue with BusinessEvent and codegen. Because it's designed to be extended directly by just adding fields that represent the published schema, and doesn't have a wapper envelope similar to DataChangeEvent, there's no way for codegen to represent the actual data defined by its event type schema. I don't think this is fixable. Consumers and producers will have to work with the JSON or a lower level representation like Jackson/Gson directly.

api/nakadi-event-bus-api.yaml
```diff
@@ -827,135 +685,118 @@ definitions:
...
 
   EventMetadata:
     type: object
-    description: |
-      Metadata for this Event.
-
-      Contains commons fields for both Business and DataChange Events. Most are enriched by Nakadi
-      upon reception, but they in general MIGHT be set by the client.
     properties:
-      metadata:
```
dehora: The `metadata` field is now declared as a property of the BusinessEvent and DataChangeEvent object. This allows the Metadata object's property fields to be direct children of the definition, making them visible to the ui and producing a better output from codegen. The definition of Metadata remain the same as do the actual JSON structures for BusinessEvent and DataChangeEvent.

api/nakadi-event-bus-api.yaml
```diff
@@ -827,135 +685,118 @@ definitions:
...
+          reception of the Event.
+        type: string
+        format: uuid
+        example: '105a76d8-db49-4144-ace7-e683e8f4ba46'
+      event_type:
+        description: |
+          The EventType of this Event. This is enriched by Nakadi on reception of the Event
+          based on the endpoint where the Producer sent the Event to.
 
-        additionalProperties: false
```
dehora: This `additionalProperties` can go; it breaks swagger tooling and there's little benefit to closing metadata for extension (we can just use must ignore processing)

api/nakadi-event-bus-api.yaml
```diff
@@ -116,28 +116,6 @@ securityDefinitions:
         Grants access for consuming Event streams.
 
 paths:
-  /metrics:
```
dehora: This path is available on the server but produces no data. It can be added back in when there's an implementation

api/nakadi-event-bus-api.yaml
```diff
@@ -827,135 +685,118 @@ definitions:
...
-    properties:
-      data:
-        type: object
-        description: |
-            The payload of the type
-    required:
-        - data
-    additionalProperties: false
+      - $ref: '#/definitions/Event'
+      - type: object
```
dehora: This removes DataChangeEventQualifier and inlines that object's properties into DataChangeEvent. DataChangeEventQualifier is never going to be used outside a DataChangeEvent, plus its properties were directly inlined and not placed under a parent property (eg the way Metadata is placed under a `metadata` field).

api/nakadi-event-bus-api.yaml
```diff
@@ -1147,8 +985,7 @@ definitions:
 
       schema:
         type: object
-        allOf:
-          - $ref: '#/definitions/EventTypeSchema'
+        $ref: '#/definitions/EventTypeSchema'
```
dehora: This ref style is used in a few places in the pr. The tooling seems generates better output compared to using allOf, and it doesn't change the JSON structure.

api/nakadi-event-bus-api.yaml
```diff
@@ -827,135 +685,118 @@ definitions:
...
-      - data_op
+      Usually represents a status transition in a Business process.    
+    allOf:  
+      - $ref: '#/definitions/Event'
+      - type: object
+        properties:
+          metadata:
+              $ref: '#/definitions/EventMetadata'
+        required:
+            - metadata
```
dehora: This way of laying out `allOf` allowed codegen to extend from Event without obliterating the fields in the subclass (apparently codegen has very limited capabillties for polymorphism). We want the data/business object to subclass from Event as that's the object used in the generated api stubs . Without this the codegen stub http callers can't be used with business and data categories or worse they'll remove data when marshalling json content.

api/nakadi-event-bus-api.yaml
```diff
@@ -90,7 +90,7 @@ info:
     parameters (number of partitions, retention times, etc), as well as their modification.
 
 
-  version: '0.5.1'
+  version: '0.6.0'
```
dehora: Bump the minor revision number because the object definitions have changed, but the API JSON is the same.

api/nakadi-event-bus-api.yaml
rcillo: @dehora actually it's already implemented. Which environment did you test it?

api/nakadi-event-bus-api.yaml
```diff
@@ -272,18 +232,6 @@ paths:
           description: Client is not authenticated
           schema:
             $ref: '#/definitions/Problem'
-        '404':
```
rcillo: Why do you consider 404 not worth mentioning?

api/nakadi-event-bus-api.yaml
```diff
@@ -681,16 +564,8 @@ paths:
...
-          description: EventType not found
-          schema:
-            $ref: '#/definitions/Problem'
-        '500':
-          description: Server error
-          schema:
-            $ref: '#/definitions/Problem'
-        '503':
-          description: Service (temporarily) unavailable
+        '401':
```
rcillo: I would say that 401 is not worth mentioning. It's clear what it's about and since all endpoints are protected by default, I think it would be fine to omit this status code.

api/nakadi-event-bus-api.yaml
rcillo: We weren't aware that this is the default by the time of writing 👍

api/nakadi-event-bus-api.yaml
dehora: from the docker containers last week. I'll recheck

api/nakadi-event-bus-api.yaml
dehora: > Why do you consider 404 not worth mentioning?

@rcillo, excellent question. If we consider the definition to be documentation _about the API_, then 404 says very little about things to inform _an API user_, ie, it's unremarkable that asking for an event type that does not exist will produce 404 (unless the API doesn't, which is a different thing). So on that front all it really does is make the documentation longer than it needs to be, by describing standard HTTP behaviour. 

In general, one can make the same why argument for almost all the http codes but that's going to make for a very long API document :) Opinions can vary on what significant is, but common client/server that appear in reality and are really about HTTP itself, are probably best skipped (400, 404-406, 411-418, 500, 502, 503, 504). As a counter-example I'd probably document `410 Gone` as that's not always seen in the wild (eg it might say something to an API user that this thing did once or recently exist for certain delete scenarios and give the API designer an opportunity to describe that). 

On the other hand, you'd include the very mundane 200 because it allows you to describe the response, also 201/202 as they describe how creates/updates work which tend to matter to API users, and you might include ones like 409, 422, 429 as they might say something interesting about the how the API handles edits, etc. For auth I'd generally include `401` as it lets the designer say something about how to authorize with the server (oauth, basic, api keys, etc), and I'd skip `403` unless there was a specific API scenario in mind that need to make a distinction between not allowed and won't ever allow. The `3xx` redirect range are often worth documenting _for APIs_ because it's not unusual for programmatic clients to not follow redirections (eg a media API might want to document it use and support of redirection) whereas it's built into general purpose clients like browsers.

But like I say, opinions vary on what matters :)

api/nakadi-event-bus-api.yaml
dehora: (to be precise, the resource was there but there was no content being returned, ie it didn't have any information.)

PR #291, opened 2016-06-16
aruha-246 fix data change event partition

rcillo:
When defining `partition_key_fields` users expect that the lookup path
to be relative to the originally defined schema, but since Nakadi
manipulates the schema for DataChangeEvent's, we need to add this
special case in order to scope the path to the nested "data" field.

This ways users no longer need to prepend "data." to
partition_key_fields.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -792,7 +792,8 @@ paths:
 
         - `hash`: Resolution of the partition follows the computation of a hash from the value of
           the fields indicated in the EventType's `partition_key_fields`, guaranteeing that Events
-          with same values on those fields end in the same partition.
+          with same values on those fields end in the same partition. Given the event type's category
+          is DataChangeEvent, field path is considered relative to "data".
       responses:
         '200':
           description: Returns a list of all partitioning strategies known to Nakadi
@@ -1158,7 +1159,7 @@ definitions:
           type: string
         description: |
           Required when 'partition_resolution_strategy' is set to 'hash'. Must be absent otherwise.
-          Indicates the fields used for evaluatiion the partition of Events of this type.
+          Indicates the fields used for evaluation the partition of Events of this type.
 
           If set it MUST be a valid required field as defined in the schema.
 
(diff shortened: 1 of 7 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/291" target="_blank">ReviewNinja

src/main/java/de/zalando/aruha/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -28,6 +29,7 @@ public String calculatePartition(final EventType eventType, final JSONObject eve
             final int hashValue = partitionKeyFields.stream()
                     // The problem is that JSONObject doesn't override hashCode(). Therefore convert it to
                     // a string first and then use hashCode()
+                    .map(pkf -> EventCategory.DATA.equals(eventType.getCategory()) ? "data." + pkf : pkf)
```
antban: Maybe we need to create constant field with value `data` just to know were exactly it's used. 
I've found only EventTypeValidation#wrapSchemaInData, and it will be hard for me to understand where else it could happen.

rcillo: @antban  👍 done https://github.com/zalando/nakadi/pull/291/commits/054d421217dfbf4215ddf1b0fd5b447d2f675276

I would not invest too much effort into refactoring all this category thing because I think it would not bring too much value now. In the future, when we focus on improving schema registry, then I would think about some generalisations in order to remove these special cases.

src/main/java/de/zalando/aruha/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -13,6 +15,8 @@
 
 public class HashPartitionStrategy implements PartitionStrategy {
 
+    private final String DATA_PATH_PREFIX = EventValidation.DATA_CHANGE_WRAP_FIELD + ".";
```
v-stepanov: static?

api/nakadi-event-bus-api.yaml
```diff
@@ -792,7 +792,8 @@ paths:
 
         - `hash`: Resolution of the partition follows the computation of a hash from the value of
           the fields indicated in the EventType's `partition_key_fields`, guaranteeing that Events
-          with same values on those fields end in the same partition.
+          with same values on those fields end in the same partition. Given the event type's category
```
v-stepanov: Nice way to avoid `if` word 😄

v-stepanov:
:shipit: as soon as tests are successful (for now the build is not finished yet)

v-stepanov:
one more :shipit:

PR #292, opened 2016-06-16
ARUHA-245 Build scripts refactored

antban:
Just removed senza.yaml file
The guy who really want to use senza can prepare this file by himself


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/292" target="_blank">ReviewNinja

build.sh
```diff
@@ -0,0 +1,114 @@
...
+fi
+
+pierone login
+if [ $? -ne 0 ]; then
+    echo "Failed to login to pierone" &>2
+    exit 2
+fi
+
+APPLICATION_ID="nakadi"
+DOCKER_IMAGE_TAG="pierone.stups.zalan.do/aruha/${APPLICATION_ID}"
```
valgog: Hm... this should not be in the public repository I suppose

deploy.sh
```diff
@@ -0,0 +1,109 @@
...
+    echo "################################################################################"
+    echo "#                                                                              #"
+    echo "# WARNING: You are deploying into the production environment                   #"
+    echo "#                                                                              #"
+    echo "################################################################################"
+    sleep 3
+fi
+
+# Check if stack configuration exists
+CONFIG_FILE="../nakadi-private/config/$STACK_NAME"
```
valgog: Is it valid?

deploy.sh
```diff
@@ -0,0 +1,109 @@
...
+    exit 4
+fi
+
+# Load configuration
+IFS=$'\n\t'
+
+SENZA_PARAMS=(
+"ApplicationId=nakadi" \
+"ApplicationVersion=${VERSION}" \
+"DockerTag=pierone.stups.zalan.do/aruha/nakadi" \
```
valgog: Private infrastructure references here as well

build.sh
antban: Yes, forgot to move it to configuration.
The problem is that we are trying to create this script in public repo instead of moving senza.yaml to private repo

deploy.sh
antban: Probably it should be used as parameter, but for now - it just exists, and I've never experienced situation where this must be done from different paths.

v-stepanov:
🚢

PR #295, opened 2016-06-20
aruha-205 fix scm source json missing

rcillo:
In order to fix the root cause of this violation type, it's required to
add the scm to the built image.

I also needed to change the base image since all others weren't available anymore.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM zalando/python:3.4.0-2
+FROM registry.opensource.zalan.do/stups/python:3.4.0-4
 
 COPY requirements.txt /
 RUN pip3 install -r /requirements.txt
@@ -9,6 +9,7 @@ ADD nakadi /nakadi
 RUN find ./nakadi | grep -E "(__pycache__|\.pyc|\.pyo$)" | xargs rm -rf
 
 ADD run.sh /nakadi/run.sh
+ADD scm-source.json /scm-source.json
 
 WORKDIR /
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .cache
 __pycache__
 .coverage
+scm-source.json
 *.pyc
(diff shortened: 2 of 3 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/295" target="_blank">ReviewNinja

Dockerfile
```diff
@@ -9,6 +9,7 @@ ADD nakadi /nakadi
 RUN find ./nakadi | grep -E "(__pycache__|\.pyc|\.pyo$)" | xargs rm -rf
 
 ADD run.sh /nakadi/run.sh
+ADD scm-source.json /scm-source.json
```
v-stepanov: Where is the file itself?

v-stepanov:
You also need to check base image for local-kafka dockerfile
https://github.com/zalando/nakadi/blob/eventstore-latest/local-test/Dockerfile.kafka#L1

rcillo:
@v-stepanov done https://github.com/zalando/nakadi/pull/295/commits/93fa9ca46284a233fe60c7b16edd8fa7a52290cc

Dockerfile
rcillo: This file is usually considered part of the build artifacts, e.g. not versioned.

v-stepanov: My only concern here is that it will be not possible to build docker image if the file is not present. (and it will be present only after we run build script from private repo)
Maybe it's possible to add that to container only if the file exists? (to avoid failure when the file doesn't exist)

antban: I think that it's ok to fail build if scm-source.json is missing. Build script (if it exists) must generate it.

v-stepanov: My only problem is that eventstore is opensource project and the build script is in our private repo.
And we require something that can be done only with help of private repo script. That's why I think that adding scm-source.json to container should be optional.

rcillo: @v-stepanov  Nakadi JVM has the same in it's Dockerfile.

v-stepanov: Nakadi JVM has means to generate it in gradle file
Here the problem is that dockerfile requires something that can be generated only using stuff from private repository.

rcillo:
Since we are not fixing these violations anymore, I'm closing this PR. If in the future we are to solve then, just look here.

PR #303, opened 2016-06-27
@Aruha 259 hila event delivered twice

v-stepanov:
@antban @rcillo @jkliff 
Please review. 
This fixes double-event bug; problem with empty batch sent at the end; problem with being not possible to commit very first event; + some acceptance tests

### DIFF
--- a/src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
+++ b/src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
@@ -1,0 +1,44 @@
+package de.zalando.aruha.nakadi.webservice.hila;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.common.collect.ImmutableList;
+import de.zalando.aruha.nakadi.domain.Cursor;
+
+import javax.annotation.concurrent.Immutable;
+import java.util.List;
+import java.util.Map;
+
+import static java.util.Collections.unmodifiableList;
+
+@Immutable
+public class StreamBatch {
+
+    private Cursor cursor;
+    private List<Map> events = ImmutableList.of();
+
+    public StreamBatch(@JsonProperty("cursor") final Cursor cursor, @JsonProperty("events") final List<Map> events) {
+        this.cursor = cursor;
+        this.events = events;
+    }
+
+    public Cursor getCursor() {
+        return cursor;
+    }
+
+    public List<Map> getEvents() {
+        return unmodifiableList(events);
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        final StreamBatch that = (StreamBatch) o;
+        return cursor.equals(that.cursor) && events.equals(that.events);
+    }
+
+    @Override
+    public int hashCode() {
+        return 31 * cursor.hashCode() + events.hashCode();
+    }
+}
--- a/src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
+++ b/src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
@@ -7,6 +7,7 @@ import de.zalando.aruha.nakadi.domain.Cursor;
 import javax.annotation.concurrent.Immutable;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import static java.util.Collections.unmodifiableList;
 
@@ -14,11 +15,11 @@ import static java.util.Collections.unmodifiableList;
 public class StreamBatch {
 
     private Cursor cursor;
-    private List<Map> events = ImmutableList.of();
+    private List<Map> events;
 
     public StreamBatch(@JsonProperty("cursor") final Cursor cursor, @JsonProperty("events") final List<Map> events) {
         this.cursor = cursor;
-        this.events = events;
+        this.events = Optional.ofNullable(events).orElse(ImmutableList.of());
     }
 
     public Cursor getCursor() {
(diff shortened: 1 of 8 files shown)


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/303" target="_blank">ReviewNinja

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
```diff
@@ -0,0 +1,48 @@
...
+
+import static java.util.Collections.unmodifiableList;
+
+@Immutable
+public class StreamBatch {
+
+    private Cursor cursor;
+    private List<Map> events = ImmutableList.of();
+
+    public StreamBatch(@JsonProperty("cursor") final Cursor cursor, @JsonProperty("events") final List<Map> events) {
```
rcillo: Is it necessary to specify `JsonProperty`?

v-stepanov: Without that it will not be possible for Jackson to deserialise. It will not understand where to pass json properties.

antban:
I like it

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
rcillo: Why isn't it necessary for other serialisable classes, like EventType, for example?

v-stepanov: For EventType deserialisation happens using setters, which doesn't require any additional markup.
For this class there are no setters as I wanted it to be immutable, so the deserialisation happens using constructor and Jackson needs to know which json property should go to which constructor parameter.

src/main/java/de/zalando/aruha/nakadi/service/subscription/KafkaClient.java
```diff
@@ -23,7 +23,7 @@ public KafkaClient(final Subscription subscription, final TopicRepository topicR
         try {
             for (final String eventType : subscription.getEventTypes()) {
                 topicRepository.materializePositions(eventType, subscription.getStartFrom()).entrySet().forEach(
-                        e -> offsets.put(new Partition.PartitionKey(eventType, e.getKey()), e.getValue()));
+                        e -> offsets.put(new Partition.PartitionKey(eventType, e.getKey()), e.getValue() - 1));
```
rcillo: Isn't there any problems with special boundary values, e.g. BEGIN?

rcillo:
Do you think it would be possible to extract offset comparison and manipulation to a single class? I see that we have some + 1 and - 1 spread over the project and sometimes it's due to the differences on how nakadi and kafka handle offsets.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
jhorstmann: > Without that it will not be possible for Jackson to deserialise. It will not understand where to pass json properties.

The jackson parameter-names and jdk8 modules should also solve this.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+import static java.util.Collections.unmodifiableList;
+
+@Immutable
+public class StreamBatch {
+
+    private Cursor cursor;
+    private List<Map> events;
```
jhorstmann: What's the generic type of this `Map`? The fields could also be `final`.

src/main/java/de/zalando/aruha/nakadi/service/subscription/KafkaClient.java
v-stepanov: no, here it will not have BEGIN value, as the value comes directly from Kafka.

v-stepanov:
@rcillo I would be more than happy to isolate that +/-1
But currently kafka offsets migrated to Hila functionality, so first we will need to do a refactoring to completely isolate kafka offsets on kafka-layer and to have string offsets everywhere else. But I believe it should be done by a separate refactoring ticket.

rcillo:
@v-stepanov ok. I created a ticket for that https://techjira.zalando.net/browse/ARUHA-263

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
rcillo: @v-stepanov I saw you commited the finals but what about the generic type?

v-stepanov: 1) Map is basically <String, Object> but I don't see much sense to specify that as the key itself can be String or List or Map. So defining a map as <String, Object> doesn't really help.
2) Added finals, good point.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
v-stepanov: @jhorstmann good point, I will check it.
But I believe adding parameter-names module should be done as a separate task, not in a context of this one.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/hila/StreamBatch.java
v-stepanov: @jhorstmann @rcillo do you think that `<String, Object>` will make it better?

antban:
:shipit:

Issue #306, opened 2016-06-28
Event type name conflict "." and "_"


rcillo:
Kafka handles dots "." as underscores "_" in the sense that "event.type.name" would colide with "event_type_name".

PR #308, opened 2016-06-29
Aruha 249 decouple event type from kafka topic

rcillo:
In order to improve overall design quality and fix some bugs, we are decoupling event type name from Kafka topic name.

Fixes https://github.com/zalando/nakadi/issues/306


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/308" target="_blank">ReviewNinja

src/acceptance-test/java/de/zalando/aruha/nakadi/repository/kafka/KafkaRepositoryAT.java
```diff
@@ -1,41 +1,36 @@
 package de.zalando.aruha.nakadi.repository.kafka;
 
-import static org.echocat.jomon.runtime.concurrent.Retryer.executeWithRetry;
```
rcillo: I killed kittens again. Sorry. Blame Intellij.

src/acceptance-test/java/de/zalando/aruha/nakadi/webservice/BaseAT.java
```diff
@@ -11,11 +22,41 @@
...
+    protected JdbcTemplate template;
+    protected ObjectMapper mapper;
+
     static {
         RestAssured.port = PORT;
         RestAssured.defaultParser = Parser.JSON;
     }
+
+    @Before
+    public void setUp() throws Exception {
```
rcillo: I'm not proud of this `setUp`. The fact is that tests used to rely on a topic called "test-topic" which is created by a bash script that initialise local storages. E.g. the tests rely on some external setup.

After the decoupling, topic name is retrieved indirectly through event type table. So the way I found to solve this was to inject in the database the proper entry, since these are acceptance tests and mocking is not an option.

rcillo:
@v-stepanov @antban we haven't discussed how the database migration should be handled yet. I'll write the required SQL and I suggest creating another ticket with another PR in order to structure migrations properly.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -177,6 +186,10 @@ private void validateSchema(final EventType eventType) throws InvalidEventTypeEx
         }
     }
 
+    private void assignTopic(final EventType eventType) {
+        eventType.setTopic(uuidGenerator.randomUUID().toString());
```
v-stepanov: I believe for debugging purposes it would be better if topic name is prefixed with event-type name. What do you think?

database/nakadi/10_data/04_tables/01_event_type.sql
```diff
@@ -1,7 +1,10 @@
 CREATE TABLE IF NOT EXISTS zn_data.event_type (
   et_name varchar(255) NOT NULL PRIMARY KEY CHECK (et_name <> ''),
+  et_topic varchar(255) NOT NULL,
```
v-stepanov: Just wonder why you decided to have it as a separate field?

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
rcillo: I thought about something like that. But topic name has some limitations which is already source of bugs:

1. underscores and dots conflicts
2. size limitation
3. other characters limitations

rcillo: In order to do so, correctly, we should add some more limitations to the event type name creation and handle those already created that doesn't fit in this constraints.

database/nakadi/10_data/04_tables/01_event_type.sql
rcillo: I even considered having a separated table just to relate both. Event types and topics should be decoupled. I'm considering removing it from the EventType POJO and adding a new method to the repository in order to retrieve it directly from the database.

rcillo: We could also go for a more coupled solution and actually add it to the event type object also exposing it to users.

v-stepanov: If you take it out from EventType POJO - you will loose caching and will have to take topic name from DB each time :)

v-stepanov: No, I think the way it's implemented now is quite good. I think we should not expose that to user.

rcillo: Yes... 😅 that's why I added it to the POJO. But I'm not sure if I like it. Every design problem can be solved by yet another layer, but I tried to avoid splitting event type object in a "user facing" object and in "internal object". But currently I think it's weird to have this "ignored" property that is only used internally.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
v-stepanov: I think we can create some translation from event-type name to topic name. Like:
`topicName = translate(eventType.name) + uniqueRandomBlah

I'm just afraid that we will have hard time matching eventType name to topic name in kafka. For example the user reports that he has some problems with eventType "order.updater.ad.12" and if we will want to check something in kafka we will first need to hack into DB with tunnel to see the topic name.

But ok, I'm fine with current implementation, I hope that will not happen toо often.

rcillo:
@v-stepanov @antban this is the SQL for data migration

```sql
ALTER TABLE zn_data.event_type ADD COLUMN et_topic varchar(255) NOT NULL DEFAULT ''

CREATE UNIQUE INDEX idx_et_topic ON zn_data.event_type (et_topic);

UPDATE zn_data.event_type
SET et_topic = subquery.et_name
FROM (SELECT et_name FROM zn_data.event_type) AS subquery
WHERE zn_data.event_type.et_name = subquery.et_name;

ALTER TABLE zn_data.event_type ALTER COLUMN et_topic DROP DEFAULT;
```

src/main/java/de/zalando/aruha/nakadi/util/UUIDGenerator.java
```diff
@@ -0,0 +1,9 @@
+package de.zalando.aruha.nakadi.util;
+
+import java.util.UUID;
+
+public class UUIDGenerator {
+    public UUID randomUUID() {
+        return UUID.randomUUID();
+    }
```
antban: What is the purpose of special class?
We have several places where we generate UUIDs with UUID.randomUUID(),
Now we have places with this generator and UUID call directly

v-stepanov: Having it as a separate class can be useful when creating tests (I see that Ricardo mocks it in the test)
So I now added usage of that class in another place as well.
The places where we generate UUID in tests I didn't touch because those are tests :)

antban:
:+1:

src/main/java/de/zalando/aruha/nakadi/util/UUIDGenerator.java
rcillo: Yes, @v-stepanov is right. I did it because I could not mock a final class.

PR #316, opened 2016-07-04
ARUHA-270: use 30 as default value for batch_flush_timeout

v1ctor:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/316" target="_blank">ReviewNinja

v1ctor:
@antban @adyach @v-stepanov

src/main/java/de/zalando/aruha/nakadi/service/EventStreamConfig.java
```diff
@@ -7,6 +7,9 @@
 
 @Immutable
 public class EventStreamConfig {
+
+    private static int BATCH_FLUSH_TIMEOUT_DEFAULT = 30;
```
v-stepanov: I don't like that we have this constant in two places now: here and additionally in controller method parameter annotation. But I can't offer a better solution now :)

The only possibility I can think about is to move all default values from controller-method to EventStreamConfig class. But for that we will need to change all parameters of controller-method to be Objects instead of primitives (Integer instead of int etc.) - because we will need them to differ between 0 and null values (if we use primitives that will not be possible)

What do you think?

v1ctor: Yes, I think it's a good idea to hide and resolve all default values inside builder. I'll try to do that.

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -16,12 +16,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.http.HttpStatus;
-import org.springframework.web.bind.annotation.PathVariable;
-import org.springframework.web.bind.annotation.RequestHeader;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RequestMethod;
-import org.springframework.web.bind.annotation.RequestParam;
-import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.bind.annotation.*;
```
v-stepanov: We have an agreement in our team not to use wildcard imports :)

src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -66,11 +56,11 @@ public EventStreamController(final TopicRepository topicRepository, final Object
...
-            @RequestParam(value = "batch_limit", required = false, defaultValue = "1") final int batchLimit,
-            @RequestParam(value = "stream_limit", required = false, defaultValue = "0") final int streamLimit,
-            @RequestParam(value = "batch_flush_timeout", required = false, defaultValue = "30") final int batchTimeout,
-            @RequestParam(value = "stream_timeout", required = false, defaultValue = "60") final int streamTimeout,
-            @RequestParam(value = "stream_keep_alive_limit", required = false, defaultValue = "0") final int streamKeepAliveLimit,
+            @RequestParam(value = "batch_limit", required = false) final Integer batchLimit,
+            @RequestParam(value = "stream_limit", required = false) final Integer streamLimit,
+            @RequestParam(value = "batch_flush_timeout", required = false) final Integer batchTimeout,
+            @RequestParam(value = "stream_timeout", required = false) final Integer streamTimeout,
+            @RequestParam(value = "stream_keep_alive_limit", required = false) final Integer streamKeepAliveLimit,
```
v-stepanov: Another convention I forgot to tell you: null handling.
In Nakadi we consider everything to be none-nullable, if there is something that can be null - it should be marked as `@Nullable`
I believe all these parameters are nullable.

src/main/java/de/zalando/aruha/nakadi/service/EventStreamConfig.java
```diff
@@ -115,39 +123,55 @@ public Builder withCursors(final Map<String, String> cursors) {
             return this;
         }
 
-        public Builder withBatchLimit(final int batchLimit) {
-            this.batchLimit = batchLimit;
+        public Builder withBatchLimit(final Integer batchLimit) {
```
v-stepanov: also please mark parameters of the builder methods `@Nullable`

v-stepanov:
:shipit:

PR #317, opened 2016-07-05
ARUHA-277: Use checkstyle

v1ctor:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/317" target="_blank">ReviewNinja

antban:
:shipit: :+1: :zapprit:

v-stepanov:
I would also love to see a check that the line is not longer than 120 chars :)
But that will also force us to fix all places we violate that. As we now have a very big branch with HILA I would not like to add that check now because the merging will be a nightmare. Maybe later.

v-stepanov:
:shipit:

v-stepanov:
@antban your approval didn't work. I think it should be a comment with just a singe emoji.

v-stepanov:
@v1ctor checkstyle task doesn't check the code in acceptance tests
Try to run `check` and it will find a lot of problems in acceptance tests

PR #319, opened 2016-07-06
ARUHA-278: add zappr config

v1ctor:


review-ninja:
<a href="https://app.review.ninja/zalando/nakadi/pull/319" target="_blank">ReviewNinja

.zappr.yaml
```diff
@@ -0,0 +1,15 @@
+approvals:
+  minimum: 1
+  ignore: pr_opener
+  pattern: "^([aA][cC][kK]|[lL][gG][tT][mM]|👌|:ok_hand:|👍|:\\+1:|:thumbsup:|:shipit:)$"
```
v-stepanov: I think there are two many. Somebody can accidentally approve. 
I think we only need `:shipit:` and ":+1:"
(and `:-1:` for disapproval)

v-stepanov:
:shipit:

PR #320, opened 2016-07-07
Crutch for closing connections in CLOSE_WAIT state

antban:
Right now AWS ELB behaves strangely with connections closed from client side:
 - In case of http ELB it leaves backend connection in ESTABLISHED state
 - In case of tcp ELB it sends FIN packet, but after that it continues to send ACKS for half open connections (which means that client is still receiving for nakadi)

In this PR special crutch introduced, it monitors /proc/net/tcp and /proc/net/tcp6 files to detect connections in closing (or unknown) states, and notifies long polling request processors about connection close.

Right now only /event-types/{}/events works in long polling, so the crutch is used only there.

Crutch is made under feature toggle, so in case of any incorrect work it can be turned of without service restart.


v-stepanov:
@antban ok, so this is basically 2 in 1, right?

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+        public String toString() {
+            return "ConnectionInfo{" +
+                    "address=" + address +
+                    ", port=" + port +
+                    '}';
+        }
+    }
+
+    static class ListenerInfo implements Comparable<ListenerInfo> {
+        private final Date addedAt;
```
adyach: Maybe it is better to use long field here, because I do not see any usage of Date class only milliseconds.

v-stepanov:
@antban can you please name the branches not only with ticket id but also add few words description like ARUHA-193-connection-problem-crutch

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+                    return val;
+                }
+            }
+            throw new IllegalArgumentException("Failed to find connection state from " + value);
+        }
+    }
+
+    static class ConnectionInfo {
+        private final InetAddress address;
+        private final int port;
```
v-stepanov: You identify connection by IP and port. What if there are several connections from a machine to Nakadi? I think it may be quite a common case that the client-app opens several connections. (or that several client-apps are running at the same machine)

src/main/java/de/zalando/aruha/nakadi/service/EventStreamConfig.java
```diff
@@ -13,7 +13,7 @@
     private static final int BATCH_LIMIT_DEFAULT = 1;
     private static final int STREAM_LIMIT_DEFAULT = 0;
     private static final int BATCH_FLUSH_TIMEOUT_DEFAULT = 30;
-    private static final int STREAM_TIMEOUT_DEFAULT = 60;
+    private static final int STREAM_TIMEOUT_DEFAULT = 0;
```
v-stepanov: I think it's fine to keep it 60. Our users already informed about that.

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
antban: Connection is uniquely identified by two pairs of ip:port. There can't be more than 1 connection with the same remote ip:port to my application.

src/main/java/de/zalando/aruha/nakadi/util/TimeBasedCache.java
```diff
@@ -0,0 +1,68 @@
+package de.zalando.aruha.nakadi.util;
+
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Function;
+
+public class TimeBasedCache<Key, Value> {
```
v-stepanov: It's cool that you decided to build your own cache :) But wasn't there anything which is already created and suits our needs?
E.g. in google guava there is a CacheBuilder https://google.github.io/guava/releases/19.0/api/docs/com/google/common/cache/CacheBuilder.html

src/test/java/de/zalando/aruha/nakadi/controller/EventStreamControllerTest.java
```diff
@@ -322,7 +331,9 @@ public void reportCurrentNumberOfConsumers() throws Exception {
             Thread.sleep(500);
 
             executeWithRetry(
-                    () -> {assertThat(counter.getCount(), equalTo((long) clients.size()));},
+                    () -> {
```
v-stepanov: why do we need braces here at all?

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+import org.apache.commons.codec.binary.Hex;
+import org.apache.commons.lang3.ArrayUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.scheduling.annotation.Scheduled;
+import org.springframework.stereotype.Service;
+
+@Service
```
v-stepanov: Why do you need this one?

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+        LOG.debug("Listening for connection to close using crutch (" + address + ":" + port + ")");
+        final Date now = new Date();
+        synchronized (this) {
+            toAdd.add(new ListenerInfo(now, new ConnectionInfo(address, port), onCloseListener));
+        }
+    }
+
+    @Scheduled(fixedDelay = 1000)
+    public void refresh() throws IOException {
+        if (!featureToggleService.isFeatureEnabled(FEATURE_NAME)) {
```
v-stepanov: I think we decided to have these constants in FeatureToggleService class

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+            return;
+        }
+        final long selector = System.currentTimeMillis() - DELAY_MILLIS;
```
v-stepanov: Why you have this 100ms threshold?

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+        synchronized (this) {
+            final Iterator<ListenerInfo> it = toAdd.iterator();
+            while (it.hasNext()) {
+                final ListenerInfo first = it.next();
+                if (first.addedAt.getTime() >= selector) {
+                    break;
+                }
+                listeners.computeIfAbsent(first.connectionInfo, info -> new ArrayList<>()).add(first.callback);
```
v-stepanov: how this `listeners` collection is cleaned if it was not a "crutch case" (connection was closed in a good way)?

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
antban: no

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
antban: ok, i'll change it

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
antban: Well, initially it was related to testing purposes, but you are right, I can remote it now

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
antban: Connection will disappear as usual

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+                it.remove();
+            }
+        }
+        final Map<ConnectionInfo, ConnectionState> currentConnections = readAllConnectionStates();
+        final Set<ConnectionInfo> keys = new HashSet<>(listeners.keySet());
+        final AtomicInteger closedCount = new AtomicInteger();
+        for (final ConnectionInfo key : keys) {
```
v-stepanov: you could be hipster here and do 
`listeners.keySet().stream().filter(...).map(...).count()`
:)

src/test/java/de/zalando/aruha/nakadi/controller/EventStreamControllerTest.java
antban: why i touched this place???, anyway, I'll remove it

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
v-stepanov: Yes, connection will disappear. I worry about the element in `listeners` map. For the connection that will be closed in a good way you will still have element in `listeners` map. And it will grow and grow... Or am I missing something?

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
v-stepanov: like
```java
final long closedNum = listeners.keySet()
        .stream()
        .filter(key -> CLOSED_STATES.contains(currentConnections.getOrDefault(key, ConnectionState.TCP_CLOSE)))
        .mapToLong(key -> {
            LOG.info("Notifying about connection close via crutch: " + key);
            return listeners.remove(key).stream().filter(BooleanSupplier::getAsBoolean).count();
        })
        .sum();
```

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+            try (FileInputStream in = new FileInputStream(f)) {
+                result.putAll(getCurrentConnections(in));
+            } catch (final FileNotFoundException e) {
+                LOG.warn("Failed to find file " + f.getName() + ", skipping");
+            }
+        }
+        return result;
+    }
+
+    Map<ConnectionInfo, ConnectionState> getCurrentConnections(final InputStream in) throws IOException {
```
v-stepanov: why it has package scope?

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+                    }
+                } catch (DecoderException | RuntimeException ex) {
+                    LOG.error("Failed to parse line, skipping: " + line, ex);
+                }
+            }
+        }
+        return connectionToState;
+    }
+
+    static InetAddress[] restoreAddresses(final String address) throws DecoderException, UnknownHostException {
```
v-stepanov: why it has package scope?

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+                        continue;
+                    }
+                    // read remote address
+                    final String[] remoteAddressPair = tokenizer.nextToken().split(":");
+                    final InetAddress[] remoteAddresses = restoreAddresses(remoteAddressPair[0]);
+                    final int remotePort = Integer.parseInt(remoteAddressPair[1], 16);
+
+                    // read connection state
+                    final ConnectionState connectionState = ConnectionState.fromCode(tokenizer.nextToken());
+                    for (final InetAddress tmp : remoteAddresses) {
```
v-stepanov: not a best name

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+                        connectionToState.put(new ConnectionInfo(tmp, remotePort), connectionState);
+                    }
+                } catch (DecoderException | RuntimeException ex) {
```
v-stepanov: final

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,294 @@
...
+                    return new InetAddress[]{result, InetAddresses.getEmbeddedIPv4ClientAddress((Inet6Address) result)};
+                }
+            }
+            return new InetAddress[]{result};
+        } else {
+            return new InetAddress[]{InetAddress.getByAddress(data)};
+        }
+    }
+
+    private static final Set<ConnectionState> CLOSED_STATES = Arrays.asList(
```
v-stepanov: You can just use `ImmutableSet.of(...)` here instead of all conversions.

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
antban: There is no way it can grow. 
Each time on refresh() call there is a for cycle, that goes thru all the connections in listeners collection. 
If connection must be removed, than it is removed with notification. If it's not - than connection is still open.

Maybe you misunderstood the way it is removed:
```
listeners.remove(key)
```

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
antban: ok

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
antban: `@VisibleForTesting`, but you don't like it

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
antban: The same

src/main/java/de/zalando/aruha/nakadi/service/EventStreamConfig.java
antban: No, I insist that we must remove it. It's in the API and users were notified that it was a temporary solution.

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
v-stepanov: ok, got it now

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
v-stepanov: It's not that I don't like the annotation itself. I don't like the idea of testing private methods making them package-level. I think that is an indication of that the design of the class could be better.
@VisibleForTesting is better than not having it and still testing a package-level method.

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
```diff
@@ -0,0 +1,238 @@
...
+        }
+        synchronized (toAdd) {
+            toAdd.forEach((conn, toAddListeners) -> {
+                listeners.computeIfAbsent(conn, c -> new ArrayList<>()).addAll(toAddListeners);
+            });
+            toAdd.clear();
+        }
+        final Map<ConnectionInfo, ConnectionState> currentConnections = readAllConnectionStates();
+        final long closedCount =
+                new HashSet<>(listeners.keySet()).stream()
```
v-stepanov: why do you need here `new HashSet<>()` ?

src/test/resources/de/zalando/aruha/nakadi/service/tcp
```diff
@@ -0,0 +1,35 @@
+  sl  local_address rem_address   st tx_queue rx_queue tr tm->when retrnsmt   uid  timeout inode
```
v-stepanov: Is this file used?

src/main/java/de/zalando/aruha/nakadi/service/ClosedConnectionsCrutch.java
antban: to avoid ConcurrentModificationException

src/test/resources/de/zalando/aruha/nakadi/service/tcp
antban: I'm checking that files are parsable

antban: But i will split test into 2 tests for each file, it will be better

v-stepanov:
:shipit:

PR #321, opened 2016-07-07
ARUHA-182: return 500 status code if doesn't exist topic in kafka for corresponding eventType

v1ctor:


src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -88,6 +95,11 @@ public StreamingResponseBody streamEvents(
 
                 // validate parameters
                 if (!topicRepository.topicExists(topic)) {
+                    Optional<EventType> eventTypeO = eventTypeRepository.findByNameO(eventTypeName);
+                    if (eventTypeO.isPresent()) {
+                        writeProblemResponse(response, outputStream,INTERNAL_SERVER_ERROR, "topic is absent in kafka");
+                        return;
+                    }
```
antban: I think that it will interfere with #308

v1ctor: I agree with you, but I think we will be able to merge it when @rcillo will come back.

src/main/java/de/zalando/aruha/nakadi/repository/EventTypeRepository.java
```diff
@@ -6,13 +6,22 @@
...
 import java.util.List;
+import java.util.Optional;
 
 public interface EventTypeRepository {
 
     void saveEventType(EventType eventType) throws InternalNakadiException, DuplicatedEventTypeNameException;
 
     EventType findByName(String name) throws InternalNakadiException, NoSuchEventTypeException;
 
+    default Optional<EventType> findByNameO(String name) throws InternalNakadiException {
```
v-stepanov: I don't really like that we now have two methods in this repository that do exactly the same but have different notations.   I can imagine that someone will create a third method that returns null because he doesn't like optionals 😄

v-stepanov:
:shipit:

PR #323, opened 2016-07-07
ARUHA-282: add toogle to switch off creation and deletion of event type

v1ctor:


src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -70,7 +77,11 @@ public EventTypeController(final EventTypeRepository eventTypeRepository,
     @RequestMapping(method = RequestMethod.POST)
     public ResponseEntity<?> createEventType(@Valid @RequestBody final EventType eventType,
                                              final Errors errors,
-                                             final NativeWebRequest nativeWebRequest) {
+                                             final NativeWebRequest nativeWebRequest)
+    {
+        if (featureToggleService.isFeatureEnabled("disable_event_type_creation")) {
```
adyach: Usually, I put it in constant

v1ctor: You're right.

v-stepanov: in parallel universe we have a long-living brach where we put these constants to FeatureToggleService class.
https://github.com/zalando/nakadi/blob/ARUHA-215-ut/src/main/java/de/zalando/aruha/nakadi/util/FeatureToggleService.java#L13

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -48,16 +51,20 @@
...
+    private final FeatureToggleService featureToggleService;
 
     @Autowired
     public EventTypeController(final EventTypeRepository eventTypeRepository,
                                final TopicRepository topicRepository,
                                final PartitionResolver partitionResolver,
-                               final Enrichment enrichment) {
+                               final Enrichment enrichment,
+                               final FeatureToggleService featureToggleService)
+    {
```
v-stepanov: Another good point for coding convention discussion :-D
Starting braces from a new line :scream:

adyach: Looks like C++ style

src/main/java/de/zalando/aruha/nakadi/repository/zookeeper/ZooKeeperHolder.java
```diff
@@ -55,13 +53,7 @@ public void init() throws Exception {
...
         if (exhibitorAddresses != null) {
             final Collection<String> exhibitorHosts = Arrays.asList(exhibitorAddresses.split("\\s*,\\s*"));
-            final Exhibitors exhibitors = new Exhibitors(exhibitorHosts, exhibitorPort,
-                    new BackupConnectionStringProvider() {
-                        @Override
-                        public String getBackupConnectionString() throws Exception {
-                            return zookeeperBrokers + zookeeperKafkaNamespace;
-                        }
-                    });
+            final Exhibitors exhibitors = new Exhibitors(exhibitorHosts, exhibitorPort, () -> zookeeperBrokers + zookeeperKafkaNamespace);
```
v-stepanov: I did such a refactoring in this place a couple of times. But then my branch was never merged to "master" branch. So this is a really bad sign to make this refactoring :laughing:

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
v1ctor: I start it from next line if only there is a line break in params. But of course we should discuss it, and better adjust checkstyle settings.

src/test/java/de/zalando/aruha/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -64,17 +66,19 @@
...
         final EventTypeController controller = new EventTypeController(eventTypeRepository, topicRepository,
-                partitionResolver, enrichment);
+                partitionResolver, enrichment, featureToggleService);
 
         final MappingJackson2HttpMessageConverter jackson2HttpMessageConverter =
             new MappingJackson2HttpMessageConverter(objectMapper);
 
         mockMvc = standaloneSetup(controller).setMessageConverters(new StringHttpMessageConverter(),
                 jackson2HttpMessageConverter).build();
+        doReturn(false).when(featureToggleService).isFeatureEnabled(any(FeatureToggleService.Feature.class));
```
v-stepanov: In general here you can just write `any()`

src/main/java/de/zalando/aruha/nakadi/util/FeatureToggleService.java
```diff
@@ -4,31 +4,54 @@
...
 
-    @Value("${nakadi.featureToggle.enableAll}")
-    private boolean forceEnableAll;
+    private static final String PREFIX = "/nakadi/feature_toggle/";
 
+    private final boolean forceEnableAll;
     private final ZooKeeperHolder zkHolder;
 
-    public FeatureToggleService(final ZooKeeperHolder zkHolder) {
+    public FeatureToggleService(boolean forceEnableAll, final ZooKeeperHolder zkHolder) {
```
v-stepanov: final for first param would be good

src/main/java/de/zalando/aruha/nakadi/util/FeatureToggleService.java
```diff
@@ -4,31 +4,54 @@
...
+            final Stat stat = zkHolder.get().checkExists().forPath(PREFIX + feature.getId());
             return stat != null;
         } catch (Exception e) {
-            LOG.warn("Error occurred when checking if feature '" + feature + "' is toggled", e);
+            LOG.warn("Error occurred when checking if feature '" + feature.getId() + "' is toggled", e);
             return false;
         }
     }
+
+    public static class Feature {
```
adyach: @v-stepanov @v1ctor 
Maybe it is good to have for particular users? In that case we can do A/B testing in live.

adyach: Why don't you want to move it to a separate class?

v1ctor: I think it would be better to do A/B testing infrastructure in different task.

What disadvantages you see in inner class?

adyach: It will grow bigger and bigger that will make FeatureToggleService less readable.
Also it looks like a class that can be used by others that's why it is not only for internal usage of FeatureToggleService.

v1ctor: It won't grow as fast as you think :)
And it can't be used without FeatureToggleService. 
Anyway I'd better do it enum the same way like Feature in SerializationConfig in Jackson.

src/main/java/de/zalando/aruha/nakadi/util/FeatureToggleService.java
```diff
@@ -4,31 +4,55 @@
...
             return false;
         }
     }
+
+    public enum Feature {
+
+        DISABLE_EVENT_TYPE_CREATION("disable_event_type_creation", false),
+        DISABLE_EVENT_TYPE_DELETION("disable_event_type_deletion", false),
+        HIGH_LEVEL_API("high_level_api", true),
+        ;
```
adyach: why do you need it here?

src/main/java/de/zalando/aruha/nakadi/util/FeatureToggleService.java
```diff
@@ -4,31 +4,55 @@
...
+            final Stat stat = zkHolder.get().checkExists().forPath(PREFIX + feature.getId());
             return stat != null;
         } catch (Exception e) {
-            LOG.warn("Error occurred when checking if feature '" + feature + "' is toggled", e);
+            LOG.warn("Error occurred when checking if feature '" + feature.getId() + "' is toggled", e);
             return false;
         }
     }
+
+    public enum Feature {
```
adyach: I meant that this class is used outside of FeatureToggleService that's why it should not be encapsulated in FeatureToggleService

v1ctor: Why not? I gave an example where it's like that.

src/main/java/de/zalando/aruha/nakadi/util/FeatureToggleService.java
v1ctor: No :)

v-stepanov: 😱

adyach: Why do you need this , and ; at the end ?

v1ctor: When you will add new value here you will get only one line diff in git. And it looks more clean.

adyach: Should we really care about diffs in git or about cleaner code ...

v1ctor: This code looks clean for me.

v-stepanov: between having nice code and having a nice diff I would choose having nice code

src/main/java/de/zalando/aruha/nakadi/util/FeatureToggleService.java
adyach: I looked at that example and you can see that in version 1.9 but in higher versions Feature was moved out

src/main/java/de/zalando/aruha/nakadi/util/FeatureToggleService.java
v1ctor: Don't you think that we are spending a lot of time arguing about code conventions? :)

src/main/java/de/zalando/aruha/nakadi/util/FeatureToggleService.java
v1ctor: https://github.com/FasterXML/jackson-core/blob/2.7/src/main/java/com/fasterxml/jackson/core/JsonFactory.java

v1ctor: I could do like you want, but I don't see any advantages.

adyach:
:shipit:

adyach:
:shipit:

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
rcillo: I see that it got merged with this comment not been addressed. Even if it's a very simple matter I don't feel comfortable with this approach. We should try to reply or address every comment.

PR #327, opened 2016-07-08
ARUHA-164: reject fields if they are not in the schema

v1ctor:


src/test/java/de/zalando/aruha/nakadi/utils/TestUtils.java
```diff
@@ -100,9 +102,16 @@ public static String resourceAsString(final String resourceName, final Class cla
     }
 
     public static EventType buildEventType(final String name, final JSONObject schema) {
+        return buildEventType(name, schema, Collections.emptyList());
+    }
+
+    public static EventType buildEventType(final String name, final JSONObject schema, List<String>
```
v-stepanov: final

v-stepanov:
I'm not sure if we can merge it right now. I think some of users will have problems with that. If I remember correctly eventlog guys were using metadata.eid as a partitioning key. (which is not part of schema of course)

src/test/java/de/zalando/aruha/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -264,9 +266,23 @@ public void whenDeleteEventTypeAndTopicDeletionExceptionThen503() throws Excepti
     }
 
     @Test
+    public void whenCreateEventTypeWithWrongPartitionKeyFieldsThen422() throws Exception {
+
+        final EventType eventType = buildEventType(randomValidEventTypeName(), new JSONObject("{ \"price\": 1000 }"),
+                Collections.singletonList("blabla"));
+
+        Mockito.doReturn(eventType).when(eventTypeRepository).findByName(eventType.getName());
+        Mockito.doThrow(new TopicDeletionException("dummy message", null)).when(topicRepository).deleteTopic(
```
rcillo: I could not understand why is this mock `doThrow` needed. Are you sure it's needed?

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -192,6 +193,13 @@ private void validateSchema(final EventType eventType) throws InvalidEventTypeEx
                 throw new InvalidEventTypeException("\"metadata\" property is reserved");
             }
 
+            List<String> absentFields = eventType.getPartitionKeyFields().stream()
```
rcillo: Is the update flow also safe? Or just creation flow?

rcillo:
@v-stepanov regarding your comment, this should no longer work for data change events, for example, since the partition key lookup is done scoped by "data" field.

So I think it's fine to release this feature.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
v1ctor: We can't change schema or partition keys in update. So it's working only for update.

src/test/java/de/zalando/aruha/nakadi/controller/EventTypeControllerTest.java
v1ctor: You right, I've copied it form other test.

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
rcillo: Are you sure it's not possible to change partition keys in updates?

v1ctor: You are right. There is a possibility to change partition keys. I will fix it.

v-stepanov:
@rcillo but this still can be a problem for business events
and I think all eventlog events are business events
@v1ctor for me it would be more comfortable if we have a feature toggle for this.

rcillo:
@v-stepanov I see. So maybe this validation should be done based on the effective schema?

v1ctor:
@v-stepanov I've added feature toggle for that

rcillo:
:shipit:

PR #328, opened 2016-07-11
ARUHA-162: When posting single event instead of array - the error message is not informative

adyach:
Added detail info in response about JSON errors


src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -75,6 +78,10 @@ private ResponseEntity postEventInternal(String eventTypeName, String eventsAsSt
         }
     }
 
+    private ThrowableProblem crateProblem(JSONException e) {
+        return Problem.valueOf(Response.Status.BAD_REQUEST, e.getMessage());
```
v1ctor: Do you know what is inside e.getMessage() ? According to API Guidelines it's not really good idea to show internal error messages in API.

v1ctor: I don't know this library. That's why I'm asking you :) For example Jackson could return really specific information about mapper configuration in such error message.

adyach: You can look at it: https://github.com/stleary/JSON-java
Also it is not clear how to implement it according to ARUHA-162 because if we show 'Your json request is not array' for every JSONException it will be wrong. We can check for the array our selfs but I think it is more cumbersome

v-stepanov:
@adyach You mixed up ticket id in your commit message :)
Should be ARUHA-162

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -75,6 +78,10 @@ private ResponseEntity postEventInternal(String eventTypeName, String eventsAsSt
         }
     }
 
+    private ThrowableProblem crateProblem(JSONException e) {
```
v-stepanov: I don't remember if I told that, but we have a rule that if something can be final - it should be final

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
v-stepanov: This is JSONException, I think in this case it's ok to expose the message to user, because this is exactly what he wants to see: what is wrong with the json he provided.

adyach:
@v-stepanov Monday morning :)

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
v1ctor: For example there could be IOException https://github.com/stleary/JSON-java/blob/25a87975beb7e826b148b5f69d77337f68ae8832/JSONWriter.java#L121

v1ctor:
:shipit:

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
adyach: I can check for the 'cause' of the exception and if it is not a JSONException then return just 'bad request'. What do you think ?

v-stepanov: @v1ctor is JSONWriter used here somehow?

v1ctor: It's the same for JSONTokener https://github.com/stleary/JSON-java/blob/25a87975beb7e826b148b5f69d77337f68ae8832/JSONTokener.java#L158

v-stepanov: As we read from String, I can't imagine situation when it can happen for us.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -75,6 +81,17 @@ private ResponseEntity postEventInternal(String eventTypeName, String eventsAsSt
         }
     }
 
+    private ResponseEntity processJSONException(final JSONException e, final NativeWebRequest nativeWebRequest) {
+        if (e.getCause() == null) {
```
v-stepanov: I'm not sure that it is correct. Why are we interested only with exceptions without cause?

adyach: Because in some case it will have information which is not relevant to the user of the request and won't be informative for the user at all. I think here we want to show the user the error which is related to the json formatting and not internal errors inside json library.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -75,6 +81,17 @@ private ResponseEntity postEventInternal(String eventTypeName, String eventsAsSt
...
+            return create(crateProblem(e), nativeWebRequest);
+        }
+        return create(Problem.valueOf(Response.Status.BAD_REQUEST), nativeWebRequest);
+    }
+
+    private ThrowableProblem crateProblem(final JSONException e) {
```
v-stepanov: why you decided to have this method separately ?

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
```diff
@@ -75,6 +81,17 @@ private ResponseEntity postEventInternal(String eventTypeName, String eventsAsSt
         }
     }
 
+    private ResponseEntity processJSONException(final JSONException e, final NativeWebRequest nativeWebRequest) {
+        if (e.getCause() == null) {
+            return create(crateProblem(e), nativeWebRequest);
```
rcillo: There is a typo in this method `crateProblem`.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
adyach: For me it looks cleaner because 'create' method is not overloaded with additional text.

src/main/java/de/zalando/aruha/nakadi/controller/EventPublishingController.java
rcillo: What do you think of having more control over the error message? Instead of using the one provided by the json library we could write our own "Publishing events expect a JSON array".

What scenarios is this `if` clause handling? I mean, is this `if` solving only the `Json array` issue or is it exposing a miriade of cryptic error messages from the json library? I'm afraid this could raise even more questions by users.

adyach: In general, JSONException contains messages about json parsing errors (we've already discussed here in this pr) but there is case when it can contain some errors about IO.

rcillo: @adyach do you think this is relevant to our discussion http://zalando.github.io/restful-api-guidelines/common-data-objects/CommonDataObjects.html#must-an-error-message-must-not-contain-the-stack-trace ?

adyach: I do not put stacktrace to the details field. I put only message from JSONException which is json parsing errors in general (https://github.com/stleary/JSON-java).

Also it is not clear how to implement it according to ARUHA-162 because if we show 'Your json request is not array' for every JSONException it will be wrong. We can check for the array our selfs but I think it is more cumbersome.

What do you think ?

rcillo: I agree that checking ourselves doesn't seem to be a good idea. Maybe we can release this way and see the reaction from users 👍

rcillo:
:shipit:

v-stepanov:
:shipit:

PR #329, opened 2016-07-11
ARUHA-182: return 500 status code if doesn't exist topic in kafka for…

v1ctor:
… corresponding eventType

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/controller/PartitionsController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/PartitionsController.java
@@ -15,6 +15,7 @@ import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.context.request.NativeWebRequest;
 import org.zalando.problem.Problem;
 
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
 import static javax.ws.rs.core.Response.Status.NOT_FOUND;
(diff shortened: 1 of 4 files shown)


v-stepanov:
:shipit:

src/main/java/de/zalando/aruha/nakadi/controller/PartitionsController.java
```diff
@@ -39,7 +40,11 @@ public PartitionsController(final EventTypeRepository eventTypeRepository, final
         try {
             final EventType eventType = eventTypeRepository.findByName(eventTypeName);
 
-            return ok().body(topicRepository.listPartitions(eventType.getTopic()));
+            if (!topicRepository.topicExists(eventType.getTopic())) {
+                return create(Problem.valueOf(INTERNAL_SERVER_ERROR, "topic is absent in kafka"), request);
```
rcillo: Sometimes I think it's valuable to log 500 errors. It could also be a good idea to monitor then afterwards.

What do you think?

rcillo: @v1ctor ?

v1ctor: @rcillo Sorry, I've missed your comment. I will add logging here in another pull request.

PR #330, opened 2016-07-11
Aruha 266 null handling strategy

v-stepanov:
Improved null handling
@antban @v1ctor @adyach @rcillo please review

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamParameters.java
+++ b/src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamParameters.java
@@ -1,5 +1,6 @@
 package de.zalando.aruha.nakadi.service.subscription;
 
+import javax.annotation.Nullable;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.IntStream;
@@ -35,9 +36,9 @@ public class StreamParameters {
     public final long commitTimeoutMillis;
 
     private StreamParameters(
-            final int batchLimitEvents, final Long streamLimitEvents, final long batchTimeoutMillis,
-            final Long streamTimeoutMillis, final Integer batchKeepAliveIterations, final int windowSizeMessages,
-            final long commitTimeoutMillis) {
+            final int batchLimitEvents, @Nullable final Long streamLimitEvents, final long batchTimeoutMillis,
+            @Nullable final Long streamTimeoutMillis, @Nullable final Integer batchKeepAliveIterations,
+            final int windowSizeMessages, final long commitTimeoutMillis) {
         this.batchLimitEvents = batchLimitEvents;
         this.streamLimitEvents = Optional.ofNullable(streamLimitEvents);
         this.batchTimeoutMillis = batchTimeoutMillis;
--- a/src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
+++ b/src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
@@ -160,7 +160,7 @@ public class StreamingContext implements SubscriptionStreamer {
             clientListChanges.refresh();
             zkClient.runLocked(() -> {
                 final Partition[] changeset = rebalancer.apply(zkClient.listSessions(), zkClient.listPartitions());
-                if (changeset != null && changeset.length > 0) {
+                if (changeset.length > 0) {
                     Stream.of(changeset).forEach(zkClient::updatePartitionConfiguration);
                     zkClient.incrementTopology();
                 }
--- a/src/main/java/de/zalando/aruha/nakadi/service/subscription/state/ClosingState.java
+++ b/src/main/java/de/zalando/aruha/nakadi/service/subscription/state/ClosingState.java
@@ -50,6 +50,9 @@ class ClosingState extends State {
     }
 
     private void onTopologyChanged() {
+        if (topologyListener == null) {
+            throw new IllegalStateException("topologyListener should not be null when calling onTopologyChanged method");
+        }
         topologyListener.refresh();
         reactOnTopologyChange();
     }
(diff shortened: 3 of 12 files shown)


src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamParameters.java
```diff
@@ -35,9 +36,9 @@
     public final long commitTimeoutMillis;
 
     private StreamParameters(
```
adyach: Isn't it too many parameters for the constructor?

antban: Yes, it is, we need to refactor it

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
```diff
@@ -160,7 +160,7 @@ private void rebalance() {
             clientListChanges.refresh();
             zkClient.runLocked(() -> {
                 final Partition[] changeset = rebalancer.apply(zkClient.listSessions(), zkClient.listPartitions());
-                if (changeset != null && changeset.length > 0) {
```
adyach: We have apache commons in our dependencies why not to use it? ArrayUtils.isNotEmpty()

antban:
:+1:

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/ClosingState.java
```diff
@@ -50,6 +50,9 @@ public void onEnter() {
     }
 
     private void onTopologyChanged() {
+        if (topologyListener == null) {
```
adyach: I think it is good to check parameters for nulls and even better if you can do something with these nulls (recreate, go other way etc). In this particular situation if we do not have this check we will have NPE latter on. 
What are the reasons to have it ?

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
```diff
@@ -134,19 +137,20 @@ public void fillEmptySubscription(final Map<Partition.PartitionKey, Long> partit
...
                 state.name()).collect(Collectors.joining(":")).getBytes(CHARSET);
     }
 
     private static Partition deserializeNode(final Partition.PartitionKey key, final byte[] data) {
         final String[] parts = new String(data, CHARSET).split(":");
         return new Partition(
                 key,
-                parts[0].isEmpty() ? null : parts[0],
-                parts[1].isEmpty() ? null : parts[1],
+                NO_SESSION.equals(parts[0]) ? null : parts[0],
```
adyach: in case parts[0] is null you will return null. is it okay ?

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamingContext.java
v-stepanov: I changed apply method so that it will never return null
https://github.com/zalando/nakadi/pull/330/files#diff-9797c80ff4effa509ddb92a7eda3c526R68
So there is no need to check additionally for null

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/ClosingState.java
```diff
@@ -50,6 +50,9 @@ public void onEnter() {
...
+            throw new IllegalStateException("topologyListener should not be null when calling onTopologyChanged method");
```
rcillo: Under which circumstances would `topologyListener` be null?

rcillo: I mean, isn't it possible to "fail early". If there is any problem when creating the listener, the error should be thrown in that moment and not afterwards when trying to use it. Makes sense?

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/ClosingState.java
v-stepanov: NullPointerException is not really informative, I think IllegalStateException is a perfect description of what happens here, the ClosingState object is in incorrect state where this method is called.

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/ClosingState.java
v-stepanov: The usecase for `topologyListener` to be null can be that `onTopologyChanged()` is executed after `onExit()` was called for this state.
Also, I don't see any way how it can fail during listener creation.

src/main/java/de/zalando/aruha/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
v-stepanov: @adyach I saw your question in email, but for some reason can't find it here. 
`parts[0]` and 'parts[1]' will never be null, as we don't write nulls there, only real values or empty string. So we can expect the same when we read data.

src/main/java/de/zalando/aruha/nakadi/service/subscription/state/ClosingState.java
rcillo: Ok, I see. That might happen because there is an scheduled task which could be executed at any time, inclusive after `onExit` which assigns `null` to `topologyListener`.

Failing during listener creation would not cover all cases, so this solution is probably more appropriate.

src/main/java/de/zalando/aruha/nakadi/service/subscription/StreamParameters.java
v-stepanov: I agree, but the intention of this PR is only to align nulls handling, so I will not do it here.

v-stepanov:
If nobody objects I will merge this PR (I already got an approve from @antban)

PR #331, opened 2016-07-11
ARUHA-283: switched on codacy back

v-stepanov:
jpm4j works again, I switched on codacy back
@adyach @v1ctor @antban @rcillo please review


adyach:
:shipit:

PR #332, opened 2016-07-12
ARUHA-225: Include the partition the event was written to in metadata

adyach:
Added partition parameter for metadata + fixed tests
@v-stepanov @v1ctor @antban @rcillo

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategy.java
+++ b/src/main/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategy.java
@@ -37,7 +37,7 @@ public class MetadataEnrichmentStrategy implements EnrichmentStrategy {
         metadata.put("received_at", dateTime.toString());
     }
 
-    public void setPartition(final JSONObject partiotion, final BatchItem batchItem) {
-        partiotion.put("partition", batchItem.getPartition());
+    public void setPartition(final JSONObject partition, final BatchItem batchItem) {
+        partition.put("partition", batchItem.getPartition());
     }
 }
--- a/src/main/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategy.java
+++ b/src/main/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategy.java
@@ -18,7 +18,7 @@ public class MetadataEnrichmentStrategy implements EnrichmentStrategy {
             setReceivedAt(metadata);
             setEventTypeName(metadata, eventType);
             setFlowId(metadata);
-            setPartiotion(metadata, batchItem);
+            setPartition(metadata, batchItem);
         } catch (final JSONException e) {
             throw new EnrichmentException("enrichment error", e);
         }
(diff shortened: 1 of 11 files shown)


src/main/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategy.java
```diff
@@ -10,13 +11,14 @@
...
-    public void enrich(final JSONObject event, final EventType eventType) throws EnrichmentException {
+    public void enrich(final BatchItem batchItem, final EventType eventType) throws EnrichmentException {
         try {
-            final JSONObject metadata = event.getJSONObject("metadata");
+            final JSONObject metadata = batchItem.getEvent().getJSONObject("metadata");
 
             setReceivedAt(metadata);
             setEventTypeName(metadata, eventType);
             setFlowId(metadata);
+            setPartiotion(metadata, batchItem);
```
rcillo: Typo: `s/setPartiotion/setPartition/`.

src/main/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategy.java
```diff
@@ -34,4 +36,8 @@ private void setReceivedAt(final JSONObject metadata) {
         final DateTime dateTime = new DateTime(DateTimeZone.UTC);
         metadata.put("received_at", dateTime.toString());
     }
+
+    public void setPartition(final JSONObject partiotion, final BatchItem batchItem) {
```
rcillo: `partiotion` type here as well.

rcillo:
@adyach I miss one extra assertion in the `MetadataEnrichmentStrategyTest` to make sure that partition is set correctly.

src/main/java/de/zalando/aruha/nakadi/service/EventPublisher.java
```diff
@@ -50,38 +50,37 @@ public EventPublisher(final TopicRepository topicRepository,
...
             InternalNakadiException {
         final EventType eventType = eventTypeCache.getEventType(eventTypeName);
-        final List<BatchItem> batch = initBatch(events);
+        final List<BatchItem> batch = BatchFactory.from(events);
 
         try {
             validate(batch, eventType);
-            enrich(batch, eventType);
             partition(batch, eventType);
+            enrich(batch, eventType);
```
v-stepanov: @rcillo Ricardo I know you are deeper into that than me. This change of order, can it break anything for us? Because I'm not sure.

src/main/java/de/zalando/aruha/nakadi/enrichment/MetadataEnrichmentStrategy.java
```diff
@@ -34,4 +36,8 @@ private void setReceivedAt(final JSONObject metadata) {
         final DateTime dateTime = new DateTime(DateTimeZone.UTC);
         metadata.put("received_at", dateTime.toString());
     }
+
+    public void setPartition(final JSONObject partition, final BatchItem batchItem) {
```
v-stepanov: I think first param should be called `metadata` as it is metadata

src/main/java/de/zalando/aruha/nakadi/service/EventPublisher.java
rcillo: @v-stepanov well observed. I think we should change the API as well https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#L1115 since some users might expect this order to be well defined. There used to be no dependency between these two steps but now it does.

rcillo:
:shipit:

PR #334, opened 2016-07-12
ARUHA-152: validate permissions to delete and update event-type

v1ctor:


src/main/java/de/zalando/aruha/nakadi/security/ClientIdResolver.java
```diff
@@ -0,0 +1,42 @@
...
+        return parameter.getParameterType().isAssignableFrom(String.class) && parameter.hasParameterAnnotation(Client.class);
+    }
+
+    @Override
+    public Object resolveArgument(MethodParameter parameter, ModelAndViewContainer mavContainer,
+                                  NativeWebRequest request, WebDataBinderFactory binderFactory) throws Exception
+    {
+        Optional<String> principal = Optional.ofNullable(request.getUserPrincipal()).map(Principal::getName);
+        if (settings.getAuthMode() == SecuritySettings.AuthMode.OFF) {
+            return principal.orElseGet(() -> Optional.ofNullable(request.getParameter(CLIENT_ID))
```
v-stepanov: I don't really understand this part. Why we take this parameter from a request?

src/main/scala/de/zalando/aruha/nakadi/OperationsScala.scala
```diff
@@ -1,6 +0,0 @@
-package de.zalando.aruha.nakadi
```
v-stepanov: please don't delete our door to scala world

src/main/java/de/zalando/aruha/nakadi/security/ClientIdResolver.java
v1ctor: When it's OFF you could specify any client_id you want for testing or it would be default if you don't specify anything.

src/main/scala/de/zalando/aruha/nakadi/OperationsScala.scala
v1ctor: It's dead code. I'd prefer to delete it. Our door to Scala world is always open :)

src/main/java/de/zalando/aruha/nakadi/managers/Result.java
```diff
@@ -0,0 +1,82 @@
+package de.zalando.aruha.nakadi.managers;
+
+import org.zalando.problem.Problem;
+
+import javax.ws.rs.core.Response;
+
+public interface  Result<T> {
```
v-stepanov: I dislike such approach: to have a universal return object.

v1ctor: Could you please describe in more details? It could either have result either problem which occurred.

v-stepanov:
I think we need a feature-toggle for this feature. I think a lot of users currently violate these restrictions.

v1ctor:
Ok. You propose to switch in on at staging?

src/main/java/de/zalando/aruha/nakadi/managers/EventTypeManager.java
```diff
@@ -0,0 +1,193 @@
...
+        } catch (final NakadiException e) {
+            LOG.error("Error deleting event type " + eventTypeName, e);
+            return Result.problem(e.asProblem());
+        }
+    }
+
+    public Result<Void> update(final String eventTypeName, final EventType eventType, final String clientId) {
+        try {
+            EventType original = eventTypeRepository.findByName(eventTypeName);
+            if (!original.getOwningApplication().equals(clientId)) {
```
v-stepanov: how it will work for auth mode OFF?

src/main/java/de/zalando/aruha/nakadi/security/ClientIdResolver.java
v-stepanov: I think the problem is that you want to return just a String. 
I think you should return some Client object that will have some behaviour.
So for the auth mode OFF you can return an object that is allowed to modify everything. 
Also I believe we will want to have possibility to modify/delete everything as Nakadi operators, so there can be special id that also have the god privileges.

v-stepanov:
@v1ctor no, it's not about staging or some specific stack. I just think we will have to spend some time  communicating this to our users. So I think we will first deploy with this feature switched off and then we will switch it on when all users are informed and performed actions to be ready for this new restrictions.

src/main/java/de/zalando/aruha/nakadi/security/ClientResolver.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Override
+    public Client resolveArgument(MethodParameter parameter, ModelAndViewContainer mavContainer,
+                                  NativeWebRequest request, WebDataBinderFactory binderFactory) throws Exception
+    {
+        if (!featureToggleService.isFeatureEnabled(CHECK_APPLICATION_LEVEL_PERMISSIONS)) {
+            return Client.PERMIT_ALL;
+        }
+
+        Optional<String> client_id = Optional.ofNullable(request.getUserPrincipal()).map(Principal::getName);
```
v-stepanov: 1) final
2) no snake case please

v1ctor: 1. Sure
2. Yes. Checkstyle have already told me that!

src/main/java/de/zalando/aruha/nakadi/security/ClientResolver.java
```diff
@@ -0,0 +1,52 @@
...
+
+        final Optional<String> client_id = Optional.ofNullable(request.getUserPrincipal()).map(Principal::getName);
+        if (client_id.filter(settings.getAdminClientId()::equals).isPresent()) {
+            return Client.PERMIT_ALL;
+        }
+        final Optional<Client> principal = client_id.map(Client.Authorized::new);
+        if (settings.getAuthMode() == SecuritySettings.AuthMode.OFF) {
+            return principal.orElseGet(() -> Client.PERMIT_ALL);
+        }
+        return principal.orElseThrow(() -> new UnauthorizedUserException("Client unauthorized"));
```
v-stepanov: I think it can be made simpler
like:
```java
return Optional.ofNullable(request.getUserPrincipal())
                .map(Principal::getName)
                .map(clientId ->
                        settings.getAuthMode() == OFF || settings.getAdminClientId().equals(clientId) ?
                                Client.PERMIT_ALL : new Client.Authorized(clientId))
                .orElseThrow(() -> new UnauthorizedUserException("Client unauthorized"));
```

src/main/java/de/zalando/aruha/nakadi/security/ClientResolver.java
```diff
@@ -0,0 +1,53 @@
...
+    {
+        final Optional<String> clientId = Optional.ofNullable(request.getUserPrincipal()).map(Principal::getName);
+
+        if (!featureToggleService.isFeatureEnabled(CHECK_APPLICATION_LEVEL_PERMISSIONS)
+                || clientId.filter(settings.getAdminClientId()::equals).isPresent())
+        {
+            return Client.PERMIT_ALL;
+        }
+        final Optional<Client> principal = clientId.map(Client.Authorized::new);
+        if (settings.getAuthMode() == SecuritySettings.AuthMode.OFF) {
```
v-stepanov: why don't you want to include this condition to your first if and always return PERMIT_ALL is auth mode is OFF?

src/main/java/de/zalando/aruha/nakadi/config/ManagersConfig.java
```diff
@@ -0,0 +1,24 @@
...
+import de.zalando.aruha.nakadi.repository.TopicRepository;
+import de.zalando.aruha.nakadi.util.UUIDGenerator;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class ManagersConfig {
+
+    @Bean
+    public EventTypeService eventTypeManager(final EventTypeRepository eventTypeRepository,
```
v-stepanov: as it is Service now, I think it makes sense also to rename the method and class

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -34,43 +19,33 @@
...
-                               final TopicRepository topicRepository,
-                               final PartitionResolver partitionResolver,
-                               final Enrichment enrichment,
-                               final FeatureToggleService featureToggleService,
-                               final UUIDGenerator uuidGenerator)
+    public EventTypeController(final EventTypeService service,
+                               final EventTypeRepository eventTypeRepository,
+                               final FeatureToggleService featureToggleService)
     {
+        this.service = service;
```
v-stepanov: I think service is just too common name. `eventTypeService` can be a better name.

v1ctor:
Fixed :)

src/main/scala/de/zalando/aruha/nakadi/OperationsScala.scala
v-stepanov: then please also delete commented scala in build file

src/main/java/de/zalando/aruha/nakadi/repository/EventTypeRepository.java
```diff
@@ -18,4 +19,14 @@
     List<EventType> list();
 
     void removeEventType(String name) throws InternalNakadiException, NoSuchEventTypeException;
+
+    default Optional<EventType> findByNameO(String eventTypeName) throws InternalNakadiException {
```
v-stepanov: final

v-stepanov: final

src/main/java/de/zalando/aruha/nakadi/repository/EventTypeRepository.java
```diff
@@ -18,4 +19,14 @@
...
+        try {
+            return Optional.of(findByName(eventTypeName));
+        } catch (NoSuchEventTypeException e) {
```
v-stepanov: final

src/main/java/de/zalando/aruha/nakadi/repository/EventTypeRepository.java
```diff
@@ -18,4 +19,14 @@
...
+            return Optional.empty();
+        } catch (InternalNakadiException e) {
```
v-stepanov: final

src/main/java/de/zalando/aruha/nakadi/security/Client.java
```diff
@@ -0,0 +1,23 @@
+package de.zalando.aruha.nakadi.security;
+
+public interface Client {
+
+    Client PERMIT_ALL = clientId -> true;
+
+    boolean is(final String clientId);
```
v-stepanov: i think in interface it doesn't make much sense to use final keyword

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -34,43 +19,33 @@
...
 
     @Autowired
-    public EventTypeController(final EventTypeRepository eventTypeRepository,
-                               final TopicRepository topicRepository,
-                               final PartitionResolver partitionResolver,
-                               final Enrichment enrichment,
-                               final FeatureToggleService featureToggleService,
-                               final UUIDGenerator uuidGenerator)
+    public EventTypeController(final EventTypeService eventTypeService,
+                               final EventTypeRepository eventTypeRepository,
```
v-stepanov: we have it now here only because of `eventTypeRepository.list()`
Can this be also moved to service layer?
Then the controller will have one dependency less.

src/main/java/de/zalando/aruha/nakadi/managers/EventTypeService.java
```diff
@@ -0,0 +1,195 @@
+package de.zalando.aruha.nakadi.managers;
```
v-stepanov: the package is still managers

v-stepanov:
@v1ctor do you see a possibility to write some test to functionality you implemented?

v1ctor:
Fixed.
There was some tests when I was using client_id but right now i have problems to invent how to mock OAuth authorization.
How I see there is no tests on OAuth yet.

src/main/scala/de/zalando/aruha/nakadi/OperationsScala.scala
v-stepanov: oh, it was already deleted :)

src/main/java/de/zalando/aruha/nakadi/services/EventTypeService.java
```diff
@@ -0,0 +1,200 @@
+package de.zalando.aruha.nakadi.services;
```
v-stepanov: Hmm, now we have two packages `service` and `services`

v-stepanov:
:shipit:

src/main/java/de/zalando/aruha/nakadi/repository/EventTypeRepository.java
```diff
@@ -18,4 +19,14 @@
     List<EventType> list();
 
     void removeEventType(String name) throws InternalNakadiException, NoSuchEventTypeException;
+
+    default Optional<EventType> findByNameO(final String eventTypeName) throws InternalNakadiException {
```
rcillo: This alternative has been considered previously but we decided to use throw/catch style. I see no advantage in introducing this now.

PR #337, opened 2016-07-13
aruha-253 version migration to fix default statistics

rcillo:
Related to #271


adyach:
:shipit:

v-stepanov:
👎

v-stepanov:
I think that for migration phase we need to have both default_statistics and default_statistic, or we have to make a Nakadi version that supports both

PR #340, opened 2016-07-13
Aruha 253 undo backward incompatible changes

rcillo:
During validation time the team decided to not release this change since it would be backwards incompatible and would require some downtime during rollout in order to execute database migrations.

We are going to provide an alternative implementation in order to smoothly rollout with no downtime.

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
+++ b/src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
@@ -98,7 +98,7 @@ public class EventTypeController {
             enrichment.validate(eventType);
             partitionResolver.validate(eventType);
             eventTypeRepository.saveEventType(eventType);
-            topicRepository.createTopic(eventType.getTopic(), eventType.getDefaultStatistics());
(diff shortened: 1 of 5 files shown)


adyach:
I do not know the logic behind it but the code looks fine 👍

src/main/java/de/zalando/aruha/nakadi/controller/EventTypeController.java
```diff
@@ -98,7 +98,7 @@ public EventTypeController(final EventTypeRepository eventTypeRepository,
             enrichment.validate(eventType);
             partitionResolver.validate(eventType);
             eventTypeRepository.saveEventType(eventType);
-            topicRepository.createTopic(eventType.getTopic(), eventType.getDefaultStatistics());
+            topicRepository.createTopic(eventType.getName(), eventType.getDefaultStatistic());
```
v-stepanov: I think it should be `eventType.getTopic()`

rcillo: 👍 just fixed it.

v-stepanov:
:shipit:

PR #342, opened 2016-07-14
ARUHA-277: add checkstyle for acceptance tests

v1ctor:


build.gradle
```diff
@@ -309,5 +309,5 @@ checkstyle {
 }
 
 task checkstyle {
-    dependsOn checkstyleMain, checkstyleTest
+    dependsOn checkstyleMain, checkstyleTest, checkstyleAcceptanceTest
```
v-stepanov: just curious, where `checkstyleAcceptanceTest` is defined?

v1ctor: Checkstyle gradle plugin has a task per every source set:
https://docs.gradle.org/current/userguide/checkstyle_plugin.html

adyach:
👍

build.gradle
v-stepanov: ah ok. That's cool.

v-stepanov:
:shipit:

PR #345, opened 2016-07-18
ARUHA-293: switch from configuration to autowired

v1ctor:


src/main/java/de/zalando/aruha/nakadi/controller/EventStreamController.java
```diff
@@ -59,6 +60,7 @@
     private final MetricRegistry metricRegistry;
     private final ClosedConnectionsCrutch closedConnectionsCrutch;
 
+    @Autowired
     public EventStreamController(final EventTypeRepository eventTypeRepository, final TopicRepository topicRepository,
```
rcillo: Which class is being injected in the first parameter?

v1ctor: \@Primary but for readability I could do it \@Cache

rcillo: I think it's fine. You already made it explicit when a `@DB` is required, so it's all fine 👍

src/main/java/de/zalando/aruha/nakadi/annotations/DB.java
```diff
@@ -0,0 +1,17 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.FIELD,
+        ElementType.METHOD,
+        ElementType.TYPE,
+        ElementType.PARAMETER})
+@Retention(RetentionPolicy.RUNTIME)
+@Qualifier
+public @interface DB {
```
v-stepanov: What exactly this annotation does? I think I don't get it.

v1ctor: I specify to inject particular EventTypeRepository (EventTypeDBRepository) in CachingEventTypeRepository

adyach:
For me it is fine 👍

rcillo:
:shipit:

adyach:
:shipit:

PR #348, opened 2016-07-19
[WIP] Schema versioning API

rcillo:
This is a draft for the schema versioning API. We considered how to update an schema by updating the whole event type, how to list versions and how to get an specific version.

There are some ongoing discussions on how to coordinate schema version updates from the producers perspective and how to roll out major version updates.


adyach:
Maybe we can somehow divide this file on several separate files in the future? For example by APIs.

api/nakadi-event-bus-api.yaml
```diff
@@ -995,11 +1014,42 @@ definitions:
...
         type: object
         $ref: '#/definitions/EventTypeSchema'
         description: |
-          The schema for this EventType. Submitted events will be validated against it.
+          The most recent schema for this EventType. Submitted events will be validated against it.
+
+      schema_versions:
+        type: array
+        items:
+          type: string
```
v-stepanov: what will be the exact content of this array? Just versions like ['1.0.2', '1.1.0', '2.0'] ?

api/nakadi-event-bus-api.yaml
```diff
@@ -741,6 +753,13 @@ definitions:
           If set by the producer Event will be rejected.
         format: date-time
         example: '1996-12-19T16:39:57-08:00'
+      schema_version:
+        type: string
+        description: |
+          Version of the schema used for validating this event. This is enriched upon reception.
+          Clients consuming events should desserialize events only when differences between their
+          schema and the one used to validate (produce) the event differs on minor level, i.e. from
+          '1.1.0' to '1.2.0'.
```
v-stepanov: I think I don't get the last sentence. I thought that if `compatibility_mode` is forward - then consumer can always read the event even if it has major changes.

api/nakadi-event-bus-api.yaml
```diff
@@ -995,11 +1014,42 @@ definitions:
         type: string
         default: 'random'
 
+      compatibility_mode:
```
v-stepanov: will Nakadi check that compatibility mode is respected when schema is updated? Or is it something that should be guaranteed by event owner?

api/nakadi-event-bus-api.yaml
```diff
@@ -1027,6 +1077,11 @@ definitions:
 
   EventTypeSchema:
     properties:
+      version:
+        type: string
+        description: |
+          Automatically generated value based on semantic versioning. Should not be modified by clients.
```
v-stepanov: if version always generated by Nakadi?

api/nakadi-event-bus-api.yaml
rcillo: It's still open for discussion. I would argue that the way you mentioned would be enough. I don't feel like introducing HATEOS now.

api/nakadi-event-bus-api.yaml
rcillo: If it's forwards, it will never have major changes, I guess, since major means backward incompatibility in semver.

api/nakadi-event-bus-api.yaml
rcillo: Nakadi will check it and ensure that no major changes are accepted if the compatibility mode is forwards.

api/nakadi-event-bus-api.yaml
rcillo: Correct. Nakadi always generate a version for users. The specific rules are yet being developed but it will follow semver.

v-stepanov: OK, I see. So when schema is updated Nakadi checks the changes and generates new version depending on the changes.

v-stepanov:
@rcillo so far so good. I like the direction it goes to 👍

api/nakadi-event-bus-api.yaml
adyach: I just wonder is it possible to encode compatibility modes in schema version?

rcillo: @adyach I cannot see a way of doing that and still follow http://semver.org/

api/nakadi-event-bus-api.yaml
rcillo: Changed for links and added a new endpoint for listing schemas.

api/nakadi-event-bus-api.yaml
```diff
@@ -546,6 +542,40 @@ paths:
...
+  '/event-types/{name}/schemas':
+    get:
+      tags:
+        - schema-registry-api
+      description: |
+        List of schemas ordered from most recent to oldest.
+    responses:
+        '200':
+          schema:
+            type: object
```
v-stepanov: shouldn't it be an array of objects?

api/nakadi-event-bus-api.yaml
```diff
@@ -1027,6 +1105,11 @@ definitions:
 
   EventTypeSchema:
     properties:
+      version:
+        type: string
```
v-stepanov: I think you can also use `format` property here like:
`format: semver`

api/nakadi-event-bus-api.yaml
rcillo: I did it this way to allow for extending with pagination.

api/nakadi-event-bus-api.yaml
rcillo: 👍 done

rcillo:
In my previous commit I removed compatibility modes and proposed a simpler way to solve this. Wdyt? https://github.com/zalando/nakadi/pull/348/commits/f7d376dfd4ff891a8d8d6f76ddfdf25b32170d35

rcillo:
The base branch no longer exists.

PR #349, opened 2016-07-19
Feature/aruha 190 subscription api

v-stepanov:
@rcillo @jkliff @v1ctor @dehora @antban @adyach @valgog 
complete HILA API
There already were PRs for some parts of it, and this is resulting PR


api/nakadi-event-bus-api.yaml
```diff
@@ -78,15 +75,21 @@ info:
 
     Notable omissions here are:
 
-    * The contract between Nakadi and clients in the context of the high-level API (i.e. the process for
-    clients to establish subscriptions).
+    For the subscription API:
 
-    * Enrichment procedure.
+    * Support for controlling, in the subscription API, specific delivery guarantees (at most once, at least once).
```
rcillo: Removing "in the subscription API" doesn't change the meaning of this item and make it shorter.

api/nakadi-event-bus-api.yaml
```diff
@@ -78,15 +75,21 @@ info:
...
+    * negotiation of sessions' parameters
```
rcillo: In the previous item we started with uppercase and used a "." to finish it. We should follow the same pattern in the following items as well for consistence.

rcillo: the apostrophe is in the wrong place. It should be "session's".

api/nakadi-event-bus-api.yaml
```diff
@@ -78,15 +75,21 @@ info:
...
+    * auto commit mode (implying at most once)
```
rcillo: "at most once" what?

api/nakadi-event-bus-api.yaml
```diff
@@ -78,15 +75,21 @@ info:
...
+    * support for controlled replaying of subscriptions (for example the initial point of streaming)
```
rcillo: Is it possible?

api/nakadi-event-bus-api.yaml
```diff
@@ -78,15 +75,21 @@ info:
...
+    * moment (creation of subscription or first stream) and offset (newest available or latest available) of initialization of cursors of subscription
+    *
 
-    * Security scopes (OAuth) for the different operations
+    relevant for managemnet of nakadi:
```
rcillo: management.

api/nakadi-event-bus-api.yaml
```diff
@@ -78,15 +75,21 @@ info:
...
-    clients to establish subscriptions).
+    For the subscription API:
 
-    * Enrichment procedure.
+    * Support for controlling, in the subscription API, specific delivery guarantees (at most once, at least once).
+    * negotiation of sessions' parameters
+    * auto commit mode (implying at most once)
+    * support for controlled replaying of subscriptions (for example the initial point of streaming)
+    * moment (creation of subscription or first stream) and offset (newest available or latest available) of initialization of cursors of subscription
+    *
```
rcillo: ✂️ *

api/nakadi-event-bus-api.yaml
rcillo: I would expect uppercase "Relevant".

api/nakadi-event-bus-api.yaml
```diff
@@ -78,15 +75,21 @@ info:
...
 
-    * Security scopes (OAuth) for the different operations
+    relevant for managemnet of nakadi:
+    - low watermark tracking of each strams' window per session
```
rcillo: type "strams".

api/nakadi-event-bus-api.yaml
```diff
@@ -78,15 +75,21 @@ info:
...
+    - identificaton of each session (client id)
 
-    * Explicit control of (in case of Kafka as underlying messaging broker) topic's creation
-    parameters (number of partitions, retention times, etc), as well as their modification.
+    Other topics:
+    - multi event type streaming
```
rcillo: Would be good to state that it's not yet supported.

api/nakadi-event-bus-api.yaml
```diff
@@ -743,6 +637,350 @@ paths:
...
+            type: array
+            items:
+              $ref: '#/definitions/Cursor'
+          required: true
+      responses:
+        '200':
+          description: Offsets were committed
+        '204':
+          description: |
+            No Content. At least one cursor which was tried to be committed is older or equal to already committed one.
```
rcillo: It's not clear if it stops trying to commit when it happens or if it continues to the end, committing whenever it's legal. Also, it's not possible to tell which one has been committed or not.

v-stepanov: this is all true. But I don't see a way to specify which were committed and which were already committed before. I don't really want to pass an array of results

api/nakadi-event-bus-api.yaml
```diff
@@ -743,6 +621,350 @@ paths:
...
+          schema:
+            $ref: '#/definitions/Problem'
+    get:
+      tags:
+        - managed-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: Lists all subscriptions that exist in a system.
+      responses:
+        '200':
```
antban: I think that we need to take into account that the count of subscriptions may be much bigger that amount of event types, it means that we will need to add pagination at some point. 
Maybe it's better to add wrapper object for subscriptions now, in order not to do it later as a breaking change?
I mean add object
```
SubscriptionList:
    items:
        type: array
        items:
            $ref: '#/definitions/Subscription'
```
Then we will be able to extend this object to add pagination and so on.

api/nakadi-event-bus-api.yaml
```diff
@@ -1083,6 +1305,67 @@ definitions:
...
+          - end
+        description: |
+          Position to start reading events from. "begin" - read from the oldest available event. "end" - read from the
+          most recent offset. Applied in the moment when client starts reading from a subscription.
+        default: 'end'
+
+    required:
+      - owning_application
+      - event_types
+
```
antban: I don't know if we need to manage api specification in exactly the same manner that it was, but I don't like idea of `if` conditions in description of Subscription.
For me it's better to split suscription type into to different types:
```
SubscriptionToCreate:
    properties:
        owning_application: ...
        event_types: ...
        consumer_group: ...
        read_from: ...

Subscription:
    allOf:
        - $ref: SubscriptionToCreate
    properties:
        id: ...
        created_at:
```

api/nakadi-event-bus-api.yaml
rcillo: @antban good point. Done in be23492

api/nakadi-event-bus-api.yaml
rcillo: I would agree if there were 3 variants of this object. Then your suggestion would make it easy to extend, for example, creating a `SubscriptionToUpdate` and so on. But since there are only 2 possible "states", I think the "ifs" doesn't harm.

api/nakadi-event-bus-api.yaml
```diff
@@ -54,15 +54,10 @@ info:
...
     tracking of the current position on a Stream.
 
-    **Note** *Currently the high level API is out of scope in this specification.  It is in the
-    short term plan to be included.*
-
 
     Scope and status of the API
     ---------------------------------
 
-    The API specification is in **draft** state and is subject to change.
```
rcillo: I decided to remove this comment since it makes no sense to have software in production for months but state that it's api is still in draft. I think people can rely on our api for been stable and evolve in a compatible manner.

rcillo:
I'm closing this PR in order to create a new one for review by the api guild.

PR #350, opened 2016-07-19
ARUHA-299: postgres database update script does not work for ubuntu user

adyach:
One user has a problem with running docker-compose.  The database update script runs sql scripts in reverse order for some reason.


v-stepanov:
:shipit:

rcillo:
:shipit:

PR #351, opened 2016-07-20
ARUHA-141 Support definition of retention time

adyach:


src/main/java/de/zalando/aruha/nakadi/domain/EventTypeOptions.java
```diff
@@ -0,0 +1,17 @@
+package de.zalando.aruha.nakadi.domain;
+
+import javax.annotation.Nullable;
+
+public class EventTypeOptions {
+
+    @Nullable
+    private Long retentionTime;
```
v1ctor: I think we are almost adopted usage of Optional

src/main/java/de/zalando/aruha/nakadi/domain/EventType.java
```diff
@@ -49,6 +51,10 @@
     @Nullable
     private EventTypeStatistics defaultStatistic;
 
+    @Valid
+    @Nullable
+    private EventTypeOptions options;
```
v1ctor: What is options? Do we have to move another fields there?

adyach: https://techjira.zalando.net/browse/ARUHA-141

v1ctor: Ok. But isn't an answer :) I don't understand what should belong to options and what shouldn't.

adyach: I agree that is not the clearest name for such field. I think it is more about internal Kafka adjustments and other properties which are related to internals of Nakadi.

v1ctor:
👍

api/nakadi-event-bus-api.yaml
```diff
@@ -1084,6 +1090,16 @@ definitions:
       - read_parallelism
       - write_parallelism
 
+  EventTypeOptions:
+    type: object
+    description: |
+      Additional parameters for tuning internal behavior of Nakadi.
+    properties:
+      retention_time:
+        type: long
```
rcillo: Are you sure it's possible to use a type `long`? I thought that we should use `type: integer` and `format: int64` https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md

src/main/resources/application.yml
```diff
@@ -24,8 +24,11 @@ spring:
     validationQuery: SELECT 1
 nakadi:
   topic:
+    min:
+      retentionMs: 86400000 # 1 days
     max:
       partitionNum: 8
+      retentionMs: 172800000 # 2 days
```
rcillo: I thought we would be using a larger `max` value. Teams are complaining that if they have a problem during the weekend, when they are back to work on monday events are already gone. We should have at least 3 days, so they have a change to fix it on Monday.

PR #355, opened 2016-07-20
aruha-262 fix date time validation once and for all

### LINKED ISSUE #274: Prefer Java 8 Time API to Joda Time

rcillo:
Overriding default everit date-time validator by a correct
implementation of RFC3339 validation fixes #274 and another non reported
bug that allowed offset string without colon to be accepted.

### DIFF
--- a/src/main/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidator.java
+++ b/src/main/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidator.java
@@ -2,6 +2,7 @@ package de.zalando.aruha.nakadi.validation;
 
 import org.everit.json.schema.FormatValidator;
 
+import java.time.OffsetDateTime;
 import java.time.format.DateTimeParseException;
 import java.util.Optional;
 import java.util.regex.Matcher;
@@ -21,7 +22,7 @@ public class RFC3339DateTimeValidator implements FormatValidator {
     @Override
     public Optional<String> validate(final String dateTime) {
         try {
-            java.time.LocalDate.parse(dateTime, ISO_OFFSET_DATE_TIME);
+            OffsetDateTime.parse(dateTime, ISO_OFFSET_DATE_TIME);
 
             // Unfortunately, ISO_OFFSET_DATE_TIME accepts offsets with seconds, which is not RFC3339 compliant. So
             // we need to do some further checks using a regex in order to be sure that it adhere to the given format.
--- a/src/main/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidator.java
+++ b/src/main/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidator.java
@@ -1,0 +1,39 @@
+package de.zalando.aruha.nakadi.validation;
+
+import org.everit.json.schema.FormatValidator;
+
+import java.time.format.DateTimeParseException;
+import java.util.Optional;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
+
+public class RFC3339DateTimeValidator implements FormatValidator {
+
+    private final String errorMessage = "must be a valid date-time";
+
+    // Valid offsets are either Z or hh:mm. The format hh:mm:ss is not valid
+    private final String dateTimeOffsetPattern = "^.*(Z|((\\+|-)\\d\\d:\\d\\d))$";
+    private final Pattern pattern = Pattern.compile(dateTimeOffsetPattern);
+    private final Optional<String> error = Optional.of(errorMessage);
+
+    @Override
+    public Optional<String> validate(final String dateTime) {
+        try {
+            java.time.LocalDate.parse(dateTime, ISO_OFFSET_DATE_TIME);
+
+            // Unfortunately, ISO_OFFSET_DATE_TIME accepts offsets with seconds, which is not RFC3339 compliant. So
+            // we need to do some further checks using a regex in order to be sure that it adhere to the given format.
+            final Matcher matcher = pattern.matcher(dateTime);
+            if (matcher.matches()) {
+                return Optional.empty();
+            } else {
+                return error;
+            }
+
+        } catch (final DateTimeParseException e) {
+            return error;
+        }
+    }
+}
--- a/src/test/java/de/zalando/aruha/nakadi/validation/JSONSchemaValidationTest.java
+++ b/src/test/java/de/zalando/aruha/nakadi/validation/JSONSchemaValidationTest.java
@@ -10,7 +10,6 @@ import java.util.Arrays;
 import java.util.Optional;
 
 import static de.zalando.aruha.nakadi.utils.IsOptional.isAbsent;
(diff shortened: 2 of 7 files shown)


src/main/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidator.java
```diff
@@ -0,0 +1,39 @@
...
+
+    // Valid offsets are either Z or hh:mm. The format hh:mm:ss is not valid
+    private final String dateTimeOffsetPattern = "^.*(Z|((\\+|-)\\d\\d:\\d\\d))$";
+    private final Pattern pattern = Pattern.compile(dateTimeOffsetPattern);
+    private final Optional<String> error = Optional.of(errorMessage);
+
+    @Override
+    public Optional<String> validate(final String dateTime) {
+        try {
+            java.time.LocalDate.parse(dateTime, ISO_OFFSET_DATE_TIME);
```
jhorstmann: Not sure if it makes a difference, but should this be `OffsetDateTime.parse` instead, which has `ISO_OFFSET_DATE_TIME` as the default format.

src/test/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidatorTest.java
```diff
@@ -0,0 +1,42 @@
...
+                "1996-10-45t16:39:57Z", // the RFC requires uppercase T, sorry
+                "1996-10-45T16:39:57z", // the RFC requires uppercase Z as well, sorry again
+                "1996-10-45T16:39:57", // offsets are required
+        };
+
+        final String validDateTimes[] = new String[] {
+                "1996-10-15T16:39:57+07:00", // just a very simple example
+                "1996-10-15T16:39:57-07:00", // just a very simple example
+                "1996-10-15T16:39:57.123+07:00", // simple example with milliseconds
+                "1996-10-15T16:39:57.1234Z", // tricky 4 milliseconds digits (yes it's valid, sorry)
```
jhorstmann: `ISO_OFFSET_DATE_TIME` allows anywhere from 0-9 fractional digits so it might make sense to add tests for all of them.

jhorstmann:
This also fixes the issue with more than 3 digits of fractional seconds, right?

src/test/java/de/zalando/aruha/nakadi/validation/JSONSchemaValidationTest.java
```diff
@@ -164,58 +163,6 @@ public void requireMetadataOccurredAt() {
     }
 
     @Test
-    public void requireMetadataOccurredAtToBeFormattedAsDateTime() {
```
AlexanderYastrebov: Why this tests were removed?

src/main/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidator.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.regex.Pattern;
+
+import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
+
+public class RFC3339DateTimeValidator implements FormatValidator {
+
+    private final String errorMessage = "must be a valid date-time";
+
+    // Valid offsets are either Z or hh:mm. The format hh:mm:ss is not valid
+    private final String dateTimeOffsetPattern = "^.*(Z|((\\+|-)\\d\\d:\\d\\d))$";
```
AlexanderYastrebov: One should always remember this:
> Some people, when confronted with a problem, think 
> “I know, I'll use regular expressions.”   Now they have two problems.

The proper way would be to have a look how `ISO_OFFSET_DATE_TIME` is defined and derive proper format based on that:
```java
        DateTimeFormatter rfc3339 = new DateTimeFormatterBuilder()
            .parseCaseInsensitive()
            .append(DateTimeFormatter.ISO_LOCAL_DATE_TIME)
            .appendOffset("+HH:MM", "Z")
            .toFormatter(Locale.getDefault(Locale.Category.FORMAT))
            .withChronology(IsoChronology.INSTANCE)
            .withResolverStyle(ResolverStyle.STRICT);
```

src/main/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidator.java
AlexanderYastrebov: No difference from parse POV as result is ignored, but should be changed to avoid confusion

AlexanderYastrebov:
Joda Time should be removed from dependencies

AlexanderYastrebov:
Joda Time should be removed from dependencies

src/main/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidator.java
rcillo: @AlexanderYastrebov I tried your suggestion but it accepted seconds in the offset. One of the test cases "1996-10-15T16:39:57+07:00:30" should have been rejected, but it wasn't. Do you have any idea on how to fix this?

src/test/java/de/zalando/aruha/nakadi/validation/JSONSchemaValidationTest.java
rcillo: This tests were all grouped in a more appropriate layer, testing the RFC validator. All test cases were preserved and new ones were added.

src/test/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidatorTest.java
rcillo: @jhorstmann done in https://github.com/zalando/nakadi/pull/355/commits/140f61c3741a916ff3de57d13f0c5f57771fda82

src/main/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidator.java
rcillo: Done in https://github.com/zalando/nakadi/pull/355/commits/b740c11e371f4ae0bae52f49928e1701880999e7

src/main/java/de/zalando/aruha/nakadi/validation/RFC3339DateTimeValidator.java
AlexanderYastrebov: @rcillo 
Yep, unfortunately this solution does not check the format but only ignores seconds.
If your main purpose is to reject second precision offsets then you can do this:
```java
       OffsetDateTime d = OffsetDateTime.parse("1996-10-15T16:39:57+07:00:30", DateTimeFormatter.ISO_OFFSET_DATE_TIME);
        if (d.get(ChronoField.OFFSET_SECONDS) % 60 != 0) {
            throw new IllegalArgumentException("Offset with seconds not supported");
        }
```

rcillo: @AlexanderYastrebov thanks for the tip 👍 

We are mostly interested in validating that the format is RFC 3339 compliant. We need to reject strings that contain seconds in the offset.

antban:
:+1:

PR #357, opened 2016-07-25
ARUHA-305: change de.zalando.aruha to org.zalando

v1ctor:


adyach:
This change removed all static imports 👍

src/test/java/org/zalando/nakadi/service/subscription/model/PartitionTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+import static java.util.Collections.singletonList;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.mock;
+
+public class PartitionTest {
```
v-stepanov: why this file is completely green?

v-stepanov:
I don't like all these changes about static imports.
It creates a huge amount of diff noise. We will have hard times checking the Git history of some affected files.

v1ctor:
Revert this.

adyach:
:shipit:

PR #359, opened 2016-07-28
ARUHA-310: removed extra logging of ATs

v-stepanov:
@antban @v1ctor @rcillo @adyach please review

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -285,8 +285,6 @@ acceptanceTest {
     testLogging {
         events "passed", "skipped", "failed"
     }
-    testLogging.showStandardStreams = true
-    testLogging.exceptionFormat = 'full'
 }
 
 task development << {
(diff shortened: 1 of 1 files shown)


v1ctor:
:shipit:

build.gradle
```diff
@@ -285,8 +285,6 @@ acceptanceTest {
     testLogging {
         events "passed", "skipped", "failed"
     }
-    testLogging.showStandardStreams = true
-    testLogging.exceptionFormat = 'full'
```
adyach: @v-stepanov maybe we can at least leave 'testLogging.exceptionFormat = 'full'? I would like to see the reason of the failing test when I run tests from console.

v-stepanov: You're right. I returned it back.

adyach:
👍

adyach:
👍

PR #360, opened 2016-07-28
ARUHA-164: fix logic for field validation and test refactoring

v1ctor:


v1ctor:
I've reused method hasReservedField but it was working only for business event types. 
There is a bug that if you pass not business event type it is always false.

adyach:
👍

src/test/java/org/zalando/nakadi/utils/EventTypeTestBuilder.java
```diff
@@ -0,0 +1,113 @@
...
+import org.zalando.nakadi.domain.EventType;
+import org.zalando.nakadi.domain.EventTypeOptions;
+import org.zalando.nakadi.domain.EventTypeSchema;
+import org.zalando.nakadi.domain.EventTypeStatistics;
+import org.zalando.nakadi.domain.ValidationStrategyConfiguration;
+import org.zalando.nakadi.partitioning.PartitionStrategy;
+
+import java.util.List;
+
+public class EventTypeTestBuilder {
```
v-stepanov: +1 for builder

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -204,7 +204,9 @@ private void validateSchema(final EventType eventType) throws InvalidEventTypeEx
         try {
             final JSONObject schemaAsJson = new JSONObject(eventType.getSchema().getSchema());
 
-            if (hasReservedField(eventType, schemaAsJson, "metadata")) {
+            if (eventType.getCategory() == EventCategory.BUSINESS
```
v-stepanov: Why you now check it only for BUSINESS event type? Is that because for DATA event schema checks only "data" field?

rcillo: As some food for thought, I would say that Business event is more confusing than Data event because it mixes user data and Nakadi data under the same "level". For example, a business event has "metadata" and all other properties in the same Json properties. But Data event "envelops" all user data inside "data" field. It seems to be a good idea to make it mandatory for all Events to have a "common" envelope. It would make everything more homogeneous. Event "undefined" should be enveloped, in my opinion, so that we could use all partition strategies with it.

PR #362, opened 2016-07-29
ARUHA-224 Support per event type definition of scopes for reading and writing

adyach:


src/main/java/org/zalando/nakadi/security/ScopeHelper.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class ScopeHelper {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ScopeHelper.class);
+
+    public static void checkScopes(final Optional<Set<String>> scope, final Client client) {
+        scope.ifPresent(scopes -> {
+            LOG.debug("scopes={}", scopes);
+            if (client.hasNoScopes(scopes)) {
+                throw new IllegalScopeException();
```
v-stepanov: I think it would be cool to inform user which scope(s) he misses

src/main/java/org/zalando/nakadi/security/ScopeHelper.java
```diff
@@ -0,0 +1,22 @@
...
+import org.zalando.nakadi.exceptions.IllegalScopeException;
+
+import java.util.Optional;
+import java.util.Set;
+
+public class ScopeHelper {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ScopeHelper.class);
+
+    public static void checkScopes(final Optional<Set<String>> scope, final Client client) {
```
v-stepanov: scope**s** ?

src/main/java/org/zalando/nakadi/security/ScopeHelper.java
```diff
@@ -0,0 +1,22 @@
...
+        scope.ifPresent(scopes -> {
+            LOG.debug("scopes={}", scopes);
```
v-stepanov: I think you should provide more context here if you want to log that.

src/main/java/org/zalando/nakadi/security/ScopeHelper.java
v-stepanov: Also, I think that is strange that the only result this method can bring is throwing an exception and this exception is unchecked.

src/main/java/org/zalando/nakadi/security/AuthorizedClient.java
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    @Override
+    public boolean is(final String clientId) {
+        return this.clientId.equals(clientId);
+    }
+
+    @Override
+    public boolean hasNoScopes(final Set<String> scopes) {
+        return this.scopes.stream().noneMatch(scope -> scopes.contains(scope));
```
v-stepanov: lambda expression can be replaced with method reference

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -74,6 +80,8 @@ public EventType(final String name, final String topic, final String owningAppli
         this.schema = schema;
         this.defaultStatistic = defaultStatistic;
         this.options = options;
+        this.writeScope = writeScope;
+        this.readScope = readScope;
```
v-stepanov: why the names are singular?
Those are sets, I think the names should be plural.

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -55,13 +57,17 @@
     @Nullable
     private EventTypeOptions options;
 
+    private Optional<Set<String>> writeScope;
+
+    private Optional<Set<String>> readScope;
```
v-stepanov: I don't really see much benefit of using Optional here.
I think empty set is good enough to describe "empty" case.

src/main/java/org/zalando/nakadi/security/ScopeHelper.java
adyach: Yes, I absolutely agree. I will change it.

src/main/java/org/zalando/nakadi/security/ScopeHelper.java
adyach: scope because it is optional (one) and also I think it is more readable on the next line. i've added unchecked exception to add for processing in ExceptionHandling.java

src/main/java/org/zalando/nakadi/security/Client.java
```diff
@@ -1,23 +1,23 @@
...
-            return this.clientId.equals(clientId);
+        public boolean hasNoScopes(final Set<String> scopes) {
+            return false;
         }
-    }
+    };
+
+    boolean is(final String clientId);
+
+    boolean hasNoScopes(final Set<String> scopes);
```
v-stepanov: I think it's better to have method like `hasAnyOfScopes`

adyach: Yes, I agree, but in that case in ScopeHelper I will use '!' and that makes code less readable.

src/main/java/org/zalando/nakadi/domain/EventType.java
adyach: They are optional, I think they'd better be singular because when you do operation on optional you get the lambda and there you have to use plural because you have a set.

src/main/java/org/zalando/nakadi/domain/EventType.java
adyach: In the further code I use ifPresent and it is useful there.

src/main/java/org/zalando/nakadi/domain/EventType.java
v-stepanov: I think I don't understand that. If the name was `readScopeSet` - that would be more or less fine. Read scope identifies that it's a single one.

src/main/java/org/zalando/nakadi/domain/EventType.java
v-stepanov: what stops you from using `readScope.isEmpty()` ?

src/main/java/org/zalando/nakadi/security/ClientResolver.java
```diff
@@ -35,20 +40,26 @@ public boolean supportsParameter(final MethodParameter parameter) {
...
+                                  final ModelAndViewContainer mavContainer,
+                                  final NativeWebRequest request,
+                                  final WebDataBinderFactory binderFactory) throws Exception {
         final Optional<String> clientId = Optional.ofNullable(request.getUserPrincipal()).map(Principal::getName);
-
         if (!featureToggleService.isFeatureEnabled(CHECK_APPLICATION_LEVEL_PERMISSIONS)
                 || clientId.filter(settings.getAdminClientId()::equals).isPresent()
-                || settings.getAuthMode() == OFF)
-        {
+                || settings.getAuthMode() == OFF) {
```
v-stepanov: Let the bracket battle begin :-D

src/main/java/org/zalando/nakadi/domain/EventType.java
adyach: nothing, just less readable. i will change it to not continue this talk

src/main/java/org/zalando/nakadi/security/ScopeHelper.java
```diff
@@ -0,0 +1,34 @@
...
+import org.zalando.nakadi.exceptions.IllegalScopeException;
+
+import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class ScopeHelper {
+
+    public static void checkScopes(final Set<String> allowedScopes, final Set<String> clientScopes) {
+        if (isNotEmptyScope(allowedScopes)){
```
v1ctor: allowedScopes.stream().filter(clientScopes::contains).collect(Collectors.toSet()).isEmpty();

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -55,14 +57,23 @@
...
+        this.readScopes = Collections.emptySet();
+    }
 
     public EventType(final String name, final String topic, final String owningApplication, final EventCategory category,
                      final List<ValidationStrategyConfiguration> validationStrategies,
                      final List<EnrichmentStrategyDescriptor> enrichmentStrategies, final String partitionStrategy,
                      final List<String> partitionKeyFields, final EventTypeSchema schema, final EventTypeStatistics defaultStatistic,
-                     final EventTypeOptions options)
-    {
+                     final EventTypeOptions options, final Set<String> writeScopes, final Set<String> readScopes) {
```
rcillo: This constructor has a lot of parameters.

adyach: I agree @rcillo but it is used only in one place for test purposes because of that fact I have not touched it.

src/main/java/org/zalando/nakadi/exceptions/IllegalScopeException.java
```diff
@@ -0,0 +1,18 @@
+package org.zalando.nakadi.exceptions;
+
+import java.util.Set;
+
+public class IllegalScopeException extends RuntimeException {
```
rcillo: @v1ctor @adyach @v-stepanov @antban I thought that the team had chosen to use checked exceptions https://docs.google.com/spreadsheets/d/1p2GTg-fgrNGQBkhdM1yzKwvWOXvFL4FyiOoGtQJqTxI/edit#gid=0

But I personally liked this approach in this case, unchecked exceptions fits nicely 👍

adyach: @rcillo I made it just to feel the taste of that and for me it looks very nice. I am not sure that this point in the google doc is about it :)

src/main/java/org/zalando/nakadi/security/ScopeHelper.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.zalando.nakadi.exceptions.IllegalScopeException;
+
+import java.util.Objects;
+import java.util.Set;
+
+public class ScopeHelper {
+
+    public static void checkScopes(final Set<String> allowedScopes, final Set<String> clientScopes) {
+        if (isEmptyScope(allowedScopes) ) {
```
v-stepanov: extra space 😃

src/main/java/org/zalando/nakadi/security/AuthorizedClient.java
```diff
@@ -0,0 +1,28 @@
...
+        this.scopes = scopes;
+    }
+
+    @Override
+    public boolean is(final String clientId) {
+        return this.clientId.equals(clientId);
+    }
+
+    @Override
+    public Set<String> getScopes() {
```
rcillo: I understood that the original idea of having this class hierarch (Client/Authorized) was to use polymorphism (method `is`) and delegate some logic to this object.

This property accessor `getScopes` contains no logic. I would expect something like `containsScopes` as another "check" method that returns boolean. It would feel more homogeneous approach. 

It would also remove the need for a static class like the `ScopeHelper`.

What do you think?

src/main/java/org/zalando/nakadi/security/ScopeHelper.java
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        allowedScopes.stream()
+                .filter(clientScopes::contains)
+                .findAny()
+                .orElseThrow(() -> new IllegalScopeException(allowedScopes));
+    }
+
+    private static boolean isEmptyScope(final Set<String> allowedScopes) {
+        return Objects.isNull(allowedScopes) || allowedScopes.isEmpty();
```
v-stepanov: 1) Why you check for null in such a strange way? :)
2) Why allowedScopes can be null? Our null handling strategy supposes that everything is none-null by default, and if something can be null is should be marked as `@Nullable`. But I really don't see a reason for allowedScopes to be null, I think in EventType class you should actually set it to empty set if the null is passed in setter.

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -151,4 +164,20 @@ public EventTypeOptions getOptions() {
     public void setOptions(final EventTypeOptions options) {
         this.options = options;
     }
+
+    public Set<String> getWriteScopes() {
+        return writeScopes;
```
v-stepanov: I think as a step to immutability it can be wrapped with `Collections.unmodifiableSet(...)`

src/main/java/org/zalando/nakadi/security/ScopeHelper.java
adyach: 1. I see nothing strange here, I think it is up to developer how to check it. Someone loves Joda style, others not and same here for this check :)

src/test/java/org/zalando/nakadi/controller/EventScopeTest.java
```diff
@@ -0,0 +1,119 @@
...
+    @Test
+    public void testNoScopeWrite() throws Exception {
+        mvcBuilder.setCustomArgumentResolvers(new TestHandlerMethodArgumentResolver(Collections.emptySet()));
+        Mockito.when(eventTypeCache.getEventType(EVENT_NAME)).thenReturn(createEventWith(SCOPE_WRITE));
+
+        try {
+            publishEvent(EVENT_BATCH);
+        } catch (Exception e) {
+            Assert.assertTrue(e.getCause() instanceof IllegalScopeException);
+        }
```
rcillo: If it doesn't throw anything this will just pass. Maybe you should check for that case as well.

src/test/java/org/zalando/nakadi/controller/EventScopeTest.java
```diff
@@ -0,0 +1,119 @@
...
+    private static final String EVENT_NAME = "my-event";
+    private static final String EVENT_BATCH = "[{\"payload\": \"My Event Payload\"}]";
+    private static final Set<String> SCOPE_WRITE = Collections.singleton("oauth2.scope.write");
+    private static final String URL = "/event-types/" + EVENT_NAME + "/events";
+
+    private final ObjectMapper objectMapper = new JsonConfig().jacksonObjectMapper();
+    private final EventTypeCache eventTypeCache;
+    private final StandaloneMockMvcBuilder mvcBuilder;
+
+    public EventScopeTest() throws Exception {
```
rcillo: I see that this test class is under "controller" package. But the controller under test here is the `EventPublishingController`. I see that the intention is to test only authorisation related aspect of the code but then we have a controller test that is not bound to a given controller. I would rather add this to the already existing class.

And overall I would try to reduce the usage of mocks if possible. It's hard to understand and ties the code too much to implementation.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -98,6 +101,8 @@ public StreamingResponseBody streamEvents(
                 final EventType eventType = eventTypeRepository.findByName(eventTypeName);
                 final String topic = eventType.getTopic();
 
+                client.authorize(eventType.getReadScopes());
```
rcillo: Now I realize yet another problem. This authorisation logic is specific to streaming and now it's coupled with the client. Maybe my suggestion wasn't the best. But anyway, I think it's good as is. When new authorisation rules are added, we can always refactor. Let's be pragmatic.

src/test/java/org/zalando/nakadi/controller/EventScopeTest.java
adyach: The reason why I moved it to another class is to have not mocked EventPublisher.

src/main/java/org/zalando/nakadi/security/Client.java
```diff
@@ -1,22 +1,29 @@
 package org.zalando.nakadi.security;
 
-public interface Client {
+import org.zalando.nakadi.exceptions.IllegalScopeException;
 
-    Client PERMIT_ALL = clientId -> true;
+import java.util.Set;
 
-    boolean is(String clientId);
+public class Client implements IClient {
```
v1ctor: I don't like the idea to have I prefix for interface.

src/main/java/org/zalando/nakadi/security/ScopeHelper.java
```diff
@@ -0,0 +1,25 @@
...
+
+import org.zalando.nakadi.exceptions.IllegalScopeException;
+
+import java.util.Objects;
+import java.util.Set;
+
+public class ScopeHelper {
+
+    public static void checkScopes(final Set<String> allowedScopes, final Set<String> clientScopes) {
+        if (areScopesEmpty(allowedScopes)) {
```
v1ctor: We've decided that they're can't be null, that's why you could just check allowedScopes.isEmpt()

adyach: It is unused class, I removed it. This logic moved to Client class.

src/main/java/org/zalando/nakadi/exceptions/IllegalScopeException.java
v1ctor: As for me I don't like idea with exception at all. 
I think it's better to return Result<T> and check if it's successful or not.
But I think it's another topic for our guidelines.

src/test/java/org/zalando/nakadi/controller/EventScopeTest.java
rcillo: Why do you think we should not mock EventPublisher here?

rcillo: I mean, the new code was added the publisher class. Maybe we should just add 1 mock test there to make sure the helper is invoked and another one in the controller to simulate a throwing condition.

src/main/java/org/zalando/nakadi/security/Client.java
v1ctor: Let's decide https://docs.google.com/spreadsheets/d/1p2GTg-fgrNGQBkhdM1yzKwvWOXvFL4FyiOoGtQJqTxI/edit#gid=0
BTW why have you renamed it?

adyach: I see everybody against it :) I know it is a bad practice to use such names for the interfaces but I thought it would straight forward here. Renaming happened because I think name AuthorizedClient is not clear here. This class is used to check clientId, in particular that client who is requesting resource is the client who is owning this resources, that implies that AuthorizedClient does authorization and the client is not authorized until we do AuthorizedClient.is(clientId). I could not find something better than IClient.

v1ctor: AuthenticatedClient ? :)

adyach: But he is not authenticated until you call 'is' method and till that moment it is just some client.

v1ctor: He is authenticated because we have checked and validated OAuth token. He is just not authorized yet do do some things.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -155,22 +160,20 @@ public StreamingResponseBody streamEvents(
...
                 LOG.error("Error while trying to stream events. Respond with SERVICE_UNAVAILABLE.", e);
                 writeProblemResponse(response, outputStream, e.asProblem());
-            }
-            catch (final InvalidCursorException e) {
+            } catch (final InvalidCursorException e) {
                 writeProblemResponse(response, outputStream, PRECONDITION_FAILED, e.getMessage());
-            }
-            catch (final Exception e) {
+            } catch (final IllegalScopeException e) {
+                LOG.error("Error while trying to stream events. Respond with FORBIDDEN.", e);
```
v-stepanov: I think it's not an error actually and even not a warning. I think we can log it with DEBUG level (if log at all)

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -55,14 +58,23 @@
...
+    private Set<String> writeScopes;
+
+    private Set<String> readScopes;
+
+    public EventType() {
+        this.validationStrategies = Lists.newArrayList();
+        this.enrichmentStrategies = Lists.newArrayList();
+        this.partitionStrategy = PartitionStrategy.RANDOM_STRATEGY;
+        this.writeScopes = Collections.emptySet();
+        this.readScopes = Collections.emptySet();
```
v-stepanov: For me it looks very strange that for initialising collections we use two different approaches within 5-line method :)
(Guava and java.util.Collections) 
I would prefer to have homogenous approach (at least within the same method) doesn't really matter which one.

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -151,4 +165,20 @@ public EventTypeOptions getOptions() {
...
+    public void setWriteScopes(final Set<String> writeScopes) {
+        this.writeScopes = Objects.isNull(writeScopes) ? Collections.emptySet() : writeScopes;
+    }
+
+    public Set<String> getReadScopes() {
+        return Collections.unmodifiableSet(readScopes);
+    }
+
+    public void setReadScopes(final Set<String> readScopes) {
+        this.readScopes = Objects.isNull(readScopes) ? Collections.emptySet() : readScopes;
```
v-stepanov: But why you use this isNull method? Why it's better than `readScopes == null` ?

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -155,22 +160,19 @@ public StreamingResponseBody streamEvents(
...
+            } catch (final NakadiException e) {
                 LOG.error("Error while trying to stream events. Respond with SERVICE_UNAVAILABLE.", e);
                 writeProblemResponse(response, outputStream, e.asProblem());
-            }
-            catch (final InvalidCursorException e) {
+            } catch (final InvalidCursorException e) {
                 writeProblemResponse(response, outputStream, PRECONDITION_FAILED, e.getMessage());
-            }
-            catch (final Exception e) {
+            } catch (final IllegalScopeException e) {
```
v-stepanov: BTW, why you catch this exception? As I see you added the handler for it to ExceptionHandling class. So it should be automatically converted to response, right?

v-stepanov:
:shipit:

v1ctor:
:shipit:

v1ctor:
:shipit:

PR #353, opened 2016-07-20
Feature/aruha 190 subscription api

rcillo:
Clients streaming through a subscription do not need to handle offsets persistence or partitions assignment. The workflow is very simple:

1. First client creates a subscription
2. This client starts streaming from this subscription (which automatically assigns all partitions to it)
3. A new client is launched and tries to create the same subscription
4. Since subscription creation is idempotent, it gets a reference to the existing subscription
5. The new client starts streaming from this subscription, which causes partitions to be reassigned (rebalanced) between clients.
6. Repeat 3-5 until you have as many clients as partitions.

Subscriptions also provide a mechanism for implementing backpressure using a window size strategy.

PR #363, opened 2016-07-29
Aruha 311 list subscriptions

v-stepanov:
@adyach @v1ctor @rcillo @antban please review


src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
```diff
@@ -1,24 +1,26 @@
 package org.zalando.nakadi.repository.db;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableSet;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.junit.Before;
+import org.junit.Test;
```
antban: Finally! You did it :)

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -75,16 +78,40 @@ public SubscriptionController(final SubscriptionDbRepository subscriptionReposit
...
+
+        if (!featureToggleService.isFeatureEnabled(HIGH_LEVEL_API)) {
+            return new ResponseEntity<>(NOT_IMPLEMENTED);
+        }
+
+        try {
+            final List<Subscription> subscriptions = owningApplication == null ?
+                    subscriptionRepository.listSubscriptions() :
+                    subscriptionRepository.listSubscriptionsForOwningApplication(owningApplication);
+            return status(OK).body(subscriptions);
```
antban: As far as I remember there was a change regarding subscription list, - we decided to use wrapper object for that...
And I do not remember owning application filter in api...

v-stepanov: Regarding wrapper object - true. Will fix it.
Regarding owning_application - it was requested during API review. I also think it makes sense.
https://github.com/zalando/nakadi/pull/353/files#diff-4492c6b26ee0a37518728ec1585420efR579

v-stepanov: @antban fixed, added list wrapper

src/main/java/org/zalando/nakadi/domain/SubscriptionListWrapper.java
```diff
@@ -0,0 +1,19 @@
+package org.zalando.nakadi.domain;
+
+import javax.annotation.concurrent.Immutable;
+import java.util.Collections;
+import java.util.List;
+
+@Immutable
+public class SubscriptionListWrapper {
+
+    final List<Subscription> data;
```
adyach: private ?

v-stepanov: fixed https://github.com/zalando/nakadi/pull/363/commits/35f8d80db4e0e82ad9a2c4c8b5a88052be8e685d

adyach:
👍

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -56,21 +59,45 @@ public Subscription createSubscription(final SubscriptionBase subscriptionBase)
...
         } catch (final EmptyResultDataAccessException e) {
             throw new NoSuchSubscriptionException("Subscription with id \"" + id + "\" does not exist", e);
+        } catch (final DataAccessException e) {
+            throw new ServiceUnavailableException("Error occurred when running database request", e);
+        }
+    }
+
+    public List<Subscription> listSubscriptions() throws ServiceUnavailableException {
+        try {
+            return jdbcTemplate.query("SELECT s_subscription_object FROM zn_data.subscription", rowMapper);
```
rcillo: What should be the expected order? Is there any concerns about this in the API definition?

src/main/java/org/zalando/nakadi/domain/SubscriptionListWrapper.java
```diff
@@ -0,0 +1,19 @@
...
+@Immutable
+public class SubscriptionListWrapper {
+
+    private final List<Subscription> data;
+
+    public SubscriptionListWrapper(final List<Subscription> data) {
+        this.data = data;
+    }
+
+    public List<Subscription> getData() {
```
rcillo: I think we should agree on either "data" or "items". I prefer "items" since data doesn't evoke for a list.

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -84,6 +111,8 @@ public Subscription getSubscription(final String owningApplication, final Set<St
             throw new InternalNakadiException("Serialization problem during getting event type", e);
         } catch (final EmptyResultDataAccessException e) {
             throw new NoSuchSubscriptionException("Subscription does not exist", e);
+        } catch (final DataAccessException e) {
+            throw new ServiceUnavailableException("Error occurred when running database request", e);
```
rcillo: Should we expose it users?

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: I don't know. Maybe by creation time?

src/main/java/org/zalando/nakadi/domain/SubscriptionListWrapper.java
v-stepanov: I don't have any preferences here. `items` is good for me. Will change to `items`.

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: Hmm, probably you are right. We should log the DataAccessException to our LOG and not give it to user.

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
rcillo: Yes, maybe we could enforce an ordering that we think is most convenient. Creation time descending?

v-stepanov: added sorting by created_at desc

src/main/java/org/zalando/nakadi/domain/SubscriptionListWrapper.java
v-stepanov: changed to items

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: fixed

src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
```diff
@@ -95,34 +99,47 @@ public void whenGetSubscriptionByKeyPropertiesThenOk() throws InternalNakadiExce
...
+    public void whenListSubscriptionsThenOk() throws ServiceUnavailableException {
 
-    private Subscription createSubscription() {
-        final Subscription subscription = new Subscription();
-        subscription.setId(randomUUID().toString());
-        subscription.setOwningApplication("myapp");
-        subscription.setEventTypes(ImmutableSet.of("my-et"));
-        subscription.setCreatedAt(new DateTime(DateTimeZone.UTC));
-        return subscription;
+        final List<Subscription> testSubscriptions = ImmutableList.of(
```
adyach: Wouldn't it be simpler to use sort method on collection?

src/main/java/org/zalando/nakadi/domain/SubscriptionListWrapper.java
```diff
@@ -0,0 +1,19 @@
+package org.zalando.nakadi.domain;
+
+import javax.annotation.concurrent.Immutable;
+import java.util.Collections;
+import java.util.List;
+
+@Immutable
+public class SubscriptionListWrapper {
```
rcillo: Maybe this pagination wrapper could be more generic upfront. We might use such wrapper for other entities.

rcillo: And what do you think about adding pagination. I know this is not in the original scope for this PR so I would be ok with adding it next.

rcillo:
:shipit:

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
```diff
@@ -41,10 +43,14 @@
...
         eventType = createEventType();
-        subscription = createSubscription(ImmutableSet.of(eventType.getName()));
+        final SubscriptionBase subscription = randomSubscription()
+                .withEventType(eventType.getName())
+                .withStartFrom(BEGIN)
+                .buildSubscriptionBase();
+        this.subscription = createSubscription(subscription);
     }
 
-    @Test(timeout = 30000)
```
adyach: I understood why you removed this timeout, but I can not understand why other tests have it :)
Method waitFor has default timeout of 10 seconds but some tests waits shorter time.
I think can remove @Test(timeout = *) for these tests.

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
```diff
@@ -41,10 +43,14 @@
...
+    @Test()
```
adyach: No brackets here

src/main/java/org/zalando/nakadi/config/SecurityConfiguration.java
```diff
@@ -60,6 +60,7 @@ public void configure(final HttpSecurity http) throws Exception {
                     .antMatchers(POST, "/event-types/**").access(hasScope(eventTypeWriteScope))
                     .antMatchers(PUT, "/event-types/**").access(hasScope(eventTypeWriteScope))
                     .antMatchers(POST, "/subscriptions/**").access(hasScope(eventStreamReadScope))
+                    .antMatchers(GET, "/subscriptions/**").access(hasScope(eventStreamReadScope))
```
adyach: Up there you have
.antMatchers(GET, "/subscriptions/*/events/**").access(hasScope(eventStreamReadScope))

I am not sure but I think this line covers this case
.antMatchers(GET, "/subscriptions/**").access(hasScope(eventStreamReadScope))

What do you think ?

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -56,21 +63,51 @@ public Subscription createSubscription(final SubscriptionBase subscriptionBase)
...
+                    "s_subscription_object->>'created_at' DESC", rowMapper);
+        } catch (final DataAccessException e) {
+            LOG.error("Database error when listing subscriptions", e);
+            throw new ServiceUnavailableException("Error occurred when running database request");
+        }
+    }
+
+    public List<Subscription> listSubscriptionsForOwningApplication(final String owningApplication)
+            throws ServiceUnavailableException {
+        final String query = "SELECT s_subscription_object FROM zn_data.subscription " +
```
adyach: We have queries like:
final String sql = "SELECT s_subscription_object FROM zn_data.subscription WHERE s_id = ?";
and like:
final String query = "SELECT s_subscription_object FROM zn_data.subscription " +
 +                "WHERE s_subscription_object->>'owning_application' = ? " +
 +                "ORDER BY s_subscription_object->>'created_at' DESC";

I mean that in one query we use  s_subscription_object->>'owning_application' = ?  but in others we use just s_id = ? instead of s_subscription_object->>'s_id' = ?.
I do not know the difference in that.

Maybe it is just about the style?

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -56,21 +63,51 @@ public Subscription createSubscription(final SubscriptionBase subscriptionBase)
...
+                "WHERE s_subscription_object->>'owning_application' = ? " +
+                "ORDER BY s_subscription_object->>'created_at' DESC";
```
rcillo: Maybe we should add an index to this field in the database.

src/main/java/org/zalando/nakadi/domain/SubscriptionListWrapper.java
v-stepanov: I would not like to touch pagination in this PR. 
I think pagination is a separate point that we should implement as a separate feature.

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
v-stepanov: `I understood why you removed this timeout` I don't :)
I think I did it accidentally. The purpose of this timeout is to limit test run time. The test can stuck not only in the place if `waitFor` but also for example when `commitCursors` method is called (and there we don't have a timeout)
So I would prefer return 30 sec timeout back.

src/main/java/org/zalando/nakadi/config/SecurityConfiguration.java
v-stepanov: Yes, we actually have some other things here which can be replaced by more general mask. But I would prefer not to do it because currently each line corresponds to an endpoint and I think it makes things more clear and easy.

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: No, it's not about the style. For id we have a separate column in DB, so we can write `WHERE s_id = ?`, all other fields we have to extract from `s_subscription_object`. That's the only reason.

src/main/java/org/zalando/nakadi/config/SecurityConfiguration.java
adyach: I think we either make then more generic or make them explicit. I mean we can use
/subscriptions/* instead of /subscriptions/** and the same for the others.

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
v-stepanov: good point 👍  Added index: https://github.com/zalando/nakadi/pull/363/commits/b40fb97f15dd1987e6fa0820ad03aab7d893d2e3

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
adyach: I thought that RestAssured has default timeout :) and I think it is better to set it

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/config/SecurityConfiguration.java
v-stepanov: `/subscriptions/* instead of /subscriptions/**`
I think there was a problem with that
1) if you make it '/subscriptions/*' - it will match only URLs with exactly one path segment after /subscriptions/
2) If you make it '/subscriptions' - it will mot match the URL ending with slash
3) If you make it '/subscriptions/' - it will not allow URL without slash at the end
:(
But I understand what you don't like here. But let's separate this discussion/refactoring as a separate task.

adyach:
👍

v-stepanov:
@adyach @rcillo guys, can you please approve again. I just merged changes from master.

adyach:
👍

v-stepanov:
@adyach @rcillo guys can you please approve one more time? :)
I merged a travis file bugfix and fixed some checkstyle issues.

adyach:
:shipit:

PR #365, opened 2016-08-01
ARUHA-312 set default retention time when it was not provided

adyach:


v1ctor:
:shipit:

rcillo:
@adyach I was expecting a slightly different approach. I thought that we would provide default option values and store it in the "event type". E.g. create a default option when initialising an event type and make it "non null". This would also allow users to check what's the retention time for that event type.

Another task would be to populate this field correctly based on the already existing topics in Kafka.

adyach:
@rcillo I understand your point but I am not totally agree with that because it looks like we change the user provided object and user knows nothing about this change.

v-stepanov:
@adyach @rcillo I also was thinking of approach Ricardo suggests, that we should store that in EventType. Because the default value we can change in the future and we will not have any information in event-types about their retention time if we don't store yet.
And I also agree that there should be a script for filling already existing event-types.
So for now I would disapprove

v-stepanov:
👎

v-stepanov:
@adyach One more question. Why you decided to have this EventTypeOptions object? Do you plan to migrate some other existing properties there?

adyach:
I added it for flexibility, it is not that big but we can add some other properties in the future.

adyach:
@v-stepanov @rcillo maybe you though about different task? I intended to fix the issue when user can provide null value for retention_time but you are talking about setting a default value for every event type which does not have a EventTypeOptions.

If we are talking about your case what to do if the user didn't provide the EventTypeOptions but we set to current default value (which for now is default) after that we change it using the script for all event types and we will still the old default value for the event for which we set it. 
Do I miss something here?

v-stepanov:
@adyach I think we are talking about the same things. What I see as a problem:

In your current implementation if the user doesn't define retention time for his event-type - the default value from settings will be used for topic creation. Problems I see here:
- When the user will do GET for his event-type he will not see what retention time is used.
- Currently we have a default RT 4 days. It will be used by default when creating new event-types. Then let's imagine at some moment we decide to change default RT to 7 days. And the new RTs will be created by default with this RT. So, for the event-types that don't have RT defined we will not have any idea what RT is used in fact (4 or 7). The only way would be to go to kafka and check.

I understand your point that you wanted just to do a quick fix here. But I think it's a good moment and place to do it in a way it should be.

api/nakadi-event-bus-api.yaml
```diff
@@ -1097,8 +1097,9 @@ definitions:
       retention_time:
         type: integer
         format: int64
+        default: 172800000
```
rcillo: Maybe add a comment at the side to make it easier to read? `4 days` right?

src/main/java/org/zalando/nakadi/domain/EventTypeOptions.java
```diff
@@ -1,10 +1,7 @@
 package org.zalando.nakadi.domain;
 
-import javax.annotation.Nullable;
-
 public class EventTypeOptions {
 
-    @Nullable
     private Long retentionTime;
```
rcillo: Why not just add a `@NotNull` annotation and assign a default value in the constructor?

It would simplify the `getTopicRetentionMs` method. Since we would be sure that there is always a value there, we could remove all logic and just return the value. It also removes the need of a `settings` to hold default values.

database/migration/aruha-312-set-default-retention-time-when-it-was-not-provided.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE zn_data.event_type
+SET et_event_type_object=jsonb_set(et_event_type_object, '{options}', '{"retention_time": 172800000}')
+WHERE et_event_type_object->>'options' = '{"retention_time": null}' OR et_event_type_object->'options' is null;
```
rcillo: Nice 👍 I think no one in the team has used postgres json manipulation properly before. Learned something.

src/main/java/org/zalando/nakadi/domain/EventTypeOptions.java
adyach: We use default values from application.yml to be able to change this values without recompiling (using env vars). Maybe I did not get your point.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -110,9 +108,8 @@ public void createTopic(final EventType eventType) throws TopicCreationException
     }
 
     private long getTopicRetentionMs(final EventType eventType) {
-        return Optional.ofNullable(eventType.getOptions())
-                .map(EventTypeOptions::getRetentionTime)
-                .orElse(settings.getDefaultTopicRetentionMs());
+        final Long retentionTime = eventType.getOptions().getRetentionTime();
+        return retentionTime == null ? settings.getDefaultTopicRetentionMs() : retentionTime.longValue();
```
rcillo: Will it ever be null?

src/main/java/org/zalando/nakadi/domain/EventTypeOptions.java
rcillo: Like this https://github.com/zalando/nakadi/blob/master/src/main/java/org/zalando/nakadi/domain/EventType.java#L41

v-stepanov:
@adyach For migration we will also need a script that will go to kafka for each event-type to check it's retention time and set it to DB object.
But probably it should not be part of this PR, but can be done in our private repo.

src/main/java/org/zalando/nakadi/domain/EventTypeOptions.java
adyach: In that case it will be hard-code but I think it is better to have possibility to change it.

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -694,6 +717,11 @@ private ResultActions putEventType(final String content, final String name, fina
         return mockMvc.perform(requestBuilder);
     }
 
+    private ResultActions getEventType() throws Exception {
```
v-stepanov: it's not really getEventType it's more like getEventType**s**

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
adyach: I considered the case when we want to use KafkaTopicRepository.create method and pass EventType without retention time.

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -626,7 +628,28 @@ public void whenOptionsRetentionTimeExist() throws Exception {
...
-    public void whenOptionsRetentionTimeBiggerThanMax() throws Exception {
+    public void whenGetOptionsRetentionTimeExist() throws Exception {
+        final EventType defaultEventType = buildDefaultEventType();
+        final EventTypeOptions eventTypeOptions = new EventTypeOptions();
+        eventTypeOptions.setRetentionTime(150L);
+        defaultEventType.setOptions(eventTypeOptions);
+
+        Mockito.doReturn(Collections.singletonList(defaultEventType)).when(eventTypeRepository).list();
+
+        getEventType()
```
v-stepanov: why you get here a list of of event types? I think it would be more straightforward if you get a single event type, then it will be possible to use something like
`.andExpect(jsonPath("$.options.retention_time", equalTo(150)))` instead of StringContains matching

src/main/java/org/zalando/nakadi/validation/EventTypeOptionsValidator.java
```diff
@@ -28,13 +30,16 @@ public void validate(final Object target, final Errors errors) {
...
     private void checkRetentionTime(final Errors errors, final EventTypeOptions options) {
-        if (Objects.nonNull(options) && Objects.nonNull(options.getRetentionTime())) {
-            final Long retentionTime = options.getRetentionTime();
-            if (retentionTime > maxTopicRetentionMs) {
-                createError(errors, "can not be more than " + maxTopicRetentionMs);
-            } else if (retentionTime < minTopicRetentionMs) {
-                createError(errors, "can not be less than " + minTopicRetentionMs);
-            }
+        final Long retentionTime = options.getRetentionTime();
+        if (retentionTime == null) {
```
v-stepanov: If it can be null please mark the getter `@Nullable`

src/main/java/org/zalando/nakadi/validation/EventTypeOptionsValidator.java
```diff
@@ -28,13 +30,16 @@ public void validate(final Object target, final Errors errors) {
...
+            options.setRetentionTime(defaultTopicRetentionMs);
```
v-stepanov: For me it looks very strange that validator updates value. I think Validator should only validate and nothing more.

adyach: I agree that it looks a bit strange but I think it is better place to set the default value otherwise we have to do it somewhere in controller or service where we will have to inject KafkaRepositorySettings to get the default value for retention time which is not very suitable as well. Also I tried approach with json deserialization which didn't work our for me. What is your point?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: Then I would say that it's callers responsibility to properly use the `create` method. It would better throw an exception than doing something by default. Allowing for silently misuse something could be harmful.

src/main/java/org/zalando/nakadi/domain/EventTypeOptions.java
```diff
@@ -4,13 +4,13 @@
...
 
     public Long getRetentionTime() {
         return retentionTime;
     }
 
+    @Nullable
```
v-stepanov: I think you put it to the wrong place. You should have Nullable on the getter and on parameter of setter.

src/main/java/org/zalando/nakadi/validation/EventTypeOptionsValidator.java
```diff
@@ -28,11 +27,11 @@ public void validate(final Object target, final Errors errors) {
     }
 
     private void checkRetentionTime(final Errors errors, final EventTypeOptions options) {
-        if (Objects.nonNull(options) && Objects.nonNull(options.getRetentionTime())) {
-            final Long retentionTime = options.getRetentionTime();
-            if (retentionTime > maxTopicRetentionMs) {
+        final Long retentionTime = options.getRetentionTime();
+        if (retentionTime != null) {
+            if (retentionTime.longValue() > maxTopicRetentionMs) {
```
v-stepanov: Why you have this `longValue()`? Were there any problems with automatic unboxing?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -105,15 +112,9 @@ public KafkaTopicRepository(final ZooKeeperHolder zkFactory, final KafkaFactory
...
-                getTopicRetentionMs(eventType),
-                settings.getDefaultTopicRotationMs());
-    }
-
-    private long getTopicRetentionMs(final EventType eventType) {
-        return Optional.ofNullable(eventType.getOptions())
-                .map(EventTypeOptions::getRetentionTime)
-                .orElse(settings.getDefaultTopicRetentionMs());
+                nakadiSettings.getDefaultTopicReplicaFactor(),
+                eventType.getOptions().getRetentionTime(),
```
v-stepanov: You should check retentionTime for null here as it is nullable. I understand that you expect it not to be null when this method is called, but as we declare it nullable - we should check it for null. If it is null - I think we should throw exception and Nakadi should return 500 (because something really went wrong if we have it null here)

src/main/java/org/zalando/nakadi/validation/EventTypeOptionsValidator.java
adyach: 1. It is a little bit faster in some cases when unboxing.
2. maxTopicRetentionMs is long and for me it is more clearer to compare primitive types rather than reference and primitive.
Maybe it is just my old habit from previous projects with the older java versions :)

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
adyach: But it can not be null here we set to default value in the controller (the whole feature is about it). If somebody is going to call this method separately then it is the responsibility of the caller. Also I think we will have some kind of exception when we provide 'null' to kafka config (I have not checked it though). Also we talked to @rcillo about it.

v-stepanov: It should not be the responsibility of caller. He sets null to nullable property, he does everything correct. He may not be aware that createTopic requires it to be not null.
Also, the null will not come to kafka, there will be NullPointerException right on this line when it will try to unbox Long to long.

src/main/java/org/zalando/nakadi/validation/EventTypeOptionsValidator.java
v-stepanov: But for me it seems strange that you don't use autounboxing but use autoboxing, here for example: https://github.com/zalando/nakadi/pull/365/files#diff-ae87fdc01827028b148be3085e54d23eR85
I'm not aware of any performance issues with auto- boxing/unboxing I think it's only about the coding style.
I think we use auto boxing/unboxing elsewhere in the project, so I don't really see much reasons to use manual unboxing in this single place.

v-stepanov:
:shipit:

PR #366, opened 2016-08-01
ARUHA-123: auth plugin API

v1ctor:
https://github.com/zalando-incubator/nakadi-plugin-api


rcillo:
Could you provide some rationale on why having this as a plugin is required?

v1ctor:
We've decided to use this approach to not push internal realisation of owning_application checking to github.

src/main/java/org/zalando/nakadi/config/NakadiConfig.java
```diff
@@ -67,4 +73,24 @@ public MetricRegistry metricRegistry() {
...
     }
 
+    @Bean
+    public SystemProperties systemProperties(final ApplicationContext context) {
+        return name -> context.getEnvironment().getProperty(name);
+    }
+
+    @Bean
+    @SuppressWarnings("unchecked")
+    public ApplicationService factory(@Value("${nakadi.auth.plugin.factory}") final String factoryName,
```
adyach: we create ApplicationService from ApplicationServiceFactory here, maybe we can change it to smth like
` public ApplicationService applicationService`

v1ctor: You're right.

adyach:
For me it looks fine and it is quite enough for starting, but I would like to have more options for method "exists" or even change return type for something generic in the future. Also possibility to implement more than one extension looks reasonable for me.

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -96,7 +97,7 @@ public void init() throws Exception {
         final EventTypeOptionsValidator eventTypeOptionsValidator = new EventTypeOptionsValidator(TOPIC_RETENTION_MIN_MS, TOPIC_RETENTION_MAX_MS);
         final EventTypeController controller = new EventTypeController(eventTypeService,
                 featureToggleService,
-                eventTypeOptionsValidator);
+                eventTypeOptionsValidator, new DefaultApplicationService());
```
rcillo: It would be nice to have a test that exercises the new check.

v1ctor: Yes you're totally right

v-stepanov:
@v1ctor For me it looks good. Have you considered dependency injection as a way to activate plugin? So that when a jar is in classpath its bean with higher priority is injected replacing the default implementation? :-D

src/main/java/org/zalando/nakadi/controller/EventTypeController.java
```diff
@@ -60,6 +66,10 @@ public EventTypeController(final EventTypeService eventTypeService,
         }
 
         ValidationUtils.invokeValidator(eventTypeOptionsValidator, eventType.getOptions(), errors);
+        if (!applicationService.exists(eventType.getOwningApplication())) {
```
v-stepanov: I think we need a future-toggle for this feature as it forbids something that was allowed in the past.

v1ctor:
@adyach @rcillo @v-stepanov @antban 
Please have a look

src/main/resources/application.yml
```diff
@@ -61,4 +61,10 @@ nakadi:
...
       eventStreamRead: nakadi.event_stream.read
-      eventStreamWrite: nakadi.event_stream.write
\ No newline at end of file
+      eventStreamWrite: nakadi.event_stream.write
+  auth:
+    plugin:
+      factory: org.zalando.nakadi.plugin.auth.DefaultApplicationServiceFactory
+      accessClientUri: https://example.com/tokeninfo
+      authTokenUri: https://example.com/tokeninfo
+      tokenId: nakadi
```
antban: Last 3 properties are used somewhere?

v1ctor: No. Only in private. I will remove them.

build.gradle
```diff
@@ -122,6 +126,7 @@ dependencies {
     compile 'org.apache.commons:commons-lang3:3.4'
     compile 'org.zalando:problem-spring-web:0.5.0'
     compile 'org.zalando:jackson-datatype-problem:0.5.0'
+    compile 'org.zalando:nakadi-plugin-api:1.0.2'
```
antban: by the way, where this library is?

antban:
:+1:

build.gradle
v1ctor: It's in the maven central already.

PR #373, opened 2016-08-08
ARUHA-316 support reading and writing scopes for subscriptions

adyach:


src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -71,7 +74,7 @@ public SubscriptionController(final SubscriptionDbRepository subscriptionReposit
         }
 
         try {
-            return createSubscription(subscriptionBase, request);
+            return createSubscription(subscriptionBase, request, client);
```
rcillo: I don't think that authorisation should be done at the service layer. I see authorisation as another level of feature. The same way that authentication is done in a AOP way, I think that decoupling authorisation for other business process is a good design principle.

I would rather have it in the controller, just to avoid mixing it with the service layer, than having it in the service.

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -115,12 +119,13 @@ public SubscriptionController(final SubscriptionDbRepository subscriptionReposit
...
 
-    private List<String> checkExistingEventTypes(final SubscriptionBase subscriptionBase)
+    private List<String> checkExistingEventTypes(final SubscriptionBase subscriptionBase, final Client client)
             throws InternalNakadiException {
         final List<String> noneExistingEventTypes = Lists.newArrayList();
         for (final String etName : subscriptionBase.getEventTypes()) {
             try {
-                eventTypeRepository.findByName(etName);
+                final EventType eventType = eventTypeRepository.findByName(etName);
+                client.checkScopes(eventType.getReadScopes());
```
rcillo: The name of the method is `checkExistingEventTypes` but there we are doing authorisation. I think it should go elsewhere. I know it's handy here because we just got the event type from the repository, but I would rather refactor it to have it in a more appropriate place.

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -143,4 +148,35 @@ private Subscription getExistingSubscription(final SubscriptionBase subscription
                 subscriptionBase.getConsumerGroup());
     }
 
+    private final class EventTypesHolder {
```
v-stepanov: I don't really see a good reason for this class to exist. Why do you need to collect missing ETs? You can always know the missing ones: just subtract found event types from the set of all event types of subscription.

src/test/java/org/zalando/nakadi/controller/SubscriptionControllerTest.java
```diff
@@ -176,8 +194,33 @@ public void whenGetNoneExistingSubscriptionThenNotFound() throws Exception {
...
 
+    @Test
+    public void whenPostSubscriptionThenWithNoReadScope() throws Exception {
+        when(eventTypeRepository.findByName("myET")).thenReturn(EventTypeTestBuilder.builder()
+                .name("myET")
+                .readScopes(Collections.singleton("oauth.read.scope"))
+                .build());
+
+        expectedException.expectCause(IsInstanceOf.instanceOf(IllegalScopeException.class));
+        expectedException.expectMessage("Client has to have scopes: [oauth.read.scope]");
```
v-stepanov: Why don't you just check which problem we have in response? Like here for example https://github.com/zalando/nakadi/pull/373/files#diff-8cb6844b0c2bfa55bfd6a5e918d7e925L176

src/test/java/org/zalando/nakadi/controller/SubscriptionControllerTest.java
```diff
@@ -176,8 +194,33 @@ public void whenGetNoneExistingSubscriptionThenNotFound() throws Exception {
...
+
+        postSubscription(createSubscription("app", ImmutableSet.of("myET")));
+    }
+
+    @Test
+    public void whenPostSubscriptionThenWithHasReadScope() throws Exception {
```
v-stepanov: I had hard times reading the name of this method

src/test/java/org/zalando/nakadi/controller/SubscriptionControllerTest.java
```diff
@@ -176,8 +194,33 @@ public void whenGetNoneExistingSubscriptionThenNotFound() throws Exception {
                 .andExpect(content().string(jsonHelper.matchesObject(expectedProblem)));
     }
 
+    @Test
+    public void whenPostSubscriptionThenWithNoReadScope() throws Exception {
```
v-stepanov: I think more suitable name will be `whenPostSubscriptionWithNoReadScopeThenForbidden`

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
adyach: I think this solution is pretty straight forward (though the class is cumbersome) and It will be more confusing to have a method which subtracting found event types from subscription event names. Also in that case we will have several loops but we have exception for that when calling findByName.

src/test/java/org/zalando/nakadi/controller/SubscriptionControllerTest.java
adyach: We do not get proper exception handling for exception which are in ExceptionHandling.java,  because we do not start the whole spring application and ExceptionHandling class is not initialized. There are ways to add this to current mvc mock but it looks clumsy for me.

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
v1ctor: As for me it's complex. I'd prefer solution without extra class. For example 
```java
Map<Boolean, List<String>> result= types.stream().collect(Collectors.partitionBy(type -> repository.exists(type)));
List<String> notExistsTypes = result.getOrDefault(false, List.singletonList());
```

adyach: @v1ctor I did not quite get your point.
What is 'repository.exists(type)' and why do we need it?
I think the simpler will be to use approach with subset:
        getExistingEventTypes().stream()
                .filter(eventType -> List<EventType> missingEventTypes = subscriptionBase.getEventTypes().contains(eventType.getName()))
                .map(EventType::getName)
                .collect(toList());

But I preferred to do it with the additional object because I think it is good to save additional cpu cycles.

v-stepanov: @adyach This class is just two lists. And one of these lists you don't really need in my opinion, because you can always get it with something like:
```java
final List<EventType> existingEventTypes = getExistingEventTypes(subscriptionBase);
final List<String> missingEventTypes = subscriptionBase.getEventTypes()
        .stream()
        .filter(et -> existingEventTypes.stream().anyMatch(existingET -> existingET.getName().equals(et)))
```

v-stepanov: @adyach Haven't seen you latest comment :)
I would really prefer something like you have in you last comment. I don't think that saving few CPU cycles will help here. Creation of subscription is relatively rare operation, and this optimisation I believe will not even save 1ms, because it's just iteration of collection with few elements.

adyach: @v-stepanov why don't I need it? actually, I need this list! and in your code you create this list. I check for existing event types on every call of this method. 
I just gather information about non existing event types in one loop.

src/test/java/org/zalando/nakadi/controller/SubscriptionControllerTest.java
v-stepanov: In spring-test 4.2 it's very easy to integrate to mockMvc:
http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/test/web/servlet/setup/StandaloneMockMvcBuilder.html#setControllerAdvice-java.lang.Object...-
You will just add setControllerAdvice(new ExceptionHandling()) when building mockMvc
I think it's a good reason to upgrade to spring-test 4.2 :)

adyach: Thanks! Super, I did not know about this new versions. I will fix it.

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
v-stepanov: Yes, of course you need the second list. What I meant is that you don't need to return two lists from the method. Because the only reason of EventTypesHolder to exist is that you want to return two lists from the method. This class is used only once only in this place and will never be used again. 
That's why I think the more clear way would be to write a method that returns only existing event-types and then you find missing event-types by comparing to lists (all - existing).

v1ctor: @adyach I thought you need two lists.
For me it better to do like that:
```json
List<String> result= types.stream().filter(type -> { 
try { 
    repository.findOne(type);
    return true;
} catch (Exception e) {
    return false;
}).collect(Collectors.toList());
```

```json
repository.exists(type)
``` 
could be just syntax sugar

src/test/java/org/zalando/nakadi/controller/SubscriptionControllerTest.java
v-stepanov: Have you tried that? Did it work?

adyach: Yes, I set controller advice and as you can see the test passes :)

src/test/java/org/zalando/nakadi/controller/SubscriptionControllerTest.java
```diff
@@ -250,11 +265,43 @@ public void whenPostSubscriptionAndExceptionThenServiceUnavailable() throws Exce
...
+                .readScopes(Collections.singleton("oauth.read.scope"))
+                .build());
+
+        final SubscriptionBase subscriptionBase = randomSubscription()
+                .withOwningApplication("app")
+                .withEventTypes(ImmutableSet.of("myET"))
+                .buildSubscriptionBase();
+        postSubscription(subscriptionBase)
+                .andExpect(status().isForbidden())
+                .andExpect(content().string(new StringContains("Client has to have scopes: [oauth.read.scope]")));
```
v-stepanov: But why you don't just check for problem here? Like here 
https://github.com/zalando/nakadi/pull/373/files#diff-8cb6844b0c2bfa55bfd6a5e918d7e925R263
It will also check that the response structure is a problem object

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -132,37 +136,42 @@ public SubscriptionController(final SubscriptionDbRepository subscriptionReposit
...
     private ResponseEntity<?> createSubscription(final SubscriptionBase subscriptionBase,
-                                                 final NativeWebRequest request)
+                                                 final NativeWebRequest request,
+                                                 final Client client)
             throws InternalNakadiException, DuplicatedSubscriptionException, ServiceUnavailableException {
-        final List<String> noneExistingEventTypes = checkExistingEventTypes(subscriptionBase);
-        if (!noneExistingEventTypes.isEmpty()) {
-            final String errorMessage = createErrorMessage(noneExistingEventTypes);
-            LOG.debug(errorMessage);
+        final List<EventType> existingEventTypes = getExistingEventTypes(subscriptionBase);
```
antban: For me all these methods makes no sence, because the idea of it is pretty strightforward - load everything that is in database, if something is missing - show error. 
```
        final Map<String, Optional<EventType>> eventTypeMapping =
                subscriptionBase.getEventTypes().stream()
                .collect(Collectors.toMap(Function.identity(), t -> eventTypeRepository.findByNameO(t)));
        if (eventTypeMapping.values().stream().filter(o -> !o.isPresent()).findAny().isPresent()) {
            return create(
                    UNPROCESSABLE_ENTITY, 
                    createErrorMessage(eventTypeMapping.entrySet().stream().filter(e -> !e.getValue().isPresent()).map(Map.Entry::getKey).collect(Collectors.toList())), 
                    request);
        }
        eventTypeMapping.values().stream().map(Optional::get).forEach(eventType -> client.checkScopes(eventType.getReadScopes()));

        // generate subscription id and try to create subscription in DB
        final Subscription subscription = subscriptionRepository.createSubscription(subscriptionBase);
        return status(HttpStatus.CREATED).body(subscription);
```

The only problem that I see is that method returning optional throws Exception (findByNameO), It should be replaced with ExecutionException, that wraps InternalNakadiException.

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -132,37 +138,36 @@ public SubscriptionController(final SubscriptionDbRepository subscriptionReposit
...
+        final List<String> missingEventTypes = eventTypeMapping.entrySet().stream()
+                .filter(entry -> !entry.getValue().isPresent())
+                .map(Map.Entry::getKey)
+                .collect(Collectors.toList());
+
+        if (!missingEventTypes.isEmpty()) {
+            return create(UNPROCESSABLE_ENTITY, createErrorMessage(missingEventTypes), request);
         }
 
+        eventTypeMapping.values().stream()
```
rcillo: Isnt't there going to be a feature toggle for that?

rcillo:
:shipit:

PR #374, opened 2016-08-08
Remove dependency on non public class

PetrGlad:
com.sun.security.auth might not be available in OpenJDK


v-stepanov:
@PetrGlad Hey, why you closed the PR? I wanted to merge it.

PetrGlad:
Because I had no idea which branches/project to use. I can push branches to nakadi-incubator but it seem to have "master" and "jvm-master" some having ~1000 commits diff with nakadi/master and hence merge conflicts. It was not  not even obvious whether it should be Python or Java :)

Next time I probably should just fork it as my personal repo (not Zalando's) before making changes.

PR #375, opened 2016-08-10
ARUHA-320 fix travis docker update

adyach:


.travis.yml
```diff
@@ -12,14 +12,9 @@ services:
...
 
-# Travis has an old version of docker and docker-compose that is why we need to load from the net and install them.
 before_install:
-  - sudo sh -c 'echo "deb https://apt.dockerproject.org/repo ubuntu-precise main" > /etc/apt/sources.list.d/docker.list'
-  - sudo apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
-  - sudo apt-get update
-  - sudo apt-key update
-  - sudo apt-get -qqy -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install docker-engine=${DOCKER_VERSION}
-  - sudo rm /usr/local/bin/docker-compose
+  - docker --version
```
v1ctor: Does travis have docker by default now?
Do we need to print version? :)

adyach: Travis had docker by default but with very old version and because of that we have to reinstall to the newer version. Right now it has the latest version 1.12.
I think it is useful to have in case we are gonna have similar issues with that in the future. I was not able to find info about current travis docker version in the docs.

v-stepanov:
For me it looks good

v-stepanov:
:shipit:

antban:
:+1:

PR #377, opened 2016-08-12
Clarify event schema definition format

PetrGlad:


rcillo:
We omit some details in the README to avoid duplicating information that is already present in the API https://github.com/zalando/nakadi/blob/bf447f0cd03186bd5ae76f575489ef7f4ee083b9/api/nakadi-event-bus-api.yaml#L175-L176

PetrGlad:
Yes I already know that this took me a while to find that out. And this will be a problem for every new user. Probably then one should remove examples from README and create complete separate tutorial instead.

PR #378, opened 2016-08-12
ARUHA-164: validate partition keys

v1ctor:


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -180,52 +181,50 @@ private void validateSchemaChange(final EventType eventType, final EventType exi
         }
     }
 
-    private void validatePartitionKeys(final EventType eventType) throws InvalidEventTypeException {
```
rcillo: The problem with removing this toggle is that currently it's impossible to change schemas but it's possible to change partition keys. But after merging this code it will be impossible to change partition keys if they are not defined in the schema.

But anyway, I think if we notify users properly, there should be no major consequences.

v1ctor: @rcillo This method has been moved below validaeSchema, and there is toggle :)

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -180,52 +181,50 @@ private void validateSchemaChange(final EventType eventType, final EventType exi
...
     }
 
     private void assignTopic(final EventType eventType) {
         eventType.setTopic(uuidGenerator.randomUUID().toString());
     }
 
-    private boolean hasReservedField(final JSONObject schemaAsJson, final String field) {
-        return schemaAsJson.optJSONObject("properties") != null
-                && schemaAsJson.getJSONObject("properties").has(field);
+    private String convertToJSONPointer(final String value) {
```
adyach: I think it is better to put in utility class

src/test/java/org/zalando/nakadi/validation/JSONSchemaValidationTest.java
```diff
@@ -57,7 +57,7 @@ public void schemaValidationShouldRespectEventTypeDefinition() {
                 equalTo("#: required key [bar] not found"));
         assertThat(validator.validate(invalidEventMissingNameField).get().getMessage(), equalTo("name is required"));
         assertThat(validator.validate(nestedSchemaViolation).get().getMessage(),
-                equalTo("#: 2 schema violations found\n#/bar: 2 schema violations found\n#/bar: required key [foo]" +
+                equalTo("#: 3 schema violations found\n#/bar: 2 schema violations found\n#/bar: required key [foo]" +
```
rcillo: Is this related to some other fix in the library?

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -180,52 +181,50 @@ private void validateSchemaChange(final EventType eventType, final EventType exi
...
+        return value.replaceAll("\\.", "/");
```
rcillo: I would not like to go into the details of JSON Pointer, since our previous path representation is simplified, but maybe we should have another ticket to avoid definition of properties with weird string, like https://tools.ietf.org/html/rfc6901#section-5

Because if we allow that, then this transformation could fail.

rcillo:
👍

rcillo:
Please, add a test case with "$ref"

src/test/java/org/zalando/nakadi/validation/JSONSchemaValidationTest.java
v1ctor: Yes.

v1ctor:
Could you please describe in more details?

rcillo:
@v1ctor  partition keys should work with schemas that have "definitions" and thus "$ref" fields.

v1ctor:
Ok, I can. But I have written this tests in library would you like me to write the same tests in nakadi?

rcillo:
If that's the case, then I see no need for that.

rcillo:
:shipit:

PR #385, opened 2016-08-22
ARUHA-124: added appD utility

adyach:


Dockerfile
```diff
@@ -10,4 +10,5 @@ ADD scm-source.json /scm-source.json
 EXPOSE 8080
 
 # run the server when a container based on this image is being run
-ENTRYPOINT java $JAVA_OPTS -Dloader.path=/plugins -Djava.security.egd=file:/dev/./urandom -jar nakadi.jar
\ No newline at end of file
+ENTRYPOINT java $JAVA_OPTS $(appdynamics-agent) -Dloader.path=/plugins -Djava.security.egd=file:/dev/./urandom -jar \
```
v1ctor: How does it work? Don't we need to have dependency?

valgog: Oh, is there a way not to have appdinamics dependency like this in our open-source project?

adyach: @v1ctor we use "zalando/openjdk" Docker base image that's why the image already has necessary env for that (at least they say that)

v1ctor: I propose to add this to JavaOpts property in nakadi-private project.

adyach: @valgog @v1ctor thanks! I will move it to private repo then.

PR #386, opened 2016-08-24
aruha-150: Close streams on event type deletion

adyach:
I've implemented only the first part of the task. the event type won't be removed anymore it will be marked as deleted.

### DIFF
--- a/database/nakadi/10_data/04_tables/01_event_type.sql
+++ b/database/nakadi/10_data/04_tables/01_event_type.sql
@@ -1,5 +1,5 @@
 CREATE TABLE IF NOT EXISTS zn_data.event_type (
-  et_name varchar(255) NOT NULL PRIMARY KEY CHECK (et_name <> ''),
+  et_name varchar(255) NOT NULL CHECK (et_name <> ''),
   et_topic varchar(255) NOT NULL,
   et_event_type_object jsonb NOT NULL,
   et_deleted boolean DEFAULT FALSE,
--- a/database/nakadi/10_data/04_tables/01_event_type.sql
+++ b/database/nakadi/10_data/04_tables/01_event_type.sql
@@ -2,7 +2,7 @@ CREATE TABLE IF NOT EXISTS zn_data.event_type (
   et_name varchar(255) NOT NULL CHECK (et_name <> ''),
   et_topic varchar(255) NOT NULL,
   et_event_type_object jsonb NOT NULL,
-  et_deleted boolean DEFAULT FALSE,
+  et_archived boolean DEFAULT FALSE,
   CHECK (et_event_type_object->>'name' = et_name)
 );
 
--- a/database/nakadi/10_data/04_tables/01_event_type.sql
+++ b/database/nakadi/10_data/04_tables/01_event_type.sql
@@ -2,6 +2,7 @@ CREATE TABLE IF NOT EXISTS zn_data.event_type (
   et_name varchar(255) NOT NULL PRIMARY KEY CHECK (et_name <> ''),
   et_topic varchar(255) NOT NULL,
   et_event_type_object jsonb NOT NULL,
+  et_deleted boolean DEFAULT FALSE,
   CHECK (et_event_type_object->>'name' = et_name)
 );
 
--- a/src/main/java/org/zalando/nakadi/repository/EventTypeRepository.java
+++ b/src/main/java/org/zalando/nakadi/repository/EventTypeRepository.java
@@ -18,7 +18,7 @@ public interface EventTypeRepository {
 
     List<EventType> list();
 
-    void setEventTypeDeleted(String name) throws InternalNakadiException, NoSuchEventTypeException;
+    void archiveEventType(String name) throws InternalNakadiException, NoSuchEventTypeException;
 
     default Optional<EventType> findByNameO(final String eventTypeName) throws InternalNakadiException {
         try {
--- a/src/main/java/org/zalando/nakadi/repository/EventTypeRepository.java
+++ b/src/main/java/org/zalando/nakadi/repository/EventTypeRepository.java
@@ -18,8 +18,6 @@ public interface EventTypeRepository {
 
     List<EventType> list();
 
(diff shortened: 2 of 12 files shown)


src/main/java/org/zalando/nakadi/repository/db/EventTypeDbRepository.java
```diff
@@ -86,7 +92,7 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
     @Override
     public List<EventType> list() {
```
v-stepanov: I think you should also adjust saving of event-type. Because with current implementation it will not be possible for user to delete event-type and then create another one with the same name (method with throw DuplicatedEventTypeNameException)

database/nakadi/10_data/04_tables/01_event_type.sql
```diff
@@ -2,6 +2,7 @@ CREATE TABLE IF NOT EXISTS zn_data.event_type (
   et_name varchar(255) NOT NULL PRIMARY KEY CHECK (et_name <> ''),
```
v-stepanov: I think as we don't really remove event types now we should remove this `PRIMARY KEY` because now it will be possible to have several event types with the same name when user removes event type and then creates another one with the same name.
And we should control the uniqueness of event type names only for event types with et_deleted = FALSE

src/main/java/org/zalando/nakadi/repository/db/EventTypeDbRepository.java
adyach: You are back, @v-stepanov :) Thanks, I will fix it!

src/main/java/org/zalando/nakadi/repository/db/EventTypeDbRepository.java
```diff
@@ -48,10 +48,15 @@ public void saveEventType(final EventType eventType) throws InternalNakadiExcept
...
     @Override
     public EventType findByName(final String name) throws NoSuchEventTypeException {
-        final String sql = "SELECT et_topic, et_event_type_object FROM zn_data.event_type WHERE et_name = ?";
+        final String sql = "SELECT et_topic, et_event_type_object, et_deleted " +
+                "FROM zn_data.event_type WHERE et_name = ?";
 
         try {
-            return jdbcTemplate.queryForObject(sql, new Object[]{name}, new EventTypeMapper());
+            final EventType eventType = jdbcTemplate.queryForObject(sql, new Object[]{name}, new EventTypeMapper());
+            if (eventType.isDeleted()) {
```
rcillo: I would say that this could be done using plain query. Maybe adding another condition `et_deleted = false`? Then it would return an empty set.

rcillo:
@adyach I couldn't find any information in the ticket about not deleting the event-type https://techjira.zalando.net/browse/ARUHA-150

Could you explain how this change will help solving the ticket.

adyach:
@rcillo I do not think this change touches directly aruha-150, but I think it is nice to have event types saved foerver because of that I decided to add it in that task.

src/main/java/org/zalando/nakadi/repository/EventTypeRepository.java
```diff
@@ -20,6 +20,8 @@
 
     void removeEventType(String name) throws InternalNakadiException, NoSuchEventTypeException;
 
+    void setEventTypeDeleted(String name) throws InternalNakadiException, NoSuchEventTypeException;
```
rcillo: From a client code perspective (invoking this method in a controller, for example) it's difficult to tell the difference from `removeEventType`. They seem to be the same thing. One would have to read the implementation to understand the difference. Also, why should we have true deletion and also "stored forever deletion" at the same time?

adyach:
@v-stepanov @rcillo @v1ctor another iteration on that pr, thank you for your comments

src/main/java/org/zalando/nakadi/repository/db/EventTypeDbRepository.java
```diff
@@ -86,20 +89,20 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
...
     public List<EventType> list() {
         return jdbcTemplate.query(
-                "SELECT et_topic, et_event_type_object FROM zn_data.event_type",
+                "SELECT et_topic, et_event_type_object, et_deleted FROM zn_data.event_type WHERE et_deleted = FALSE",
                 new EventTypeMapper());
     }
 
     @Override
-    public void removeEventType(final String name) throws NoSuchEventTypeException, InternalNakadiException {
+    public void setEventTypeDeleted(final String name) throws InternalNakadiException, NoSuchEventTypeException {
```
rcillo: I would suggest changing this name to "archive".

adyach: @rcillo https://github.com/zalando/nakadi/pull/386/commits/351ee0c7db54ca659e6760a5118cb1899284bbd3

src/main/java/org/zalando/nakadi/repository/db/EventTypeDbRepository.java
rcillo: @adyach what has been done regarding @v-stepanov comment?

adyach: > I think you should also adjust saving of event-type. Because with current implementation it will not be possible for user to delete event-type and then create another one with the same name (method with throw DuplicatedEventTypeNameException)

I fixed it like that https://github.com/zalando/nakadi/pull/386/files#diff-7bd69958c0e0e2d02f44804aacc370d8R35
and findByNameO selects only queries are not archived

adyach: and test https://github.com/zalando/nakadi/pull/386/commits/71d5574c65915a82f86caed7b2a2886c9b1d1516

adyach:
I closed this PR since it was opened for too long time without any blames/approves :(

PR #387, opened 2016-08-25
Clarify on required docker versions.

mtesseract:
Nakadi doesn't work with some popular docker versions.
The documentation should mention that you need very recent versions of docker.


adyach:
@mtesseract hi!
Thank you for helping out with the project!

I think it is better to specify docker / docker-compose verisons directly.
docker-compose >= 1.7.0

adyach:
👍

PR #392, opened 2016-08-29
aruha-124: simplified docker file + added support for providing docker file path for gradle task

adyach:
It relates to this one https://github.bus.zalan.do/aruha/nakadi-private/pull/40


Dockerfile
```diff
@@ -4,10 +4,7 @@ MAINTAINER Team Aruha, team-aruha@zalando.de
...
 ADD build/libs/nakadi.jar nakadi.jar
-ADD plugins plugins/
-ADD scm-source.json /scm-source.json
 
 EXPOSE 8080
 
-# run the server when a container based on this image is being run
-ENTRYPOINT java $JAVA_OPTS -Dloader.path=/plugins -Djava.security.egd=file:/dev/./urandom -jar nakadi.jar
\ No newline at end of file
+ENTRYPOINT java -jar nakadi.jar
```
v1ctor: How I understood Slava it was need for faster startup. Have you tested it?

adyach: @v1ctor On local machine it was started up just fine. I left it for private part. In case you will experience issues with that we can add it in the future.

v-stepanov: I checked it now. Currently I don't experience any problems with that. So I think we can remove that 
(if the problem happens again in the future we can return it back)

v-stepanov:
:shipit:

Dockerfile
```diff
@@ -4,10 +4,7 @@ MAINTAINER Team Aruha, team-aruha@zalando.de
 
 WORKDIR /
 ADD build/libs/nakadi.jar nakadi.jar
-ADD plugins plugins/
-ADD scm-source.json /scm-source.json
```
rcillo: Why removing this from the public repo?

PR #394, opened 2016-08-30
aruha-339: added description for write/read scopes

adyach:


api/nakadi-event-bus-api.yaml
```diff
@@ -1024,6 +1024,26 @@ definitions:
         description: |
           Provides ability to set internal Nakadi parameters.
 
+      write_scopes:
+        type: array
+        items:
+          type: string
+        description: |
+          Provides ability to set oauth2 scopes to be able to publish an Event. The event publisher is cheked for
```
rcillo: What do you think of these suggestions:

"This field is used to authorise clients for publishing to this event type. Nakadi only authorises publishers whose token contains at least one of the scopes in this list."

rcillo: But keeping "If no scopes provided then anyone can publish to this event type."

adyach: I think it is good enough! My point was to describe it as wide as possible to cut additional questions in the chat room.
I tweaked it a bit:
This field is used to authorise clients for publishing to this event type. Nakadi authorises publishers whose token contains at least one of the scopes in this list. Publishers in their turn has to request scopes from event type owner.

rcillo: I think the idea is clever, but what about our manual http://zalando.github.io/nakadi-manual/? I would say that the API definition should merely describe the semantics of fields and the manual should have a more extensive usage description.

rcillo:
As you mentioned in https://github.com/zalando/nakadi/pull/394#discussion_r76779670 users might have many question on how to operate this. It would be great to see another PR with some more explanation in the manual 😄

rcillo:
:shipit:

adyach:
@rcillo already in progress :)

PR #395, opened 2016-08-30
Aruha 338 stream parameters change

rcillo:
This PR is part of bigger task. I'm merging it against a feature branch, so that I could get some early feedback and make reviewing easier. It includes:

## rename window_size and change default value

Renaming window_size to max_uncommitted_size makes the API easier to
understand since it doesn't introduce any new terminology (window
concept is not present anywhere in the api so far).

We are also reducing the default amount to 10 since there are some
potentially large messages and a default of 100 would mean that
several megabytes would be streamed at once.

## remove commit_timeout

Allowing clients to control the commit_timeout implied in potentially
unlimited resource allocation in the server, leading to security and
operational problem.

We removed this parameter from the API but kept it as an application
configuration.

The default value was also changed to 60 seconds for launching locally
but as 5 seconds for testing purposes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -223,6 +223,17 @@ task pushDockerImage {
     }
 }
 
+task testInitNakadi(type: Exec) {
+    environment "NAKADI_STREAM_DEFAULT_COMMITTIMEOUT", "5"
+    commandLine "bash", "-c", "docker-compose up -d --build"
+    doLast {
+        println "Waiting for Nakadi to start up"
+        long start = System.currentTimeMillis()
+        waitForNakadi()
+        println "Nakadi is fully started in " + (System.currentTimeMillis() - start) + " ms"
+    }
+}
+
 task initNakadi(type: Exec) {
     commandLine "bash", "-c", "docker-compose up -d --build"
     doLast {
@@ -267,12 +278,12 @@ task stopStorages(type: Exec) {
 }
 
 task fullTest(type: GradleBuild) {
-    tasks = ['clean', 'bootRepackage', 'initNakadi', 'test', 'acceptanceTest']
+    tasks = ['clean', 'bootRepackage', 'testInitNakadi', 'test', 'acceptanceTest']
     finalizedBy stopNakadi
 }
 
 task fullAcceptanceTest(type: GradleBuild) {
-    tasks = ['clean', 'bootRepackage', 'initNakadi', 'acceptanceTest']
+    tasks = ['clean', 'bootRepackage', 'testInitNakadi', 'acceptanceTest']
     finalizedBy stopNakadi
 }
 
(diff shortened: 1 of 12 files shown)


docker-compose.yml
```diff
@@ -10,12 +10,13 @@ services:
      - zookeeper
      - kafka
     environment:
-      NAKADI_OAUTH2_MODE: "OFF"
-      NAKADI_FEATURETOGGLE_DEFAULT: "true"
-      NAKADI_ZOOKEEPER_BROKERS: localhost:2181
-      SPRING_DATASOURCE_URL: jdbc:postgresql://localhost:5432/local_nakadi_db
-      SPRING_DATASOURCE_USERNAME: nakadi
-      SPRING_DATASOURCE_PASSWORD: nakadi
+      - NAKADI_STREAM_DEFAULT_COMMITTIMEOUT
```
adyach: It looks strange that env variable has no value here. Maybe we can define it here like 60?

build.gradle
```diff
@@ -223,6 +223,17 @@ task pushDockerImage {
     }
 }
 
+task testInitNakadi(type: Exec) {
```
adyach: Can we set env variable inside fullTest/fullAcceptanceTest to not duplicate the code ?

docker-compose.yml
rcillo: It's a feature from docker compose. It just transport this environment variable from the shell, if defined, into the docker container.

The default value is in the application.yml from the application.

One needs to keep in mind the requirements: (1) 5 seconds for tests, (2) 60 seconds for local execution.

build.gradle
rcillo: I tried that and failed. I'm not so experienced with gradle =/

Any ideas?

rcillo: Here is an improvement https://github.com/zalando/nakadi/pull/395/commits/9a1c210aa5d71aa0b2855c685980958d8f5ed4da

I managed to use a different profile, but in order to do so I still need to have this environment variable set.

The advantage of the latest approach is that it's now easy to define any new values by just updating a configuration file. There will be no further changes to the gradle file.

adyach: Cool! I'd better use
environment "SPRING_PROFILES_ACTIVE", "acceptanceTest"
because intention is more clear.

v-stepanov:
For me it's :shipit: 
But maybe you would like to apply what Andrey suggest in his comment (SPRING_PROFILES_ACTIVE)

build.gradle
rcillo: @adyach done https://github.com/zalando/nakadi/pull/395/commits/8b47e09ab4f3c312ca822cf14bc1b487da8e7770

adyach:
👍

PR #396, opened 2016-08-31
Aruha 335 adjust list subscriptions

v-stepanov:
Adjusted according to new API https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#L601

I will also make few tweaks to this endpoint API:
1) Min value of "limit" should be 1
2) There should be one more possible response 400
This will come as a separate PR

Please review @adyach @rcillo @v1ctor @antban


src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -110,17 +115,32 @@ public SubscriptionController(final SubscriptionDbRepository subscriptionReposit
...
+            @Nullable @RequestParam(value = "event_type", required = false) final Set<String> eventTypes,
+            @RequestParam(value = "limit", required = false, defaultValue = "20") final int limit,
+            @RequestParam(value = "offset", required = false, defaultValue = "0") final int offset,
             final NativeWebRequest request) {
 
         if (!featureToggleService.isFeatureEnabled(HIGH_LEVEL_API)) {
             return new ResponseEntity<>(NOT_IMPLEMENTED);
         }
 
+        if (limit < 1 || limit > 1000) {
```
v1ctor: I don't know if Spring supports javax.validation or org.hibrnate.validation, but it would be nice to use annotations Min and Max or Range.

v-stepanov:
@rcillo I'm very against absolute URIs. I don't want Nakadi to know its own URL, because currently it can be accessed by several URLs. For example, currently we can access Nakadi by stack-name ("nakadi-live") and by stack-name+version ("nakadi-live-18").

rcillo:
@v-stepanov I see. This argument haven't been raised. I also agree that it would complicate implementation and release process. I think it's important to understand if it's really a problem for client to compose the full URI or it's just a matter of maximize convenience.

src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
```diff
@@ -108,26 +111,65 @@ public void whenListSubscriptionsThenOk() throws ServiceUnavailableException {
...
+                Optional.empty(), 0, 10);
+        assertThat(subscriptions, equalTo(expectedSubscriptions));
+    }
+
+    @Test
+    public void whenListSubscriptionsLimitAndOffsetAreRespected() throws ServiceUnavailableException {
+        final List<Subscription> testSubscriptions = createRandomSubscriptions(10);
         testSubscriptions.forEach(this::insertSubscriptionToDB);
 
         final List<Subscription> expectedSubscriptions = testSubscriptions.stream()
```
adyach: Maybe like here https://github.com/zalando/nakadi/pull/396/files#diff-2cda0fafb983703597e45858ee3ef415R111 ?

v-stepanov:
@rcillo I think for the client it should be easier because he already knows Nakadi URL (as he managed to make the first request)

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -110,17 +115,32 @@ public SubscriptionController(final SubscriptionDbRepository subscriptionReposit
     @RequestMapping(method = RequestMethod.GET)
     public ResponseEntity<?> listSubscriptions(
             @Nullable @RequestParam(value = "owning_application", required = false) final String owningApplication,
+            @Nullable @RequestParam(value = "event_type", required = false) final Set<String> eventTypes,
+            @RequestParam(value = "limit", required = false, defaultValue = "20") final int limit,
+            @RequestParam(value = "offset", required = false, defaultValue = "0") final int offset,
             final NativeWebRequest request) {
 
         if (!featureToggleService.isFeatureEnabled(HIGH_LEVEL_API)) {
```
adyach: Do we still want to have this toggle ?

src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
```diff
@@ -108,26 +111,65 @@ public void whenListSubscriptionsThenOk() throws ServiceUnavailableException {
...
-                .filter(sub -> "myapp1".equals(sub.getOwningApplication()))
                 .sorted(SUBSCRIPTION_CREATION_DATE_DESC_COMPARATOR)
                 .collect(toList());
+        expectedSubscriptions.subList(0, 2).clear();
```
adyach: Could you explain that a bit ?

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
v-stepanov: I checked that. Theoretically it is possible using [MethodValidationPostProcessor](http://docs.spring.io/autorepo/docs/spring-framework/4.1.8.RELEASE/javadoc-api/org/springframework/validation/beanvalidation/MethodValidationPostProcessor.html). But it will not work in our controller test setup :(
So I will keep it as it is.

src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
v-stepanov: Good point
Fixed https://github.com/zalando/nakadi/pull/396/commits/52b346da40c6ce04e6502d5ad1747d98c3a5b7ff

src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
v-stepanov: In this method I test pagination with offset=2 and limit=3.
So in expected list I first remove first two elements (because this is what offset=2 will do) and then I remove the tail of the list, keeping only 3 elements (because the limit=3). And in that way I get a list which I expect to get from repository method.

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
v-stepanov: I think we should keep the toggle till we finish the HILA and will be ready to show that to our users on live.

src/main/java/org/zalando/nakadi/util/SubscriptionsUriHelper.java
```diff
@@ -0,0 +1,50 @@
...
+        Optional<PaginationLinks.Link> nextLink = Optional.empty();
+        if (currentPageItemCount >= limit) {
+            final String nextUri = createSubscriptionListUri(owningApplication, eventTypes, offset + 1, limit);
+            nextLink = Optional.of(new PaginationLinks.Link(nextUri));
+        }
+
+        return new PaginationLinks(prevLink, nextLink);
+    }
+
+    public static String createSubscriptionListUri(final Optional<String> owningApplication,
```
adyach: Just a consideration to use: https://hc.apache.org/httpcomponents-client-4.2.x/httpclient/apidocs/org/apache/http/client/utils/URIBuilder.html

src/main/java/org/zalando/nakadi/util/SubscriptionsUriHelper.java
```diff
@@ -0,0 +1,50 @@
...
+                                                   final Set<String> eventTypes, final int offset, final int limit) {
+        final StringBuilder urlBuilder = new StringBuilder("/subscriptions?");
+        if (!eventTypes.isEmpty()) {
```
adyach: I think you do not need this check here because if Stream is empty it wont do opearations which are defined on it

src/main/java/org/zalando/nakadi/domain/PaginationLinks.java
```diff
@@ -0,0 +1,46 @@
...
+
+    private final Optional<Link> prev;
+
+    private final Optional<Link> next;
+
+    public PaginationLinks() {
+        this(Optional.empty(), Optional.empty());
+    }
+
+    public PaginationLinks(final Optional<Link> prev, final Optional<Link> next) {
```
adyach: What do you think about using LInk in constructor but wrap it in Optional inside constructor or mark links as nullable? I think it will help use to support our 'null policy' because right now we can pass null in constructor.

src/main/java/org/zalando/nakadi/util/SubscriptionsUriHelper.java
adyach: Or even better
http://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/UriComponentsBuilder.html

v-stepanov: Good point. I changed that to using UriComponentsBuilder. Now the code is much cleaner.

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/util/SubscriptionsUriHelper.java
v-stepanov: not relevant any more as I migrated to using UriComponentsBuilder

src/main/java/org/zalando/nakadi/domain/PaginationLinks.java
v-stepanov: As we decided that we prefer Optionals over nulls I would choose to keep the current implementation.

> because right now we can pass null in constructor.

No, we can't, because the parameters are not marked as `@Nullable`

PR #401, opened 2016-08-31
aruha-190 renaming event type property

rcillo:
There is no way to be over specific on the meaning of this field. We use
the same approach in other parts of the API.


v-stepanov:
:shipit:

PR #403, opened 2016-08-31
aruha-190 define offsets format

rcillo:
Up to now Nakadi Cursors were advertised to be opaque entities and
clients should not make assumptions about its structure and meaning.

But we've been observing clients making assumptions and coding based on
them. This could lead to a major confusion in the future if we no longer
hold those assumptions true.

In order to mitigate this risk, we are committing to a well defined
structure beforehand. It's inspired by Postgres WAL format
https://www.postgresql.org/docs/9.2/static/wal-internals.html and we aim
to address two issues with this format:

1. Make Nakadi operations easier (this feature would have been
tremendously useful for cluster migration or implementing partition
rebalacing)
2. Provide a well defined set of guarantees for consumers to rely one.

**update**

As for why 27 characters = 9 (max integer 7FFFFFFF) + separator + 17 (max long 7FFFFFFFFFFFFFFF).

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1191,10 +1191,10 @@ definitions:
             000000001.
           - the remaining 17 characters defines the index, a strictly increasing sequence in the scope of a timeline.
             This maps Kafka offsets just by converting to hexadecimals.
+          - offsets for empty partitions have a special value of BEGIN.
           - offsets are fully ordered and comparable lexicographicaly.
           - it's not possible to determine how many events there are between two offsets from different timelines.
           - timeline and index are always padded by zeros.
-          - the first event in a log has a special offset of BEGIN.
   SubscriptionCursor:
     allOf:
       - $ref: '#/definitions/Cursor'
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1184,8 +1184,14 @@ definitions:
       offset:
         type: string
         description: |
-          Offset of the event being pointed to.
+          Offset of the event being pointed to. Regarding its format and guarantees:
 
+          - offsets have a fixed size of 26 characters ranging from 0 to F (Hexadecimals).
+          - the first 9 characters are used to define the timeline, a monotonically increasing sequence.
+          - the remaining 17 characters defines the index, a strictly increasing sequence in the scope of a timeline.
+            This maps Kafka offsets just by converting to hexadecimals.
+          - offsets are fully ordered and comparable lexicographicaly.
+          - it's not possible to determine how many events there are between two offsets from different timelines.
   SubscriptionCursor:
     allOf:
       - $ref: '#/definitions/Cursor'
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1197,7 +1197,7 @@ definitions:
           - it's not possible to determine how many events there are between two offsets. Check
             SubscriptionEventTypeStats for that.
           - timeline and index are zero padded.
-        example: "0000000AB/0123456789ABCDEF"
+        example: "0000000AB/0123456789ABCDEFF"
   SubscriptionCursor:
     allOf:
       - $ref: '#/definitions/Cursor'
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1192,6 +1192,7 @@ definitions:
             This maps Kafka offsets just by converting to hexadecimals.
           - offsets are fully ordered and comparable lexicographicaly.
           - it's not possible to determine how many events there are between two offsets from different timelines.
+          - timeline and index are always padded by zeros.
   SubscriptionCursor:
     allOf:
       - $ref: '#/definitions/Cursor'
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1186,7 +1186,7 @@ definitions:
         description: |
           Offset of the event being pointed to. Regarding its format and guarantees:
 
-          - offsets have a fixed size of 27 characters ranging from 0 to F (Hexadecimals).
+          - offsets have a fixed size of 27 characters.
           - the first 9 characters are used to define the timeline, a monotonically increasing sequence. Starts with
             000000001.
           - its followed by the separator "/".
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1194,8 +1194,9 @@ definitions:
             This maps Kafka offsets just by converting to hexadecimals.
           - offsets for empty partitions have a special value of BEGIN.
           - offsets are fully ordered and comparable lexicographicaly.
-          - it's not possible to determine how many events there are between two offsets from different timelines.
-          - timeline and index are always padded by zeros.
+          - it's not possible to determine how many events there are between two offsets. Check
+            SubscriptionEventTypeStats for that.
+          - timeline and index are zero padded.
   SubscriptionCursor:
     allOf:
       - $ref: '#/definitions/Cursor'
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1194,6 +1194,7 @@ definitions:
           - offsets are fully ordered and comparable lexicographicaly.
           - it's not possible to determine how many events there are between two offsets from different timelines.
           - timeline and index are always padded by zeros.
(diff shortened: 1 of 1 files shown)


api/nakadi-event-bus-api.yaml
```diff
@@ -1184,8 +1184,17 @@ definitions:
...
+
+          - offsets have a fixed size of 26 characters ranging from 0 to F (Hexadecimals).
+          - the first 9 characters are used to define the timeline, a monotonically increasing sequence. Starts with
+            000000001.
+          - the remaining 17 characters defines the index, a strictly increasing sequence in the scope of a timeline.
+            This maps Kafka offsets just by converting to hexadecimals.
+          - offsets are fully ordered and comparable lexicographicaly.
+          - it's not possible to determine how many events there are between two offsets from different timelines.
+          - timeline and index are always padded by zeros.
+          - the first event in a log has a special offset of BEGIN.
```
v-stepanov: why? I think the first event has offset zero

v-stepanov: Virtual event before the first one has offset BEGIN

rcillo:
@valgog @dehora @antban @v-stepanov @v1ctor @adyach what are your thoughts on this?

I would say that it's going to be hard to implement. 

1. in the code there are many direct casts from String to Long which would be broken.
2. not been able to count how many events there are between two cursors from different timelines would break the functionality from SubscriptionStats

On the other hand, we could just trim the timeline part and keep the implementation as is, delaying the cost of a real implementation.

v-stepanov:
Don't we want to separate the parts by separator, as we advertise cursor as a string anyway ?

rcillo:
@v-stepanov one could argue that a separator would be redundant since the sizes are fixed and defined. On the other hand, it could make implementation less error prone, since splitting by a separator is more guaranteed than splitting by index.

v-stepanov:
@rcillo I think as those are two separate logical parts It doesn't make much sense to merge it. I would choose the implementation with separator. But let's see what others say.

rcillo:
I've added a separator and removed the possibility for doing arithmetics.

api/nakadi-event-bus-api.yaml
```diff
@@ -1186,15 +1186,17 @@ definitions:
         description: |
           Offset of the event being pointed to. Regarding its format and guarantees:
 
-          - offsets have a fixed size of 26 characters ranging from 0 to F (Hexadecimals).
+          - offsets have a fixed size of 27 characters ranging from 0 to F (Hexadecimals).
```
valgog: Formally speaking after introduction of a separator `/` these 27 characters are not ranging from 0 to F. One of them is `/`

valgog: maybe a regex would be an easier as a documentation? 

`^([0-F]{9})/([0-F]{17})$`

api/nakadi-event-bus-api.yaml
```diff
@@ -1184,8 +1184,19 @@ definitions:
       offset:
```
adyach: @rcillo I think it is good to see an example here

api/nakadi-event-bus-api.yaml
rcillo: Fixed in https://github.com/zalando/nakadi/pull/403/commits/de05a98ad2fe2fa3323cb2b6e0e39afc00362ff7

I think that after adding an example it became more clear. I would say that the regex could just make it more scary to some people.

api/nakadi-event-bus-api.yaml
rcillo: Done 👍 https://github.com/zalando/nakadi/pull/403/commits/663d602655f23f56a97a86296a009dcec4fd3bf2

tlossen:
hmmmm ... i see the convenience of having offsets scoped by timeline. how about introducing timelines in a more backwards-compatible way?

a long is 64 bit, which we could divide into timeline (16 bit) plus offset (48 bit). in hex, that would be 16 characters, for example 000100000000babe (timeline 0001, offset 00000000babe).

this way, Long.parseLong(value, 16) would still give me a usable long offset that can be compared arithmetically, not only lexicographically.

48 bit is plenty! i did a quick calculation, it would last 8500 years if you publish 1000 events per second. 65K timelines should also be enough for now.

rcillo:
@tlossen we've seem some clients doing arithmetics with offsets. Something like `number_of_events_to_be_consumed = latest_offset_available - current_consumed_offset`. If we make it "backward compatible" (I quoted it because such guarantess have never been in the contract), it would just be delaying errors that should be spotted as soon as possible.

One of the motivations for this change is to push clients to fix their implementations and stop making dangerous assumptions over offsets.

tkrop:
Do you really want to expose Kafka implementation details in this kind. Where is the sugar Nakadi may be able to add to Kafka in the future, if you "fix" it this way?

rcillo:
@tkrop could you please be more specific?

api/nakadi-event-bus-api.yaml
```diff
@@ -1184,8 +1184,20 @@ definitions:
...
-          Offset of the event being pointed to.
-
+          Offset of the event being pointed to. Regarding its format and guarantees:
+
+          - offsets have a fixed size of 27 characters.
+          - the first 9 characters are used to define the timeline, a monotonically increasing sequence. Starts with
+            000000001.
+          - its followed by the separator "/".
+          - the remaining 17 characters defines the index, a strictly increasing sequence in the scope of a timeline.
+            This maps Kafka offsets just by converting to hexadecimals.
```
bocytko: Can you clarify that hex thing with an example?

rcillo: @bocytko example added bellow.

tlossen:
minor nitpick: max integer if 7FFFFFFF (= 8 hex chars), max long is 7FFFFFFFFFFFFFFF (= 16 hex chars).

api/nakadi-event-bus-api.yaml
```diff
@@ -1184,8 +1184,20 @@ definitions:
...
+          - offsets for empty partitions have a special value of BEGIN.
+          - offsets are fully ordered and comparable lexicographicaly.
+          - it's not possible to determine how many events there are between two offsets. Check
+            SubscriptionEventTypeStats for that.
+          - timeline and index are zero padded.
+        example: "0000000AB/0123456789ABCDEF"
```
bocytko: The length is now 26 (1 less than before).

api/nakadi-event-bus-api.yaml
bocytko: The length is now 26 (1 less than before).

rcillo: Bad `wc` usage options. My mistake. Fixed.

bocytko:
> we've seem some clients doing arithmetics with offsets. Something like `number_of_events_to_be_consumed = ...`

This is a natural concequence of exposing numeric cursors.

If clients do it now, why not provide the data they need directly within the API?
- number of available events since `last_offset`
- cursor for `last_offset`
- cursor for `next_offset`

Clients can use different operations:
a) Give me N events since `last_offset`
b) Give me the offset `last_offset + N`
c) Give me up to N events beginning with `next_offset` (long-poll).

The offsets could then be non-numeric strings like `uuids`, similar to database cursors where you have special functions for manipulating and fetching data from.

tlossen:
@rcillo ok, let me rephrase my proposal. do we really need 92 bit for a simple offset? or can we maybe fit it into 64 bit, as described above? 

my main concern is ease of use / overhead when passing it around or storing it, as the largest base types in many systems or languages are 64 bit (for example long in java, bigint in postgres).

jhorstmann:
Do we need to define this format at all in this detail? My only usecase as a library author is described at https://github.com/zalando-incubator/fahrschein/blob/0.4.4/src/main/java/org/zalando/fahrschein/CursorManager.java#L50 and any format that can be compared lexicographically (maybe after padding to the same length) would support that usecase. What are the usecases that other libraries or users have?

rcillo:
After @tkrop and @bocytko suggestions, we are inclined to ditch this proposal in favor of a completely opaque cursor, which would contain only encoded data.

It would make room for further extensions at the cost of opaqueness. On the other hand, we would be adding endpoints for cursor manipulation, filling the gap left by opaque cursors.

tlossen:
@rcillo in that case, it would be good if nakadi could provide something that represents the (per-partition) order of the stream, like the SequenceNumber in kinesis.

rcillo:
We are closing this PR after collecting all your feedback. We'll be working on another proposal for opaque cursors.

@tlossen when working on this new proposal we'll study the viability of adding this SequenceNumber 👍

PR #405, opened 2016-09-01
Aruha 340 subscription cursor

v-stepanov:
I wanted just to change cursor structure according to latest API but that caused a lot of changes

@adyach @rcillo @v1ctor @antban please review


src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -50,43 +51,53 @@
...
-        topicRepository.validateCommitCursors(eventType.getTopic(), cursors);
 
         boolean allCommitted = true;
-        for (final Cursor cursor : cursors) {
-            final boolean cursorCommitted = commitCursor(subscriptionId, eventType.getTopic(), cursor);
-            allCommitted = allCommitted && cursorCommitted;
+        for (final String eventTypeName : subscription.getEventTypes()) {
+
+            final EventType eventType = eventTypeRepository.findByName(eventTypeName);
+            createSubscriptionInZkIfNeeded(subscription);
```
rcillo: Why is it required to do it for every event type in the subscription?

v-stepanov: It's a bug :)
Fixed https://github.com/zalando/nakadi/pull/405/commits/aebaaae51282d7c29865b46fe249970cbee22c8d

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -157,19 +160,36 @@ private void streamToOutput() {
...
 
+    private String serializeBatch(final Partition.PartitionKey partitionKey, final String offset,
+                                  final List<String> events) throws JsonProcessingException {
+
+        final String eventType = getContext().getEventTypesForTopics().get(partitionKey.topic);
+        final String token = getContext().getCursorTokenService().generateToken();
+        final SubscriptionCursor cursor = new SubscriptionCursor(partitionKey.partition, offset, eventType, token);
+        final String cursorSerialized = getContext().getObjectMapper().writeValueAsString(cursor);
+
+        final StringBuilder builder = new StringBuilder()
```
rcillo: Is it done with string builder for efficiency? Why not use json object mapper and serialize a Domain object?

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -50,43 +51,52 @@
...
         createSubscriptionInZkIfNeeded(subscription);
-        topicRepository.validateCommitCursors(eventType.getTopic(), cursors);
 
         boolean allCommitted = true;
-        for (final Cursor cursor : cursors) {
-            final boolean cursorCommitted = commitCursor(subscriptionId, eventType.getTopic(), cursor);
-            allCommitted = allCommitted && cursorCommitted;
+        for (final String eventTypeName : subscription.getEventTypes()) {
+
+            final EventType eventType = eventTypeRepository.findByName(eventTypeName);
```
rcillo: Why not to iterate over cursors?

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -50,43 +51,52 @@
...
-        final EventType eventType = eventTypeRepository.findByName(eventTypeName);
-
         createSubscriptionInZkIfNeeded(subscription);
-        topicRepository.validateCommitCursors(eventType.getTopic(), cursors);
 
         boolean allCommitted = true;
-        for (final Cursor cursor : cursors) {
-            final boolean cursorCommitted = commitCursor(subscriptionId, eventType.getTopic(), cursor);
-            allCommitted = allCommitted && cursorCommitted;
+        for (final String eventTypeName : subscription.getEventTypes()) {
```
rcillo: Why not to iterate over cursors?

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
v-stepanov: No, it's not because of efficiency (actually on my benchmark jackson showed better results than making that by hand with string builder)

The thing is that the events are already serialized json strings, so we can't put events to some domain object and then serialize the object because the events will be serialized twice in that case. That's why we have to do this manual work with string builder. 
(Another possibility could be to deserialize events, put them to some batch-object and then serialize the whole object, but I'm afraid that will be very time-consuming)

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: Yes, it really makes sense to iterate by cursors. I changed that, now the code is more clear.

v-stepanov: https://github.com/zalando/nakadi/pull/405/commits/b316b0f3e74ca930098b4fec7c008aeb0c649418

rcillo:
:shipit:

rcillo:
:shipit:

PR #406, opened 2016-09-01
aruha-338 add metadata debugging info to first and last batch

rcillo:
In order to provide more information for users debug their
implementation, we are writting the reason for closing a connection in
the last batch.

We could introduce more information to this stream metadata in the
future, so that clients would be notified of important events such as
rebalances or even help then distinguish between empty logs or blocking
commits.

To make sure that clients won't be suprised by such field at the very
last batch, we are also sending a welcome message in the fist batch, so
that if someone is implementing in a fragile way, their code will break
early so that can fix.


src/acceptance-test/java/org/zalando/nakadi/webservice/hila/StreamBatch.java
```diff
@@ -17,11 +18,14 @@
 
     private final Cursor cursor;
     private final List<Map> events;
+    private final JsonNode metadata;
 
     public StreamBatch(@JsonProperty("cursor") final Cursor cursor,
-                       @Nullable @JsonProperty("events") final List<Map> events) {
+                       @Nullable @JsonProperty("events") final List<Map> events,
+                       @JsonProperty("metadata") final JsonNode metadata) {
```
v-stepanov: I think for consistency it should have type Map. (or events should have type List<JsonNode>)

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -72,6 +73,9 @@ private void checkCommitTimeout() {
 
     private void shutdownGracefully(final String reason) {
         getLog().info("Shutting down gracefully. Reason: {}", reason);
+        offsets.entrySet().stream().findFirst()
+                .ifPresent(pk -> flushData(pk.getKey(), new TreeMap<>(), Optional.of(reason)));
```
v-stepanov: So you will send that for random topic-partition, right?

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -156,13 +164,14 @@ private void streamToOutput() {
...
-    private void flushData(final Partition.PartitionKey pk, final SortedMap<Long, String> data) {
-        final String evt = EventStream.createStreamEvent(
-                pk.partition,
-                String.valueOf(offsets.get(pk).getSentOffset()),
-                new ArrayList<>(data.values()),
-                Optional.empty());
+    private void flushData(final Partition.PartitionKey pk, final SortedMap<Long, String> data,
+                           final Optional<String> metadata) {
         try {
+            final String evt = EventStream.createStreamEvent(
```
v-stepanov: please merge with master, it's different now.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
rcillo: Yes. I though about changing the API in order to send a batch with only metadata. Wdyt?

v-stepanov: For me it's fine if you send it with random partition.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
rcillo: Done

jhorstmann:
Is the new property supposed to be named `metadata` as in the code or `info` as in the swagger docs?

rcillo:
@jhorstmann it's going to be named metadata. The branch with the API is still under some minor adjustments, sorry.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -186,6 +195,9 @@ private String serializeBatch(final Partition.PartitionKey partitionKey, final S
             events.stream().forEach(event -> builder.append(event).append(","));
             builder.deleteCharAt(builder.length() - 1).append("]");
         }
+        if (metadata.isPresent()) {
```
adyach: For me this one looks cleaner:
metadata.ifPresent(data->builder.append(",\"metadata\":{\"debug\":\"").append(data).append("\"}"))

rcillo:
@v1ctor @v-stepanov @antban @adyach could you please review again. I did some changes because there was a bug causing some tests to fail.

src/acceptance-test/java/org/zalando/nakadi/domain/StreamMetadata.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public StreamMetadata(@Nullable @JsonProperty("debug") final String debug) {
+        this.debug = debug;
+    }
+
+    public String getDebug() {
+        return debug;
+    }
+
+    public void setDebug(final String debug) {
```
v-stepanov: I think in our radar we decided to be immutable if possible. Is it possible to make this class immutable?

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
```diff
@@ -72,9 +72,11 @@ public void whenOffsetIsCommittedNextSessionStartsFromNextEventAfterCommitted()
                 .create(URL, subscription.getId(), "stream_limit=2")
                 .start();
         waitFor(() -> assertThat(client.getBatches(), hasSize(2)));
-
         assertThat(client.getBatches().get(0), equalToBatchIgnoringToken(singleEventBatch("0", "0", eventType.getName(),
-                ImmutableMap.of("blah", "foo0"))));
+                ImmutableMap.of("blah", "foo0"), "Stream started")));
+        System.out.println(client.getBatches().get(1));
+        System.out.println(singleEventBatch("0", "1", eventType.getName(),
```
v-stepanov: These two System.out.println can be removed, right?

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
```diff
@@ -207,7 +211,7 @@ public void whenConnectionIsClosedByClientNakadiRecognizesIt() throws Exception
         waitFor(() -> assertThat(client.getBatches(), hasSize(1)));
 
         client.close();
-        Thread.sleep(2000);
+        Thread.sleep(10000);
```
v-stepanov: why it was increased so significantly? As I understand in this PR you just added new property to the batch, I expected that it will not increase any timeouts etc.

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/StreamBatch.java
```diff
@@ -36,30 +40,54 @@ public SubscriptionCursor getCursor() {
         return unmodifiableList(events);
     }
 
+    public StreamMetadata getMetadata() {
```
v-stepanov: If it can be null - the getter should be marked as `@Nullable` or it can be Optional.

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/StreamBatch.java
```diff
@@ -36,30 +40,54 @@ public SubscriptionCursor getCursor() {
...
     @Override
     public boolean equals(final Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
+
         final StreamBatch that = (StreamBatch) o;
-        return cursor.equals(that.cursor) && events.equals(that.events);
+
+        if (!cursor.equals(that.cursor)) return false;
+        if (events != null ? !events.equals(that.events) : that.events != null) return false;
```
v-stepanov: events can't be null

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/StreamBatch.java
```diff
@@ -36,30 +40,54 @@ public SubscriptionCursor getCursor() {
...
+        return metadata != null ? metadata.equals(that.metadata) : that.metadata == null;
+
     }
 
     @Override
     public int hashCode() {
-        return 31 * cursor.hashCode() + events.hashCode();
+        int result = cursor.hashCode();
+        result = 31 * result + (events != null ? events.hashCode() : 0);
```
v-stepanov: events can't be null

src/acceptance-test/java/org/zalando/nakadi/domain/StreamMetadata.java
```diff
@@ -0,0 +1,42 @@
...
+    @Override
+    public int hashCode() {
+        return debug != null ? debug.hashCode() : 0;
+    }
+
+    public StreamMetadata(@Nullable @JsonProperty("debug") final String debug) {
+        this.debug = debug;
+    }
+
+    public String getDebug() {
```
v-stepanov: this should be `@Nullable` as well

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
rcillo: 😓 forgot those...

src/test/java/org/zalando/nakadi/service/EventStreamTest.java
```diff
@@ -294,6 +295,11 @@ private static NakadiKafkaConsumer endlessDummyConsumer() throws NakadiException
 
     private static String jsonBatch(final String partition, final String offset,
                                     final Optional<List<String>> eventsOrNone) {
+        return jsonBatch(partition, offset, eventsOrNone, null);
+    }
+
+    private static String jsonBatch(final String partition, final String offset,
+                                    final Optional<List<String>> eventsOrNone, final String metadata) {
```
v-stepanov: If you expect that metadata can be null - the parameter should be @Nullable

However for me it makes more sense to make it Optional

src/acceptance-test/java/org/zalando/nakadi/domain/StreamMetadata.java
rcillo: Done https://github.com/zalando/nakadi/pull/406/commits/88b912d67ab1850c76bcff58ba4ae30140ad5068

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
rcillo: Done https://github.com/zalando/nakadi/pull/406/commits/3cb83c86c51e58d6623a2dbfd9a736bd1dac918e

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/StreamBatch.java
rcillo: That's not what the API says. It's optional.

src/acceptance-test/java/org/zalando/nakadi/domain/StreamMetadata.java
rcillo: Isn't it enough to mark the constructor parameter as nullable?

src/test/java/org/zalando/nakadi/service/EventStreamTest.java
rcillo: 👍 https://github.com/zalando/nakadi/pull/406/commits/7684ef47c40a0f7103b70d2f1d07583c2d0abbcd

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/StreamBatch.java
rcillo: Isn't it enough to mark the constructor parameter as `@Nullable`?

v-stepanov: I think no. I think we should think from the perspective of the user who uses the class. When he calls `getMetadata` he should be aware that he can get null. (the fact that it is Nullable in constructor will not really help in this situation)

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/StreamBatch.java
v-stepanov: But I mean the implementation, not the API.
It's implemented in a way that it would be empty if there are no events https://github.com/zalando/nakadi/pull/406/files#diff-f5f9dae0f2568532a30e93bda9f32536L28

src/acceptance-test/java/org/zalando/nakadi/domain/StreamMetadata.java
```diff
@@ -0,0 +1,38 @@
+package org.zalando.nakadi.domain;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import javax.annotation.Nullable;
+
+public class StreamMetadata {
+    private String debug;
```
adyach: final ?

rcillo: done https://github.com/zalando/nakadi/pull/406/commits/d042b4d720db3072c10282250330d93886164638

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/StreamBatch.java
adyach: @v-stepanov @rcillo according to the code events can not be null but it is marked as @Nullable. 
Maybe we can fix this divergence?

src/acceptance-test/java/org/zalando/nakadi/domain/StreamMetadata.java
antban: @rcillo I think if we are using this kind of annotation, than it's not enough.
Whenever I call method, I do not think about the way how the value was built.

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/StreamBatch.java
```diff
@@ -26,7 +26,7 @@
 
     public StreamBatch(@JsonProperty("cursor") final SubscriptionCursor cursor,
                        @Nullable @JsonProperty("events") final List<Map> events,
-                       @Nullable @JsonProperty("metadata") final StreamMetadata metadata) {
+                       @Nullable @JsonProperty("info") final StreamMetadata metadata) {
         this.cursor = cursor;
         this.events = Optional.ofNullable(events).orElse(ImmutableList.of());
         this.metadata = metadata;
```
adyach: @rcillo do you consider to rename field as well ?

v-stepanov:
:shipit:

PR #407, opened 2016-09-02
ARUHA-341: move logic from subscription controller to subscription service

adyach:
In the previous commits I've added SubscriptionService.java and was not able to move everything from SubscriptionController.java because there were too much PRs for that one class. I think it will be naturally to move remained logic to service.


src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -1,41 +1,199 @@
...
                                final TopicRepository topicRepository,
                                final EventTypeRepository eventTypeRepository) {
+        this.subscriptionRepository = subscriptionRepository;
         this.zkSubscriptionClientFactory = zkSubscriptionClientFactory;
         this.topicRepository = topicRepository;
         this.eventTypeRepository = eventTypeRepository;
     }
 
-    public List<SubscriptionEventTypeStats> createSubscriptionStat(final Subscription subscription) {
+    public Result<Subscription> createSubscription(final SubscriptionBase subscriptionBase, final Client client)
```
rcillo: Usually a service layer is considered a domain layer. Returning HTTP entities doesn't make sense since it's specific to the outer controller layer. But creating all this separation seems artificial. That's why I don't care of having code in the controller.

adyach: @rcillo true, but we do not use internal entities yet.

src/main/java/org/zalando/nakadi/service/WrappableResult.java
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.context.request.NativeWebRequest;
+import org.zalando.problem.spring.web.advice.Responses;
+
+import static org.springframework.http.HttpStatus.OK;
+import static org.springframework.http.ResponseEntity.status;
+
+public class WrappableResult {
+
+    public interface Wrappable {
```
v1ctor: Supplier

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -1,41 +1,199 @@
...
+        } catch (final NoSuchSubscriptionException e) {
+            LOG.debug("Failed to find subscription: {}", subscriptionId, e);
+            return Result.problem(e.asProblem());
+        } catch (final ServiceUnavailableException e) {
+            LOG.error("Error occurred when trying to get subscription: {}", subscriptionId, e);
+            return Result.problem(e.asProblem());
+        }
+    }
+
+    public Result<ItemsWrapper<List<SubscriptionEventTypeStats>>> getSubscriptionStat(final String subscriptionId) {
```
v-stepanov: Hmm...
Should not it be `Result<ItemsWrapper<SubscriptionEventTypeStats>>` ?

adyach: it must be that!

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -1,41 +1,199 @@
...
+    }
+
+    public UriComponents getSubscriptionUri(final Subscription subscription) {
+        final UriComponents path = SUBSCRIPTION_PATH.buildAndExpand(subscription.getId());
+        return path;
+    }
+
+    public Result listSubscriptions(final String owningApplication, final Set<String> eventTypes, final int limit,
+                                    final int offset) {
+        if(limit< 1||limit >1000) {
```
v-stepanov: code formatting is bad on this line

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -1,41 +1,199 @@
...
+            final Problem problem = Problem.valueOf(Response.Status.BAD_REQUEST,
+                    "'limit' parameter should have value from 1 to 1000");
+            return Result.problem(problem);
+        }
+        if(offset< 0) {
```
v-stepanov: some spaces missing here

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -1,41 +1,199 @@
...
+
+    private Subscription getExistingSubscription(final SubscriptionBase subscriptionBase)
+            throws NoSuchSubscriptionException, InternalNakadiException, ServiceUnavailableException {
+        return subscriptionRepository.getSubscription(
+                subscriptionBase.getOwningApplication(),
+                subscriptionBase.getEventTypes(),
+                subscriptionBase.getConsumerGroup());
+    }
+
+    private String createErrorMessage(final List<String> missingEventTypes) {
```
v-stepanov: I think the method name is too generic.

adyach:
@rcillo @v1ctor @v-stepanov @antban guys, your comments are welcomed!

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -97,30 +62,40 @@ public SubscriptionController(final SubscriptionDbRepository subscriptionReposit
...
-        } catch (final NakadiException e) {
-            LOG.error("Error occurred during subscription creation", e);
-            return create(e.asProblem(), request);
+            return prepareLocationResponse(result.getValue());
+        } catch (DuplicatedSubscriptionException e) {
+            final Result<Subscription> result = subscriptionService.processDuplicatedSubscription(subscriptionBase);
+            if (!result.isSuccessful()) {
+                return Responses.create(result.getProblem(), request);
+            }
+            final Subscription existingSubscription =  result.getValue();
```
v1ctor: two spaces

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -163,85 +117,20 @@ public SubscriptionController(final SubscriptionDbRepository subscriptionReposit
...
-            return status(OK).body(subscription);
-        } catch (final NoSuchSubscriptionException e) {
-            LOG.debug("Failed to find subscription: {}", subscriptionId, e);
-            return create(e.asProblem(), request);
-        } catch (final ServiceUnavailableException e) {
-            LOG.error("Error occurred when trying to get subscription: {}", subscriptionId, e);
-            return create(e.asProblem(), request);
-        }
+
+        return WrappableResult.resultWrapper(() ->
```
v1ctor: for me it's too long. Maybe Result.wrap() ?

src/main/java/org/zalando/nakadi/service/WebResult.java
```diff
@@ -0,0 +1,21 @@
...
+import org.zalando.problem.spring.web.advice.Responses;
+
+import java.util.function.Supplier;
+
+import static org.springframework.http.HttpStatus.OK;
+import static org.springframework.http.ResponseEntity.status;
+
+public class WebResult {
+
+    public static ResponseEntity<?> wrap(final Supplier<Result> supplier, final NativeWebRequest request) {
```
v-stepanov: I don't really like these Result and WebResult.
In more complex cases (when we not just return 200) it will not work and we will anyway have to write code in controller.
And I don't really understand the desire to have 1-liners in controller methods.

v-stepanov:
:shipit:

src/main/java/org/zalando/nakadi/service/WebResult.java
adyach: @v-stepanov agree, i also do not like it much but for now I just wrap the same exceptions in one method it reduces some amount of duplicated code. I think a right way is to use RuntimeException and handle it in @ControllerAdvice. It will give us less duplicated code and have one place where we handle all exceptions.

adyach:
@v-stepanov @rcillo @v1ctor merged with master, reapprove, please

v-stepanov:
:shipit:

PR #409, opened 2016-09-05
ARUHA-336: refactor exception wrapping and validate cursor read scopes

v1ctor:
@adyach @v-stepanov @rcillo @antban 
Tests in progress


src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -31,38 +38,48 @@
...
                                            @RequestBody final List<SubscriptionCursor> cursors,
-                                           final NativeWebRequest request) {
+                                           final NativeWebRequest request,
+                                           final Client client) {
         if (!featureToggleService.isFeatureEnabled(HIGH_LEVEL_API)) {
             return new ResponseEntity<>(NOT_IMPLEMENTED);
         }
 
         try {
+            validateSubscriptionReadScopes(client, subscriptionId);
```
adyach: Why do we have to check read scopes when we commit cursor ?

adyach: Ah, I've got an idea.

src/main/java/org/zalando/nakadi/exceptions/NakadiRuntimeException.java
```diff
@@ -0,0 +1,17 @@
...
+public class NakadiRuntimeException extends RuntimeException {
+
+    private final Exception exception;
+
+    public NakadiRuntimeException(final Exception exception) {
+        super(exception);
+        this.exception = exception;
+    }
+
+    public Exception getException() {
```
adyach: Throwable has getCause() method. Maybe we can use it ?

v1ctor: There are logic in CleanupState which expect Exception.
And I just rename ExceptionWrapper to NakadiRuntimeException. And port Scala Try to be more functioal :)

v-stepanov:
I think I don't understand. Is it only ARUHA-336? Or is it both 336 and 337 ?

src/acceptance-test/java/org/zalando/nakadi/webservice/SubscriptionAT.java
```diff
@@ -146,7 +147,7 @@ public void testOffsetsCommit() throws Exception {
                 "\",\"cursor_token\":\"abc\"}]";
         commitCursors(subscription, cursor)
                 .then()
-                .statusCode(HttpStatus.SC_NO_CONTENT);
+                .statusCode(HttpStatus.SC_OK);
```
v-stepanov: There's a comment few lines above. It should be adjusted according to your changes.

src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -31,44 +39,83 @@
...
 
         try {
-            final boolean allCommitted = cursorsService.commitCursors(subscriptionId, cursors);
-            return allCommitted ? ok().build() : noContent().build();
+            validateSubscriptionReadScopes(client, subscriptionId);
+            final Map<SubscriptionCursor, Boolean> result = cursorsService.commitCursors(subscriptionId, cursors);
+            final List<CursorCommitResult> items = result.entrySet().stream()
+                    .map(entry -> new CursorCommitResult(entry.getKey(), entry.getValue()))
+                    .collect(Collectors.toList());
+            return ok(new ItemsWrapper<>(items));
```
v-stepanov: According to API, it should be 200 only if at least one cursor wasn't successfully committed.
Otherwise it should be 204.

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -34,15 +34,16 @@ public String calculatePartition(final EventType eventType, final JSONObject eve
                     // The problem is that JSONObject doesn't override hashCode(). Therefore convert it to
                     // a string first and then use hashCode()
                     .map(pkf -> EventCategory.DATA.equals(eventType.getCategory()) ? DATA_PATH_PREFIX + pkf : pkf)
-                    .map(wrapFunction(okf -> traversableJsonEvent.get(okf).toString().hashCode()))
+                    .map(Try.wrap(okf -> traversableJsonEvent.get(okf).toString().hashCode()))
```
v-stepanov: Just in few words, what is the advantage of switching to this Try instead of exception wrapper?
Currently I see one disadvantage that the code became more verbose as we always call `Try::getOrThrow`.

v1ctor: I think it more flexible, and give us more opportunities. 
For example we could collect all exceptions instead of first one:
```java
List<Exception> exceptions = callableList.stream().map(callable -> Try.cons(callable)).filter(Try::isFailed).map(Try::getException).collect(Collections.toList());
```

src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -31,44 +39,85 @@
...
-            final boolean allCommitted = cursorsService.commitCursors(subscriptionId, cursors);
-            return allCommitted ? ok().build() : noContent().build();
+            validateSubscriptionReadScopes(client, subscriptionId);
+            final Map<SubscriptionCursor, Boolean> result = cursorsService.commitCursors(subscriptionId, cursors);
+            final List<CursorCommitResult> items = result.entrySet().stream()
+                    .map(entry -> new CursorCommitResult(entry.getKey(), entry.getValue()))
+                    .collect(Collectors.toList());
+            final boolean allCommited = result.values().stream().reduce(Boolean::logicalAnd).orElseGet(() -> true);
+            final ItemsWrapper<CursorCommitResult> body = new ItemsWrapper<>(items);
+            return allCommited ? status(201).body(body) : ok(body);
```
v-stepanov: We always use constants in controller. Why you decided to use the code itself?

v-stepanov: Other two questions
1) Why it's 201? In API it's 204.
2) In a case when all cursors are committed the body should be empty.

v1ctor: Fixed.

src/test/java/org/zalando/nakadi/controller/CursorsControllerTest.java
```diff
@@ -2,37 +2,47 @@
...
-import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
 import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
 import static org.springframework.test.web.servlet.setup.MockMvcBuilders.standaloneSetup;
+import static org.zalando.nakadi.util.FeatureToggleService.Feature.CHECK_PARTITIONS_KEYS;
+import static org.zalando.nakadi.utils.TestUtils.buildDefaultEventType;
 import static org.zalando.problem.MoreStatus.UNPROCESSABLE_ENTITY;
 
 public class CursorsControllerTest {
```
v-stepanov: There is no test for the case when it's partial success and we return list of commit results.

v1ctor: There is acceptance test. Isn't it ok?

v-stepanov: I think we decided to have unit tests everywhere where it is possible. So there should be a good reason for not having unit test. 
And for acceptance tests we only test very simple straightforward scenarios.

v1ctor: I've added test for this case. :)

v-stepanov: I can't find it :)
Have you pushed it ?

v1ctor: Fixed. Pushed to another branch :(

src/test/java/org/zalando/nakadi/controller/CursorsControllerTest.java
```diff
@@ -60,27 +71,40 @@ public CursorsControllerTest() throws Exception {
...
 
     @Test
-    public void whenCommitOldCursorsThenNoContent() throws Exception {
-        when(cursorsService.commitCursors(any(), any())).thenReturn(false);
+    public void whenCommitInvalidCursorsThenOk() throws Exception {
+        when(cursorsService.commitCursors(any(), any()))
+                .thenReturn(Collections.singletonMap(any(), false));
         putCursors(DUMMY_CURSORS)
-                .andExpect(status().isNoContent());
+                .andExpect(status().isOk());
```
v-stepanov: I think it would be interesting to test not only the status code but also the body, as it should contain the list of commit results.

v-stepanov:
:shipit:

PR #410, opened 2016-09-06
aruha-123 fix user journey tests

rcillo:
These tests are manually executed against staging and sandbox
environment after deploying a new release.

Since we toggled this feature on in staging, this test has been broken
due to missing owning application.


v-stepanov:
:shipit:

PR #412, opened 2016-09-09
ARUHA-344: validate stream id when commit cursors

v1ctor:


rcillo:
Looks fine 👍 once tests are fixed I would approve it.

src/acceptance-test/java/org/zalando/nakadi/webservice/SubscriptionAT.java
```diff
@@ -131,9 +132,14 @@ public void testOffsetsCommit() throws Exception {
 
         final Subscription subscription = createSubscriptionForEventType(etName);
 
+        final TestStreamingClient client = TestStreamingClient
+                .create(URL, subscription.getId(), "")
+                .start();
+        Thread.sleep(1000);
```
v-stepanov: We decided not to use this approach with Thread.sleep if possible, because it's unclean and sometimes can lead to very random behavior. 

Instead you can do something like:
`TestUtils.waitFor(() -> assertThat(client.getSessionId(), not(equalTo("UNKNOWN"))));`

So that you will wait for what you really need, not to abstract 1 second.

src/acceptance-test/java/org/zalando/nakadi/webservice/SubscriptionAT.java
```diff
@@ -155,12 +161,18 @@ public void testOffsetsCommit() throws Exception {
...
         final String etName = createEventType().getName();
         final Subscription subscription = createSubscriptionForEventType(etName);
         final String cursor = "[{\"partition\":\"0\",\"offset\":\"25\",\"event_type\":\"" + etName +
                 "\",\"cursor_token\":\"abc\"}]";
-        commitCursors(subscription, cursor)
+
+        final TestStreamingClient client = TestStreamingClient
+                .create(URL, subscription.getId(), "")
+                .start();
+        Thread.sleep(1000);
```
v-stepanov: please see my comment above about Thread.sleep

src/acceptance-test/java/org/zalando/nakadi/webservice/SubscriptionAT.java
```diff
@@ -218,4 +231,11 @@ private EventType createEventType() throws JsonProcessingException {
         return eventType;
     }
 
+    private static void publishEvent(final String eventType, final String event) {
```
v-stepanov: is it used?

v-stepanov:
:shipit:

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -64,30 +61,48 @@ public CursorsService(final ZooKeeperHolder zkHolder,
...
         return cursors.stream().collect(Collectors.toMap(Function.identity(),
                 Try.<SubscriptionCursor, Boolean>wrap(cursor -> processCursor(subscriptionId, cursor))
                         .andThen(Try::getOrThrow)));
     }
 
+    private void validateCursors(final ZkSubscriptionClient subscriptionClient,
+                                 final List<SubscriptionCursor> cursors,
+                                 final String streamId) throws InvalidCursorException
+    {
+        final Map<Partition.PartitionKey, Partition> partitions = Arrays.stream(subscriptionClient.listPartitions())
```
antban: On this step there may be NoNodeException (because inside client there is no check for that)

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -64,30 +61,48 @@ public CursorsService(final ZooKeeperHolder zkHolder,
...
+                .collect(Collectors.toMap(Partition::getKey, Function.identity()));
+        final List<SubscriptionCursor> invalidCursors = cursors.stream().filter(cursor -> Try.cons(() ->
+                eventTypeRepository.findByName(cursor.getEventType())).getO().map(eventType -> {
+            final Partition.PartitionKey key = new Partition.PartitionKey(eventType.getTopic(), cursor.getPartition());
+            final Partition partition = partitions.get(key);
+            return partition == null || !streamId.equals(partition.getSession());
```
antban: If there is no sessions in session list, than partitions will be assigned to last remembered state (which probably is not the expected behavior)

antban: And last thing - calls to listSessions() and listPartitions() should be made under `zkClient.runLocked()`, because they must operate with consistent data.

PR #416, opened 2016-09-13
Aruha 357 fix response code when failed

v-stepanov:
@rcillo @antban @adyach @v1ctor please review

### DIFF
--- a/src/main/java/org/zalando/nakadi/domain/BatchItem.java
+++ b/src/main/java/org/zalando/nakadi/domain/BatchItem.java
@@ -14,8 +14,7 @@ public class BatchItem {
         this.response = new BatchItemResponse();
         this.event = event;
 
-        Optional
-                .ofNullable(event.optJSONObject("metadata"))
+        Optional.ofNullable(event.optJSONObject("metadata"))
                 .map(e -> e.optString("eid", null))
                 .ifPresent(this.response::setEid);
     }
--- a/src/main/java/org/zalando/nakadi/domain/BatchItem.java
+++ b/src/main/java/org/zalando/nakadi/domain/BatchItem.java
@@ -2,6 +2,7 @@ package org.zalando.nakadi.domain;
 
 import org.json.JSONObject;
 
+import javax.annotation.Nullable;
 import java.util.Optional;
 
 public class BatchItem {
@@ -13,8 +14,8 @@ public class BatchItem {
         this.response = new BatchItemResponse();
         this.event = event;
 
-        Optional.
-                ofNullable(event.optJSONObject("metadata"))
+        Optional
+                .ofNullable(event.optJSONObject("metadata"))
                 .map(e -> e.optString("eid", null))
                 .ifPresent(this.response::setEid);
     }
@@ -27,6 +28,7 @@ public class BatchItem {
         this.partition = partition;
     }
 
+    @Nullable
     public String getPartition() {
         return this.partition;
     }
--- a/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -188,26 +188,26 @@ public class KafkaTopicRepository implements TopicRepository {
             final boolean isSuccessful = done.await(kafkaSettings.getKafkaSendTimeoutMs(), TimeUnit.MILLISECONDS);
 
             if (!isSuccessful) {
-                failBatch(batch, "timed out");
+                failUnpublished(batch, "timed out");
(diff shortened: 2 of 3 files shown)


src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -185,19 +188,26 @@ public void syncPostBatch(final String topicId, final List<BatchItem> batch) thr
...
         }
     }
 
-    private void failBatch(final List<BatchItem> batch, final String reason) {
-        for (final BatchItem item : batch) {
-            item.updateStatusAndDetail(EventPublishingStatus.FAILED, reason);
-        }
+    private void failUnpublished(final List<BatchItem> batch, final String reason) {
+        batch.stream()
+                .filter(item -> !(item.getResponse().getPublishingStatus().equals(EventPublishingStatus.SUBMITTED)))
```
adyach: I think it looks cleaner:
item.getResponse().getPublishingStatus() != EventPublishingStatus.SUBMITTED

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -168,6 +169,8 @@ public void syncPostBatch(final String topicId, final List<BatchItem> batch) thr
         final CountDownLatch done = new CountDownLatch(batch.size());
 
         for (final BatchItem item : batch) {
+            Preconditions.checkNotNull(item.getPartition(),
```
rcillo: It should have caused an error during partition phase, which happens previous to this one.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -185,19 +188,26 @@ public void syncPostBatch(final String topicId, final List<BatchItem> batch) thr
...
 
             if (!isSuccessful) {
-                failBatch(batch, "timed out");
+                failUnpublished(batch, "timed out");
                 throw new EventPublishingException("Timeout publishing events");
             }
+
+            final boolean atLeastOneFailed = batch.stream()
+                    .anyMatch(item -> item.getResponse().getPublishingStatus() == EventPublishingStatus.FAILED);
+            if (atLeastOneFailed) {
```
rcillo: There is a scenario when a callback would be invoked concurrently between lines 195 and 197. This is a corner case when a failed event would go unnoticed.

rcillo: This corner case would be removed once we have proper timeout configuration in place. Then, there would be no need for "failUnpublished" as well.

src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -13,8 +14,8 @@ public BatchItem(final JSONObject event) {
         this.response = new BatchItemResponse();
         this.event = event;
 
-        Optional.
-                ofNullable(event.optJSONObject("metadata"))
+        Optional
+                .ofNullable(event.optJSONObject("metadata"))
```
v1ctor: As for me there is enough space to keep it on single line.

rcillo:
Sharing what we discussed moments ago, once we fixed the timeout configuration we should also remove the "failUnpublished" calls, since it would no longer be necessary. That would also remove the concurrency problem.

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
v-stepanov: Fixed. I actually forgot that those are not strings and it's possible to do just !=

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
v-stepanov: Well, I just wanted to be null-safe.
The partitioning phase can be changed in future and the null can appear here.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
v-stepanov: No, there is not corner case here, because here we will come only in a case when isSuccessful == true, which means that all callbacks were called.

src/main/java/org/zalando/nakadi/domain/BatchItem.java
v-stepanov: fixed

rcillo:
:shipit:

PR #420, opened 2016-09-14
ARUHA-334: fix subscription creation

v1ctor:
@rcillo @v-stepanov


v-stepanov:
Looks good for me. But you opened PR against master. Should not it be against release-branch?

PR #421, opened 2016-09-14
ARUHA-334: fix subscription cretion endpoint

v1ctor:


v-stepanov:
:shipit:

v-stepanov:
@v1ctor this requires DB migration script.

PR #426, opened 2016-09-14
ARUHA-365: returned back pseudo random session id generation

v-stepanov:
Since we removed "-Djava.security.egd=file:/dev/./urandom" from dockerfile endtrypoint it sometimes takes a lot of time for Nakadi to start in Docker (at least in my environment):

`INFO: Creation of SecureRandom instance for session ID generation using [SHA1PRNG] took [23,655] milliseconds.`

I suggest to return it back.


adyach:
👍

PR #427, opened 2016-09-14
ARUHA-333: HILA API: create stats endpoint

adyach:


src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -197,9 +199,11 @@ public Result listSubscriptions(final String owningApplication, final Set<String
 
     private List<SubscriptionEventTypeStats> createSubscriptionStat(final Subscription subscription)
             throws ServiceUnavailableException {
+        final String subscriptionId = subscription.getId();
         final ZkSubscriptionClient zkSubscriptionClient =
-                zkSubscriptionClientFactory.createZkSubscriptionClient(subscription.getId());
-        final Partition[] partitions = zkSubscriptionClient.listPartitions();
+                zkSubscriptionClientFactory.createZkSubscriptionClient(subscriptionId);
+        final ZkSubscriptionNode zkSubscriptionNode = zkSubscriptionClient.getZkSubscriptionNode();
```
v1ctor: Dmitry told that we have to lock when requesting data from subscription.

adyach: @v1ctor I am locking it inside this method call https://github.com/zalando/nakadi/pull/427/files/ec2e3f7ad4170463e46576fa2abb1386c3bf6f86#diff-b07971694c861bf3a96d82f76f5459bdR299

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -227,25 +233,39 @@ public Result listSubscriptions(final String owningApplication, final Set<String
...
-                                                                 final TopicPartition topicPartition)
+                                                                 final TopicPartition topicPartition,
+                                                                 final boolean hasSessions)
             throws NakadiException {
         final String partitionName = partition.getKey().getPartition();
-        final String partitionState = partition.getState().getDescription();
-        final String partitionSession = partition.getSession();
+        final String partitionState;
+        final String partitionSession;
+        if (hasSessions){
```
v1ctor: space

src/main/java/org/zalando/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
```diff
@@ -290,4 +291,23 @@ public void transfer(final String sessionId, final Collection<Partition.Partitio
             throw new NakadiRuntimeException(e);
         }
     }
+
+    @Override
+    public ZkSubscriptionNode getZkSubscriptionNode() {
```
rcillo: The name of the method is `get` but if I understood it correctly it updates the node. Maybe it should be renamed or operations should be separated between side effects and no side effects. Wdyt?

rcillo: Now I got it. It's just build the object.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -227,25 +233,39 @@ public Result listSubscriptions(final String owningApplication, final Set<String
...
+            partitionSession = partition.getSession();
+        } else {
+            partitionState = Partition.State.UNASSIGNED.getDescription();
+            partitionSession = "";
+        }
+
         final long clientOffset = zkSubscriptionClient.getOffset(partition.getKey());
-        final long total = Long.valueOf(topicPartition.getNewestAvailableOffset());
-        final long unconsumedEvents = total - clientOffset;
+        Long unconsumedEvents = null;
```
v-stepanov: with this you break null handling policy as you pass this to constructor where it is not marked as Nullable

rcillo: Shouldn't it be zero?

adyach: @v-stepanov I will fix it.
@rcillo According to API https://github.com/zalando/nakadi/blob/nakadi-jvm/api/nakadi-event-bus-api.yaml#L1523 we have to hide this field when it was not read yet.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -214,10 +218,12 @@ public Result listSubscriptions(final String owningApplication, final Set<String
 
         return eventTypes.stream()
                 .map(eventType -> {
-                    final Set<SubscriptionEventTypeStats.Partition> statPartitions = Arrays.stream(partitions)
+                    final Set<SubscriptionEventTypeStats.Partition> statPartitions =
+                            Arrays.stream(zkSubscriptionNode.getPartitions())
+                            .sorted((p1, p2) -> p1.getKey().getPartition().compareTo(p2.getKey().getPartition()))
```
v-stepanov: These will be sorted as strings. So partition 10 will come before partition 2.

src/main/java/org/zalando/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
```diff
@@ -290,4 +291,23 @@ public void transfer(final String sessionId, final Collection<Partition.Partitio
...
+            runLocked(() -> {
+                subscriptionNode.setPartitions(listPartitions());
+                subscriptionNode.setSessions(listSessions());
+            });
+        } catch (final NakadiRuntimeException nre) {
+            final Exception cause = nre.getException();
+            if (!(cause instanceof KeeperException.NoNodeException)) {
+                throw new NakadiRuntimeException(cause);
+            }
+            log.info("No data about provided subscription {} in ZK", subscriptionId);
```
v-stepanov: I don't exactly understand what we will show to user in this case, because ZkSubscriptionNode has empty array of partitions in this case.

adyach: something like that:
`{
  "items": [
    {
      "event_type": "test.ARUHA_NAKADI",
      "partitions": []
    }
  ]
}`

v-stepanov: I think for user it sounds like "event type doesn't have partitions"
I think it's better just to return some specific status code in this situation and an empty body

adyach:
@rcillo @v-stepanov @v1ctor @antban thank you for comments

src/main/java/org/zalando/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
v-stepanov: Or maybe it would be better just to grab list of partitions from Kafka in this case, because displaying empty list of partitions is very unnatural

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
rcillo: Just reminded we did not padded it... it almost hurt ourselves. Added a ticket to remind.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -49,6 +52,8 @@
 
     private static final Logger LOG = LoggerFactory.getLogger(SubscriptionService.class);
     private static final UriComponentsBuilder SUBSCRIPTION_PATH = UriComponentsBuilder.fromPath("/subscriptions/{id}");
+    private static final Comparator<? super Partition> PARTITIONS_COMPARATOR = (p1, p2) ->
+            Integer.valueOf(p1.getKey().getPartition()).compareTo(Integer.valueOf(p2.getKey().getPartition()));
```
antban: Comparator.<Partition, Integer>comparing (p-> Integer.valueOf(p.getKey().getPartition()))

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -227,25 +235,39 @@ public Result listSubscriptions(final String owningApplication, final Set<String
     private Stream<SubscriptionEventTypeStats.Partition> filterTopicPartitions(
```
antban: is it really filter ? For me this method is more like mapToPartitionView

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -227,25 +235,39 @@ public Result listSubscriptions(final String owningApplication, final Set<String
...
+        final String partitionSession;
+        if (hasSessions) {
+            partitionState = partition.getState().getDescription();
+            partitionSession = partition.getSession();
+        } else {
+            partitionState = Partition.State.UNASSIGNED.getDescription();
+            partitionSession = "";
+        }
+
         final long clientOffset = zkSubscriptionClient.getOffset(partition.getKey());
```
antban: Will throw NakadiRuntimeException in case if subscription is not started.

src/main/java/org/zalando/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
antban: For me it makes sense to return Optional<ZkSubscriptionNode>, return empty() if there is NoNodeException and rethrow in all other cases.

And in `createSubscriptionStat` method the base for result must be received from topicrepository, while ZkSubscriptionNode must be used only as supplementary node to get more information

src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionClient.java
```diff
@@ -128,4 +128,6 @@
      * @param partitions topic ids and partition ids of transferred data.
      */
     void transfer(String sessionId, Collection<Partition.PartitionKey> partitions);
+
+    ZkSubscriptionNode getZkSubscriptionNode();
```
antban: Please, add `Locked` postfix, if it is locked.

adyach:
@rcillo @v-stepanov @v1ctor @antban fixed comments, review, please

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -214,39 +217,54 @@ public Result listSubscriptions(final String owningApplication, final Set<String
...
+        final Partition partition = Arrays.stream(zkSubscriptionNode.getPartitions())
+                .filter(p -> p.getKey().getPartition().equals(topicPartition.getPartitionId()))
+                .findFirst()
+                .orElse(null);
+
+        return createPartition(zkSubscriptionClient, partition, topicPartition, hasSessions);
     }
 
     private SubscriptionEventTypeStats.Partition createPartition(final ZkSubscriptionClient zkSubscriptionClient,
                                                                  final Partition partition,
```
v1ctor: @Nullable ?

src/main/java/org/zalando/nakadi/service/subscription/zk/CuratorZkSubscriptionClient.java
adyach: @v-stepanov fixed 6ddba11

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
adyach: @v1ctor Yes, it will be null, should I mark it as @Nullable? I can not remember that rule.

antban:
:+1:

v-stepanov:
:shipit:

antban:
:+1:

PR #430, opened 2016-09-16
ARUHA-344: add migration

v1ctor:


rcillo:
I see that there are some other files in this directory. These migrations were already executed but they were not placed under their release folder.

The way it is today, we risk running then again. So I would suggest to organize the folders to make it clear that we should run only this newly added file.

database/migration/aruha-334-rename-subscription-fields.sql
```diff
@@ -0,0 +1,7 @@
+UPDATE zn_data.subscription
+SET s_subscription_object = jsonb_set(s_subscription_object, '{consumer_group}', '"default"')
```
adyach: Are you sure we can use it for 9.4? I could not managed to find it https://www.postgresql.org/docs/9.4/static/functions-json.html

v-stepanov:
:shipit:

PR #436, opened 2016-09-20
Aruha 370 fix commit endpoint and api

v-stepanov:
@adyach @v1ctor @rcillo @antban 
please review

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -568,28 +568,25 @@ paths:
           schema:
             $ref: '#/definitions/Subscription'
           headers:
-             Location:
-               description: |
-                 The absolute URI for this subscription resource.
-               type: string
-               example: 'https://example.org/api/subscriptions/105a76d8-db49-4144-ace7-e683e8f4ba46'
+            Location:
+              description: |
+                The relative URI for this subscription resource.
+              type: string
         '201':
           description: Subscription was successfuly created. Returns subscription object that was created.
           schema:
             $ref: '#/definitions/Subscription'
           headers:
-             Location:
-               description: |
-                 The absolute URI for the created resource.
-               type: string
-               example: 'https://example.org/api/subscriptions/105a76d8-db49-4144-ace7-e683e8f4ba46'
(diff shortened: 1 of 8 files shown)


v1ctor:
👍

src/main/java/org/zalando/nakadi/domain/SubscriptionCursor.java
```diff
@@ -40,8 +43,8 @@ public boolean equals(final Object o) {
     @Override
     public int hashCode() {
         int result = super.hashCode();
-        result = 31 * result + eventType.hashCode();
-        result = 31 * result + cursorToken.hashCode();
+        result = 31 * result + (eventType != null ? eventType.hashCode() : 0);
```
adyach: if you check it for null here why don't you check it in `equals` method?

v-stepanov: I check it for null here only because the validation for some reason calls hashCode before it check properties for null.

adyach: @v-stepanov understood, I think it is better to have equals and hashcode methods consistent.

v-stepanov: This check for null in hashCode is a hack for validation to work. If I put that check to equals as well - then why shouldn't I put the null check to all other places?
That's why I don't want to put this check to any other places.

adyach:
:shipit:

src/main/java/org/zalando/nakadi/domain/SubscriptionCursor.java
v-stepanov: @adyach I added comment in hashCode method as we discussed

api/nakadi-event-bus-api.yaml
```diff
@@ -773,8 +770,11 @@ paths:
             $ref: '#/definitions/Problem'
         '429':
           description: Too many requests
-          schema:
```
rcillo: Do you think it would be fine to add it later?

PR #437, opened 2016-09-20
fix typo in default statistics (#411)

### LINKED ISSUE #411: Field default_statistics: API description does not match the validation procedure
When creating an event, according to the API specification there can be a field "default_statistics". When that field is provided, you will get the following error message:

The error response says, it should be "default_statistic"

`{"type":"http://httpstatus.es/400","title":"Bad Request","status":400,"detail":"Unrecognized field \"default_statistics\" (class org.zalando.nakadi.domain.EventType), not marked as ignorable (11 known properties: \"schema\", \"default_statistic\", \"partition_strategy\", \"name\", \"owning_application\", \"category\", \"write_scopes\", \"read_scopes\", \"enrichment_strategies\", \"partition_key_fields\", \"options\"])\n at [Source: java.io.PushbackInputStream@4c7c16dd; line: 1, column: 1053] (through reference chain: org.zalando.nakadi.domain.EventType[\"default_statistics\"])"}`

jmhofer:
This is an attempt to fix #411.

As I assume it should be "default_statistics" (swagger is correct), I tried to fix the code accordingly.


v-stepanov:
@jmhofer unfortunately it can't be solved in that way. Because it will cause a short downtime in Nakadi when switching versions and migrating DB. 
We need 2-phase migration here:
current version that works with default_statistic --> intermediate version that works with both default_statistics and default_statistic --> final version that works with default_statistics.

jmhofer:
I see...

rcillo:
We have done it in the past https://github.com/zalando/nakadi/pull/271 and https://github.com/zalando/nakadi/pull/337/files but then it was reverted when we realised what @v-stepanov just mentioned.

jmhofer:
was about to fix this, now you closed it?

rcillo:
@jmhofer sorry, I understood it would be a different PR since the approach is different. I can reopen it, no problem.

jmhofer:
So, this is not very pretty, but it takes both "default_statistics" and "default_statistic" now (it prefers "default_statistics" internally).

adyach:
@rcillo for me this one looks good! just to be clear the migration plan is:
- release this PR change first
- update database default statistics field using https://github.com/zalando/nakadi/pull/337/files
- revert this PR change

jmhofer:
Merged (it was a very simple merge), so this PR should be up-to-date again with your master.

adyach:
@jmhofer however, there could be some deeper issues regarding the wrong naming at all as well as there are users who use 'default_statistic' naming, it stops us to merge this PR.

jmhofer:
@adyach Sorry, I don't really understand what you mean by that.

adyach:
@jmhofer there is a possibility that some users already use this naming ('default_statistic'). Also name 'default_statistic' itself is not right for what it does.

jmhofer:
@adyach With this PR anyone who wants to can still keep using `default_statistic`, though.
However, I see your point about the general naming. "default statistics" is not very accurate, probably.

adyach:
@jmhofer with this PR yes, but it also won't be correct to keep both parameters' names for the same functionality and at some point we will have to remove it. I will close this PR for now.

PR #438, opened 2016-09-20
ARUHA-369: Fixed handling of first offset when partition doesn't have…

v-stepanov:
@antban @adyach @v1ctor @rcillo please review


antban:
:+1:

PR #443, opened 2016-09-26
ARUHA-349 Recreate producer in specific cases

antban:


src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -8,22 +15,73 @@
...
+        } finally {
+            rwLock.readLock().unlock();
+        }
+        rwLock.writeLock().lock();
+        try {
+            if (null != activeProducer) {
+                useCount.get(activeProducer).incrementAndGet();
+                return activeProducer;
+            }
+            activeProducer = new KafkaProducer<>(kafkaLocationManager.getKafkaProducerProperties());
```
rcillo: It would make life easier if we logged something when a new broker is created and also when it's deleted in line 70.

src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -29,18 +28,22 @@ public void setPartition(final String partition) {
 
     @Nullable
     public String getPartition() {
-        return this.partition;
+        return partition;
```
adyach: Why did you remove 'this' here but left it for other methods?

v-stepanov:
I wanted to take a look, so I will put a block for now :)

v-stepanov:
👎

v-stepanov:
I think there should be tests for new functionality

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -8,22 +15,73 @@
...
+            useCount.put(activeProducer, new AtomicInteger(1));
+            return activeProducer;
+        } finally {
+            rwLock.writeLock().unlock();
+        }
+    }
+
+    public void releaseProducer(final Producer<String, String> producer) {
+        final int newUseCount = useCount.get(producer).decrementAndGet();
+        assert newUseCount >= 0;
```
adyach: Usually, assertions are used during development. Maybe you will remove it or replace with throwing exception?

antban: I wll just use different if condition

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
antban: Makes sence

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -8,22 +15,73 @@
...
-
 @Component
 @Profile("!test")
 public class KafkaFactory {
+
     private final KafkaLocationManager kafkaLocationManager;
-    private final KafkaProducer<String, String> kafkaProducer;
+    @Nullable
+    private Producer<String, String> activeProducer;
+    private final Map<Producer<String, String>, AtomicInteger> useCount = new ConcurrentHashMap<>();
```
adyach: AFAIU, it is possible to replace Map with AtomicInteger because you use only one producer all the time. What do you think ?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -8,22 +15,73 @@
...
+            activeProducer = new KafkaProducer<>(kafkaLocationManager.getKafkaProducerProperties());
+            useCount.put(activeProducer, new AtomicInteger(1));
+            return activeProducer;
+        } finally {
+            rwLock.writeLock().unlock();
+        }
+    }
+
+    public void releaseProducer(final Producer<String, String> producer) {
+        final int newUseCount = useCount.get(producer).decrementAndGet();
```
adyach: This value can inconsistence in a way if `takeProducer` is be called after `releaseProducer`.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -164,43 +165,89 @@ public boolean partitionExists(final String topic, final String partition) throw
...
-                    .anyMatch(item -> item.getResponse().getPublishingStatus() == EventPublishingStatus.FAILED);
-            if (atLeastOneFailed) {
-                failUnpublished(batch, "internal error");
-                throw new EventPublishingException("Error publishing message to kafka");
+            final CompletableFuture<Void> multiFuture = CompletableFuture.allOf(
+                    sendFutures.values().toArray(new CompletableFuture<?>[sendFutures.size()]));
+            multiFuture.get(createSendTimeout(), TimeUnit.MILLISECONDS);
+
+            // Now lets check for errors
+            final boolean needReset = sendFutures.entrySet().stream().filter(
```
adyach: For me this one is cleaner:
            `sendFutures.entrySet().stream().filter(
                    entry -> isExceptionShouldLeadToReset(entry.getValue().getNow(null)))
                    .findAny()
                    .ifPresent(entry -> kafkaFactory.terminateProducer(producer));`

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -164,43 +165,89 @@ public boolean partitionExists(final String topic, final String partition) throw
...
+            item.updateStatusAndDetail(EventPublishingStatus.FAILED, "internal error");
+            throw new EventPublishingException("Error publishing message to kafka", e);
+        } catch (final RuntimeException e) {
+            item.updateStatusAndDetail(EventPublishingStatus.FAILED, "internal error");
+            throw new EventPublishingException("Error publishing message to kafka", e);
+        }
+    }
 
-            try {
-                kafkaProducer.send(record, kafkaSendCallback(item, done));
```
adyach: Unused method in the code detected `kafkaSendCallback`

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -164,43 +165,89 @@ public boolean partitionExists(final String topic, final String partition) throw
...
-            } catch (InterruptException | SerializationException | BufferExhaustedException e) {
-                item.updateStatusAndDetail(EventPublishingStatus.FAILED, "internal error");
-                throw new EventPublishingException("Error publishing message to kafka", e);
-            }
+    private static boolean isExceptionShouldLeadToReset(
+            @Nullable final Exception exception) {
```
adyach: Why not on the same line?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -164,43 +165,89 @@ public boolean partitionExists(final String topic, final String partition) throw
...
+                    item.updateStatusAndDetail(EventPublishingStatus.FAILED, "internal error");
+                    result.complete(exception);
+                } else {
+                    item.setStep(EventPublishingStep.PUBLISHED);
+                    item.updateStatusAndDetail(EventPublishingStatus.SUBMITTED, "");
+                    result.complete(null);
+                }
+            }));
+            return result;
+        } catch (final InterruptException e) {
```
adyach: Could you tell me who throws the exception in this method?

src/main/java/org/zalando/nakadi/domain/BatchItem.java
antban: @adyach Because it was the only `this` usage

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
antban: No, because of several reasons - Producers are taken and released under read lock, which means, that they are accessed from multiple threads simultaneously. That means, that it is impossible to use useCount variable without additional locking (that is why AtomicInteger)
`put` call to useCount map by itself is made with writeLock, while getting data is made without lock at all, and removing from this map is made with readLock. That means, that map by itself should be somehow synchronized (that is why `ConcurrentHashMap`)

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
antban: `multifuture.get()`

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
antban: Added javadocs: https://github.com/zalando/nakadi/pull/443/commits/a19fe4943cad21e3c7b8b9fa2fe506cee4428766

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
antban: Added logging https://github.com/zalando/nakadi/pull/443/commits/a19fe4943cad21e3c7b8b9fa2fe506cee4428766

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
antban: Well, for me this step is too important to hide it somewhere in call chain

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
antban: Fixed https://github.com/zalando/nakadi/pull/443/commits/a19fe4943cad21e3c7b8b9fa2fe506cee4428766

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
antban: Fixed: https://github.com/zalando/nakadi/pull/443/commits/a19fe4943cad21e3c7b8b9fa2fe506cee4428766

src/main/java/org/zalando/nakadi/domain/BatchItem.java
adyach: @antban I did not get it. you still use `this` in `updateStatusAndDetail`.

src/main/java/org/zalando/nakadi/domain/EventPublishingStep.java
```diff
@@ -1,5 +1,11 @@
...
 
 public enum EventPublishingStep {
-    NONE, VALIDATING, ENRICHING, PARTITIONING, PUBLISHING
+    NONE,
+    VALIDATING,
+    ENRICHING,
+    PARTITIONING,
+    PUBLISHING,
+    PUBLISHED,
+    FAILED_PUBLISH
```
v-stepanov: We can't change this. This can break a lot of clients which expect only the specific values. This will be backward-incompatible change.
https://github.com/zalando/nakadi/blob/master/api/nakadi-event-bus-api.yaml#L1597

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -1,29 +1,132 @@
...
+                useCount.get(activeProducer).incrementAndGet();
+                return activeProducer;
+            } else if (canCreate) {
+                activeProducer = new KafkaProducer<>(kafkaLocationManager.getKafkaProducerProperties());
+                useCountMetric.inc();
+                useCount.put(activeProducer, new AtomicInteger(1));
+                LOG.info("New producer instance created: " + activeProducer);
+                return activeProducer;
+            } else {
+                return null;
```
v-stepanov: I don't understand this branch. Why just not to create it in this case? Because this is exactly what you do outside of this method. You call takeUnderLock(false) and if it returns null you again call takeUnderLock(true).

src/main/java/org/zalando/nakadi/domain/BatchItem.java
antban: Fixed https://github.com/zalando/nakadi/pull/443/commits/f793a564f0541cba936210948d5d2707cb1e474b

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
antban: And fixed implementation: https://github.com/zalando/nakadi/pull/443/commits/f793a564f0541cba936210948d5d2707cb1e474b

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
antban: ok, it is not there, it is Exception from kafka, which they rethrow without interrupting current thread back.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
antban: Because in first case it takes readLock (e.g. many threads can call this method at the same time). Second call is under write lock, which means, that only one thread is creating this producer. Thus, I guarantee, that I will now spawn several producers at the same time

src/main/java/org/zalando/nakadi/domain/EventPublishingStep.java
antban: Wow! Fixed https://github.com/zalando/nakadi/pull/443/commits/f7f405d972ece4b7cd56a77f56df44a08ee3007e

antban:
@v-stepanov Tests are added, please, revote :)

src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -29,19 +28,23 @@ public void setPartition(final String partition) {
...
+    }
+
+    public EventPublishingStep getStep() {
+        return response.getStep();
     }
 
-    public synchronized void updateStatusAndDetail(final EventPublishingStatus publishingStatus, final String detail) {
-        this.response.setPublishingStatus(publishingStatus);
-        this.response.setDetail(detail);
+    public void updateStatusAndDetail(final EventPublishingStatus publishingStatus, final String detail) {
```
rcillo: There is no need for this combined update status and details anymore. It used to be like that due to a race condition protected by synchronized.

src/main/java/org/zalando/nakadi/domain/EventPublishingStep.java
```diff
@@ -1,5 +1,9 @@
 package org.zalando.nakadi.domain;
 
 public enum EventPublishingStep {
-    NONE, VALIDATING, ENRICHING, PARTITIONING, PUBLISHING
+    NONE,
+    VALIDATING,
+    ENRICHING,
+    PARTITIONING,
+    PUBLISHING,
```
adyach: ,

adyach:
For me it looks good 👍

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -164,43 +164,89 @@ public boolean partitionExists(final String topic, final String partition) throw
...
+
+            // Now lets check for errors
+            final Optional<Exception> needReset = sendFutures.entrySet().stream()
+                    .filter(entry -> isExceptionShouldLeadToReset(entry.getValue().getNow(null)))
+                    .map(entry -> entry.getValue().getNow(null))
+                    .findAny();
+            if (needReset.isPresent()) {
+                LOG.info("Terminating producer while publishing to topic " + topicId +
+                        " because of unrecoverable exception", needReset.get());
+                kafkaFactory.terminateProducer(producer);
```
v-stepanov: Wasn't the initial plan to retry publishing after the producer is recreated?

v-stepanov:
:shipit:

PR #447, opened 2016-09-28
ARUHA-355: Block abusive producers/consumers

adyach:


src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -135,6 +143,13 @@ public StreamingResponseBody streamEvents(
             SubscriptionStreamer streamer = null;
             final SubscriptionOutputImpl output = new SubscriptionOutputImpl(response, outputStream);
             try {
+                if  (subscriptionDbRepository.getSubscription(subscriptionId).getEventTypes().stream()
+                        .map(etName -> floodService.isConsumptionBlocked(etName)).findFirst().orElse(false)) {
+                    response.setStatus(HttpStatus.TOO_MANY_REQUESTS.value());
```
dehora: It look like if the floodService throws exceptions or fails internally, access will "fail closed" and the server will throw 500s; is that the intent?

src/main/java/org/zalando/nakadi/service/FloodService.java
```diff
@@ -0,0 +1,177 @@
...
+
+    public String getRetryAfterStr() {
+        return String.valueOf(RETRY_AFTER_SEC);
+    }
+
+    public void blockFlooder(final Flooder flooder) throws RuntimeException {
+        try {
+            zooKeeperHolder.get().create().creatingParentsIfNeeded().forPath(createFlooderPath(flooder));
+        } catch (final KeeperException.NoNodeException nne) {
+            LOG.debug("Flooder does not exist {}", flooder.getName());
```
dehora: Wondering why is it ok for the this exception to be swallowed, and not thrown.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -92,6 +95,12 @@ public StreamingResponseBody streamEvents(
 
         return outputStream -> {
 
+            if  (floodService.isConsumptionBlocked(eventTypeName)) {
```
adyach: Yes, I think so. There is a possibility to block by event type or owning application.

src/main/java/org/zalando/nakadi/controller/NakadiController.java
```diff
@@ -0,0 +1,74 @@
...
+import org.zalando.nakadi.domain.ItemsWrapper;
+import org.zalando.nakadi.security.Client;
+import org.zalando.nakadi.service.FloodService;
+import org.zalando.nakadi.util.FeatureToggleService;
+
+import static org.zalando.nakadi.security.Client.FULL_ACCESS;
+
+@RestController
+@RequestMapping(value = "/nakadi")
+public class NakadiController {
```
adyach: Implementation supports access only for specified oauth uid, so called application id. https://github.com/zalando/nakadi/blob/master/src/main/resources/application.yml#L59
Only the owning team is allowed to request ouath token for that application.

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
adyach: Generally speaking, when I thought about situations in which the exceptions are thrown are critical for the whole nakadi component. Maybe it will be a good approach just to swallow all exceptions within the service level.

src/main/java/org/zalando/nakadi/service/FloodService.java
```diff
@@ -0,0 +1,177 @@
...
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+@Component
+public class FloodService {
```
adyach: I agree I think it could be even inappropriate to use zk for such loads but this task is about to have a way to block an aggressive producer/consumer constantly until team will decide to unblock it. This feature was introduced due to the last experiences on live as a fast solution.

dehora: >fast solution.

Got it. +1 on the basis this will get revisited.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
dehora: 🆗

src/main/java/org/zalando/nakadi/service/FloodService.java
adyach: I believe this is a product of testing. I will remove it. Thanks!

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
dehora: This is a hard one. Fail closed means problems with the rate limiter will stop good writes from succeeding; fail open means bad actors could succeed in taking you and the rest of your customers out. As per `due to the last experiences on live as a fast solution` you could ship this now, but it's really really worth a think-through of the consequences pretty soon (a classic cascading failure mode is suppressing errors). So maybe :shipit: but revisit soon.

src/main/java/org/zalando/nakadi/controller/NakadiController.java
dehora: >Implementation supports access only for specified oauth uid, so called application id. https://github.com/zalando/nakadi/blob/master/src/main/resources/application.yml#L59
Only the owning team is allowed to request ouath token for that application.

That sounds la Nakadi user (producer/consumer) can't ever use these APIs and only an operator can. If so, :ok:

adyach: Yes, it was developed to be managed only by a Nakadi owning team.

src/main/java/org/zalando/nakadi/service/FloodService.java
adyach: > Got it. +1 on the basis this will get revisited.

Do you think it is really harmful to use Curator's TreeCache for a big number of read operations? 
We will write directly to zk in very rare cases when we really need to block/unblock a particular client.

dehora: > Do you think it is really harmful to use Curator's TreeCache for a big number of read operations? 

My observation is that extra load is being potentially being placed into a ZooKeeper that exist to run Kafka. It isn't just this feature, it's also being used iirc to store offsets for the hila (an approach modern Kafka has deliberately moved away from, by moving offsets into a Kafka topic). When running Kafka clusters at any kind of load of load or scale, the default operational approach is to give the Zk cluster exclusively to the Kafka system.

Put another way - if rate limiting was required by a service, would a ZooKeeper cluster be the automatic choice, or is it being used because it happens to be there and close to hand? 

Again - I do understand the urgency. I want to be sure a larger problem isn't being created for the future. So maybe this can ship, but there should be ongoing thinking whether operational stability is ultimately achieved in the mid/long term by placing multiple orthogonal workloads into ZooKeeper.

src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
```diff
@@ -54,6 +54,7 @@
     private static final int KAFKA_REQUEST_TIMEOUT = 30000;
     private static final int KAFKA_BATCH_SIZE = 1048576;
     private static final long KAFKA_LINGER_MS = 0;
+    public static final long RETRY_AFTER = 300;
```
antban: What is the measurement unit?

src/main/java/org/zalando/nakadi/controller/NakadiController.java
```diff
@@ -0,0 +1,74 @@
...
+    @Autowired
+    public NakadiController(final FloodService floodService,
+                            final FeatureToggleService featureToggleService) {
+        this.floodService = floodService;
+        this.featureToggleService = featureToggleService;
+    }
+
+    @RequestMapping(path = "/flooders", method = RequestMethod.GET)
+    public ResponseEntity<?> getFlooders(final Client client) {
+        if (client != FULL_ACCESS) {
```
v-stepanov: I think this is not a correct check. As I see in `ClientResolver`, one of the use-cases for getting FULL_ACCESS is when feature CHECK_APPLICATION_LEVEL_PERMISSIONS is not enabled. So when we deploy your code on live it will be possible for everyone to call these endpoints.

src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
adyach: @antban this is seconds. fixed 34338d5

src/main/java/org/zalando/nakadi/controller/NakadiController.java
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.web.bind.annotation.RestController;
+import org.zalando.nakadi.domain.ItemsWrapper;
+import org.zalando.nakadi.security.Client;
+import org.zalando.nakadi.service.FloodService;
+import org.zalando.nakadi.util.FeatureToggleService;
+
+import static org.zalando.nakadi.security.Client.FULL_ACCESS;
+
+@RestController
+@RequestMapping(value = "/nakadi")
```
v-stepanov: why `/nakadi` ?
We already have "nakadi" in our URL in the very beginning. I think it should be more meaningful. `/nakadi` means everything and nothing at the same time.

adyach: @v-stepanov what is your suggestions?

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
v-stepanov: @adyach I can't really find where you block by application. Or am I missing something?

src/main/java/org/zalando/nakadi/service/FloodService.java
```diff
@@ -0,0 +1,205 @@
...
+            return isBlocked(path, eventType.getOwningApplication());
+        } catch (final NakadiException ne) {
+            LOG.error(ne.getMessage(), ne);
+        }
+        return false;
+    }
+
+    private boolean isBlocked(final String path, final String name) {
+        try {
+            return floodersCache.getCurrentData(path + "/" + name) == null ? false : true;
```
v-stepanov: it can be just `return floodersCache.getCurrentData(path + "/" + name) != null`

src/main/resources/application.yml
```diff
@@ -49,14 +49,11 @@ nakadi:
...
     brokers: 127.0.0.1:2181
-#    exhibitor:
-#      brokers: 127.0.0.1,127.0.0.2
-#      port: 8181
     sessionTimeoutMs: 30000
     connectionTimeoutMs: 10000
   oauth2:
     mode: BASIC
-    adminClientId: nakadi
+    adminClientId: stups_nakadi
```
v-stepanov: I think we should not use real value here. We can set some dummy id here and then override it in private configuration.

adyach: @v-stepanov fixed b23babc

src/main/java/org/zalando/nakadi/service/FloodService.java
adyach: @v-stepanov fixed b23babc, thanks!

src/main/java/org/zalando/nakadi/controller/NakadiController.java
adyach: @v-stepanov i do not the logic behind that but it seems like it is simpler to have check for feature in one place. i chose the easiest way to fix it for my implementation. maybe we can refactor it in another PR. b23babc

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
adyach: @v-stepanov d6297cf

src/main/java/org/zalando/nakadi/controller/NakadiController.java
v-stepanov: my suggestion is `/settings` or `/configuration`

src/acceptance-test/java/org/zalando/nakadi/webservice/EventPublishingAT.java
```diff
@@ -0,0 +1,49 @@
...
+public class EventPublishingAT extends BaseAT {
+
+    @Test
+    public void whenPublishingToBlockedEventTypeThen429() throws IOException {
+        final EventType eventType = NakadiTestUtils.createEventType();
+
+        given()
+                .body("[{\"blah\":\"bloh\"}]")
+                .contentType(ContentType.JSON)
+                .post(MessageFormat.format("/event-types/{0}/events", eventType.getName()))
```
v-stepanov: I think you can extract above four lines to a separate method. And then use it in all 3 places.

src/acceptance-test/java/org/zalando/nakadi/webservice/EventPublishingAT.java
```diff
@@ -0,0 +1,49 @@
...
+import org.zalando.nakadi.service.FloodService;
+import org.zalando.nakadi.webservice.utils.NakadiTestUtils;
+import org.zalando.problem.MoreStatus;
+
+import java.io.IOException;
+import java.text.MessageFormat;
+
+import static com.jayway.restassured.RestAssured.given;
+
+public class EventPublishingAT extends BaseAT {
```
v-stepanov: I think it's more like BlockPublishingBAT

src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
```diff
@@ -301,6 +302,68 @@ public void whenInvalidCursorsThenPreconditionFailed() {
...
+                .param("stream_timeout", "2")
+                .param("batch_flush_timeout", "2")
+                .when()
+                .get(STREAM_ENDPOINT)
+                .then()
+                .statusCode(HttpStatus.OK.value());
+    }
+
+    @Test(timeout = 10000)
+    public void whenReadEventsConsumerIsBlocked() throws Exception {
```
v-stepanov: How this test works? What exactly it tests?

src/acceptance-test/java/org/zalando/nakadi/webservice/NakadiControllerAT.java
```diff
@@ -0,0 +1,114 @@
...
+                .get(FLOODERS_URL)
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .content(JSON_HELPER.matchesObject(new HashedMap() {
+                    {
+                        put("consumers", new HashMap<String, Set<String>>() {{
+                            put("event_types", Collections.singleton(eventType.getName()));
+                            put("apps", Collections.emptySet());
+                        }});
+                        put("producers", new HashMap<String, Set<String>>() {{
```
v-stepanov: `{{`

src/acceptance-test/java/org/zalando/nakadi/webservice/NakadiControllerAT.java
```diff
@@ -0,0 +1,114 @@
...
+                new FloodService.Flooder(eventType.getName(), FloodService.Type.CONSUMER_ET);
+        blockFlooder(flooder);
+
+        given()
+                .contentType(ContentType.JSON)
+                .get(FLOODERS_URL)
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .content(JSON_HELPER.matchesObject(new HashedMap() {
+                    {
```
v-stepanov: `{{`

src/acceptance-test/java/org/zalando/nakadi/webservice/NakadiControllerAT.java
```diff
@@ -0,0 +1,114 @@
...
+                        put("consumers", new HashMap<String, Set<String>>() {{
```
v-stepanov: ahh ok, now I see why you have {{ :)
I think it would be easier to compare the actual json output, not to create these hashmaps.

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
```diff
@@ -280,4 +285,43 @@ public void testGetSubscriptionStat() throws Exception {
...
+    @Test(timeout = 10000)
+    public void whenConsumerIsBlocked429() throws Exception {
+        final FloodService.Flooder flooder =
+                new FloodService.Flooder(eventType.getName(), FloodService.Type.CONSUMER_ET);
+        NakadiControllerAT.blockFlooder(flooder);
+
+        TestStreamingClient client = TestStreamingClient
+                .create(URL, subscription.getId(), "")
+                .start();
+        Thread.sleep(2000);
```
v-stepanov: We decided not to use these plain sleeps but rather wait to specific thing. 
Please take a look at other tests in the class. You can do something like
`waitFor(() -> assertThat(MoreStatus.TOO_MANY_REQUESTS.getStatusCode(), client.getResponseCode()));`

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
```diff
@@ -280,4 +285,43 @@ public void testGetSubscriptionStat() throws Exception {
...
+        Assert.assertEquals(MoreStatus.TOO_MANY_REQUESTS.getStatusCode(), client.getResponseCode());
+        Assert.assertEquals("300", client.getHeaderValue("Retry-After"));
+
+        NakadiControllerAT.unblockFlooder(flooder);
+
+        client = TestStreamingClient
+                .create(URL, subscription.getId(), "")
+                .start();
+        Thread.sleep(2000);
```
v-stepanov: here also `waitFor(...)` can be used instead of plain sleep.

src/acceptance-test/java/org/zalando/nakadi/webservice/utils/TestStreamingClient.java
```diff
@@ -58,8 +63,12 @@ public void run() {
             final String url = format("{0}/subscriptions/{1}/events?{2}", baseUrl, subscriptionId, params);
             final HttpURLConnection conn = (HttpURLConnection) new URL(url).openConnection();
             token.ifPresent(token -> conn.setRequestProperty("Authorization", "Bearer " + token));
-            if (conn.getResponseCode() != HttpURLConnection.HTTP_OK) {
-                throw new IOException("Response code is " + conn.getResponseCode());
+            responseCode = conn.getResponseCode();
+            conn.getHeaderFields().entrySet().stream()
+                    .filter(entry -> entry.getKey() != null)
```
v-stepanov: Is it really the case when the header name is null?
Because if not - you can simply do: `headers.putAll(conn.getHeaderFields())`

src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
v-stepanov: ah ok, so you test that the stream quits earlier than test timeout.

src/acceptance-test/java/org/zalando/nakadi/webservice/utils/TestStreamingClient.java
adyach: @v-stepanov yeah, this is the real case. `conn.getHeaderFields()` returns `GET / HTTP/1.1` as a header value with `null` name 😕

src/acceptance-test/java/org/zalando/nakadi/webservice/utils/TestStreamingClient.java
```diff
@@ -128,4 +138,16 @@ public boolean isRunning() {
...
     }
+
+    public int getResponseCode() {
+        return responseCode;
+    }
+
+    public String getHeaderValue(final String name) {
+        final List<String> values = headers.get(name);
+        if (values == null) {
+            return null;
```
adyach: @v-stepanov 4b1549c

src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
adyach: @v-stepanov actually, i test that the consumer was blocked during consumption but to test that I use test timeout, right.

src/acceptance-test/java/org/zalando/nakadi/webservice/utils/TestStreamingClient.java
dehora: Alternative is to make `getHeaderValue` return an Optional:

```java
    public Optional<String> getHeaderValue(final String name) {
       final List<String> values = headers.get(name);
       if (values == null) {
           return Optional.empty();
       }
       return Optional.of(values.get(0));
    }
```
and then:

```java
conn.getHeaderFields().entrySet().stream().filter(Optional::isPresent);
```

There's probably a way to flatten/simplify `getHeaderValue`'s internals but this would make it clear header values are not assured on the method definition.

(also `Optional.of(values.get(0))` picks the first value only; I'm assuming that's ok for Nakadi?)

rcillo:
@adyach where is this feature documented? Has it been added to the API definition yaml?

adyach:
@rcillo it is describe here https://techjira.zalando.net/browse/ARUHA-355 and we had this ealier (but removed) https://github.com/zalando/nakadi/blob/feature/aruha-355/api/nakadi-event-bus-api.yaml#L857, I will update api.

rcillo:
@adyach I mean, where is it documented how to use this newly added endpoint?

adyach:
@rcillo 2e6cd5d

api/nakadi-event-bus-api.yaml
```diff
@@ -954,6 +976,107 @@ paths:
...
+      parameters:
+        - name: flooder
+          in: body
+          description: Abusive client to be blocked
+          schema:
+            $ref: '#/definitions/Flooder'
+          required: true
+      responses:
+        '204':
+          description: Abusive client was successfully blocked.
```
rcillo: Besides a client it could also be an event type, right?

api/nakadi-event-bus-api.yaml
```diff
@@ -954,6 +976,107 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+  /settings/flooders:
```
rcillo: How can we mention the authorisation constraints for these endpoints?

src/main/java/org/zalando/nakadi/service/FloodService.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.Map;
+import java.util.Set;
+
+@Component
+public class FloodService {
+
+    private static final Logger LOG = LoggerFactory.getLogger(FloodService.class);
+    private static final String PATH_FLOODER = "/nakadi/flooders";
+
+    private final EventTypeCache eventTypeCache;
```
antban: Not used

src/main/java/org/zalando/nakadi/service/FloodService.java
```diff
@@ -0,0 +1,199 @@
...
+        return new HashedMap() {
+            {
+                put("consumers", new HashMap<String, Set<String>>() {{
+                    put("event_types", getChildren(Type.CONSUMER_ET));
+                    put("apps", getChildren(Type.CONSUMER_APP));
+                }});
+                put("producers", new HashMap<String, Set<String>>() {{
+                    put("event_types", getChildren(Type.PRODUCER_ET));
+                    put("apps", getChildren(Type.PRODUCER_APP));
+                }});
```
antban: I do not like this method.
The way you create anonymous classes in it is bad. This method can be called 100 times per second, and each time it will create class with static initializer (new class, not instance). That means, that after 10 minutes class list in application will contain 10 * 60 * 100 = 60000 classes (and multiplied by 3, because there are 3 static initializers -> 180K classes), and I'm not sure if it is possible to unload it.

jhorstmann: > (new class, not instance)

new instance actually, the class file will be generated by javac just as any other anonymous class. But I agree that this is a bad use of anonymous classes, just to save one line of code.

adyach: @antban
1. there was no intention to call this method 100 times per second.
2. i did not know about that problem, thanks!
fixed 1af2a29

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
dehora: >> `due to the last experiences on live as a fast solution`
>
> ...So maybe :shipit: but revisit/iterate ideas soon,

It's 14d since that comment. At this point this pr has gone beyond a quick patch and is something more fundamental. I think we should challenge the fail-closed policy of stopping access when the rate limiter fails. There are two scenarios I'm thinking about:

 - traffic level is fine. The rate limiter fails - system unavailable

- traffic level is not fine. The rate limiter fails - system unavailable

I'm worried that the intent of the limiter is to help maintain availability in the latter case of excess traffic, but has consequences for the former where the core system is working under acceptable traffic conditions.  A fail open policy with an alarm/escalation avoids the former scenario where the compensating mechanism takes out the system unnecessarily.

api/nakadi-event-bus-api.yaml
adyach: Maybe we can add description like:
`The oauth resource owner username has to be equal to 'nakadi.oauth2.adminClientId' property to be able to access this endpoint`

rcillo: 👍

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
adyach: Right now, check for blocking implemented in the fail-open manner. From my point of view, it is important not to stop serving clients are not under block and if the problem with so called rate limiter other clients should not be affected by that instead we have to fix the issues faster.

rcillo:
Even if we have some open discussions regarding stability, I would say that this PR should be merged since it adds some essential protection that currently is not in place.

The usage of ZK for many tasks is indeed a latent problem and we need to work on it. We are committed to do it in the near future.

rcillo:
:shipit:

PR #448, opened 2016-09-29
ARUHA-380: Remove 429 status code from HILA endpoints

adyach:


rcillo:
:shipit:

dehora:
For context, these were specifically added to the API definition to signal that a client should be prepared to handle a rate limit error, even though the server didn't implement them. I think it's ok to leave them in place.

v-stepanov:
@dehora We asked at API guild, they told it's ok to add this status code to API in the future (when that will actually be implemented). So we decided just to remove what is not there yet.

rcillo:
But it's a good question. Why does it hurt as is?

adyach:
I think it is fine as it is but inconsistent with the implementation.

v-stepanov:
@rcillo because it can be added later without breaking compatibility. I don't really see a good reason to have it beforehand.

rcillo:
One argument is that if it's in the API, clients implementation would be prepared to handle such scenario once we roll it out.

The problem with that is that we MUST not change the definition. So, if for example, we figure out that it's complicated to implement Retry-after, we have no way of changing, since it would be backward incompatible.

v-stepanov:
@rcillo yeah, I also wanted to mention this point but forgot :)
I would not like to announce something in which I'm not sure yet.

rcillo:
:shipit:

dehora:
🆗 👍

PR #450, opened 2016-09-29
ARUHA-387: "Stream started" message is repeated

adyach:


v-stepanov:
In fact you don't need the counter. The boolean value would be enough :) 
But anyway the implementation is good for me.

v-stepanov:
:shipit:

PR #451, opened 2016-09-30
Aruha 371 hila documentation

v-stepanov:
@dehora @antban @v1ctor @adyach @rcillo 
This is a basic documentation for people starting using Subscription API.


README.md
```diff
@@ -428,6 +437,203 @@ HTTP/1.1 200 OK
...
+The value should be the same as you get in `X-Nakadi-StreamId` header when opening a stream of events. 
+Also, each client can commit only the batches that were sent to him.
+
+The possible success answers from commit endpoints are:
+
+- `204`: cursors were successfully committed and offset was increased;
+- `200`: cursors were committed but at least one of the cursors didn't increase the offset as it was 
+less or equal to already committed one. In a case of this response code user will get a json in a 
+response body with a list of cursors and the results of their commits.
+
```
BGehrels: Please also describe the behaviour in case of errors: What happens if a cursor
- is not committed for a long time while the client keeps reading
- is not committed and the connection gets closed
- is not committed and a second client starts reading the same subscription (what happens if a partition with uncommitted cursors gets reassigned to another client)?

antban: @BGehrels As far as I understand, this information must be provided in api specification. Here is a short introduction to subscriptions (line 472).

README.md
```diff
@@ -428,6 +437,203 @@ HTTP/1.1 200 OK
...
+{
+  "items": [
+    {
+      "event_type": "order.ORDER_RECEIVED",
+      "partitions": [
+        {
+          "partition": "0",
+          "state": "reassigning",
+          "unconsumed_events": 2115,
+          "client_id": "b75c3102-98a4-4385-a5fd-b96f1d7872f2"
```
antban: Isn't it a bug? Header named `X-Nakadi-StreamId`, and here we see `client_id`

README.md
```diff
@@ -428,6 +437,203 @@ HTTP/1.1 200 OK
...
+To delete subscription one should run DELETE request on /subscriptions/{subscriptionId}
+```sh
+curl -v -X DELETE "http://localhost:8080/subscriptions/038fc871-1d2c-4e2e-aa29-1579e8f2e71f"
+```
+
+#### Getting and Listing Subscriptions
+Getting single subscription by id:
+```sh
+curl -v -X GET "http://localhost:8080/subscriptions/038fc871-1d2c-4e2e-aa29-1579e8f2e71f"
+```
```
antban: Probably it's better to provide example answer here

README.md
v-stepanov: yes, I think it makes sense to rename it to stream_id

README.md
```diff
@@ -428,6 +437,203 @@ HTTP/1.1 200 OK
...
+2) start reading events from your subscription;
+
+3) commit the cursors you get in event batches;
+
+If you closed the connection and after some time started reading again - you get events from the 
+point of your latest commit.
+Also, it is not necessary to commit each batch you get. When you commit the cursor, all events that
+are before this cursor will also be considered committed.
+
+If you need more that one client for your subscription to distribute the load - you can start
```
v-stepanov: then -> than

README.md
v-stepanov: @antban I think it makes sense to describe what @BGehrels mentioned, because it is not described anywhere. I'm not sure if it should be described here or not. But for now I will describe it here, and later we can move it to more appropriate place.

README.md
v-stepanov: @antban so for me there are two ways:
1) Specify the example answer for every endpoint
2) Skip the example answer for endpoints where the answer is obvious

README.md
v-stepanov: fixed

README.md
v-stepanov: @antban I added example answer for all endpoints

README.md
v-stepanov: @BGehrels I added description of the cases you mentioned. I hope it's more clear now.

BGehrels: Great, yes, makes things much clearer :-)

README.md
```diff
@@ -428,6 +437,284 @@ HTTP/1.1 200 OK
...
+than the total number of all partitions of the event-types of your subscription. 
+
+_E.g. if you have a subscription for two event-types A and B having 2 and 4 partitions accordingly
+then if you start reading events by a single client - you will get events from all 6 partitions
+by this client. If you connect by a second client - 3 partitions will be transferred from first client
+to a second client and each client will be getting data from 3 partitions. So the maximum possible 
+number of clients for this subscription is 6 (in that case each one will read from one subscription)_
+
+Bellow please find basic examples of Subscription API usage. For more detailed description and advanced
+configuration please take a look at Nakadi [swagger](api/nakadi-event-bus-api.yaml) file.
```
BGehrels: I would add some remarks on delivery semantics here and how they are different from a clients perspective. (the two phase commit problems). Something like

> When implementing a client, be aware that 
> - events of different partitions wil be multiplexed into one stream. That implies that you should not assume that messages will be delivered to you in the order they were sent.
> - [in case of errors during cursor commit](#committing-cursors) (network connectivity, timeouts, ...) it is not guarenteed that you receive each message only once. 
> - as a result of the later, we may even deliver messages of the same partition in a non-chronological order in case the cursor could not be successfull committed.
> 
> For data replication use cases, a typical solution for this is to send idempotent messages by adding a strictly monotonically increasing message id. The change will then only be applied by the receiver if the message id is greater than the last processed one for each entity.

README.md
```diff
@@ -428,6 +437,290 @@ HTTP/1.1 200 OK
...
+  ]
+}
+```
+
+#### Deleting Subscription
+To delete subscription one should run DELETE request on /subscriptions/{subscriptionId}
+```sh
+curl -v -X DELETE "http://localhost:8080/subscriptions/038fc871-1d2c-4e2e-aa29-1579e8f2e71f"
+```
+Success answer:
```
antban: Successful

antban:
:+1:

README.md
v-stepanov: @BGehrels Ok, thanks for raising these points. 

I added some description here https://github.com/zalando/nakadi/pull/451/commits/ae861e209b677b71f2d8e8ab3b3340ac40fdbcc1

However I didn't describe here that `different partitions will be multiplexed into one stream` because it's not subscription-specific behavior, we had the same behavior in Low Level API, so I think that should be described in more general place. Also I'm not really agree regarding `non-chronological order`, but maybe I just misunderstood what you mean.

PR #452, opened 2016-10-04
ARUHA-356: Throttling

v1ctor:
Tests are in progress


src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -72,9 +77,13 @@ private ResponseEntity postEventInternal(final String eventTypeName,
             final JSONArray eventsAsJsonObjects = new JSONArray(eventsAsString);
 
             final int eventCount = eventsAsJsonObjects.length();
-            eventTypeMetrics.reportSizing(eventCount, eventsAsString.length());
```
v1ctor: There were bug, because String.length() doesn't correspond amount of bytes in string.

src/main/java/org/zalando/nakadi/security/ClientResolver.java
```diff
@@ -27,11 +28,14 @@
 
     private final SecuritySettings settings;
     private final FeatureToggleService featureToggleService;
+    private final String defaultId;
 
     @Autowired
-    public ClientResolver(final SecuritySettings settings, final FeatureToggleService featureToggleService) {
+    public ClientResolver(final SecuritySettings settings, final FeatureToggleService featureToggleService,
```
v1ctor: I've refactored this class because I need to now clientId to throttle. By default if AuthMode is OFF we are using adminClientId.

src/main/java/org/zalando/nakadi/throttling/SampledTotal.java
```diff
@@ -0,0 +1,97 @@
...
+public class SampledTotal {
+
+    private int current = 0;
+
+    private List<Sample> samples;
+    private final long windowMs;
+    private final int samplesCount;
+    private final long expireAge;
+
+    public SampledTotal(final long windowMs, final int samplesAmount) {
```
v1ctor: Keep last 'samplesAnount' samples. Period of sample aggreagtion equals 'windowMs' millisecons

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -62,7 +68,13 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName,
         }
     }
 
-    private ResponseEntity postEventInternal(final String eventTypeName,
+    @RequestMapping(value = "/event-types/{eventTypeName}/events", method = HEAD)
+    public ResponseEntity<?> headEvent(@PathVariable final String eventTypeName, final Client client) {
```
v-stepanov: I think this endpoint is questionable.
It was not reviewed by the guild and I think it doesn't align to API guidelines: http://zalando.github.io/restful-api-guidelines/http/Http.html#head

src/main/resources/application.yml
```diff
@@ -68,6 +68,13 @@ nakadi:
   auth:
     plugin:
       factory: org.zalando.nakadi.plugin.auth.DefaultApplicationServiceFactory
+  throttling:
+    bytesLimit: 10737418240 # 10 GB per 15 minutes
+    messagesLimit: 1000000 # million per 15 minutes
+    batchesLimit: 1000000000 # million per 15 minutes
```
v-stepanov: Why the batchesLimit is bigger than messagesLimit? I would expect the opposite.

src/main/java/org/zalando/nakadi/config/WebConfig.java
```diff
@@ -23,6 +19,10 @@
 import org.springframework.web.servlet.config.annotation.AsyncSupportConfigurer;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurationSupport;
 import org.springframework.web.servlet.mvc.method.annotation.RequestMappingHandlerMapping;
+import org.zalando.nakadi.metrics.MonitoringRequestFilter;
+import org.zalando.nakadi.security.ClientResolver;
+import org.zalando.nakadi.util.FlowIdRequestFilter;
+import org.zalando.nakadi.util.GzipBodyRequestFilter;
```
v-stepanov: I think if you only rearranged imports it makes sense to revert this file

rcillo:
Rate-limiting is done based on 2 parameters:

1. application id (token based)
2. resource (since we are restful, endpoints and resources are correlated 1-to-1)

That said, I think it would be possible to implement a generic solution in a non intrusive manner. E.g. an aspect (or Filter in spring terms), that matches endpoints and applications.

Other endpoints also need to be rate-limited and pushing towards a generic solution seems to be valuable.

src/main/resources/application.yml
```diff
@@ -65,12 +65,11 @@ nakadi:
   auth:
     plugin:
       factory: org.zalando.nakadi.plugin.auth.DefaultApplicationServiceFactory
-  flood.retry.after: 300 # seconds
----
-
-spring:
-  profiles: acceptanceTest
```
rcillo: Some tests required this profile. That might explain why some tests are broken now.

src/acceptance-test/java/org/zalando/nakadi/webservice/ThrottlingAT.java
```diff
@@ -0,0 +1,23 @@
...
+import org.zalando.nakadi.domain.EventType;
+
+import static java.util.stream.IntStream.rangeClosed;
+import static org.zalando.nakadi.webservice.utils.NakadiTestUtils.createEventType;
+import static org.zalando.nakadi.webservice.utils.NakadiTestUtils.publishEvent;
+
+public class ThrottlingAT extends BaseAT {
+
+    @Test(timeout = 30000)
+    public void throttleAfter10Messages() throws Exception {
```
v-stepanov: Why it should throttle after 10 messages? Or am I missing something?

v1ctor: This param in application.yaml in acceptanceTest section.

src/main/resources/application.yml
adyach: @v1ctor ?

src/acceptance-test/java/org/zalando/nakadi/webservice/ThrottlingAT.java
```diff
@@ -0,0 +1,23 @@
...
+public class ThrottlingAT extends BaseAT {
+
+    @Test(timeout = 30000)
+    public void throttleAfter100Messages() throws Exception {
+        final EventType eventType = createEventType();
+        rangeClosed(1, 200)
+                .forEach(x -> publishEvent(eventType.getName(), "{\"blah\":\"foo" + x + "\"}")
+                        .then().statusCode(200));
+
+        publishEvent(eventType.getName(), "{\"blah\":\"foo\"}")
```
adyach: It will be nice to see asserts for expected and actual number of limits

src/main/java/org/zalando/nakadi/throttling/SampledTotal.java
```diff
@@ -0,0 +1,41 @@
+package org.zalando.nakadi.throttling;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class SampledTotal {
+
+    private final ConcurrentHashMap<Long, AtomicLong> samples;
+    private final long windowMs;
+    private final long expireAge;
```
adyach: Maybe we can use expirationTime / liveTime for clarity?

src/main/java/org/zalando/nakadi/throttling/SampledTotal.java
```diff
@@ -0,0 +1,41 @@
+package org.zalando.nakadi.throttling;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class SampledTotal {
+
+    private final ConcurrentHashMap<Long, AtomicLong> samples;
```
adyach: I think there is no need in AtomicLong

src/main/java/org/zalando/nakadi/throttling/SampledTotal.java
```diff
@@ -0,0 +1,41 @@
+package org.zalando.nakadi.throttling;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicLong;
+
+public class SampledTotal {
```
valgog: I would implement it a little bit differently, though your implementation is cool. I would store the tuple of three counts (to have more consistent sums).

The basic idea, is that your SampleCollector can look something like:
```java
SampleCollector {
    ReadWriteLock rwl;

    CounterBin currentBin;
    long currentBinSwitchTime;

    HashMap<Long, CounterBin> previousBins;
    long previousBinsSwitchTime;
    ...

}
```

And `currentBin` contains your three [`LongAdder`](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/LongAdder.html) values for collecting the values.

You 1) ReadLock, 2) If current time is higher, then `currentBinSwitchTime`, you 3) WriteLock, 4) create a new `currentBin` 4) add old bin to the map (actually it can be a simple `HashMap` in this case), 5) proclaim new `currentBin` and update the `currentBinSwitchTime` 5) release the locks.

See http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.html for examples in case have needed before.

This operation will be done only once per `windowDuration` time and lock for the time of the swap and map update.

Clean up of the previous bins can be done using the same approach, by swapping it away during the `currentBin` swap operation or from a separate thread.

Just my 2 cents :D

src/main/java/org/zalando/nakadi/throttling/ThrottleResult.java
```diff
@@ -0,0 +1,60 @@
...
+    public ThrottleResult(final long bytesLimit, final long bytesRemaining, final long messagesLimit,
+                          final long messagesRemaining, final long batchesLimit, final long batchesRemaining,
+                          final Instant reset, final boolean throttled) {
+        this.bytesLimit = bytesLimit;
+        this.bytesRemaining = bytesRemaining;
+        this.messagesLimit = messagesLimit;
+        this.messagesRemaining = messagesRemaining;
+        this.batchesLimit = batchesLimit;
+        this.batchesRemaining = batchesRemaining;
+        this.reset = reset;
```
valgog: What does `reset` mean here? Is it a reset time?

src/main/java/org/zalando/nakadi/throttling/ThrottlingService.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+@Component
+public class ThrottlingService {
+
+    private static final String BYTES_LIMIT = "nakadi.throttling.bytesLimit";
+    private static final String MESSAGES_LIMIT = "nakadi.throttling.messagesLimit";
+    private static final String BATCHES_LIMIT = "nakadi.throttling.batchesLimit";
+    private final long windowSize;
```
valgog: This is time window? If it is time, then in English it will be `windowLength` or `windowDuration`.

src/main/java/org/zalando/nakadi/util/ZkConfigurationService.java
```diff
@@ -0,0 +1,59 @@
...
+                });
+    }
+
+    private String getValue(final String key) {
+        try {
+            final byte[] data = zooKeeperHolder.get().getData().forPath(PREFIX + key);
+            return new String(data, CHARSET);
+        } catch (Exception e) {
+            LOG.error("Can't obtain value from ZK", e);
+            return context.getEnvironment().getProperty(key);
```
v1ctor: Right now it will return null, but we define this properties in application.yml file and it's already a fallback.

src/main/java/org/zalando/nakadi/util/ZkConfigurationService.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    public Long getLong(final String key) {
+        try {
+            return Long.parseLong(cache.getUnchecked(key));
+        } catch (NumberFormatException e) {
+            LOG.error("Can't obtain value from ZK", e);
+            throw new IllegalArgumentException(e);
```
v1ctor: I think request will fail. I will add fallback params. But it's strange to have fallback, on fallback :)

src/main/java/org/zalando/nakadi/throttling/SampledTotal.java
```diff
@@ -0,0 +1,41 @@
...
+        this.samples = new ConcurrentHashMap<>();
+        this.expireAge = samplesAmount * windowMs;
+    }
+
+    public void record(final Long value, final long timeMs) {
+        final AtomicLong sample = current(timeMs);
+        sample.addAndGet(value);
+    }
+
+    public Long measure(final long now) {
```
v1ctor: `measure()` is called on every request.

rcillo:
For me the state looks like this

```
                                       +------------------------------+
                            Yes        |                              |     No
                                       | Is a single application      |
                          +------------+ overloading Nakadi and       +------+
                          |            | compromising ser^ice levels? |      |
                          |            |                              |      |
                          |            +------------------------------+      |
                          |                                                  |
             +------------v--------------+                     +-------------v-----------+
             |                           |                     |  What's killing Nakadi? |
       +-----+Is it fine if we rate limit+---+                 |                         |
 Yes   |     |this application?          |   | No              |                         |
       |     |                           |   |                 +-------------------------+
       |     +---------------------------+   |
       |                                     |
+------v-------------------+   +-------------v----+
| At which level it starts |   | We need to scale |
| to cause problems?       |   |                  |
+----------+---------------+   +------------------+
           |
           |
     +-----v-----+
     |Rate limit!|
     +-----------+
```

We implemented rate limiting without going through the left most path until the end and now we don't know what to do with this PR.

We need to got through the questions:

1. Is a single application killing Nakadi?
2. Is it fine to rate limit it?
3. At which level we are rate limiting it?

valgog:
@rcillo for me this PR is only about possibility to rate limit and not about the total concept. 

I know, that @v1ctor will be again angry with me about this, but I think we can surely say, that one instance can limit publishing rate to one partition (not topic) at least to 1000 Req/s. To be more data driven we actually know how much our Kafka can receive per partition in MB/s (there are results of the benchmarking somewhere). Setting this limit per partition from the instance is definitely a good idea and a first step. Rate limiting per partition is used for Kynesis (2MB/s for them, but we have measured much higher speeds for our appliance) and can be used as a starting point for us as well. 

After we understand our bottlenecks and the real throtteling concept is there, we can move here fast. 

Your ASCII graph is really cool! I like it, and it should part of the real throtteling concept

rcillo:
My diagram is just a way to make questions that will tell us if we can or cannot do it in the way we planned.

I'm afraid that such rate limit would break shop-updater causing out-of-sync solr product information.

Shop updater batches have in average several megabytes. I think that 2 MB/s could make their task take much longer than it already takes. I don't know if they can live with that. Link to more data about batch sizes https://docs.google.com/spreadsheets/d/1iznXHOR3ebUimVs90P_AxBEznOIlkwgW6XTxvKQYfsU/edit#gid=119423282

@tmuehl is it fine if we rate-limit the capacity of shop updater to push only 2 MB/s?

If we do so and publisher application receives a 429 response (which wasn't in the API previously) would it retry or adapt to such limitation?

valgog:
@rcillo just to clarify, I have never said that 2 MB/s per partition should be our limit.

tmuehl:
> @tmuehl is it fine if we rate-limit the capacity of shop updater to push only 2 MB/s per partition?

@rcillo You ask me if it's ok to break the fashion store? :anguished:

Seriously, you can (should be able to) handle the traffic by the Shop Update easily, it's traffic is predictable and totally valid business. Why do you want to downgrade it?

rcillo:
@tmuehl sorry for providing little context. That's exactly my point. I'm against merging this PR exactly because I think we shouldn't downgrade the updater. ;)

rcillo:
👎

PR #462, opened 2016-10-10
Add missing OAuth2 to registry

mastersimon:
- add oauth2 to _registry/validation-strategies_
- add oauth2 to _registry/enrichment-strategies_
- add oauth2 to _registry/partition-strategies_

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -897,33 +897,6 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
-  '/registry/validation-strategies':
-    get:
-      tags:
-        - schema-registry-api
-      security:
-        - oauth2: ['nakadi.event_stream.read']
-      description: |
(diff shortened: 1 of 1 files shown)


adyach:
@mastersimon thank you for contribution! I think having `nakadi.event_stream.read` for reading some configuration data is not the right approach. I do not know exactly what was the intention for `nakadi.config` scopes but I propose to use `nakadi.config.read` for `registry` endpoint. Also there is a possibility that someone is already using `nakadi.event_stream.read` scope then we do not have many choices here. WDYT, @antban, @rcillo, @v1ctor, @v-stepanov ?

mastersimon:
Hi @adyach,

my approach was to add OAuth to the endpoints - because you need it. Per attention I added `nakadi.event_stream.read`. Of course I can change it to whatever makes sense.

adyach:
@mastersimon at least it shows current situation with the api, please, resolve conflicts and I will merge it. thanks!

adyach:
👍

PR #463, opened 2016-10-11
ARUHA-398: Endpoint is not implemented: /registry/validation-strategies

adyach:


antban:
:+1:

valgog:
How difficult it is to implement it? We only have a couple of validation strategies. Or do I miss something?

valgog:
@adyach if implementation is not not easy, probably creating a ticket to implement it would make sense. Otherwise we will forget to put it back :D

adyach:
@valgog I would not say that it is difficult to implement but it involves changes to EventType object. Also some validation strategies have to be applied every time for specific event type category (e.g. https://github.com/zalando/nakadi/blob/master/src/main/java/org/zalando/nakadi/validation/EventValidation.java#L25, `business` or `data` category has to be validated against `metadata validation strategy` and not necessary for `undefined` category)

we had a short discussion here [ARUHA-398](https://techjira.zalando.net/browse/ARUHA-398)

rcillo:
I would argue that it's more valuable to push the product in the direction of removing options than implementing them. These differences between categories are hard to explain and brings very few value to the product.

Besides, there's not much to configure. We either validate or not. The semantics of validation is embedded in the category, so it's not possible to have business event without metadata validation, for example.

Making this explicit, as a configuration, would make it redundant.

PR #464, opened 2016-10-13
Aruha 116 schema versioning api

rcillo:
The document with the requirements for compatible schema evolution is here https://docs.google.com/document/d/1FmGHxI9oklxN16LKZe1UG_U21jtQFVCtLtRiC2W1pAk/edit#

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -16,9 +16,6 @@ info:
     * The **schema** of the Event of this EventType. The schema defines the accepted format of
     Events of an EventType and will be, if so desired, enforced by Nakadi. Usually Nakadi will
     respect the schema for the EventTypes in accordance to how an owning Application defines them.
-    **Note:** *Currently the specification of the schema must be pushed into Nakadi on EventType
-    creation; in the future, assuming that Applications will expose the schema for its owned
-    resources, Nakadi might support fetching the schema directly from them.*
 
     * The expected **validation** and **enrichment** procedures upon reception of an Event.
     Validation define conditions for the acceptance of the incoming Event and are strictly enforced
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1585,7 +1585,7 @@ definitions:
         type: array
         items:
           type: string
-        description: List of references to previous schemas.
+        description: List of links to previous schemas.
 
       partition_key_fields:
         type: array
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1573,12 +1573,6 @@ definitions:
         description: |
           The most recent schema for this EventType. Submitted events will be validated against it.
 
-      schema_versions:
-        type: array
-        items:
-          type: string
-        description: List of links to previous schemas.
-
       partition_key_fields:
         type: array
         items:
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1567,9 +1567,6 @@ definitions:
           When in compatible mode, it's allowed to add new optional properties and definitions to an existing schema,
           but no other changes are allowed.
         type: string
-        enum:
-          - none
-          - compatible
         default: compatible
 
       schema:
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1168,7 +1168,7 @@ definitions:
         readOnly: true
         description: |
           Version of the schema used for validating this event. This is enriched upon reception.
-          Clients consuming events should desserialize events only when differences between their
+          Clients consuming events should deserialize events only when differences between their
           schema and the one used to validate differs on minor level, for example, from '1.1.0' to
           '1.2.0'.
       parent_eids:
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1553,11 +1553,10 @@ definitions:
 
           There are 2 compatibility modes:
 
-          - 'none': Every schema change is allowed. Coordination between producers and consumers
-            must be done accordingly when backward incompatible changes are to be rolled out.
+          - 'fixed': Schema changes are not allowed.
 
-          - 'compatible': Changes never break existing consumers. Every event published since the first version is still
-            valid based on the newest schema.
+          - 'compatible': Schema changes never break existing consumers. Every event published since the first version
+            is still valid based on the newest schema.
 
           When in compatible mode, it' allowed to add new optional properties to an existing schema, but no other
           changes are allowed.
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1555,6 +1555,8 @@ definitions:
 
           - 'legacy': Schema changes are not allowed. This is used only by legacy event types that are not compatible.
 
+          - 'incompatible': Any schema modification is accepted, even if it might break existing producers or consumers.
+
           - 'compatible': Schema changes never break existing consumers. Every event published since the first version
             is still valid based on the newest schema.
 
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -486,7 +486,7 @@ paths:
           schema:
             type: object
             properties:
-              items:
+              data:
                 $ref: '#/definitions/EventTypeSchema'
         '401':
           description: Client is not authenticated
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1639,8 +1639,10 @@ definitions:
           The schema as string in the syntax defined in the field type. Failure to respect the
           syntax will fail any operation on an EventType.
 
-          To have a generic, undefined schema it is possible to define the schema as `"schema":
-          "{\"additionalProperties\": true}"`.
+          When defining a schema, every property should be explicitly defined with its type. It means that,
+          additionalProperties true is not accepted.
+
+          The complete list of not allowed attributes is: additionalProperties, not and patternProperties.
     required:
       - type
       - schema
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -498,8 +498,8 @@ paths:
       tags:
         - schema-registry-api
       description: |
-        Retrieves a given schema version as JSON. A special `{version}` key named 'latest' is
-        provided for convenience.
+        Retrieves a given schema version. A special `{version}` key named 'latest' is provided for
+        convenience.
     responses:
         '200':
           description: |
@@ -1167,8 +1167,8 @@ definitions:
         description: |
           Version of the schema used for validating this event. This is enriched upon reception.
           Clients consuming events should desserialize events only when differences between their
-          schema and the one used to validate (produce) the event differs on minor level, i.e. from
-          '1.1.0' to '1.2.0'.
+          schema and the one used to validate differs on minor level, for example, from '1.1.0' to
+          '1.2.0'.
       parent_eids:
         type: array
         items:
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1165,6 +1165,7 @@ definitions:
         example: '1996-12-19T16:39:57-08:00'
       version:
         type: string
+        readOnly: true
         description: |
           Version of the schema used for validating this event. This is enriched upon reception.
           Clients consuming events should desserialize events only when differences between their
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -484,10 +484,9 @@ paths:
     responses:
         '200':
           schema:
-            type: object
-            properties:
-              data:
-                $ref: '#/definitions/EventTypeSchema'
+            type: array
+            items:
+              $ref: '#/definitions/EventTypeSchema'
         '401':
           description: Client is not authenticated
           schema:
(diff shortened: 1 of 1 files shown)


api/nakadi-event-bus-api.yaml
```diff
@@ -1567,13 +1648,14 @@ definitions:
           future there could be others.
       schema:
         type: string
-        $ref: '#/definitions/EventTypeSchema'
         description: |
           The schema as string in the syntax defined in the field type. Failure to respect the
           syntax will fail any operation on an EventType.
 
-          To have a generic, undefined schema it is possible to define the schema as `"schema":
-          "{\"additionalProperties\": true}"`.
```
rcillo: This breaks the contract. Maybe we should let it working for "legacy" types?

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1539,39 @@ definitions:
...
+          inadvertently breaking existent consumers.
+
+          Once defined, the compatibility mode is fixed, since otherwise it would break a predefined contract,
+          jeopardizing the value of this feature.
+
+          There are 2 compatibility modes:
+
+          - 'legacy': Schema changes are not allowed. This is used only by legacy event types that are not compatible.
+
+          - 'incompatible': Any schema modification is accepted, even if it might break existing producers or consumers.
```
adyach: > I am not 100% sure if we should have this option.

If we create schema versioning which has to be followed by the users I do not think we have to create some kind of workarounds. If you need an incompatible schema then create new event type (btw, you event have https://github.com/zalando/nakadi/pull/464/files#diff-4492c6b26ee0a37518728ec1585420efR1553)
👎

api/nakadi-event-bus-api.yaml
adyach: Unless it is restricted to create new event types with "legacy" compatible mode.

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1539,39 @@ definitions:
...
+
+          It's designed to be flexible enough so that producers can evolve their schemas while not
+          inadvertently breaking existent consumers.
+
+          Once defined, the compatibility mode is fixed, since otherwise it would break a predefined contract,
+          jeopardizing the value of this feature.
+
+          There are 2 compatibility modes:
+
+          - 'legacy': Schema changes are not allowed. This is used only by legacy event types that are not compatible.
```
adyach: Maybe we can specify that creation of such mode is not allowed for users?

rcillo: done https://github.com/zalando/nakadi/pull/464/commits/a9fa489b65da60077e37a0b2dfeb3a1d009fb3ea

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1539,40 @@ definitions:
         type: string
         default: 'random'
 
+      compatibility_mode:
+        description: |
+          Compatibility modes provides a mean for event owners to evolve their schema, given changes
```
v-stepanov: **mode provides** or **modes provide**

api/nakadi-event-bus-api.yaml
```diff
@@ -479,7 +475,40 @@ paths:
...
+  '/event-types/{name}/schemas':
+    get:
+      tags:
+        - schema-registry-api
+      description: |
+        List of schemas ordered from most recent to oldest.
+    responses:
+        '200':
+          schema:
+            type: array
```
rcillo: I'm uncomfortable with that because I think that every endpoint has it's peculiarities. I think that this extensibility decisions should be done only if there are envisioned extensions for the near future. Making it extensible just to be safe could lead to accidental complexity.

Are you aware of anything we might add to this endpoint in the future?

api/nakadi-event-bus-api.yaml
```diff
@@ -1132,6 +1161,13 @@ definitions:
           If set by the producer Event will be rejected.
         format: date-time
         example: '1996-12-19T16:39:57-08:00'
+      version:
```
v-stepanov: readOnly: true 
?

api/nakadi-event-bus-api.yaml
```diff
@@ -1132,6 +1161,13 @@ definitions:
...
+        type: string
+        description: |
+          Version of the schema used for validating this event. This is enriched upon reception.
+          Clients consuming events should desserialize events only when differences between their
+          schema and the one used to validate differs on minor level, for example, from '1.1.0' to
+          '1.2.0'.
```
v-stepanov: What if it differs on PATCH level? (e.g. 1.1.0 and 1.1.2)
I think in that case it's also safe to deserialize event.

api/nakadi-event-bus-api.yaml
```diff
@@ -1132,6 +1161,13 @@ definitions:
           If set by the producer Event will be rejected.
         format: date-time
         example: '1996-12-19T16:39:57-08:00'
+      version:
+        type: string
+        description: |
+          Version of the schema used for validating this event. This is enriched upon reception.
+          Clients consuming events should desserialize events only when differences between their
```
v-stepanov: deserialize

api/nakadi-event-bus-api.yaml
```diff
@@ -479,7 +475,40 @@ paths:
...
+      tags:
+        - schema-registry-api
+      description: |
+        Retrieves a given schema version. A special `{version}` key named 'latest' is provided for
+        convenience.
+    responses:
+        '200':
+          description: |
+            Schema as JSON.
+        '401':
```
v-stepanov: API guidelines now allow to skip the common failure codes (like 401) in swagger file
See here http://zalando.github.io/restful-api-guidelines/http/Http.html#must-provide-error-documentation

api/nakadi-event-bus-api.yaml
```diff
@@ -479,7 +475,40 @@ paths:
...
+
+  '/event-types/{name}/schemas/{version}':
+    get:
+      tags:
+        - schema-registry-api
+      description: |
+        Retrieves a given schema version. A special `{version}` key named 'latest' is provided for
+        convenience.
+    responses:
+        '200':
```
rcillo: Done b3d581e

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1539,40 @@ definitions:
...
+
+          - 'compatible': Schema changes never break existing consumers. Every event published since the first version
+            is still valid based on the newest schema.
+
+          When in compatible mode, it's allowed to add new optional properties and definitions to an existing schema,
+          but no other changes are allowed.
+        type: string
+        enum:
+          - none
+          - compatible
```
v-stepanov: the possible values of enum don't correspond to the description

api/nakadi-event-bus-api.yaml
rcillo: Done 480cacf

v-stepanov: Sorry, forget about my comment here. This is event metadata it is not supposed to be submitted by user anyway.

api/nakadi-event-bus-api.yaml
rcillo: Done 7332f87

api/nakadi-event-bus-api.yaml
```diff
@@ -1558,6 +1635,11 @@ definitions:
 
   EventTypeSchema:
     properties:
+      version:
```
v-stepanov: readOnly: true
?

api/nakadi-event-bus-api.yaml
rcillo: Yes it is. Do you we might better make it explicit? I thought it would be implicit since PATCH is within MINOR.

api/nakadi-event-bus-api.yaml
rcillo: Done 30014c5

api/nakadi-event-bus-api.yaml
```diff
@@ -1567,13 +1652,14 @@ definitions:
...
         type: string
-        $ref: '#/definitions/EventTypeSchema'
         description: |
           The schema as string in the syntax defined in the field type. Failure to respect the
           syntax will fail any operation on an EventType.
 
-          To have a generic, undefined schema it is possible to define the schema as `"schema":
-          "{\"additionalProperties\": true}"`.
+          When defining a schema, every property should be explicitly defined with its type. It means that,
+          additionalProperties true is not accepted.
```
v-stepanov: what will we do with already existing event types that have "additionalProperties": true ?

api/nakadi-event-bus-api.yaml
rcillo: True. But maybe it would be more clear if we made all of then readOnly.

api/nakadi-event-bus-api.yaml
rcillo: Since extending enums is not backward compatible by the api guidelines, maybe we should not explicitly state the enum values. I'm afraid we might add new valued to this enum in the future, when we design a solution for incompatible schema evolution.

api/nakadi-event-bus-api.yaml
rcillo: I also marked other metadata properties as readOnly.

api/nakadi-event-bus-api.yaml
rcillo: Not all of them actually. I marked the ones that are readOnly.

v-stepanov: I don't know. As we don't expect this object to be submitted by the user - I think it doesn't make sense to mark fields as readOnly.

api/nakadi-event-bus-api.yaml
v-stepanov: I'm not a big expert of semver so for me it sounded that only MINOR level changes allow to deserialize :)

api/nakadi-event-bus-api.yaml
v-stepanov: ok, fine

api/nakadi-event-bus-api.yaml
rcillo: Done 184415f

api/nakadi-event-bus-api.yaml
rcillo: @v-stepanov they will be marked with `compatibility_mode: deprecated` and will keep working as usual. Upgrading them would be possible since that would mean that events from the same event type could have incompatible fields.

Event if additionalProperties true is not explicitly defined, it's the default semantics for validation. So I don't believe we can solve this. New event types with proper schema and compatibility modes must be created for every event type.

But that just sounds terrible. So, we might open an exception and allow an "upgrade" in the compatibility mode, given everybody involved is aware that old (not versioned) event types are going to be incompatible with new (1.0 on) events.

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1543,37 @@ definitions:
...
+
+          List of available compatibility modes:
+
+          - 'deprecated': Schema changes are not allowed. This is used only by legacy event types that are not
+            compatible. This option is not available for new event types. Additional properties are allowed.
+
+          - 'incompatible': Any schema modification is accepted, even if it might break existing producers or consumers.
+
+          - 'compatible': Schema changes never break existing consumers. Every event published since the first version
+            is still valid based on the newest schema.
```
v-stepanov: does it mean that 'compatible' schema will never go out of version 1.x.x ?

rcillo: Exactly. By the way that's the meaning of MAJOR version on semver.

> MAJOR version when you make incompatible API changes,

v-stepanov:
@rcillo one more question. How the update of schema will be performed? By doing a PUT to `/event-types/{my-et}`?
If yes - do we want to extend the endpoint with some conflict-response in a case when schema changes don't correspond to compatibility mode?

api/nakadi-event-bus-api.yaml
```diff
@@ -479,7 +475,42 @@ paths:
...
+      description: |
+        Retrieves a given schema version. A special `{version}` key named 'latest' is provided for
+        convenience.
+    responses:
+        '200':
+          schema:
+            type: object
+          description: |
+            Schema as JSON.
+        '401':
```
v-stepanov: My comment got lost somewhere so I will repeat it :)

API guidelines now allow to skip the common failure codes (like 401) in swagger file
See here http://zalando.github.io/restful-api-guidelines/http/Http.html#must-provide-error-documentation

We actually already used this approach when defining Subscription API (we skipped all 401, 403, 503 etc.)

rcillo:
@v-stepanov yes, the schema update will be done by PUT to event-types/name.

I thought about using 422 for unprocessable schema changes. Do you think 409 would better classify the error?

api/nakadi-event-bus-api.yaml
rcillo: Done dbf3d5c

v-stepanov:
@rcillo no I think 422 is pretty ok.

v-stepanov:
:shipit:

rcillo:
Now that we have an approval, I'll send it to API guild review under a new pristine PR.

api/nakadi-event-bus-api.yaml
```diff
@@ -479,7 +475,34 @@ paths:
...
+        - schema-registry-api
+      description: |
+        Retrieves a given schema version. A special `{version}` key named 'latest' is provided for
+        convenience.
+    responses:
+        '200':
+          schema:
+            type: object
+          description: |
+            Schema as JSON.
```
valgog: @v-stepanov @rcillo is it consistent with the type we use to accept schema definition when creating an `enent-type`?

Do not we use `string` there? 

I think we should at least document the difference if there is any.

PR #465, opened 2016-10-16
ARUHA-411 Improve gradle support in travis-ci

valgog:
Do some improvements in gradle usage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,15 @@ services:
 addons:
   hostname: localhost
 
+before_cache:
+  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
+  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
+
+cache:
+  directories:
+    - $HOME/.gradle/caches/
+    - $HOME/.gradle/wrapper/
+
 before_install:
   - docker --version
   # install docker-compose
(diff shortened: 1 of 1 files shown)


valgog:
@zalando/aruhacore please review

.travis.yml
```diff
@@ -12,30 +12,45 @@ services:
...
   - sudo mv docker-compose /usr/local/bin
-
-before_script:
-  - sudo /etc/init.d/postgresql stop
   - pip install --user codecov
   - curl http://www.jpm4j.org/install/script | sudo sh
   - sudo jpm install com.codacy:codacy-coverage-reporter:assembly
 
+install:
+  - ./gradlew compileJava compileTestJava assemble
```
adyach: Could you elaborate on that? I don't understand why we need this step.

.travis.yml
```diff
@@ -12,30 +12,45 @@ services:
...
+cache:
+  directories:
+    - $HOME/.gradle/caches/
+    - $HOME/.gradle/wrapper/
+    - $HOME/.cache/pip
+
 before_install:
   - docker --version
   # install docker-compose
   - curl -L https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > docker-compose
```
adyach: Can we cache it as well?

.travis.yml
valgog: default `install` step is calling `gradlew assemble`, that leads to `compileJava` but not `compileTestJava`. see https://docs.travis-ci.com/user/languages/java/#Projects-Using-Gradle

adding `compileTestJava` to the `install` phase ensures, that the complication of test resources is done before running the tests themselves and, hopefully, improves performance of the later matrix script executions. Though I do not understand completely how much of the compilation results is shared between matrix jobs.

.travis.yml
valgog: Yes, I thought about it, but it is also a download from S3 most probably (the same as the cache), so do not know if this changes anything.

.travis.yml
adyach: @valgog what I meant is that we do compilation in the matrix script anyway and it seems like the assemble task is redundant because we do not use it results.

adyach:
:shipit:

v-stepanov:
@valgog Can you please briefly describe what are these improvements about?

Another point is that your commit messages don't have ticket id, so that will produce 6 violations.

v-stepanov:
👎

valgog:
@v-stepanov is it better now?

PR #469, opened 2016-10-19
Aruha 116 schema versioning api

rcillo:
The document with requirements for compatible schema evolution is here https://docs.google.com/document/d/1FmGHxI9oklxN16LKZe1UG_U21jtQFVCtLtRiC2W1pAk/edit#


api/nakadi-event-bus-api.yaml
```diff
@@ -1567,13 +1651,14 @@ definitions:
...
         type: string
-        $ref: '#/definitions/EventTypeSchema'
         description: |
           The schema as string in the syntax defined in the field type. Failure to respect the
           syntax will fail any operation on an EventType.
 
-          To have a generic, undefined schema it is possible to define the schema as `"schema":
-          "{\"additionalProperties\": true}"`.
+          When defining a schema, every property should be explicitly defined with its type. It means that,
+          additionalProperties true is not accepted.
```
conorclifford: We have a need to have 2 constructs that may have problems here - can you confirm that we will be able to work with this, or are we effected.

Firstly, we use `additionalProperties` to define a `Map[String, SomeDefinition]`:

```json
       "some_map": {
          "additionalProperties": {
            "$ref": "#/definitions/SomeOtherDefinition"
          }
        }
```

Secondly, we need to have a property within the event (deep nested in fact), that has a type of simply `object` - the value of this property is a well formatted JSON object, the contents of which are fully validated prior to event submission, but this content depends on a flexible data model that can change over time (however, we also have careful compatible change control within this own data modelling construct.)

Actually, the "property" mentioned is in several definitions, and is handled by having the equivalent of the following definition "mixed" in with other definitions (using `allOf` composition): 

```json
"FlexibleModelledData": {
      "type": "object",
      "required": [
        "data_modeled_and_validated_elsewhere"
      ],
      "properties": {
        "data_modeled_and_validated_elsewhere": {
          "type": "object"
        }
      }
```

rcillo: @conorclifford I think your example will help us better define our goal with this feature. I think the principle of open for extension, closed for change would fit nicely here.

I see no problems with simple `object` properties. Yes, it's extensible, since you can always add something more inside it. But if we allow further changes to this `object`, by defining some property, it changes the previously defined "contract". Let me give you an example:

```javascript
{
  "properties": {
    "extensible_object": {
      "type": "object";
     }
  }
}
```

And later, it changes to 

```javascript
{
  "properties": {
    "extensible_object": {
      "type": "object";
      "properties": {
        "some_very_specific_constraint": {
          "type": "string";
        }
     }
  }
}
```

What it means is that, for some time, this event type might have accepted `some_very_specific_constraint` property with any `type`, but after some point, it defined that the `type` is a string.

That would not fulfil one of the goals - [backward compatibility](https://docs.google.com/document/d/1FmGHxI9oklxN16LKZe1UG_U21jtQFVCtLtRiC2W1pAk/edit#heading=h.hov7tfyvg4a9).

That all said, I think things are clearer now. What we need to guarantee here is there is an "open definition" it should never change.

Just to make it very clear: if "additionalProperties" is allowed, it might not be able to change this part of the schema in the future.

Would it work for you?

conorclifford: @rcillo That sounds good (and sensible - if starting as "extensible" then it makes perfect sense that that cannot then later be constrained (as you say, constraint may not hold for older data).

I'm not sure I follow your point on `additionalProperties` though - is it that we will it be possible to use `additionalProperties` to define a map of some specific type, but, that map's definition cannot change after that? Or something else?

rcillo: @conorclifford correct. I would be possible to use `additionalProperties` to define a map of some specific type, but once defined, it won't be possible to change it. So if additionalProperties is used within an object, this object is closed for schema changes, since additionalProperties means it's extensible.

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1535,37 @@ definitions:
...
+
+          It's designed to be flexible enough so that producers can evolve their schemas while not
+          inadvertently breaking existent consumers.
+
+          Once defined, the compatibility mode is fixed, since otherwise it would break a predefined contract,
+          jeopardizing the value of this feature.
+
+          List of available compatibility modes:
+
+          - 'deprecated': Schema changes are not allowed. This is used only by legacy event types. This option is not
```
dehora: I think this could be called `fixed`. Deprecated means something different. I'm not sure how this will be implemented but it sounds like it will be a date check on the schema creation and something that a producer will never be allowed to set themselves. I'm wondering if it's actually needed. 

(gr)

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1535,37 @@ definitions:
...
+            available for new event types. Schema validation is lenient and allow fields that are not defined. Also,
+            it's possible to use the full json schema specification for defining schemas.
+
+          - 'incompatible': Any schema modification is accepted, even if it might break existing producers or consumers.
```
dehora: `'incompatible'` I think this could be called `none`. That describes the nature of the check being made by the server.

(gr)

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1535,37 @@ definitions:
...
+
+          - 'compatible': Schema changes never break existing consumers. Every event published since the first version
```
dehora: This overloads multiple kinds of compatibility which might make the feature harder to evolve. What's the reason not to use forwards/backward/all, which were the ones identified in the early design. That is it should be possible to map `compatible` to one of `forwards/backward/all`.

(gr)

api/nakadi-event-bus-api.yaml
```diff
@@ -1550,6 +1608,18 @@ definitions:
           contains at least one of the scopes in this list.
           If no scopes provided then anyone can consume from this event type.
 
+      created_at:
+        type: string
+        pattern: date-time
+        description: |
+          Date and time when this event type was created.
```
dehora: :+1: to the date - is the name `created_at` consistent with the rest of the API?

(gr)

api/nakadi-event-bus-api.yaml
```diff
@@ -1550,6 +1608,18 @@ definitions:
...
+
+      updated_at:
```
dehora: :+1: to the date - is the name `updated_at` consistent with the rest of the API?

(gr)

api/nakadi-event-bus-api.yaml
```diff
@@ -1558,6 +1628,20 @@ definitions:
 
   EventTypeSchema:
     properties:
+      version:
```
dehora: A few things here:

- semver allows trailing information as well as the three digit scheme. Is that allowed here?

- semver treats 0.y.z different to 1.y.z; the former (pre-1) range is allowed to have breaking changes. Is that allowed here?

- is semver definitely adequate to express what's needed for server versioning?

(gr)

api/nakadi-event-bus-api.yaml
dehora: whether type bound additionalProperties are allowed needs to be decided

(gr)

api/nakadi-event-bus-api.yaml
rcillo: Besides been "fixed" in the sense that schema changes are not allowed, validation of events sent to these topics is "lenient". By default, schema validation allows additional properties even when not explicitly stated. It's not just about schema evolution but also about important changes to the way events are going to be validated from now on.

Also, it might include some feature that are no longer accepts, such as "not" attributes and "patternProperties".

Our goal is to move these events from deprecated to the other possible values. We think that allowing lenient schema validation is harmful and should be avoided.

Having something called "fixed" would indicate that it's a viable option, when it's not.

api/nakadi-event-bus-api.yaml
rcillo: 👍

api/nakadi-event-bus-api.yaml
rcillo: In a nutshell, our design diverged from Avro's because Avro compatibility mode only applies to reading events, not writing.

When we decided to allow schemas to be evolved in a non coordinated manner (not notifying producers or consumers), we constrained the problem, reducing the possibilities of having several modes.

The current solution is simpler to understand, train and debug. It works as `all` compatibility addressing all requirements.

api/nakadi-event-bus-api.yaml
rcillo: 1. This field is managed by Nakadi and the three digit scheme suffices.
2. Version starts on 1.0 which means it doesn't allow breaking changes (under compatibility mode it always upgrades MINOR or PATCH level, under incompatible mode it might change MAJOR).
3. Semver is widely adopted and understood by developers. It captures all the necessary semantics for versioning as far as we could map.

api/nakadi-event-bus-api.yaml
```diff
@@ -479,7 +475,34 @@ paths:
...
+      tags:
+        - schema-registry-api
+      description: |
+        List of schemas ordered from most recent to oldest.
+    responses:
+        '200':
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/EventTypeSchema'
```
rcillo: 👍  b2f265c

api/nakadi-event-bus-api.yaml
```diff
@@ -1126,12 +1149,21 @@ definitions:
...
         format: date-time
         example: '1996-12-19T16:39:57-08:00'
+      version:
+        type: string
+        readOnly: true
+        description: |
+          Version of the schema used for validating this event. This is enriched upon reception.
+          Clients consuming events should deserialize events only when differences between their
+          schema and the one used to validate differs on minor level, for example, from '1.1.0' to
+          '1.2.0'.
```
rcillo: The meaning of the version is well defined as following semver https://github.com/zalando/nakadi/pull/469/files#diff-4492c6b26ee0a37518728ec1585420efR1641

So users are safe to interpret it. Actually, as by our experience, every field that is human readable will be interpreted.

This description gives a hint on how to implement clients in a defensive mindset, since there is no independent channel to notify about changes, we encourage clients to be aware of changes by inspecting this field.

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1535,37 @@ definitions:
...
+      compatibility_mode:
+        description: |
+          Compatibility modes provide a mean for event owners to evolve their schema, given changes
+          respect the semantics defined by this field.
+
+          It's designed to be flexible enough so that producers can evolve their schemas while not
+          inadvertently breaking existent consumers.
+
+          Once defined, the compatibility mode is fixed, since otherwise it would break a predefined contract,
+          jeopardizing the value of this feature.
```
rcillo: 👍 f888986

api/nakadi-event-bus-api.yaml
rcillo: 👍 it follows the same pattern used by subscription API.

rcillo:
@dehora we've addressed all comments.

Compatibility modes have been clarified, additionalProperties behaviour as well. Version string is semver and return arrays have been converted to objects.

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1541,37 @@ definitions:
...
+
+          - 'deprecated': Schema changes are not allowed. This is used only by legacy event types. This option is not
+            available for new event types. Schema validation is lenient and allow fields that are not defined. Also,
+            it's possible to use the full json schema specification for defining schemas.
+
+          - 'none': Any schema modification is accepted, even if it might break existing producers or consumers.
+
+          - 'compatible': Schema changes never break existing consumers. Every event published since the first version
+            is still valid based on the newest schema. When in compatible mode, it's allowed to add new optional
+            properties and definitions to an existing schema, but no other changes are allowed. If additionalProperties
```
unlaut: Couldn't we prohibit additionalProperties altogether here? In the end, if events are to be archived in the Data Lake additionalProperties can become a huge pain. (E.g. types can change freely any time. And it is hard to imagine how this would integrate with a system that stores security labels or documentation for event attributes.) Is there really a good use case for this?

rcillo: @unlaut we could do that. Actually, I even developed this feature of blocking additionalProperties. But we have some users that rely on that for defining schemas that seems reasonable.

For example, let's suppose I have a schema that holds a Map<String, String>. Let's suppose it's the description of a product with its attributes.

```javascript
{
  "sku": "CM01-R",
  "product_attributes": {
    "color": "red",
    "size": "M"
  }
}
```

That could be achieved by specifying a schema like:

```javascript
{
  "properties": {
    "sku": {
      "type": "string",
    },
    "product_attributes": {
       "additionalProperties": {
         "type": "string"
       }
    }
  }
}
```

If we block additionalProperties altogether (which would make my life easier and also data lake more sane), user would be required to model such events in different ways. One alternative could be to use arrays, like:

```javascript
{
  "sku": "CM01-R",
  "product_attributes": [
    {
      "attribute_name": "color",      
      "attribute_value": "red"
    },
    {
      "attribute_name": "size",      
      "attribute_value": "M",      
    }
  }
}
```

Which schema would look like:

```javascript
{
  "properties": {
    "sku": {
      "type": "string",
    },
    "product_attributes": {
       "type": "array",
       "items": {
         "type": "object",
         "properties": {
           "attribute_name": {
             "type": "string"
           },
           "attribute_value": {
             "type": "string"
           }
         }
       }
    }
  }
}
```

That said, I think it would be completely possible to remove additionalProperties.

rcillo: @conorclifford do you think that it would be possible to remodel your event types using my suggestion above?

unlaut: @rcillo Thanks -- but I am somehow worried about the array based approach as well: 

1. It shares the structural problem of `additionalProperties =true` --- new attributes can sneak in without being explicitly defined
2. We have even less control over data types (all values have to be strings)
3. It's harder to read and process.

Why couldn't the schema for your example be as follows?

`{
  "type": "object",
  "properties": {
    "sku": {
      "type": "string"
    },
    "product_attributes": {
      "type": "object",
      "properties": {
        "color": {
          "type": "string"
        },
        "size": {
          "type": "string"
        }
      },
      "additionalProperties": false
    }
  },
  "additionalProperties": false
}`

rcillo: @unlaut that would be ideal for us. But I'm afraid that it might not be possible. Imagine that attribute name "color" was generated dynamically by some process with human interaction.

It's a hard decision to make. In terms of control, I would prefer more control and explicitness, but on the other hand, it might force users to create "blob" fields to model extensible fields.

rcillo:
@antban @adyach @v-stepanov @dehora I've updated this PR with the results from our latest discussions.

I think we should not be over documenting in the API since this is a very complex subject. I'm going to write a complete section in the gitbook for that. I suggest in the API we focus on the HTTP aspect of it and not so much in the behaviour.

Could you please give a 👍 or 👎

api/nakadi-event-bus-api.yaml
```diff
@@ -1503,11 +1539,39 @@ definitions:
...
+
+          It's designed to be flexible enough so that producers can evolve their schemas while not
+          inadvertently breaking existent consumers.
+
+          Once defined, the compatibility mode is fixed, since otherwise it would break a predefined contract,
+          declared by the producer.
+
+          List of compatibility modes:
+
+          - 'deprecated': Schema changes are not allowed. This is used only by legacy event types. This option is not
```
dehora: i'm wondering if there's a way to not have this option; it's very specific to the state of a deployment. Would it possible to default existing schemas to none and allow them to stay on none or move up to compatible (if the schema in place meets the compatibility rules). 

Alternatively if it's being kept it could be renamed to "fixed" instead of "deprecated".

api/nakadi-event-bus-api.yaml
```diff
@@ -479,7 +475,40 @@ paths:
...
+            properties:
+              items:
+                description: list of schemas.
+                type: array
+                items:
+                  $ref: '#/definitions/EventTypeSchema'
+            required:
+              - items
+
+  '/event-types/{name}/schemas/{version}':
```
rcillo: @dehora every event will be enriched with the version string. For example "1.0.0".

api/nakadi-event-bus-api.yaml
rcillo: > Would it <be> possible to default existing schemas to none and allow them to stay on none

It's not. The `none` mode allows for schema changes indiscriminately. That differs from the `deprecated` which doesn't allow changes. It means that consumers relying on fixed schemas would be broken by changes.

> or move up to compatible (if the schema in place meets the compatibility rules)

That's also not possible since currently, when `additionalProperties` is not used, the validator behaves as if it's `additionalProperties=true`. The new `compatible` mode changes that and makes sure that the behaviour will be like `additionalProperties=false`. So, even if the schema is "upgradable", the event validation behaviour would change, possibly breaking producers.

It will be possible to upgrade the event, but producers should make a PUT to the event type once they are sure it won't break their publishing routine.

> it could be renamed to "fixed" instead of "deprecated"

If it's just a matter of renaming, then I'm fine with that.

But the `deprecated` (or `fixed`) option won't be available for new event types. It's just a way to mark event types that are in the process of been moved to the new event types. All new event types should be either  `compatible` or `none`.

rcillo: Renamed b95fcae

dehora:
👍 
(gr)

would suggest a rebase after 51 commits, but up to you

dehora:
:+1:

PR #470, opened 2016-10-21
ARUHA-406: Avoid Kafka unavailability from cascading to Nakadi

adyach:
This change only for publishing events.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -147,7 +147,6 @@ dependencies {
         exclude module: "json"
     }
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-joda:2.4.0'
-//    compile 'com.netflix.hystrix:hystrix-core:1.5.6'
 
     // tests
     testCompile 'org.hamcrest:hamcrest-all:1.3'
--- a/build.gradle
+++ b/build.gradle
@@ -109,7 +109,7 @@ dependencies {
     compile 'org.springframework.boot:spring-boot-starter-actuator:1.3.1.RELEASE'
     compile 'org.zalando.zmon:zmon-actuator:0.9.3'
 
-    compile 'org.springframework.cloud:spring-cloud-starter-hystrix:1.2.1.RELEASE'
+    compile 'org.springframework.cloud:spring-cloud-starter-hystrix:1.1.3.RELEASE'
 
     // storage
     compile 'org.springframework.boot:spring-boot-starter-jdbc:1.3.1.RELEASE'
--- a/build.gradle
+++ b/build.gradle
@@ -109,6 +109,8 @@ dependencies {
     compile 'org.springframework.boot:spring-boot-starter-actuator:1.3.1.RELEASE'
     compile 'org.zalando.zmon:zmon-actuator:0.9.3'
 
+    compile 'org.springframework.cloud:spring-cloud-starter-hystrix:1.2.1.RELEASE'
+
     // storage
     compile 'org.springframework.boot:spring-boot-starter-jdbc:1.3.1.RELEASE'
     compile 'org.postgresql:postgresql:9.4.1207'
@@ -145,6 +147,7 @@ dependencies {
         exclude module: "json"
     }
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-joda:2.4.0'
+//    compile 'com.netflix.hystrix:hystrix-core:1.5.6'
 
     // tests
     testCompile 'org.hamcrest:hamcrest-all:1.3'
--- a/src/main/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommand.java
+++ b/src/main/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommand.java
@@ -1,0 +1,193 @@
+package org.zalando.nakadi.repository.kafka;
+
+import com.google.common.base.Preconditions;
+import com.netflix.hystrix.HystrixCommand;
+import com.netflix.hystrix.HystrixCommandGroupKey;
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.errors.InterruptException;
+import org.apache.kafka.common.errors.NetworkException;
+import org.apache.kafka.common.errors.NotLeaderForPartitionException;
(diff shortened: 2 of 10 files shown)


build.gradle
```diff
@@ -145,6 +147,7 @@ dependencies {
         exclude module: "json"
     }
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-joda:2.4.0'
+//    compile 'com.netflix.hystrix:hystrix-core:1.5.6'
```
v1ctor: I think you have to remove it then?

adyach: 7d2995b

src/main/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommand.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Command is aimed to prevent calls in cases Kafka is not available.
+ */
+public class KafkaSyncPostCommand extends HystrixCommand<KafkaSyncPostCommand.CommandResult> {
```
v-stepanov: What exactly Hystrix will do here?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommand.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(KafkaSyncPostCommand.class);
+    private static final String GROUP_KEY_SUFFIX = "kafka-sync-post-";
```
rcillo: Is it s suffix or a prefix?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommand.java
```diff
@@ -0,0 +1,195 @@
...
+    private static final int HYSTRIX_COMMAND_TIMEOUT_DELTA = 5000;
```
rcillo: Isn't it better to extract these configurations to application yml?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommand.java
```diff
@@ -0,0 +1,195 @@
...
+            if (hasKafkaConnectionException(exceptions)) {
+                throw eventPublishingException;
+            }
+            return new CommandResult(eventPublishingException);
+        }
+        return new CommandResult();
+    }
+
+    private boolean hasKafkaConnectionException(final List<Exception> exceptions) {
+        return exceptions.stream().filter(ex ->
```
rcillo: What about `anyMatch`?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommand.java
```diff
@@ -0,0 +1,195 @@
...
+            throw new EventPublishingException("Error publishing message to kafka", e);
+        } catch (final RuntimeException e) {
+            item.updateStatusAndDetail(EventPublishingStatus.FAILED, "internal error");
+            throw new EventPublishingException("Error publishing message to kafka", e);
+        }
+    }
+
+    public static class CommandResult {
+
+        private final EventPublishingException exception;
```
rcillo: What about using subclassing to differentiate between a successful command from a failed one? We could have abstract CommandResult and then SuccessfulCommandResult and FailedCommandResult. What do you think?

adyach: @rcillo I thought about it but there are no so many cases for that. It is very specific case when we have to return one exception and throw another. I think it is easier to understand with less code.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommand.java
adyach: Thanks, fixed 5a403ff

src/main/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommand.java
adyach: Yep, you are right, thanks 5a403ff

src/main/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommand.java
adyach: fixed 5a403ff

src/test/java/org/zalando/nakadi/repository/kafka/KafkaSyncPostCommandTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+    @Test
+    public void testUnknownTopicOrPartitionException() {
+        final BatchItem batchItem = new BatchItem(new JSONObject());
+        batchItem.setPartition("0");
+        mockedKafkaProducer.setException(new UnknownTopicOrPartitionException());
+
+        final KafkaSyncPostCommand.CommandResult result = createCommand(batchItem).execute();
+
+        Assert.assertTrue(result.isFailed());
```
rcillo: When does hystrix increase the failure counter? Is it when an exception is launched in the `execute` command?

rcillo: Ok, got it. It considers a failure when an exception is launched during execution.

rcillo:
As we discussed previously, we have some concerns on the impact of this feature since it fails entire topics, even if they have many partitions in different brokers.

We analysed the possibility to group errors on a per broker basis, but it seems to require a bigger effort. We also considered grouping per topic-partition which is also hard to implement.

So I would like to test how it behaves first to make sure we already got some improvement over the status quo, which is dramatically failing everything.

rcillo:
:shipit:

adyach:
@rcillo I've fixed tests parameters, pls, approve again.

rcillo:
:shipit:

PR #471, opened 2016-10-25
Revert "ARUHA-406: Avoid Kafka unavailability from cascading to Nakadi"

adyach:
Reverts zalando/nakadi#470

We decided do not merge into the master right now but do testing first.


v-stepanov:
:shipit:

PR #472, opened 2016-10-28
ARUHA-414: update Kafka client to 0.9.0.1

adyach:


v-stepanov:
@adyach What is the difference from the previous version? Will it help us?

adyach:
For example in this particular release they fixed:
[KAFKA-3159] - Kafka consumer 0.9.0.0 client poll is very CPU intensive under certain conditions
[KAFKA-2877] - Messages sometimes not delivered by new consumer after Kafka restart
Here is the full [change log](https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12311720&version=12333395)

> Will it help us?

@v-stepanov in general, I do not think it will help us significantly but we have unexplainable mysterious issues with Kafka and do not use the latest stable updates.

rcillo:
We run kafka server version 0.9.0.1 but client older than that. As I understood, this is a way to make versions match, right?

adyach:
@rcillo yes, this is one of the points as well!
https://github.bus.zalan.do/aruha/bubuku-appliance/blob/master/Dockerfile#L4

rcillo:
:shipit:

PR #473, opened 2016-10-31
aruha-355: fixed admin access when check owning app feature is on

adyach:


rcillo:
:shipit:

rcillo:
:shipit:

rcillo:
:shipit:

PR #475, opened 2016-11-01
Aruha 400 update schema - part 1

rcillo:
This PR introduces the validation of forbidden features, such as "not" and "patternProperties".

From now on it won't be possible to create event types whose schemas contains these attributes.

The reasoning behind these decisions is part of the concept https://docs.google.com/document/d/1FmGHxI9oklxN16LKZe1UG_U21jtQFVCtLtRiC2W1pAk/edit

Next steps:

- introduce compatibility mode attribute to event types
- introduce version
- introduce schema update with compatibility checks


src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Stack;
+
+@Component
+public class SchemaCompatibilityChecker {
+
+    final private List<SchemaConstraint> CONSTRAINTS = Lists.newArrayList();
+
+    public SchemaCompatibilityChecker() {
+        CONSTRAINTS.add(new NotSchemaConstraint());
+        CONSTRAINTS.add(new PatternPropertiesConstraint("patternProperties"));
```
v-stepanov: Why not to inject these classes?
For example to make a list of SchemaConstraint a Spring bean and to inject it in constructor.
The way you made it makes  SchemaCompatibilityChecker very bound with NotSchemaConstraint and PatternPropertiesConstraint. For example in your unit test for SchemaCompatibilityChecker you actually also test these two classes.

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
```diff
@@ -0,0 +1,116 @@
...
+            }
+        }
+
+        if (schema instanceof ObjectSchema) {
+            recursiveCheckConstraints((ObjectSchema) schema, jsonPath, schemaIncompatibilities);
+        } else if (schema instanceof ArraySchema) {
+            recursiveCheckConstraints((ArraySchema) schema, jsonPath, schemaIncompatibilities);
+        } else if (schema instanceof ReferenceSchema) {
+            recursiveCheckConstraints((ReferenceSchema) schema, jsonPath, schemaIncompatibilities);
+        } else if (schema instanceof CombinedSchema) {
```
v-stepanov: why do you need all these 4 checks and castings? Will not it just work with:
`recursiveCheckConstraints(schema, jsonPath, schemaIncompatibilities);`

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
rcillo: 👍 good idea. Gonna do that.

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
rcillo: Actually I didn't try. This is the result of my latest refactoring. I didn't realise that maybe I could just remove the casting.

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
```diff
@@ -0,0 +1,116 @@
...
+
+        return incompatibilities;
+    }
+
+    private void recursiveCheckConstraints(
+            final Schema schema,
+            final Stack<String> jsonPath,
+            final List<SchemaIncompatibility> schemaIncompatibilities) {
+
+        for (final SchemaConstraint constraint : CONSTRAINTS) {
```
rcillo: I lost the interest because of the lack of functional data structures. For example, list manipulation in functional languages is a breeze, but in Java it's ackward. I even decided to use a `Stack` to keep recursion state =/

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
v-stepanov: No, sorry, It will not work. It seems that in Java overloading resolution is performed at compile time. :(

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
```diff
@@ -0,0 +1,116 @@
...
+
+    final private List<SchemaConstraint> CONSTRAINTS = Lists.newArrayList();
+
+    public SchemaCompatibilityChecker() {
+        CONSTRAINTS.add(new NotSchemaConstraint());
+        CONSTRAINTS.add(new PatternPropertiesConstraint("patternProperties"));
+    }
+
+    public List<SchemaIncompatibility> checkConstraints(final Schema schema) {
+        final List<SchemaIncompatibility> incompatibilities = new ArrayList<SchemaIncompatibility>();
```
v-stepanov: can be just diamond
`new ArrayList<>();`

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
```diff
@@ -0,0 +1,116 @@
...
+
+        recursiveCheckConstraints(schema, new Stack<String>(), incompatibilities);
```
v-stepanov: also a diamond can be used

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
rcillo: Done 64b79f6

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -215,6 +221,16 @@ private void validateSchema(final EventType eventType) throws InvalidEventTypeEx
         }
     }
 
+    private void validateJsonSchemaConstraints(final Schema schema) throws InvalidEventTypeException {
+        final List<SchemaIncompatibility> incompatibilities = schemaCompatibilityChecker.checkConstraints(schema);
+
+        if (!incompatibilities.isEmpty()) {
+            final String erroMessage = incompatibilities.stream().map(Object::toString)
```
rcillo: Done ca8f2d8

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
rcillo: Actually I think it's not possible because of the way Java binds these methods. If I didn't do the casts, it would bind to this same method signature and stack overflow.

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
rcillo: Done 9a4021a

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
rcillo: Done 53b5d63

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
rcillo: I wish I had pattern matching and functional data structures...

rcillo:
@v-stepanov thanks for the review. I addressed all comments. Looks better now?

v-stepanov:
@rcillo Yeah, sorry, forgot to approve. Now everything is great.

v-stepanov:
👍

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
```diff
@@ -0,0 +1,114 @@
...
+            final List<SchemaIncompatibility> schemaIncompatibilities) {
+
+        for (final SchemaConstraint constraint : constraints) {
+            final Optional<SchemaIncompatibility> incompatibility = constraint.validate(jsonPath, schema);
+            if (incompatibility.isPresent()) {
+                schemaIncompatibilities.add(incompatibility.get());
+            }
+        }
+
+        if (schema instanceof ObjectSchema) {
```
adyach: From brief look: isn't it possible to use polymorphism here?
Every schema knows how to check constraints:
`schema.recursiveCheckConstraints(jsonPath, schemaIncompatibilities)`

src/main/java/org/zalando/nakadi/validation/schema/PatternPropertiesConstraint.java
```diff
@@ -0,0 +1,24 @@
...
+import org.everit.json.schema.ObjectSchema;
+import org.everit.json.schema.Schema;
+import org.zalando.nakadi.validation.SchemaIncompatibility;
+
+import java.util.List;
+import java.util.Optional;
+
+public class PatternPropertiesConstraint extends SchemaConstraint {
+
+    static final String ATTRIBUTE = "patternProperties";
```
adyach: private ?

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
rcillo: It's not. We do not own "Schema" class since it's from a third library. We could wrap then to some Nakadi owned class, but I considered it would be too much indirection for a small gain.

src/main/java/org/zalando/nakadi/validation/schema/PatternPropertiesConstraint.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @Override
+    public Optional<SchemaIncompatibility> validate(final List<String> jsonPath, final Schema schema) {
+        if (schema instanceof ObjectSchema) {
+            final ObjectSchema objectSchema = (ObjectSchema) schema;
+            if (!objectSchema.getPatternProperties().isEmpty()) {
+                return Optional.of(new ForbiddenAttributeIncompatibility(jsonPathString(jsonPath), this.ATTRIBUTE));
```
adyach: without `this` or PatternPropertiesConstraint.ATTRIBUTE

src/main/java/org/zalando/nakadi/validation/schema/PatternPropertiesConstraint.java
rcillo: Fixed afabd8d

src/main/java/org/zalando/nakadi/validation/schema/PatternPropertiesConstraint.java
rcillo: Fixed afabd8d

rcillo:
@v-stepanov @adyach I need a new 👍 after changes.

src/main/java/org/zalando/nakadi/validation/SchemaCompatibilityChecker.java
adyach: I think there are two ways:
1. extend/wrap schema library objects, but it won't solve the issue of casting I will have to cast anyway.
2. the second approach is to add a little bit of code to the library and build it.
All approaches have their own downsides. Maybe we can solve it later if we need support for more constraints/schema types.

adyach:
:shipit:

PR #477, opened 2016-11-02
aruha-355: changed api according to guidlines and task validation

adyach:
I would like to send this api to guild but first it will be nice to have an agreement within the team. So you are welcome to comment.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -390,6 +390,8 @@ paths:
               description: |
                 Number of seconds the client ought to wait before making a follow-up request.
               type: number
+          schema:
+            $ref: '#/definitions/Problem'
 
     get:
       tags:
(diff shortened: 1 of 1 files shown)


api/nakadi-event-bus-api.yaml
```diff
@@ -1054,7 +1082,7 @@ paths:
                   $ref: '#/definitions/Feature'
             required:
               - items
-    post:
+    put:
```
antban: I don't think that it's a good idea

rcillo:
👍 looks great.

v-stepanov:
I actually thought that the /settings is our internal API and it should not be in swagger file
If we want to expose that then I'm strongly against word "flooder" it's misleading.

v-stepanov:
Looks good for me now

v-stepanov:
👍

v-stepanov:
👍

rcillo:
👍

PR #478, opened 2016-11-02
ARUHA-349 Log exceptions from kafka

antban:
... and correctly calculate producer use count

### DIFF
--- a/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -177,6 +177,7 @@ public class KafkaTopicRepository implements TopicRepository {
 
             producer.send(kafkaRecord, ((metadata, exception) -> {
                 if (null != exception) {
+                    LOG.warn("Failed to publish to kafka topic {}", topicId, exception);
                     item.updateStatusAndDetail(EventPublishingStatus.FAILED, "internal error");
                     result.complete(exception);
                 } else {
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -177,6 +177,7 @@ public boolean partitionExists(final String topic, final String partition) throw
 
             producer.send(kafkaRecord, ((metadata, exception) -> {
                 if (null != exception) {
+                    LOG.warn("Failed to publish to kafka topic " + topicId, exception);
```
adyach: LOG.warn("Failed to publish to kafka topic {}", topicId, exception);

antban: I thought that error is really error. 
Now it may be just information about leader change and so on... For now it is just warning

antban: If the warning is present for a long time - than it's an error.

adyach:
For me looks fine 👍

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -177,6 +177,7 @@ public boolean partitionExists(final String topic, final String partition) throw
 
             producer.send(kafkaRecord, ((metadata, exception) -> {
                 if (null != exception) {
+                    LOG.warn("Failed to publish to kafka topic {}", topicId, exception);
```
antban: @rcillo It seems that I lost initial comment.
My point is that for now it's just a warning, because it may be normal system behavior (for example transferring leadership).
In case if warning is present for a long time, than it's an error.

v1ctor:
👍

rcillo:
👍

PR #481, opened 2016-11-09
ARUHA-406: Kafka requests isolation concept

adyach:
https://github.com/zalando/nakadi/wiki/Kafka-requests-isolation-concept


src/main/resources/application.yml
```diff
@@ -22,6 +22,12 @@ spring:
     maxActive: 20
     testOnBorrow: true
     validationQuery: SELECT 1
+
+hystrix.command.default.execution.isolation.strategy: SEMAPHORE
+hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds: 15000
+hystrix.command.default.circuitBreaker.requestVolumeThreshold: 20
+hystrix.command.default.circuitBreaker.sleepWindowInMilliseconds: 5000
```
v-stepanov: What these params mean exactly? That if 20 errors happen withing 5 sec - then it will break the circuit for 15 sec?

v-stepanov:
@adyach @v1ctor @antban @rcillo 
Sorry for bringing this discussion back again, but I think we need to reconsider if this will bring more value than problems. When we were discussing this, as an example we had a "slow broker" problem and this implementation will help to handle this problem pretty well.

But I would like to bring to discussion another problem which happens much more often in our case. When rebalance happens we have quite often the kafka errors for the topic that changes a leader. The amount of such errors during single leader switch is enough to trigger circuit breaker (if the event publishing rate was high enough for that topic) - and that will block the whole broker for some seconds, so the single leader switch of one topic can make it impossible to publish to 100 other topics for many seconds. And when we add a new node to the cluster we have a lot of leader changes.  

I don't know, maybe it is possible to avoid this problem with more precise parameters selection for Hystrix. But I think we should discuss that again.
Or do I miss something maybe?

adyach:
@v-stepanov 
I would like to clarify how I process this and how Hystrix works in that way:
In simple words Hystrix counts number of failures by thrown exceptions from HystrixCommand.run().
According to concept I process only these exceptions:
https://github.com/zalando/nakadi/wiki/Kafka-requests-isolation-concept#short-circuit-criteria
In the ProducerSendCommand there is a method hasKafkaConnectionException which checks for the exceptions and throws exception from Hystrix command only in case we have connection issues.

Do you think it is not enough and we should implement other cases?

v-stepanov:
@adyach ahh, ok I see now. So for Hystrix it will be an error only in a case of timeout or network problem. So in a case of error during leader switch for Hystrix it will be 'ok'.
Ok, so problem I described will not happen. Good.

src/main/resources/application.yml
adyach: @v-stepanov this property is overwritten during command init to Nakadi send + Kafka send timeout.

src/main/java/org/zalando/nakadi/repository/kafka/ProducerSendCommand.java
```diff
@@ -0,0 +1,123 @@
...
+    private final KafkaFactory kafkaFactory;
+    private final String topicId;
+    private final BatchItem batchItem;
+    private final long timeout;
+
+    protected ProducerSendCommand(final KafkaFactory kafkaFactory,
+                                  final String topicId,
+                                  final BatchItem batchItem,
+                                  final long timeout) {
+        super(HystrixCommandGroupKey.Factory.asKey(GROUP_PREFIX + batchItem.getBrokerId()), (int) timeout);
```
v-stepanov: so the timeout for Hystrix command is exactly the same as for CompletableFuture. Which one of them you expect to trigger earlier?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -164,103 +156,44 @@ public boolean partitionExists(final String topic, final String partition) throw
...
-        }
-        return Stream.of(NotLeaderForPartitionException.class, UnknownTopicOrPartitionException.class).
-                filter(clazz -> clazz.isAssignableFrom(exception.getClass()))
-                .findAny().isPresent();
-    }
-
     @Override
-    public void syncPostBatch(final String topicId, final List<BatchItem> batch) throws EventPublishingException {
-        batch.forEach(item -> Preconditions.checkNotNull(
+    public void syncPostBatch(final String topicId, final List<BatchItem> batches) throws RuntimeException {
```
antban: It is not batches, it is batch items.

src/main/resources/application.yml
adyach: I thought it will be useful to have it because we can create other commands and their timeout is taken from that property it will be more clear. maybe I should remove it for now

v-stepanov: Ah ok, so you keep it as a global setting for general hystrix command. 
For me it's fine if you keep it.

src/main/java/org/zalando/nakadi/repository/kafka/ProducerSendCommand.java
adyach: There is not much difference who is gonna be triggered first because in anyway command is timed out. I thought about adding some kind of `timeout + delta` but it adds additional complexity to the code and not so much value.

src/main/java/org/zalando/nakadi/repository/kafka/ProducerSendCommand.java
```diff
@@ -0,0 +1,123 @@
...
+                if (exception != null) {
+                    if (isExceptionShouldLeadToReset(exception)) {
+                        LOG.warn("Terminating producer while publishing to topic {}", topicId, exception);
+                        kafkaFactory.terminateProducer(producer);
+                    }
+
+                    if (hasKafkaConnectionException(exception)) {
+                        LOG.error("Kafka timeout: {} on broker {}", topicId, batchItem.getBrokerId(), exception);
+                        batchItem.updateStatusAndDetail(EventPublishingStatus.FAILED, "timed out");
+                        throw new Exception("Kafka timeout exception");
```
v-stepanov: Why you throw plain Exception ?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -164,103 +156,44 @@ public boolean partitionExists(final String topic, final String partition) throw
...
-                .anyMatch(item -> item.getResponse().getPublishingStatus() == EventPublishingStatus.FAILED);
-        if (atLeastOneFailed) {
-            failUnpublished(batch, "internal error");
-            throw new EventPublishingException("Error publishing message to kafka");
-        }
+
+        batches.stream().forEach(batch -> batch.setBrokerId(partitionInfos.stream()
+                .filter(pi -> pi.partition() == Integer.valueOf(batch.getPartition()))
+                .map(pi -> pi.leader().id())
+                .findFirst().get()));
```
antban: Sorry, I'am against these lines. PartitionInfos is an array with indexes. Index is a partition number
```
batches.forEach(item -> item.setBrokerId(partitionInfos[batch.getPartition()].leader().id())
```

src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -1,13 +1,15 @@
...
 import org.json.JSONObject;
 
+import javax.annotation.Nullable;
+import java.util.Optional;
+
 public class BatchItem {
     private final BatchItemResponse response;
     private final JSONObject event;
     private String partition;
+    private int brokerId;
```
antban: Batch item is a view object, and with creation of brokerId in this object we are linking to kafka implementation.

src/main/java/org/zalando/nakadi/repository/kafka/ProducerSendCommand.java
```diff
@@ -0,0 +1,123 @@
...
+            batchItem.updateStatusAndDetail(EventPublishingStatus.FAILED, "internal error");
+            LOG.error("Error publishing message to kafka", ex);
+        } catch (final InterruptedException ex) {
+            Thread.currentThread().interrupt();
+            batchItem.updateStatusAndDetail(EventPublishingStatus.FAILED, "interrupted");
+            LOG.error("Error publishing message to kafka", ex);
+        } catch (final java.util.concurrent.TimeoutException ex) {
+            batchItem.updateStatusAndDetail(EventPublishingStatus.FAILED, "timed out");
+            LOG.error("Kafka timeout: {} on broker {}", topicId, batchItem.getBrokerId(), ex);
+            throw new Exception("Kafka timeout exception");
```
antban: Why thread is not interrupted?

src/main/java/org/zalando/nakadi/repository/kafka/ProducerSendCommand.java
```diff
@@ -0,0 +1,123 @@
...
+                    topicId,
+                    KafkaCursor.toKafkaPartition(batchItem.getPartition()),
+                    batchItem.getPartition(),
+                    batchItem.getEvent().toString());
+
+            final Producer<String, String> producer = kafkaFactory.takeProducer();
+            try {
+                final NakadiCallback callback = new NakadiCallback();
+                producer.send(kafkaRecord, callback);
+                final Exception exception = callback.result.get(timeout, TimeUnit.MILLISECONDS);
```
antban: That means that commands executed one by one, or histryx will create several threads to process commands?
In case of second approach - how many threads? There are batches of 600 events...

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -164,103 +156,44 @@ public boolean partitionExists(final String topic, final String partition) throw
...
             kafkaFactory.releaseProducer(producer);
         }
-        final boolean atLeastOneFailed = batch.stream()
-                .anyMatch(item -> item.getResponse().getPublishingStatus() == EventPublishingStatus.FAILED);
-        if (atLeastOneFailed) {
-            failUnpublished(batch, "internal error");
-            throw new EventPublishingException("Error publishing message to kafka");
-        }
+
+        batches.stream().forEach(batch -> batch.setBrokerId(partitionInfos.stream()
```
rcillo: It would be more accurate to replace  `batch` by `batchItem` or just `item`.

rcillo: Instead of going through the list of `partitionInfos` for every `batchItem`, I would suggest to build a Map of partition -> brokerId only once. This code is executed a lot and I think it would be more performant to access a Map instead of loop through for every batch item.

rcillo: Well, if @antban comment bellow is true, then it's even better. This array can be used as "Map" where index is the partition and value is the broker.

adyach:
@antban @rcillo @v-stepanov addressed your comments. working on tests.

src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -31,6 +34,22 @@ public String getPartition() {
...
+
+    public String getBrokerId() {
+        return brokerId;
+    }
+
+    public String getTopic() {
+        return topic;
+    }
+
+    public void setTopic(String topic) {
```
v-stepanov: classic comment about "final" :)

antban:
+1

antban:
:+1:

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -164,103 +157,83 @@ public boolean partitionExists(final String topic, final String partition) throw
...
+                Preconditions.checkNotNull(
+                        item.getPartition(), "BatchItem partition can't be null at the moment of publishing!");
+                item.setBrokerId(partitionToBroker.get(item.getPartition()));
+                item.setTopic(topicId);
+            });
+
+            // send message to Kafka in asynch mode, it will fail sending in circuit is opened
+            final List<ProducerSendCommand.NakadiCallback> callbacks = batchItems.stream()
+                    .peek(item -> item.setStep(EventPublishingStep.PUBLISHING))
+                    .map(item -> new ProducerSendCommand(producer, item).execute())
```
rcillo: @adyach I know this might be too late feedback. I'm afraid that creating one `ProducerSendCommand` per item might cause short circuit too often and too early. We have some batches with hundreds of events. If one batch times out, we would have an open circuit caused by a single request.

If that's the case, is it possible for us to disable Hystrix? I mean, if my hypothesis is proven to be true and we have open circuits when we don't want it to open, is it possible to disable hystrix?

rcillo: I got it. We just open the circuit on Kafka timeouts. :shipit:

adyach: Explained details to Ricardo in offline.

adyach:
Reopened in #493

PR #482, opened 2016-11-09
aruha-400 prepare for unknown attributes

rcillo:
In order to introduce some new fields to event types (compatibility mode
and schema versioning), I plan to first run a migration that would do
so.

But once we run this database migration that adds such fields to event
types, it would break the running version, since it would introduce new
fields that are not known to the current stack.

In order to avoid this incompatibilities, I will first release this
feature first, which ignores unknow attributes instead of breaking.


v-stepanov:
What about a simple test to verify that it works? (and that it will not broken in the future)

rcillo:
@v-stepanov done. Added a test case to demonstrate that it actually works.

In the process I found out a test that wasn't testing what it was supposed to test and fixed it in the Subscription workflow.

v-stepanov:
👍

PR #484, opened 2016-11-10
Aruha 400 schema compatibility modes

rcillo:
This PR introduces the concept of a schema compatibility mode and schema versioning.

The actual rules for changing a schema are not yet implemented in this PR. That's the next thing to be done in `CompatibleSchemaChangeConstraint` class.

Here you'll find some new validations when creating and updating an event type, as well as a new table for storing event type schemas.

I considered having two schema classes, one for creation/update and another for exhibition, much like it's done with `Subscription` and `SubscriptionBase` but it would be much harder here since it's nested inside `EventType` and such change would imply in the creation of `EventTypeBase` as well.

So I preferred to have some "nulls" around (for version and created_at fields) and handle them properly.


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -211,6 +208,11 @@ private void validateSchema(final EventType eventType) throws InvalidEventTypeEx
                 throw new InvalidEventTypeException("\"metadata\" property is reserved");
             }
 
+            if (eventType.getCompatibilityMode() == CompatibilityMode.DEPRECATED) {
+                throw new InvalidEventTypeException(
+                        "\"compatibility_mode\" should be either \"compatible\" or \"none\"");
```
v-stepanov: this is done because only old schemas will be "deprecated", and the new ones should never become "deprecated", right?

src/main/java/org/zalando/nakadi/validation/schema/DeprecatedSchemaChangeConstraint.java
```diff
@@ -0,0 +1,20 @@
+package org.zalando.nakadi.validation.schema;
+
+import org.zalando.nakadi.domain.CompatibilityMode;
+import org.zalando.nakadi.domain.EventType;
+
+import java.util.Optional;
+
+public class DeprecatedSchemaChangeConstraint implements SchemaEvolutionConstraint {
+    private final String errorMessage = "schema from an event types with deprecated compatibility mode cannot not be " +
```
v-stepanov: cannot not

database/nakadi/10_data/04_tables/03_event_type_schema.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE IF NOT EXISTS zn_data.event_type_schema (
+  ets_event_type_name varchar(255) NOT NULL REFERENCES zn_data.event_type (et_name),
+  ets_schema_object jsonb NOT NULL
+);
+
+CREATE INDEX ON zn_data.event_type_schema USING gin (ets_schema_object);
+
+CREATE INDEX ON zn_data.event_type_schema ((ets_schema_object->>'version'));
+CREATE INDEX ON zn_data.event_type_schema ((ets_schema_object->>'created_at'));
```
v-stepanov: I think you can also create unique index for pair (event-type-name, version). That will guarantee the uniqueness of that pair from DB point of view.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
rcillo: @v-stepanov correct. Also, deprecated schemas should not be upgradable to compatible neither become "none". At least in this PR. There is another ticket to allow for upgrading a deprecated schema into one of the other two options. It will be done next and will require some additional checks.

src/main/java/org/zalando/nakadi/repository/db/EventTypeDbRepository.java
```diff
@@ -70,6 +85,13 @@ public void update(final EventType eventType) throws InternalNakadiException {
         }
     }
 
+    private void insertEventTypeSchema(final EventType eventType) throws JsonProcessingException {
+        jdbcTemplate.update(
+                "INSERT INTO zn_data.event_type_schema (ets_event_type_name, ets_schema_object ) VALUES (?, ?::jsonb)",
```
v-stepanov: extra space before the closing bracket 
`ets_schema_object )`

src/main/java/org/zalando/nakadi/repository/db/EventTypeDbRepository.java
```diff
@@ -31,14 +35,18 @@ public EventTypeDbRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper
     }
 
     @Override
+    @Transactional
```
rcillo: Added to remove fd4709b

My goal with this annotation is to make sure that we do not have "half-done" operations. I mean, that insertion into event_type works but then insertion into "schema" fails. With this annotation it should commit only after both operations are done or none (rollback).

database/nakadi/10_data/04_tables/03_event_type_schema.sql
rcillo: Done 75cc73f

src/acceptance-test/java/org/zalando/nakadi/repository/db/AbstractDbRepositoryTest.java
```diff
@@ -20,10 +20,10 @@
     protected JdbcTemplate template;
     protected Connection connection;
     protected ObjectMapper mapper;
-    protected String repositoryTable;
+    protected String[] dependencyTables;
 
-    public AbstractDbRepositoryTest(final String repositoryTable) {
-        this.repositoryTable = repositoryTable;
+    public AbstractDbRepositoryTest(final String[] dependencyTables) {
```
rcillo: I am Java coder from 2000's 👴

src/acceptance-test/java/org/zalando/nakadi/webservice/UserJourneyAT.java
```diff
@@ -164,6 +173,17 @@ public void userJourneyM1() throws InterruptedException, IOException {
                 .statusCode(NOT_FOUND.value());
     }
 
+    private String getUpdateEventType() throws IOException {
```
v-stepanov: What's the story here? Why you had to create this method? Is it because created_at should not be changed?

rcillo: @v-stepanov correct.

v-stepanov: Can user just skip this field when doing an update?

src/test/java/org/zalando/nakadi/utils/EventTypeTestBuilder.java
```diff
@@ -119,10 +127,23 @@ public EventTypeTestBuilder readScopes(final Set<String> readScopes) {
...
+        this.compatibilityMode = compatibilityMode;
+        return this;
+    }
+
+    public EventTypeTestBuilder version(final String version) {
+        final EventTypeSchema schema = new EventTypeSchema();
+        schema.setSchema(this.schema.getSchema());
+        schema.setVersion(new Version(version));
+        this.schema = schema;
+        return this;
```
v-stepanov: why not just
```java
this.schema.setVersion(new Version(version));
return this;
```

src/acceptance-test/java/org/zalando/nakadi/webservice/UserJourneyAT.java
rcillo: @v-stepanov done 83735e3

src/test/java/org/zalando/nakadi/validation/SchemaEvolutionServiceTest.java
```diff
@@ -0,0 +1,78 @@
...
+
+        verify(evolutionConstraint).validate(oldEventType, newEventType);
+    }
+
+    @Test
+    public void checksJsonSchemaConstraints() throws Exception {
+        final JSONArray invalidTestCases = new JSONArray(
+                readFile("org/zalando/nakadi/validation/invalid-json-schema-examples.json"));
+
+        for(final Iterator<Object> i = invalidTestCases.iterator(); i.hasNext();) {
```
v-stepanov: Hm... I don't know if using iterator makes sense here, because usual foreach loop will do pretty fine here, and using iterator just makes code more complicated.

src/test/java/org/zalando/nakadi/utils/EventTypeTestBuilder.java
rcillo: @v-stepanov you are right, actually it was unnecessary. Just removed it 2c5e72d

src/test/java/org/zalando/nakadi/validation/SchemaEvolutionServiceTest.java
rcillo: @v-stepanov done. I just actually needed to do some casting 8e1eb3d

I hate manipulating json in java.

src/main/java/org/zalando/nakadi/domain/EventTypeSchema.java
```diff
@@ -37,6 +46,24 @@ public void setSchema(final String schema) {
         this.schema = schema;
     }
 
+    @Nullable
```
rcillo: @v-stepanov Correct. I think it should be like so. If a null shows up it means that there is a bug in the code. It should not be treated. NullPointer whould raise and we should fix the root cause.

v-stepanov: If this property being a null ia a bug situation, then why is it marked as `@Nullable` ?

rcillo: @v-stepanov I tried to explain in the PR description. EventTypeSchema might have `null` `version` and `createdAt` fields when updating. If a user is publishing a new schema, it doesn't have a version yet, because Nakadi will automatically generate such version for the user.

So, even though getting this attribute might lead to `null`, hence the `@Nullable`, it should never be so in actual usages, because we would have set it to some value automatically. That's what I mean by if a null shows up, then it's a bug.

It doesn't make sense to treat "nulls" since that would be silently covering a bug (if we failed to automatically set the version, for example).

rcillo: I'll give a try on implementing an EventTypeBase and EventTypeSchemaBase

database/migration/aruha-400-schema-evolution.sql
```diff
@@ -0,0 +1,24 @@
...
+CREATE INDEX ON zn_data.event_type_schema USING gin (ets_schema_object);
+
+CREATE INDEX ON zn_data.event_type_schema ((ets_schema_object->>'version'));
+CREATE INDEX ON zn_data.event_type_schema ((ets_schema_object->>'created_at'));
+
+CREATE UNIQUE INDEX ON zn_data.event_type_schema ((ets_schema_object->>'version'),
+                                                  (ets_event_type_name));
+
+UPDATE zn_data.event_type
+SET et_event_type_object  = jsonb_set(et_event_type_object, '{schema,version}', '"1.0.0"', false);
```
adyach: jsonb_set is only for postgres 9.5. in production we use 9.4 :(

src/main/java/org/zalando/nakadi/domain/Version.java
```diff
@@ -0,0 +1,51 @@
+package org.zalando.nakadi.domain;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+public class Version {
+    private final int major;
```
adyach: I agree that it is more clear to have versions as integers but from other side you never use them in the code and all these housekeeping work requires resources + we claim that we use `semantic versioning` which support string at patch level. We can just use String inside that object and extend it if necessary, wdyt?

rcillo: I think it's going to help in the next story which will manipulate the version number. We already use for implementing comparison. If at the end we see that it was not useful, I can easily change it.

database/migration/aruha-400-schema-evolution.sql
rcillo: Hun, that's a problem. I just checked AWS documentation and they claim to have 9.5. Maybe we are using an outdated Postgres? I prefer to upgrade it than change the code =)

src/main/java/org/zalando/nakadi/validation/schema/ChangeVersionAndCreatedAtConstraint.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.Optional;
+
+public class ChangeVersionAndCreatedAtConstraint implements SchemaEvolutionConstraint {
+
+    @Override
+    public Optional<SchemaEvolutionIncompatibility> validate(final EventType original, final EventType eventType) {
+        if (original.getSchema().equals(eventType.getSchema())) {
+            if (!original.getSchema().getVersion().equals(eventType.getSchema().getVersion())) {
+                return Optional.of(new SchemaEvolutionIncompatibility("changing schema version is not allowed"));
```
rcillo: Sounds good. I'm gonna give a try.

database/migration/aruha-400-schema-evolution.sql
rcillo: @adyach I'm going to update it so =) it's about time.

src/main/java/org/zalando/nakadi/validation/schema/ChangeVersionAndCreatedAtConstraint.java
rcillo: @adyach I end up removing this class in my latest change. I introduced two new classes EventTypeBase and EventTypeSchemaBase which removes the need for this class, since it made it impossible for users to change attributes we don't want them to change.

adyach: @rcillo I meant in general for classes like *Constraint.java :)

PR #487, opened 2016-11-21
aruha-400 validate schema evolution

rcillo:
This PR contains the actual validation of what should be allowed when evolving a json-schema.


src/main/java/org/zalando/nakadi/validation/schema/PartitionKeyFieldsConstraint.java
```diff
@@ -0,0 +1,17 @@
+package org.zalando.nakadi.validation.schema;
+
+import org.zalando.nakadi.domain.EventType;
+import org.zalando.nakadi.domain.EventTypeBase;
+
+import java.util.Optional;
+
+public class PartitionKeyFieldsConstraint implements SchemaEvolutionConstraint {
```
rcillo: @v-stepanov I wasn't sure about the classification of this constraint. One could say that it's about the event type not about the schema, but it references schema properties, so I considered it to be a schema constraint.

The reason why it's not allowed to be changed is to not break order guarantees.

src/main/java/org/zalando/nakadi/validation/schema/PartitionStrategyConstraint.java
```diff
@@ -0,0 +1,17 @@
+package org.zalando.nakadi.validation.schema;
+
+import org.zalando.nakadi.domain.EventType;
+import org.zalando.nakadi.domain.EventTypeBase;
+
+import java.util.Optional;
+
+public class PartitionStrategyConstraint implements SchemaEvolutionConstraint {
```
rcillo: All these constraints are related to event type properties related to the schema. I couldn't think about a better name. Related to the this comment as well https://github.com/zalando/nakadi/pull/487#discussion_r89524324

v-stepanov:
👍

PR #488, opened 2016-11-22
Aruha 455

v1ctor:
Now it will look like this "[2016-11-22 11:27:00.071Z] [INFO] [D33T0mwjKsZCeb4gGngode4G] [http-nio-8080-exec-4] [org.zalando.nakadi.filters.LoggingFilter] --- GET "/event-types/express_shipment_offered" "curl/7.49.1" "-" 404 61 ms"

### DIFF
--- a/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
+++ b/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
@@ -23,20 +23,26 @@ public class LoggingFilter extends OncePerRequestFilter {
                                     final HttpServletResponse response, final FilterChain filterChain)
             throws ServletException, IOException {
         final long start = System.currentTimeMillis();
-
+        final String userAgent = Optional.ofNullable(request.getHeader("User-Agent")).orElse("-");
+        final String user = Optional.ofNullable(request.getUserPrincipal()).map(Principal::getName).orElse("-");
+        final String method = request.getMethod();
+        final String path = request.getRequestURI();
+        LOG.info("{} \"{}\" \"{}\" \"{}\"",
+                method,
+                path,
+                userAgent,
+                user);
         //execute request
         try {
             filterChain.doFilter(request, response);
         } finally {
             final long time = System.currentTimeMillis();
             final Long timing = time - start;
-            final Optional<String> userAgent = Optional.ofNullable(request.getHeader("User-Agent"));
-            final Optional<String> user = Optional.ofNullable(request.getUserPrincipal()).map(Principal::getName);
             LOG.info("{} \"{}\" \"{}\" \"{}\" {} {} ms",
-                    request.getMethod(),
-                    request.getRequestURI(),
-                    userAgent.orElse("-"),
-                    user.orElse("-"),
+                    method,
+                    path,
+                    userAgent,
+                    user,
                     response.getStatus(),
                     timing);
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -0,0 +1,45 @@
...
+    protected void doFilterInternal(final HttpServletRequest request,
+                                    final HttpServletResponse response, final FilterChain filterChain)
+            throws ServletException, IOException
+    {
+        final long start = System.currentTimeMillis();
+
+        //execute request
+        try {
+            filterChain.doFilter(request, response);
+        } finally {
```
v-stepanov: how it will log our async consumption requests? In the beginning of request or when the stream is finished?

v1ctor: Our streaming consumption really is one big http request and it will log at the end, but I don't how to deal with it and if I have to.

v-stepanov: I see. The problem I have with that is that somebody can create streams and stream for few hours/days and we will see it in our logs only after he closes the connection.
My point is: maybe we should before `filterChain.doFilter`? The only reason to log after is to have the timing, right?

src/main/resources/application.yml
```diff
@@ -9,7 +9,7 @@ logging:
   level:
     ROOT: INFO
     org.springframework.web: INFO
-    org.springframework.security.web.access.intercept: DEBUG
+    org.springframework.security.web.access.intercept: INFO
```
v-stepanov: I think you can just remove this line now

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.security.Principal;
+import java.util.Optional;
+
+@Component
+public class LoggingFilter extends OncePerRequestFilter {
```
v-stepanov: Does it work without additional registration?
Because in WebConfig.java we register other filters we have and define the order in which the filters will be invoked https://github.com/zalando/nakadi/blob/master/src/main/java/org/zalando/nakadi/config/WebConfig.java#L54
Do you think it will make sense to add LoggingFilter there also?

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
v1ctor: And also status code.

src/main/resources/application.yml
v1ctor: Ok, I will

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
v1ctor: I've annotated it with ```java @Component``` anotation. It is working.

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
antban: In our specific case we should add 2 records I think - at the begining and after completion

antban: Timing is not so important now, I think...

v-stepanov:
@v1ctor as we now log two times - can we use some sort of keyword for start-log and end-log? Because I believe that the most common use-case will be that we try to search in our log-aggregator. I mean something like ACCESS_LOG_BEGIN, ACCESS_LOG_END or something similar

v-stepanov:
👍

v-stepanov:
@v1ctor Please unmerge :) 
We changed release process. Now we merge to master only after the ticket is validated. Because we decided to release every time the master is updated.

PR #489, opened 2016-11-22
Aruha 456 block consumers

v-stepanov:
@antban @rcillo @adyach @v1ctor @sergkam please review
(It was not yet tested properly, i will create a separate stack and will test it there, so if you notice some bug in the code - let me know :)

I have no idea how this small feature could grow to 700 lines on code :(

Concept document moved to wiki: https://github.com/zalando/nakadi/wiki/Too-much-consumers-from-a-single-application

### DIFF
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingCleaningServiceAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingCleaningServiceAT.java
@@ -11,6 +11,7 @@ import org.junit.Test;
 import org.mockito.Mockito;
 import org.zalando.nakadi.config.JsonConfig;
 import org.zalando.nakadi.repository.zookeeper.ZooKeeperHolder;
+import org.zalando.nakadi.repository.zookeeper.ZooKeeperLockFactory;
 import org.zalando.nakadi.service.ConsumerLimitingCleaningService;
 import org.zalando.nakadi.service.ConsumerLimitingService;
 import org.zalando.nakadi.webservice.utils.ZookeeperTestUtils;
@@ -32,7 +33,8 @@ public class ConsumerLimitingCleaningServiceAT extends BaseAT {
     public ConsumerLimitingCleaningServiceAT() {
         final ZooKeeperHolder zkHolder = Mockito.mock(ZooKeeperHolder.class);
         when(zkHolder.get()).thenReturn(CURATOR);
-        final ConsumerLimitingService limitingService = new ConsumerLimitingService(zkHolder, 5);
+        final ZooKeeperLockFactory zkLockFactory = new ZooKeeperLockFactory(zkHolder);
+        final ConsumerLimitingService limitingService = new ConsumerLimitingService(zkHolder, zkLockFactory, 5);
 
         objectMapper = new JsonConfig().jacksonObjectMapper();
         cleaningService = new ConsumerLimitingCleaningService(zkHolder, objectMapper, limitingService);
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingCleaningServiceAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingCleaningServiceAT.java
@@ -1,0 +1,116 @@
+package org.zalando.nakadi.webservice;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.zookeeper.KeeperException;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+import org.zalando.nakadi.config.JsonConfig;
+import org.zalando.nakadi.repository.zookeeper.ZooKeeperHolder;
+import org.zalando.nakadi.service.ConsumerLimitingCleaningService;
+import org.zalando.nakadi.service.ConsumerLimitingService;
+import org.zalando.nakadi.webservice.utils.ZookeeperTestUtils;
+
+import java.nio.charset.Charset;
+
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.CoreMatchers.nullValue;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.mockito.Mockito.when;
+
+public class ConsumerLimitingCleaningServiceAT extends BaseAT {
+
+    private static final CuratorFramework CURATOR = ZookeeperTestUtils.createCurator(ZOOKEEPER_URL);
+
+    private ConsumerLimitingCleaningService cleaningService;
+    private ObjectMapper objectMapper;
+
+    public ConsumerLimitingCleaningServiceAT() {
+        final ZooKeeperHolder zkHolder = Mockito.mock(ZooKeeperHolder.class);
+        when(zkHolder.get()).thenReturn(CURATOR);
+        final ConsumerLimitingService limitingService = new ConsumerLimitingService(zkHolder, 5);
+
+        objectMapper = new JsonConfig().jacksonObjectMapper();
+        cleaningService = new ConsumerLimitingCleaningService(zkHolder, objectMapper, limitingService);
+    }
+
+    @Before
+    public void before() throws Exception {
+        deleteConsumersData();
+
+        // "hanging" node
+        CURATOR.create().creatingParentsIfNeeded().forPath("/nakadi/consumers/connections/hanging");
+
+        // "normal" node with child
+        CURATOR.create().creatingParentsIfNeeded().forPath("/nakadi/consumers/connections/normal/some_child");
+    }
+
+    @After
+    public void deleteConsumersData() throws Exception {
+        try {
+            CURATOR.delete().deletingChildrenIfNeeded().forPath("/nakadi/consumers");
+        } catch (final KeeperException.NoNodeException e) {
+            // this is fine
+        }
+    }
+
+    @Test
+    public void whenItsNotTimeToRunCleaningThenNothingIsCleaned() throws Exception {
+        // set latest cleaning as two hours ago
+        createLatestCleanNode(2);
+
+        cleaningService.cleanHangingNodes();
+
+        assertThat("the 'hanging' node should not be deleted",
+                CURATOR.checkExists().forPath("/nakadi/consumers/connections/hanging"),
+                not(nullValue()));
+        assertThat("node with children nodes should not be deleted",
+                CURATOR.checkExists().forPath("/nakadi/consumers/connections/normal"),
+                not(nullValue()));
+    }
+
+    @Test
+    public void whenCleaningAlreadyPerformedOnAnotherNodeThenNothingIsCleaned() throws Exception {
+        // set latest cleaning as 30 hours ago
+        createLatestCleanNode(30);
+
+        // another Nakadi instance already performs cleaning
+        CURATOR.create().creatingParentsIfNeeded().forPath("/nakadi/consumers/cleaning/lock");
+
+        cleaningService.cleanHangingNodes();
+
+        assertThat("the 'hanging' node should not be deleted",
+                CURATOR.checkExists().forPath("/nakadi/consumers/connections/hanging"),
+                not(nullValue()));
+        assertThat("node with children nodes should not be deleted",
+                CURATOR.checkExists().forPath("/nakadi/consumers/connections/normal"),
+                not(nullValue()));
+    }
+
+    @Test
+    public void whenCleaningThenOk() throws Exception {
+        // set latest cleaning as 30 hours ago
+        createLatestCleanNode(30);
+
+        cleaningService.cleanHangingNodes();
+
+        assertThat("the 'hanging' node should be deleted",
+                CURATOR.checkExists().forPath("/nakadi/consumers/connections/hanging"),
+                nullValue());
+        assertThat("node with children nodes should not be deleted",
+                CURATOR.checkExists().forPath("/nakadi/consumers/connections/normal"),
+                not(nullValue()));
+    }
+
+    private void createLatestCleanNode(final int hoursAgo) throws Exception {
+        final DateTime now = new DateTime(DateTimeZone.UTC);
+        final DateTime twoHoursAgo = now.minusHours(hoursAgo);
+        final byte[] data = objectMapper.writeValueAsString(twoHoursAgo).getBytes(Charset.forName("UTF-8"));
+        CURATOR.create().creatingParentsIfNeeded().forPath("/nakadi/consumers/cleaning/latest", data);
+    }
+
+}
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingServiceAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingServiceAT.java
@@ -72,18 +72,18 @@ public class ConsumerLimitingServiceAT extends BaseAT {
 
     @Test(expected = NoConnectionSlotsException.class)
     public void whenNoFreeSlotsThenException() throws Exception {
-        final String parition = "0";
+        final String partition = "0";
 
         range(0, 5).forEach(x -> {
             try {
-                final String path = zkPathForConsumer(parition) + "/" + randomUUID();
+                final String path = zkPathForConsumer(partition) + "/" + randomUUID();
                 CURATOR.create().creatingParentsIfNeeded().forPath(path);
             } catch (Exception e) {
                 throw new AssertionError("Error occurred when accessing Zookeeper");
             }
         });
 
-        limitingService.acquireConnectionSlots(client, eventType, ImmutableList.of(parition));
+        limitingService.acquireConnectionSlots(client, eventType, ImmutableList.of(partition));
     }
 
     @Test
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingServiceAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingServiceAT.java
@@ -52,7 +52,7 @@ public class ConsumerLimitingServiceAT extends BaseAT {
         assertThat("4 connection slots were created in ZK", connectionSlots, hasSize(4));
 
         for (final String partition : partitions) {
-            final String path = format("/nakadi/consumers/connections/{0}|{1}|{2}", client, eventType, partition);
+            final String path = zkPathForConsumer(partition);
 
             assertThat("Node for partition should be created",
                     CURATOR.checkExists().forPath(path),
@@ -74,8 +74,7 @@ public class ConsumerLimitingServiceAT extends BaseAT {
 
         range(0, 5).forEach(x -> {
             try {
-                final String path = format("/nakadi/consumers/connections/{0}|{1}|{2}/{3}", client, eventType, parition,
-                        randomUUID());
+                final String path = zkPathForConsumer(parition) + "/" + randomUUID();
                 CURATOR.create().creatingParentsIfNeeded().forPath(path);
             } catch (Exception e) {
                 throw new AssertionError("Error occurred when accessing Zookeeper");
@@ -90,18 +89,22 @@ public class ConsumerLimitingServiceAT extends BaseAT {
         final String connectionId = randomUUID();
         final String partition = "0";
 
-        final String path = format("/nakadi/consumers/connections/{0}|{1}|{2}/{3}", client, eventType, partition,
-                connectionId);
-        CURATOR.create().creatingParentsIfNeeded().forPath(path);
+        final String partitionPath = zkPathForConsumer(partition);
+        final String connectionPath = partitionPath + "/" + connectionId;
+        CURATOR.create().creatingParentsIfNeeded().forPath(connectionPath);
 
         final ImmutableList<ConnectionSlot> connectionSlots =
                 ImmutableList.of(new ConnectionSlot(client, eventType, partition, connectionId));
 
         limitingService.releaseConnectionSlots(connectionSlots);
 
-        assertThat("Zk node should be deleted",
-                CURATOR.checkExists().forPath(path),
+        assertThat("partition and connection Zk nodes should be deleted",
+                CURATOR.checkExists().forPath(partitionPath),
                 nullValue());
+    }
+
+    private String zkPathForConsumer(final String partition) {
+        return format("/nakadi/consumers/connections/{0}|{1}|{2}", client, eventType, partition);
     }
 
 }
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingServiceAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingServiceAT.java
@@ -7,7 +7,6 @@ import org.junit.Test;
 import org.mockito.Mockito;
 import org.zalando.nakadi.exceptions.NoConnectionSlotsException;
 import org.zalando.nakadi.repository.zookeeper.ZooKeeperHolder;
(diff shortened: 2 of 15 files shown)


adyach:
@v-stepanov are you going to move concept document to wiki and post the link here?

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,175 @@
...
+import org.zalando.nakadi.repository.zookeeper.ZooKeeperHolder;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.Random;
+import java.util.stream.Collectors;
+
+import static java.text.MessageFormat.format;
+import static org.echocat.jomon.runtime.concurrent.Retryer.executeWithRetry;
```
adyach: Didn't we agree on not to use static imports ? :)

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,175 @@
...
+            LOG.error("Zookeeper error when trying delete consumer node", e);
+        }
+    }
+
+    private Optional<ConnectionSlot> acquireConnectionSlot(final String client, final String eventType,
+                                                           final String partition)
+            throws ConnectionSlotOccupiedException {
+
+        final String parent = zkPathForConsumer(client, eventType, partition);
+        List<String> children = ImmutableList.of();
```
adyach: Maybe we can init it only in case of KeeperException.NoNodeException ?

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,175 @@
...
+        if (children.size() >= maxConnections) {
+            return Optional.empty();
+        }
+
+        final List<String> occupiedSlots = children;
+        final List<String> availableSlots = SLOT_NAMES.stream()
+                .filter(slot -> !occupiedSlots.contains(slot))
+                .collect(Collectors.toList());
+        final int slotIndex = new Random().nextInt(availableSlots.size());
+        final String slot = availableSlots.get(slotIndex);
```
adyach: The issue with this approach that the change of nakadi.stream.maxConnections to value grater than 5 will cause an exception.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,175 @@
...
+                                                       final List<String> partitions)
+            throws NoConnectionSlotsException, ServiceUnavailableException {
+
+        final List<Optional<ConnectionSlot>> slots = new ArrayList<>();
+        try {
+            // acquire slots for all partitions
+            for (final String partition : partitions) {
+                Optional<ConnectionSlot> connectionSlot = Optional.empty();
+                try {
+                    connectionSlot = executeWithRetry(
```
adyach: for me it is not clear why should we retry here

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
v-stepanov: Hell no! :)
We had a balance of -1 in our radar which doesn't allow neither adopt nor forbid that.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
v-stepanov: Good point. Fixed https://github.com/zalando/nakadi/pull/489/commits/37f2d3105b6d225ec9548e1c1ac6b793119f7a48

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
v-stepanov: Very good catch. I fixed it https://github.com/zalando/nakadi/pull/489/commits/dedca7c5a89bb40f5d8b78f5407819e0f14a8aae

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
v-stepanov: As you see the implementation doesn't use locks - because that puts additional load on ZK and also lock nodes are not cleaned after. So to avoid of creation of more than 5 connections the implementation uses named slots ("1", "2", "3", "4", "5"). And the possible situation is that two instances will try to capture the same free slot at the same time. In that case we should retry to try capture another free slot. 
(Maybe not the best explanation but feel free to approach me so that I can describe how it works on board)

v-stepanov:
@adyach I moved now the concept document to project wiki: https://github.com/zalando/nakadi/wiki/Too-much-consumers-from-a-single-application

src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
```diff
@@ -327,6 +334,52 @@ private Response readEvents() {
     }
 
     @Test(timeout = 10000)
+    public void whenExceedMaxConsumersNumThen429() throws IOException, InterruptedException, ExecutionException,
```
rcillo: @v-stepanov it would be good to have a test case for when a connection is closed. It should release one slot.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,182 @@
...
+
+import static java.text.MessageFormat.format;
+import static org.echocat.jomon.runtime.concurrent.Retryer.executeWithRetry;
+
+@Service
+public class ConsumerLimitingService {
+
+    public static final String CONNECTIONS_ZK_PATH = "/nakadi/consumers/connections";
+
+    private static final Logger LOG = LoggerFactory.getLogger(CursorsService.class);
```
rcillo: s/CursorsService/ConsumerLimitingService/

v-stepanov: Good point. I fixed that now https://github.com/zalando/nakadi/pull/489/commits/d82978a4dcc862cb2a7cd853286eac9bb06d1e63

src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
v-stepanov: I extended this test. Now it also checks if one slot is released when connection is closed.
https://github.com/zalando/nakadi/pull/489/commits/c8f0987d08e942a014fd0c2ebe7611fcc34a6ea2

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,182 @@
...
+
+        final List<Optional<ConnectionSlot>> slots = new ArrayList<>();
+        try {
+            // acquire slots for all partitions
+            for (final String partition : partitions) {
+                Optional<ConnectionSlot> connectionSlot = Optional.empty();
+                try {
+                    connectionSlot = executeWithRetry(
+                            () -> acquireConnectionSlot(client, eventType, partition),
+                            new RetryForSpecifiedCountStrategy<Optional<ConnectionSlot>>(5)
```
rcillo: The number of retries should be at least `maxConnections` and not a fixed 5.

v-stepanov: why?

rcillo:
👍

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
v-stepanov: It's not exactly like you described. The client tries to capture one of FREE slots (so he doesn't try blindly), and also he chooses a random free slot (not the lowest one).
But anyway I see your point. Will change the number of retries to maxConnection.

v-stepanov: changed https://github.com/zalando/nakadi/pull/489/commits/fde52a94d12d1c74329c411a9286939a590668f1

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,182 @@
...
+                    .delete()
+                    .guaranteed()
+                    .forPath(zkPath);
+            deletePartitionNodeIfPossible(parent);
+        } catch (final Exception e) {
+            LOG.error("Zookeeper error when deleting consumer connection node", e);
+        }
+    }
+
+    public void deletePartitionNodeIfPossible(final String nodePath) {
```
antban: Well, this method can be probably renamed to `delete any zk node`

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,182 @@
...
+                    connectionSlot = executeWithRetry(
+                            () -> acquireConnectionSlot(client, eventType, partition),
+                            new RetryForSpecifiedCountStrategy<Optional<ConnectionSlot>>(maxConnections)
+                                    .withExceptionsThatForceRetry(ConnectionSlotOccupiedException.class)
+                                    .withWaitBetweenEachTry(0L, 100L));
+                } catch (ConnectionSlotOccupiedException e) {
+                    LOG.info(format("Failed to capture consuming connection slot after 5 tries for client '{0}', " +
+                            "event-type '{1}', partition {2}", client, eventType, partition));
+                }
+                slots.add(connectionSlot);
```
antban: What is the reason to continue iteration if slot is occupied?

antban:
In general I don't like that service for counting connections works with global zk paths, and I think that it must be changed.

rcillo:
@antban what do you mean by global zk paths?

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
v-stepanov: I do that because I want to have a full list of partitions that don't have free slots - to give user the full information.

v-stepanov:
@antban how not to work with global ZK paths?
I think we work like that everywhere in our project.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
v-stepanov: :)

v-stepanov:
@antban I changed what you requested

v-stepanov:
@antban @rcillo approve or blame! :)

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -189,6 +211,7 @@ public StreamingResponseBody streamEvents(
                 writeProblemResponse(response, outputStream, INTERNAL_SERVER_ERROR, e.getMessage());
             } finally {
                 connectionReady.set(false);
+                consumerLimitingService.releaseConnectionSlots(connectionSlots);
```
rcillo: We should have a feature toggle here as well.

src/main/java/org/zalando/nakadi/util/FeatureToggleService.java
```diff
@@ -27,7 +27,8 @@
         HIGH_LEVEL_API("high_level_api"),
         CHECK_APPLICATION_LEVEL_PERMISSIONS("check_application_level_permissions"),
         CHECK_PARTITIONS_KEYS("check_partitions_keys"),
-        CHECK_OWNING_APPLICATION("check_owning_application");
+        CHECK_OWNING_APPLICATION("check_owning_application"),
+        LIMIT_CONSUMERS_NUMBER("limit_consumers_number");;
```
rcillo: Double `;;`

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
v-stepanov: Why? I think first feature toggle is enough.

v-stepanov: If the feature is disabled then the list `connectionSlots` will be empty and nothing will happen.
But if the feature was enabled when connection started and got disabled before the connection was closed - it's better to release the slots that were acquired.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -84,6 +88,7 @@ public EventStreamController(final EventTypeRepository eventTypeRepository, fina
         this.closedConnectionsCrutch = closedConnectionsCrutch;
         this.blacklistService = blacklistService;
         this.consumerLimitingService = consumerLimitingService;
+        this.featureToggleService = featureToggleService;
```
adyach: It is strange we need feature toggle for that feature.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingCleaningService.java
```diff
@@ -0,0 +1,141 @@
...
+
+    @Autowired
+    public ConsumerLimitingCleaningService(final ZooKeeperHolder zkHolder, final ObjectMapper objectMapper,
+                                           final ConsumerLimitingService limitingService) {
+        this.zkHolder = zkHolder;
+        this.objectMapper = objectMapper;
+        this.limitingService = limitingService;
+    }
+
+    @Scheduled(fixedRate = HANGING_NODES_CLEAN_PERIOD_H * 3600000)
```
adyach: Maybe just put the multiplication in the constant and use it here with name like PERIOD_HANGING_NODES_CLEAN.

PERIOD_HANGING_NODES_CLEAN = 6 * 60 * 60 * 1000;

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
v-stepanov: As I understand the idea was that we will be able to switch it off if it puts too high load on Zookeeper.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingCleaningService.java
v-stepanov: Yes, makes sense. I changed it now.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,187 @@
...
+            for (final String partition : partitions) {
+                Optional<ConnectionSlot> connectionSlot = Optional.empty();
+                try {
+                    connectionSlot = executeWithRetry(
+                            () -> acquireConnectionSlot(client, eventType, partition),
+                            new RetryForSpecifiedCountStrategy<Optional<ConnectionSlot>>(maxConnections)
+                                    .withExceptionsThatForceRetry(ConnectionSlotOccupiedException.class)
+                                    .withWaitBetweenEachTry(0L, 100L));
+                } catch (ConnectionSlotOccupiedException e) {
+                    LOG.info("Failed to capture consuming connection slot after 5 tries for client '{}', " +
```
v-stepanov: Yes, thanks. Fixed it.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingCleaningService.java
```diff
@@ -0,0 +1,141 @@
...
+            // try to create node for the case if it doesn't exist
+            zkHolder.get()
+                    .create()
+                    .creatingParentsIfNeeded()
+                    .forPath(CLEANING_ZK_PATH + "/latest");
+        } catch (final KeeperException.NodeExistsException e) {
+            // "latest" node may already exist - this is ok
+        }
+        // update the latest timestamp of cleaning
+        final DateTime now = new DateTime(DateTimeZone.UTC);
```
v-stepanov: I would like to have this date in readable format in Zookeeper (to be able to read the date by human directly accessing ZK).

rcillo:
I see no more pending comments.

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+        final ConnectionSlot acquiredSlot = new ConnectionSlot(client, eventType, partition, slot);
+        ACQUIRED_SLOTS.add(acquiredSlot);
+        return Optional.of(acquiredSlot);
+    }
+
+    private List<String> getChildrenCached(final String zkPath) {
+        try {
+            PathChildrenCache cache = SLOTS_CACHES.getOrDefault(zkPath, null);
```
adyach: Why don't you want to use curator TreeCache for that? In that case you won't store the list of caches.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,231 @@
...
+            if (cache == null) {
+                cache = new PathChildrenCache(zkHolder.get(), zkPath, false);
+                cache.start(BUILD_INITIAL_CACHE);
+                SLOTS_CACHES.put(zkPath, cache);
```
adyach: In case of two threads are in this `if` block It is possible that you put one cache into the map then other thread get it in line 205 but after another thread in `if` block will replace it with new object.
I suggest to use putIfAbsent or even replace `SLOTS_CACHES.getOrDefault(zkPath, null)` with computeIfAbsent.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,231 @@
...
+        }
+    }
+
+    private void deleteCacheIfPossible(final ConnectionSlot slot) throws IOException {
+        final boolean hasMoreConnectionsToPartition = ACQUIRED_SLOTS.stream()
+                .anyMatch(s -> s.getPartition().equals(slot.getPartition())
+                        && s.getClient().equals(slot.getClient())
+                        && s.getEventType().equals(slot.getEventType()));
+        if (!hasMoreConnectionsToPartition) {
+            final String consumerPath = zkPathForConsumer(slot.getClient(), slot.getEventType(), slot.getPartition());
```
adyach: Do you think it is gonna be problem If another thread will acquire slot at this point?

v-stepanov: I think I don't completely get it.
>> The node exists in the ZK but not in the ACQUIRED_SLOTS
Why you think it will not be in ACQUIRED_SLOTS ?

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
v-stepanov: I don't really need to cache all the nodes' children in "/nakadi/consumers/connections"
I'm only interested in nodes's children for connections that exist in this Nakadi instance. 
So the TreeCache is too much for me in this case, it will create a huge amount of watchers which I really don't need and that will put a very high load on ZK.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
adyach: In general, I did not get why you need ACQUIRED_SLOTS. You can get acquired slots from cache because ACQUIRED_SLOTS do not give you any synchronisation between cache and ZK.

v-stepanov: The only reason why i need ACQUIRED_SLOTS is to be able to delete the cache which I don't need any more at current Nakadi instance. Without knowing what are the current acquired slots at this instance I will not know if I can delete the cache already or if I still need it.

src/main/java/org/zalando/nakadi/service/ConsumerLimitingService.java
```diff
@@ -0,0 +1,230 @@
...
+            final PathChildrenCache cache = SLOTS_CACHES.getOrDefault(consumerPath, null);
```
adyach: You can do only remove since it returns removed value or null

v-stepanov: Good catch! Fixed it here: https://github.com/zalando/nakadi/pull/489/commits/85ead49391bcd797aa5cbc979e94f7c5230a713c

adyach:
:shipit:

v-stepanov:
@antban @rcillo @adyach @lmontrieux 
This ticket is ready for a review again.
With latest changed I added a lock when creating a connection for client_id|event-type pair

src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingServiceAT.java
```diff
@@ -0,0 +1,112 @@
...
+                    hasSize(1));
+
+            final ConnectionSlot expectedSlot = new ConnectionSlot(client, eventType, partition, children.get(0));
+            assertThat(connectionSlots, hasItem(expectedSlot));
+        }
+    }
+
+    @Test(expected = NoConnectionSlotsException.class)
+    public void whenNoFreeSlotsThenException() throws Exception {
+        final String parition = "0";
```
rcillo: s/parition/partition/

src/main/java/org/zalando/nakadi/service/ConsumerLimitingCleaningService.java
```diff
@@ -0,0 +1,140 @@
...
+import static org.zalando.nakadi.service.ConsumerLimitingService.CONNECTIONS_ZK_PATH;
+
+@Service
+public class ConsumerLimitingCleaningService {
+
+    public static final String CLEANING_ZK_PATH = "/nakadi/consumers/cleaning";
+
+    private static final Charset CHARSET_UTF8 = Charset.forName("UTF-8");
+    private static final Logger LOG = LoggerFactory.getLogger(ConsumerLimitingCleaningService.class);
+    private static final int HANGING_NODES_CLEAN_PERIOD_MS = 6 * 60 * 60 * 1000; // 6 hours
```
rcillo: What do you think of moving this number to the application config file?

rcillo: I see. That's not possible because of the usage on annotation.

src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingCleaningServiceAT.java
```diff
@@ -0,0 +1,118 @@
...
+        objectMapper = new JsonConfig().jacksonObjectMapper();
+        cleaningService = new ConsumerLimitingCleaningService(zkHolder, objectMapper, limitingService);
+    }
+
+    @Before
+    public void before() throws Exception {
+        deleteConsumersData();
+
+        // "hanging" node
+        CURATOR.create().creatingParentsIfNeeded().forPath("/nakadi/consumers/connections/hanging");
```
rcillo: it wasn't clear for me that `hanging` is the name of a client. It took me some time to realise that it's the name of a client which has no consumers anymore. That's why you clean it. Changing it to `consumer_with_no_open_streams` could make if more explicit. This is just a suggestion to improve clarity.

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/service/ConsumerLimitingCleaningService.java
v-stepanov: Yes, exactly. And even without that, I don't see much sense to make this property configurable.

src/acceptance-test/java/org/zalando/nakadi/webservice/ConsumerLimitingCleaningServiceAT.java
rcillo: As I said, it's a suggestion to improve clarity. I took some time to understand it. I understand that for the machine it doesn't matter. But that's just me.

rcillo:
:shipit:

rcillo:
:shipit:

PR #490, opened 2016-11-28
ARUHA-450: decreased DB connections to 2

adyach:


src/main/resources/application.yml
```diff
@@ -19,9 +19,14 @@ spring:
     password: nakadi
     driverClassName: org.postgresql.Driver
     initialSize: 2
-    maxActive: 20
+    maxActive: 2
```
v-stepanov: this property will be overridden with senza env variables we have in our private repo

src/main/resources/application.yml
```diff
@@ -19,9 +19,14 @@ spring:
...
+    minIdle: 1
+    maxIdle: 1
     testOnBorrow: true
     validationQuery: SELECT 1
+    testWhileIdle: true
+    timeBetweenEvictionRunsMillis: 5000
+    minEvictableIdleTimeMillis: 60000
```
v-stepanov: What these three properties are for?

src/main/resources/application.yml
adyach: @v-stepanov changed

antban: Lets change initial size too.

antban: Ignore my previous comment

antban:
:+1:

PR #494, opened 2016-11-29
Aruha 400: detect major, minor and patch changes

rcillo:
This PR adapts the schema evolution behaviour to the latest changes besides minor fixes to migrations.

1. every change had its semantics defined: MAJOR, MINOR or PATCH level change.
2. incompatible event types can evolve their schema on the MAJOR level.
3. compatible event types can evolve only up to MINOR level.

You can check the classification of each change in the `ValidatorConfi` class.

There has been some refactoring in order to extract the changes detection from the evolution rules.


database/migration/aruha-400-schema-evolution.sql
```diff
@@ -12,13 +12,19 @@ CREATE UNIQUE INDEX ON zn_data.event_type_schema ((ets_schema_object->>'version'
                                                   (ets_event_type_name));
 
 UPDATE zn_data.event_type
-SET et_event_type_object  = jsonb_set(et_event_type_object, '{schema,version}', '"1.0.0"', false);
+SET et_event_type_object  = jsonb_set(et_event_type_object, '{schema,version}', '"0.1.0"', false);
```
v-stepanov: AFAIR there is a problem with using jsonb_set in a version of Postgres we use.

database/migration/aruha-400-schema-evolution.sql
```diff
@@ -12,13 +12,19 @@ CREATE UNIQUE INDEX ON zn_data.event_type_schema ((ets_schema_object->>'version'
...
+
+UPDATE zn_data.event_type
+SET et_event_type_object  = jsonb_set(et_event_type_object, '{updated_at}', '"2016-11-09T19:32:00Z"', false);
 
 UPDATE zn_data.event_type
 SET et_event_type_object  = jsonb_set(et_event_type_object, '{schema,created_at}', '"2016-11-09T19:32:00Z"', false);
 
 UPDATE zn_data.event_type
-SET et_event_type_object  = jsonb_set(et_event_type_object, '{compatibility_mode}', '"deprecated"', false);
+SET et_event_type_object  = jsonb_set(et_event_type_object, '{compatibility_mode}', '"fixed"', false);
```
v-stepanov: I think I missed this moment. Was "deprecated" renamed to "fixed"

rcillo: yes. It was one of guild requests when reviewing this feature.

src/main/java/org/zalando/nakadi/util/FeatureToggleService.java
```diff
@@ -26,7 +26,6 @@
         DISABLE_EVENT_TYPE_DELETION("disable_event_type_deletion"),
         HIGH_LEVEL_API("high_level_api"),
         CHECK_APPLICATION_LEVEL_PERMISSIONS("check_application_level_permissions"),
-        CHECK_PARTITIONS_KEYS("check_partitions_keys"),
```
v-stepanov: is this already enabled on live?

src/main/java/org/zalando/nakadi/validation/schema/SchemaDiff.java
```diff
@@ -0,0 +1,331 @@
...
+            final Schema update,
+            final Stack<String> jsonPath,
+            final List<SchemaChange> changes) {
+
+        if (original == null && update == null) {
+            return;
+        }
+
+        if (original == null || update == null) {
+            addChange(SCHEMA_REMOVED, jsonPath, changes);
```
v-stepanov: Here we have two situations
1) original == null, update != null
2) original != null, update == null

Both of that considered to be SCHEMA_REMOVED ?

src/main/java/org/zalando/nakadi/validation/schema/SchemaDiff.java
```diff
@@ -0,0 +1,331 @@
...
+import static org.zalando.nakadi.domain.SchemaChange.Type.NUMBER_OF_ITEMS_CHANGED;
+import static org.zalando.nakadi.domain.SchemaChange.Type.PROPERTIES_ADDED;
+import static org.zalando.nakadi.domain.SchemaChange.Type.PROPERTY_REMOVED;
+import static org.zalando.nakadi.domain.SchemaChange.Type.REQUIRED_ARRAY_CHANGED;
+import static org.zalando.nakadi.domain.SchemaChange.Type.SCHEMA_REMOVED;
+import static org.zalando.nakadi.domain.SchemaChange.Type.SUB_SCHEMA_CHANGED;
+import static org.zalando.nakadi.domain.SchemaChange.Type.TITLE_CHANGED;
+import static org.zalando.nakadi.domain.SchemaChange.Type.TYPE_CHANGED;
+
+public class SchemaDiff {
```
v-stepanov: For me it feels that this class wants to be stateful class with `jsonPath` and `changes` as internal properties.

src/main/java/org/zalando/nakadi/validation/schema/SchemaDiff.java
rcillo: Actually, I could not think of a scenario when the first `original == null` would happen. Changed here 1d74868

src/main/java/org/zalando/nakadi/validation/schema/SchemaDiff.java
```diff
@@ -0,0 +1,331 @@
...
+            recursiveCheck((ArraySchema) original, (ArraySchema) update, jsonPath, changes);
+        } else if (original instanceof ReferenceSchema) {
+            recursiveCheck((ReferenceSchema) original, (ReferenceSchema) update, jsonPath, changes);
+        }
+    }
+
+    private void recursiveCheck(final StringSchema stringSchemaOriginal, final StringSchema stringSchemaUpdate,
+                                final Stack<String> jsonPath, final List<SchemaChange> changes) {
+        if (!Objects.equals(stringSchemaOriginal.getMaxLength(), stringSchemaUpdate.getMaxLength())) {
+            jsonPath.push("maxLength");
```
rcillo: `jsonPath` is a `Stack` object where I push the attribute under inspection. Since it has recursion, once it comes out of recursion, I pop the attribute.

It is used to keep a memory of the path of the recursion for reporting the where in the JSON the invalid attribute was found.

I could create an auxiliary method that would receive a `keyword` as string and a lambda, would push the `keyword` execute the lambda and pop the `keyword`. That could reduce the chances of forgetting to pop something.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -225,17 +222,8 @@ private void validateJsonSchemaConstraints(final JSONObject schema) throws Inval
         }
     }
 
-    private void validatePartitionKeys(final Optional<Schema> schemaO, final EventTypeBase eventType)
+    private void validatePartitionKeys(final Schema schema, final EventTypeBase eventType)
             throws InvalidEventTypeException, JSONException, SchemaException {
-        if (!featureToggleService.isFeatureEnabled(CHECK_PARTITIONS_KEYS)) {
```
adyach: Why did you remove it?

rcillo: This feature needs to be released. Right now, we do not check if a given partition key exists in the schema which could lead to NO-OP topics: a topic that expects events to contain a given key when partitioning but this same key is not allowed because it's not in the schema definition.

I have no clue why it wasn't turned on. I can keep it but then we need to turn it on altogether.

src/main/java/org/zalando/nakadi/validation/schema/SchemaDiff.java
adyach: I understand that it is quite difficult to do it other way but this class looks quite massive with a lot of recursion calls and loops in on file :)

adyach:
Good job, Ricardo 👍

src/main/java/org/zalando/nakadi/validation/schema/SchemaDiff.java
rcillo: Do you have any suggestions?

adyach: I've already suggested in the previous PR to extend schema classes from library, but I am not sure It will look less massive than that. Sorry, we already discussed it.

rcillo: It would break it into smaller classes but the code would be essentially the same.

I thought about other approaches to this problem, but navigating the `Schema` objects is the less error prone in my opinion, since it means that we are inspecting an already interpreted json schema object, and this interpretation is the one from the library we use for validation.

Json schema is full of ambiguities. The everit library puts some opinion on how to interpret that (it even does some kind of "type inference" when some fields are missing). So I would not be willing to move away from it.

Translating these classes to our own domain classes would mean that we would have more than 10 new classes. This fragmentation can lead to even harder to understand code.

This code doesn't look beautiful, but it's surprisingly organized. Every kind of schema has its own method. We could extract these methods to different classes to reduce the overall file size, but I feel it would not improve the overall design.

src/main/java/org/zalando/nakadi/util/FeatureToggleService.java
rcillo: @v-stepanov it's not. But we cannot release this feature without enabling it first. More details here https://github.com/zalando/nakadi/pull/494/files#r91092420

Do you remember why is there a feature toggle for it?

adyach: @rcillo https://github.com/zalando/nakadi/pull/378#r74594874

src/main/java/org/zalando/nakadi/validation/schema/diff/StringSchemaDiff.java
```diff
@@ -0,0 +1,23 @@
...
+import org.everit.json.schema.StringSchema;
+import org.zalando.nakadi.domain.SchemaChange;
+
+import java.util.List;
+import java.util.Objects;
+import java.util.Stack;
+
+import static org.zalando.nakadi.domain.SchemaChange.Type.ATTRIBUTE_VALUE_CHANGED;
+
+class StringSchemaDiff extends SchemaDiff {
```
adyach: What is the point to extend the SchemaDiff? I think it better to not extend it as you did for other classes because you use only static methods and there is no `is a` relation.

rcillo: It was mistakenly committed. Fixed in ad34236

adyach:
:shipit:

PR #495, opened 2016-12-05
Event version tag

rcillo:
In this PR we are tagging every event that has metadata (business and data categories) with the schema version used to validate such event.

We also implemented a strict validation as defined in the [concept](https://docs.google.com/document/d/1FmGHxI9oklxN16LKZe1UG_U21jtQFVCtLtRiC2W1pAk/edit#heading=h.tvwrqalwl0nc).


src/main/java/org/zalando/nakadi/validation/JsonSchemaLoader.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        switch (eventType.getCategory()) {
+            case BUSINESS: return addMetadata(schema, eventType);
+            case DATA: return wrapSchemaInData(schema, eventType);
+            default: return schema;
+        }
+    }
+
+    private void enforceStrictValidation(final JSONObject schema) {
```
rcillo: This method forces strict validation for events. We describe in the concept document (linked in the description) the reasons for doing so. What strict means is that, only attributes that are explicitly defined in the json schema are going to be allowed. Anything else will cause events to be rejected.

Strict validation is only used by "compatible" event types.

src/test/java/org/zalando/nakadi/validation/JSONSchemaValidationTest.java
```diff
@@ -18,105 +17,13 @@
...
-        assertThat(validator.validate(invalidEventMissingBar).get().getMessage(),
-                equalTo("#: required key [bar] not found"));
-        assertThat(validator.validate(invalidEventMissingNameField).get().getMessage(), equalTo("name is required"));
-        assertThat(validator.validate(nestedSchemaViolation).get().getMessage(),
-                equalTo("#: 3 schema violations found\n#/bar: 2 schema violations found\n#/bar: required key [foo]" +
-                        " not found\n#/bar: required key [bar] not found\n#: required key [foo] not found"));
-    }
-
-    @Test
-    public void schemaValidationShouldRespectIgnoreConfigurationMatchRegular() {
```
rcillo: This test no longer makes sense since this code has been finally deleted. It was never used in production.

src/test/java/org/zalando/nakadi/validation/JSONSchemaValidationTest.java
```diff
@@ -18,105 +17,13 @@
...
-        final JSONObject invalidEvent = new JSONObject(
-                "{\"event-type\" : \"X\", \"extra\": \"i should be no problem\", \"name\": \"12345\"}");
-        assertThat(validator.validate(event), isAbsent());
-        assertThat(validator.validate(eventDelete), isAbsent());
-        assertThat(validator.validate(invalidEvent).get().getMessage(), equalTo("#: required key " +
-                "[field-that-will-not-be-found] not found"));
-    }
-
-    @Test
-    public void schemaValidationShouldRespectIgnoreConfigurationMatchQualified() {
```
rcillo: This test no longer makes sense since this code has been finally deleted. It was never used in production.

src/test/java/org/zalando/nakadi/validation/JSONSchemaValidationTest.java
```diff
@@ -18,105 +17,13 @@
...
     static {
-        ValidationStrategy.register(EventBodyMustRespectSchema.NAME, new EventBodyMustRespectSchema());
+        ValidationStrategy.register(EventBodyMustRespectSchema.NAME, new EventBodyMustRespectSchema(
+                new JsonSchemaLoader()));
         ValidationStrategy.register(EventMetadataValidationStrategy.NAME, new EventMetadataValidationStrategy());
         ValidationStrategy.register(FieldNameMustBeSet.NAME, new FieldNameMustBeSet());
     }
 
     @Test
-    public void schemaValidationShouldRespectEventTypeDefinition() {
```
rcillo: Schema validation is a task of the everit library. We are not going to test it again in our project. It already has hundreds of tests in the library.

src/test/resources/org/zalando/nakadi/validation/strict-validation.json
```diff
@@ -0,0 +1,261 @@
+[
```
rcillo: Lot's of tests for checking strict validation. It illustrates all scenarios I could think of when we are able to enforce strict validation.

src/test/resources/product-json-schema.json
```diff
@@ -73,7 +73,6 @@
     },
 
     "Value": {
-      "type": "object",
```
rcillo: This is an incorrect schema example. If your schema is defined by references, it cannot be at the same time defined by a direct type. We are going to have a very hard time to explain all these details... eventually we are going to see ourselves forced to used something less broken than json schema draft v4.

src/main/java/org/zalando/nakadi/validation/JsonSchemaLoader.java
```diff
@@ -0,0 +1,191 @@
...
+public class JsonSchemaLoader {
+    public static final String DATA_CHANGE_WRAP_FIELD = "data";
+
+    private static final String ADDITIONAL_PROPERTIES = "additionalProperties";
+    private static final String ADDITIONAL_ITEMS = "additionalItems";
+    private final List<String> keywordsWithNestedSchemas = Lists.newArrayList("definitions", "dependencies",
+            "properties");
+    private final List<String> keywordsInferObjectSchema = Lists.newArrayList("required", "minProperties",
+            "maxProperties");
+    private final List<String> keywordsInferArraySchema = Lists.newArrayList("minItems", "maxItems", "uniqueItems");
```
rcillo: Sometimes, schema objects don't have explicit types. So we need to infer based on some properties. Everit library does the same https://github.com/everit-org/json-schema/blob/36b7d3f407b32066f380e91e91a2dc6e84ac66ff/core/src/main/java/org/everit/json/schema/loader/SchemaLoader.java#L349-L359

src/test/resources/product-json-schema.json
adyach: Maybe it is a good idea to collect answer to common questions in some place to provide link in the future in order not explain the same things many times.

adyach:
Seems like a lot of redundant and legacy code was removed, I like this approach 👍

src/main/java/org/zalando/nakadi/validation/JsonSchemaLoader.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+public class JsonSchemaLoader {
+    public static final String DATA_CHANGE_WRAP_FIELD = "data";
+
+    private static final String ADDITIONAL_PROPERTIES = "additionalProperties";
+    private static final String ADDITIONAL_ITEMS = "additionalItems";
+    private final List<String> keywordsWithNestedSchemas = Lists.newArrayList("definitions", "dependencies",
```
adyach: Is this list changeable? I suppose not, maybe we can make it static. also makes sense to use set vs list.

src/main/java/org/zalando/nakadi/validation/JsonSchemaLoader.java
```diff
@@ -0,0 +1,187 @@
...
+import org.zalando.nakadi.domain.CompatibilityMode;
+import org.zalando.nakadi.domain.EventType;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+public class JsonSchemaLoader {
```
adyach: Or it it is EffectiveSchema -> JsonSchemaEffecter :)

rcillo: Done 2df7603

adyach:
:shipit:

PR #496, opened 2016-12-06
ARUHA-401: List event type schemas

adyach:


src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Autowired
+    public SchemaRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper objectMapper) {
+        super(jdbcTemplate, objectMapper);
+    }
+
+    public List<EventTypeSchema> getSchemas(final String name, final int offset, final int limit) {
+        return jdbcTemplate.query(
+                "SELECT ets_schema_object FROM zn_data.event_type_schema " +
+                        "WHERE ets_event_type_name = ? ORDER BY ets_schema_object->>'version' DESC LIMIT ? OFFSET ? ",
```
v-stepanov: Are you sure that the ordering will be correct? I think it will order wrongly for the following schemas:
"1.0.2"
"2.10.3"
"10.0.0"

src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -0,0 +1,37 @@
...
+    @Autowired
+    public SchemaController(final SchemaService schemaService) {
+        this.schemaService = schemaService;
+    }
+
+    @RequestMapping("/event-types/{name}/schemas")
+    public ResponseEntity<?> getSchemas(
+            @PathVariable("name") final String name,
+            @RequestParam(value = "offset", required = false, defaultValue = "0") final int offset,
+            @RequestParam(value = "limit", required = false, defaultValue = "20") final int limit,
```
lmontrieux: Would it make sense to cherry-pick the commit(s?) related to the documentation changes for schema evolution?

src/main/java/org/zalando/nakadi/service/PaginationService.java
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        return new PaginationLinks(prevLink, nextLink);
+    }
+
+    public interface ItemsSupplier {
+
+        List query(final int offset, final int limit);
+
+        default List  queryOneMore(final int offset, final int limit) {
```
v-stepanov: aaaa! double space

src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.context.request.NativeWebRequest;
+import org.zalando.nakadi.service.Result;
+import org.zalando.nakadi.service.SchemaService;
+import org.zalando.problem.spring.web.advice.Responses;
+
+@RestController
+public class SchemaController {
```
v-stepanov: unit test for controller is missing

adyach: I have SchemaControllerAT test. Would it be enough for you if I test problem cases(bad request) inside that test?

src/main/java/org/zalando/nakadi/config/ValidatorConfig.java
```diff
@@ -39,7 +38,6 @@ public SchemaEvolutionService schemaEvolutionService() throws IOException {
 
         final List<SchemaEvolutionConstraint> schemaEvolutionConstraints = Lists.newArrayList(
                 new CompatibilityModeChangeConstraint(),
-                new CompatibleSchemaChangeConstraint(),
```
v-stepanov: Why you removed this? o_0

src/main/java/org/zalando/nakadi/service/PaginationService.java
adyach: and check style did not catch it :(

src/main/java/org/zalando/nakadi/controller/SchemaController.java
adyach: I think it better to do rebase for the upstream branch and update the downstream.

src/main/java/org/zalando/nakadi/config/ValidatorConfig.java
adyach: I also removed the test because of it. the upstream branch is not up to date with the latest changes.
@rcillo can elaborate on that :)

src/main/java/org/zalando/nakadi/controller/SchemaController.java
v-stepanov: My vision of tests in Nakadi is different.
I think you should have only one very straightforward acceptance test for this endpoint and you should have a lot of unit tests for controller to test all corner cases.
I think this is the approach we agreed on some time ago. To test everything that is possible with a unit tests and to have very few acceptance tests. To follow the approach of test pyramid.

(The only exception I see are the cases when testing with unit test will lead to mock-nightmare and testing with AT is very simple - then AT is preferable)

adyach: 89c25fe

src/main/java/org/zalando/nakadi/service/PaginationService.java
```diff
@@ -0,0 +1,68 @@
...
+                    .queryParam("offset", offset + limit).queryParam("limit", limit).build().toString();
+            nextLink = Optional.of(new PaginationLinks.Link(nextUri));
+        }
+
+        return new PaginationLinks(prevLink, nextLink);
+    }
+
+    public interface ItemsSupplier {
+
+        List query(final int offset, final int limit);
```
adyach: 89c25fe

src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.stereotype.Component;
+import org.zalando.nakadi.domain.EventTypeSchema;
+
+import java.io.IOException;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.util.List;
+
+@Component
+public class SchemaRepository extends AbstractDbRepository {
```
adyach: 89c25fe

src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
adyach: 89c25fe

src/acceptance-test/java/org/zalando/nakadi/repository/db/SchemaRepositoryTest.java
```diff
@@ -0,0 +1,70 @@
...
+        final EventTypeSchemaBase schemaBase = new EventTypeSchemaBase(EventTypeSchemaBase.Type.JSON_SCHEMA, "schema");
+        final EventType eventType = EventTypeTestBuilder.builder()
+                .name("test_et_name_schemarepositorytest")
+                .schema(new EventTypeSchema(schemaBase, "1.0.2", DateTime.now()))
+                .build();
+        insertEventType(eventType);
+        insertSchema(eventType);
+        eventType.setSchema(new EventTypeSchema(schemaBase, "2.10.3", DateTime.now()));
+        insertSchema(eventType);
+        eventType.setSchema(new EventTypeSchema(schemaBase, "10.0.0", DateTime.now()));
```
rcillo: I'm afraid this test could fail if it runs too fast. If the precision of `DateTime.now()` is too low, it could sometimes generate the same value on lines 35, 39 or 41. Makes sense?

rcillo: Maybe it's better not to generate the values dynamically, but actually make sure that one value is bigger than the other.

adyach: Actually, there is at least 100ms difference event if you do not insert in DB. I will change it to be more clear what we are comparing there.

adyach: 27a2100

src/main/java/org/zalando/nakadi/config/ValidatorConfig.java
rcillo: 👍 my mistake. The base branch I sent to @adyach didn't contain this change, which is necessary to test this feature.

src/main/java/org/zalando/nakadi/service/PaginationService.java
```diff
@@ -0,0 +1,68 @@
...
+        final PaginationLinks paginationLinks;
+        if (items.isEmpty() && offset != 0 && limit != 0) {
+            final int count = countSupplier.get();
+            int latestOffset = count / limit;
+            if (offset >= latestOffset)
+                latestOffset = 0;
+            paginationLinks = createLinks(path, latestOffset, limit, items.size());
+        } else {
+            paginationLinks = createLinks(path, offset, limit, items.size());
+            if (items.size() > limit)
```
rcillo: In which scenario is this condition true?

adyach: to proper show the link to the next page I request +1 element from db that's why I check here that if we have more than we requested I will remove the last element from the response if it is there.

rcillo: Why is it necessary to request +1?

adyach: For example you have 10 elements, offset=0, limit=10. How can you identify that you should or should not show to show the link to the next page?
My approach is to request one more element from db to understand if we have more elements to show on the next page.

rcillo: Another approach would be to run a count query and check how many elements there are. I would say that for building pagination links it's not needed to load the specific elements.

adyach:
@v-stepanov @rcillo if you are fine with PR lets merge it

rcillo:
Looks good. Nice abstraction with the pagination.

rcillo:
:shipit:

rcillo:
:shipit:

PR #497, opened 2016-12-07
Aruha 402 view specific schema version

lmontrieux:
This PR adds an endpoint to GET any version of an event type's schema. Using 'latest' as the version number will return the most recent schema.


src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
```diff
@@ -28,6 +34,29 @@ public SchemaRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper obje
                 new SchemaRowMapper());
     }
 
+    public EventTypeSchema getSchemaVersion(final String name, final String version)
+            throws NoSuchSchemaException, IllegalVersionNumberException, InternalNakadiException {
+        final Pattern versionPattern = Pattern.compile("\\d+\\.\\d+\\.\\d+");
```
adyach: It can be static class field in order not to compile it on every call.

src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
```diff
@@ -28,6 +34,29 @@ public SchemaRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper obje
                 new SchemaRowMapper());
     }
 
+    public EventTypeSchema getSchemaVersion(final String name, final String version)
```
adyach: Also I usually put the logic related things inside service layer. In that case repository only does fetch of data.

lmontrieux: Good point. Fixed in https://github.com/zalando/nakadi/pull/497/commits/c88ad820765bc55cf1e7a79af81be1bfffd02c17

adyach:
:shipit:

src/test/java/org/zalando/nakadi/controller/SchemaControllerTest.java
```diff
@@ -37,8 +44,52 @@ public void testFailure() {
...
     }
 
+    @Test
+    public void testGetLatestSchemaVersionThen200() {
+        final EventType eventType = buildDefaultEventType();
+        Mockito.when(eventTypeService.get(eventType.getName())).thenReturn(Result.ok(eventType));
+        final ResponseEntity<?> result =
+                new SchemaController(schemaService, eventTypeService)
+                        .getSchemaVersion(eventType.getName(), "latest", nativeWebRequest);
+        Assert.assertEquals(HttpStatus.OK, result.getStatusCode());
```
rcillo: I think it's fine to use assertEquals, but we use extensively hamcrest matchers with `assertThat` in other test classes.

src/test/java/org/zalando/nakadi/controller/SchemaControllerTest.java
```diff
@@ -7,28 +7,35 @@
...
         nativeWebRequest = Mockito.mock(NativeWebRequest.class);
+        eventTypeService = Mockito.mock(EventTypeService.class);
     }
 
     @Test
     public void testSuccess() {
         Mockito.when(schemaService.getSchemas("et_test_name", 0, 1)).thenReturn(Result.ok(null));
         final ResponseEntity<?> result =
-                new SchemaController(schemaService).getSchemas("et_test_name", 0, 1, nativeWebRequest);
+                new SchemaController(schemaService, eventTypeService)
```
rcillo: @lmontrieux Now I realised something. We use MockMVC for testing controllers. We do that because it tests some other functionality that is left behind when interacting with controller directly. Please take a look at some other controller class for examples on how to use it.

src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -35,4 +45,21 @@ public SchemaService(final SchemaRepository schemaRepository,
                         () -> schemaRepository.getSchemasCount(name)));
     }
 
+    public Result<EventTypeSchema> getSchemaVersion(final String name, final String version) {
+        final Matcher versionMatcher = VERSION_PATTERN.matcher(version);
```
rcillo: I would argue that data validation is responsibility of the layer above, in the controller. You could use some `@Pattern` validation on the controller directly. It would look like:

```java
public ResponseEntity<?> getSchemaVersion(@PathVariable("name") @Pattern("...") @Valid final String name
```

adyach: In my point of view controller has to be as thin as possible.

rcillo: Spring will generate the 400 for you if you use controller validation.

src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
```diff
@@ -28,6 +31,24 @@ public SchemaRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper obje
...
+            throws NoSuchSchemaException, InternalNakadiException {
+        final String sql = "SELECT ets_schema_object FROM zn_data.event_type_schema " +
+                "WHERE ets_event_type_name = ? AND ets_schema_object ->> 'version' = ?";
+
+        try {
+            final List<EventTypeSchema> schemas =
+                    jdbcTemplate.query(sql, new Object[]{name, version}, new SchemaRowMapper());
+            if (schemas.size() != 1)
```
rcillo: It's not necessary to check this condition since we have a `unique` constraint on the database.

src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
```diff
@@ -28,6 +31,24 @@ public SchemaRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper obje
...
     }
 
+    public EventTypeSchema getSchemaVersion(final String name, final String version)
+            throws NoSuchSchemaException, InternalNakadiException {
+        final String sql = "SELECT ets_schema_object FROM zn_data.event_type_schema " +
+                "WHERE ets_event_type_name = ? AND ets_schema_object ->> 'version' = ?";
+
+        try {
+            final List<EventTypeSchema> schemas =
+                    jdbcTemplate.query(sql, new Object[]{name, version}, new SchemaRowMapper());
```
rcillo: It seems that `queryForObject` would be a better fit here, since we are looking for a single row result. We have an example in the `EventTypeDbRepository.findByName`.

src/main/java/org/zalando/nakadi/service/SchemaService.java
rcillo: @adyach since we are using Spring, I would just follow the Spring way of validating format. It supports these annotations which goes into the controller. Service layer is for business rules and data format is understood by spring to be a user interfacing responsibility, e.g. controller stuff.

adyach: @rcillo I totally agree we should use the framework in the full force 👍

rcillo:
:shipit:

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
lmontrieux: Fixed in https://github.com/zalando/nakadi/commit/b28dac092bd04dd7c9b43f66d7664351359f4ede

src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
lmontrieux: Fixed in https://github.com/zalando/nakadi/commit/4e66d08012a9acae2fd5b6cdbb2c04ab7f59ec45

src/main/java/org/zalando/nakadi/service/SchemaService.java
lmontrieux: https://github.com/zalando/nakadi/commit/d1aea1cf333888fe78b40bbf7cc37a80c8909ad7 fixes the status code but doesn't use Spring validation (yet)

PR #498, opened 2016-12-07
aruha-404 allow for compatibility mode upgrade

rcillo:
This PR introduces rules for upgrading an existing FIXED event type schema to a COMPATIBLE schema.

### DIFF
--- a/src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
+++ b/src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
@@ -102,12 +102,12 @@ public class SchemaEvolutionService {
                                                     final List<SchemaChange> changes) throws InvalidEventTypeException {
         if (original.getCompatibilityMode() == CompatibilityMode.FIXED
             && eventType.getCompatibilityMode() == CompatibilityMode.COMPATIBLE) {
-            final List<SchemaChange.Type> changesTypes = changes.stream().map(SchemaChange::getType)
+
+            final List<SchemaChange> forbiddenChanges = changes.stream()
+                    .filter(change -> !FIXED_TO_COMPATIBLE_ALLOWED_CHANGES.contains(change.getType()))
                     .collect(Collectors.toList());
-            if (!FIXED_TO_COMPATIBLE_ALLOWED_CHANGES.containsAll(changesTypes)) {
-                final String errorMessage = changes.stream()
-                        .filter(change -> !FIXED_TO_COMPATIBLE_ALLOWED_CHANGES.contains(change.getType()))
-                        .map(this::formatErrorMessage)
+            if (!forbiddenChanges.isEmpty()) {
+                final String errorMessage = forbiddenChanges.stream().map(this::formatErrorMessage)
                         .collect(Collectors.joining(", "));
                 throw new InvalidEventTypeException("Invalid schema: " + errorMessage);
             }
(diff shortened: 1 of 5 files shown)


src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
```diff
@@ -77,19 +86,42 @@ public EventType evolve(final EventType original, final EventTypeBase eventType)
...
+        return this.bumpVersion(original, eventType, changeLevel);
+    }
+
+    private void validateCompatibilityModeMigration(final EventType original, final EventTypeBase eventType,
+                                                    final List<SchemaChange> changes) throws InvalidEventTypeException {
+        if (original.getCompatibilityMode() == CompatibilityMode.FIXED
+            && eventType.getCompatibilityMode() == CompatibilityMode.COMPATIBLE) {
+            final List<SchemaChange.Type> changesTypes = changes.stream().map(SchemaChange::getType)
+                    .collect(Collectors.toList());
+            if (!FIXED_TO_COMPATIBLE_ALLOWED_CHANGES.containsAll(changesTypes)) {
```
adyach: It is up to developer how to do it this check, but here we can avoid creating a list by using allMatch.

adyach: Or filter all changes types and check if the list is not empty then you can use this list inside `if`to create error message

src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
```diff
@@ -77,19 +86,42 @@ public EventType evolve(final EventType original, final EventTypeBase eventType)
...
             }
-            return this.bumpVersion(original, eventType, changeLevel);
+        }
+    }
+
+    private void validateCompatibleModeChanges(final EventType original, final List<SchemaChange> changes,
+                                               final Version.Level changeLevel) throws InvalidEventTypeException {
+        if (original.getCompatibilityMode() == CompatibilityMode.COMPATIBLE && changeLevel == MAJOR) {
+            final String errorMessage = changes.stream()
+                    .filter(change -> MAJOR.equals(changeToLevel.get(change.getType())))
```
adyach: Why do you only show major incompatibilities? AFAIU there could be other levels.

adyach: got it! CompatibilityMode.COMPATIBLE only possible for minor and patch.

adyach: It is not in this scope of work but it looks like you go through the changes several times and it is possible to do only once:
1. you get level in semanticOfChange
2. you check here for MAJOR.

Instead, you could go thorough changes if CompatibilityMode.COMPATIBLE and filter major at the same time.

src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
rcillo: Done a52dfc5

src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
rcillo: It's not a problem from a performance point of view. This collection is small and this endpoint is used very rarely. The most important aspect of this code is to be the easiest to understand as possible. If you think that changing this could help make it easier to understand than I would say that we should change it.

adyach:
:shipit:

rcillo:
@adyach I fixed some tests here https://github.com/zalando/nakadi/pull/498/commits/d61e29db78b89cd1bf880a8294fa567b1b753442 could you please approve again?

adyach:
this test worked fine some time ago. thanks for the fix!

adyach:
:shipit:

PR #502, opened 2016-12-08
aruha-400 fix migration

rcillo:
It was incorrectly using `create_missing` parameter as `null`. This migration should set this value even if it's not present.


v-stepanov:
:shipit:

PR #505, opened 2016-12-09
Aruha 402-1 bugfix view specific schema version

lmontrieux:
In this PR:
* return an EventTypeSchema instead of only the schema
* return 404 not found instead of 500 if the event type or the schema version does not exist
* return 400 bad request with details if the version number is invalid
* improved query handling in SchemaRepository

### DIFF
--- a/src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
+++ b/src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
@@ -37,7 +37,7 @@ public class SchemaRepository extends AbstractDbRepository {
                 "WHERE ets_event_type_name = ? AND ets_schema_object ->> 'version' = ?";
 
         try {
-            return jdbcTemplate.query(sql, new Object[]{name, version}, new SchemaRowMapper()).get(0);
+            return jdbcTemplate.queryForObject(sql, new Object[]{name, version}, new SchemaMapper());
         } catch (EmptyResultDataAccessException e) {
             throw new NoSuchSchemaException("EventType \"" + name
                     + "\" has no schema with version \"" + version + "\"", e);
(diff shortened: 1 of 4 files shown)


src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
```diff
@@ -37,12 +37,7 @@ public EventTypeSchema getSchemaVersion(final String name, final String version)
                 "WHERE ets_event_type_name = ? AND ets_schema_object ->> 'version' = ?";
 
         try {
-            final List<EventTypeSchema> schemas =
-                    jdbcTemplate.query(sql, new Object[]{name, version}, new SchemaRowMapper());
-            if (schemas.size() != 1)
-                throw new InternalNakadiException(
```
adyach: you do not need InternalNakadiException  on line https://github.com/zalando/nakadi/pull/505/files#diff-96bb991c5b6b4a062777cd0554da6b32L35

adyach:
:shipit:

src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -48,18 +47,16 @@ public SchemaService(final SchemaRepository schemaRepository,
...
         try {
             final EventTypeSchema schema = schemaRepository.getSchemaVersion(name, version);
             return Result.ok(schema);
         } catch (final NoSuchSchemaException e) {
             LOG.debug("Could not find EventTypeSchema version: {} for EventType: {}", version, name);
-            return Result.problem(e.asProblem());
-        } catch (final InternalNakadiException e) {
-            LOG.error("Problem loading event type schema version " + version + " for EventType " + name, e);
-            return Result.problem(e.asProblem());
+            return Result.problem(Problem.valueOf(Response.Status.NOT_FOUND,
```
lmontrieux: Done in https://github.com/zalando/nakadi/commit/25ef9b2d8c714d0c6d384f5b42e52ef5a05904e0

src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
lmontrieux: Done in https://github.com/zalando/nakadi/commit/41ec53ac33040bd758b145dce52c1545a3353cba

PR #509, opened 2016-12-13
Aruha 400 make compatibility mode required

rcillo:
Some users have the bad habit of deleting and recreating event types. If they did it after this release, it would inadvertently create a `compatible` event type, which has strict validation, and could break their process.

Making `compatibility_mode` required, at least will make them aware of the changes, since it will force them to read the docs about this new field.


rcillo:
It has been argued that some teams programatically create event types and that changing the set of required fields for an event type would break such clients.

PR #511, opened 2016-12-13
ARUHA-479 Preload event types on startup

antban:


rcillo:
@antban it makes 1 query per event type, so during start up it's going to make some 1k queries. I think it's not a problem, but didn't check. Have you considered performing just a single query and somehow cache everything?

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -37,6 +40,21 @@ public EventTypeCache(final EventTypeRepository eventTypeRepository, final Curat
...
         this.validatorCache = setupInMemoryValidatorCache(eventTypeCache);
         this.cacheSync = setupCacheSync(zkClient);
+        preloadEventTypes(eventTypeRepository);
+    }
+
+    private void preloadEventTypes(final EventTypeRepository eventTypeRepository) {
+        final long start = System.currentTimeMillis();
+        eventTypeRepository.list().stream().map(EventType::getName).forEach((name) -> {
+            try {
+                this.getEventType(name);
```
v-stepanov: Is it possible to provide a solution that will not get event-types from DB one-by-one?

antban:
@rcillo @v-stepanov Changed not to load same data several times

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -81,71 +121,69 @@ public void removed(final String name) throws Exception {
...
+    public EventType getEventType(final String name) throws NoSuchEventTypeException, InternalNakadiException {
+        return getCached(name).map(CachedValue::getEventType).orElse(null);
+    }
 
-        cacheSync.getListenable().addListener(listener);
+    public EventTypeValidator getValidator(final String name) throws InternalNakadiException, NoSuchEventTypeException {
+        return getCached(name).map(CachedValue::getEventTypeValidator).orElse(null);
     }
 
     private PathChildrenCache setupCacheSync(final CuratorFramework zkClient) throws Exception {
```
lmontrieux: I know you haven't changed this, but would it make sense to throw something more specific than Exception?

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -81,71 +121,69 @@ public void removed(final String name) throws Exception {
...
+        try {
+            zkClient.create()
+                    .creatingParentsIfNeeded()
+                    .withMode(CreateMode.PERSISTENT)
+                    .forPath(ZKNODE_PATH);
+        } catch (KeeperException.NodeExistsException expected) {
+            // silently do nothing since it means that the node is already there
```
lmontrieux: Should we perhaps log this, in case someone ever has to debug this?

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
antban: @lmontrieux Well, There is no more specific exception. Zk methods all throw this kind of exception, and I can't change it to something more valuable.

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
antban: Well, I think that we do not want it. That just means that initialization was already made some time ago. Even if we will log it, it won't say anything to us.

lmontrieux:
:shipit:

v-stepanov:
I wanted to take a look, so I will block it for now

v-stepanov:
👎

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -3,63 +3,103 @@
...
-    }
-
-    public EventType getEventType(final String name) throws NoSuchEventTypeException, InternalNakadiException {
+    /**
+     * Preloads cache.
+     * The problem with preload is that notifications about modifications may be skipped while initializing cache.
+     * That is why we are using lock for updates. In normal case (without preload) this is covered by LoadingCache
+     * by itself
+     *
+     * @param eventTypeRepository
```
v-stepanov: if you don't describe param - it's better to remove it from doc

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -81,71 +121,69 @@ public void removed(final String name) throws Exception {
...
-            private void invalidateCacheKey(final PathChildrenCacheEvent event) {
-                final String path[] = event.getData().getPath().split("/");
-                final String key = path[path.length - 1];
-
-                validatorCache.invalidate(key);
-                eventTypeCache.invalidate(key);
-            }
-        };
+    public EventType getEventType(final String name) throws NoSuchEventTypeException, InternalNakadiException {
+        return getCached(name).map(CachedValue::getEventType).orElse(null);
```
v-stepanov: If it can return null - the method should be declared as `@Nullable`, and it should be properly handled in the places where it is called. The other possibility here will be to use `orElseThrow(...)` instead.

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -81,71 +121,69 @@ public void removed(final String name) throws Exception {
...
+    }
 
-        cacheSync.getListenable().addListener(listener);
+    public EventTypeValidator getValidator(final String name) throws InternalNakadiException, NoSuchEventTypeException {
```
v-stepanov: If it can return null - the method should be declared as `@Nullable`, and it should be properly handled in the places where it is called. The other possibility here will be to use `orElseThrow(...)` instead.

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -3,63 +3,103 @@
...
+import java.util.stream.Collectors;
 
 public class EventTypeCache {
 
     public static final String ZKNODE_PATH = "/nakadi/event_types";
     public static final int CACHE_MAX_SIZE = 100000;
+    private static final Logger LOG = LoggerFactory.getLogger(EventTypeCache.class);
+
+    private static class CachedValue {
+        final EventType eventType;
```
v-stepanov: private?

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -3,63 +3,103 @@
...
+        final EventTypeValidator eventTypeValidator;
```
v-stepanov: private?

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
antban: It doesn't make any sense to make private variable inside nested private class. Nothing will be changed.

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
antban: The same here

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
antban: Ok

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
antban: ok

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
antban: ok

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
v-stepanov: For me the difference is very big.
Currently your class is mutable. If you make these fields private - it will become immutable.

Actually I don't know a single good case to make a property with package visibility.

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
v-stepanov: same here

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
antban: @v-stepanov Let's discuss it offline

v-stepanov:
👍

rcillo:
:shipit:

PR #512, opened 2016-12-14
ARUHA-456: fixed error message for client; added logging;

v-stepanov:
@lmontrieux @adyach @rcillo @antban please review.
Here I fixed error message and added logging.


lmontrieux:
:shipit:

PR #515, opened 2016-12-16
Aruha 473 check size of events

lmontrieux:
This PR checks the size of all events in a batch. If there is at least one event that is over the maximum size *after enrichment*, then the entire batch is rejected with a 422.


src/main/java/org/zalando/nakadi/domain/EventPublishingStep.java
```diff
@@ -4,6 +4,7 @@
     NONE,
     VALIDATING,
     ENRICHING,
+    VALIDATING_SIZE,
```
v-stepanov: Unfortunately you can't change this. Because this enum is defined in our API and if you add values here that will mean that we add backwards incompatible changes.
https://github.com/zalando/nakadi/blob/master/api/nakadi-event-bus-api.yaml#L1817

Why do you want a separate step for that? I thought I can be part of VALIDATING step.

lmontrieux: If we make it part of VALIDATING, then we will check the size of the event before enriching. I think it makes more sense to check the size after enriching, since the enrichment strategy is chose by the user who created the event type

lmontrieux: I guess I could either check the size during validation, in which case we ignore whatever is added by the enrichment step; or, go back to validation after enriching, but that may be very confusing to users

src/main/java/org/zalando/nakadi/service/EventPublisher.java
```diff
@@ -76,6 +83,9 @@ public EventPublishResult publish(final JSONArray events, final String eventType
         } catch (final EnrichmentException e) {
             LOG.debug("Event enrichment error: {}", e.getMessage());
             return aborted(EventPublishingStep.ENRICHING, batch);
+        } catch (final EventSizeValidationException e) {
+            LOG.debug("Event size validation error: ){", e.getMessage());
```
adyach: ){ -> {} ?

src/main/resources/application.yml
```diff
@@ -78,6 +78,7 @@ nakadi:
   auth:
     plugin:
       factory: org.zalando.nakadi.plugin.auth.DefaultApplicationServiceFactory
+  event.max.bytes: 1000000
```
lmontrieux: I went for the default value in Kafka, which is 1,000,000, but I'm happy to go for a real 1MB if you think it is better

adyach: Also there is a problem with that, if you occasionally want to change producer's property message.max.bytes(and others related), let say to 3MB, it won't work because you restrict it to 1MB and you always have to remember about event.max.bytes.

adyach: https://kafka.apache.org/090/documentation.html#brokerconfigs
it is different according to docs :)
message.max.bytes= 1000012

v-stepanov: > Also there is a problem with that, if you occasionally want to change producer's property message.max.bytes(and others related), let say to 3MB, it won't work because you restrict it to 1MB and you always have to remember about event.max.bytes.

Why is it a problem?

adyach: because you increased kafka broker message size to 3MB and expect that it is gonna be 3MB, but actually it only 1MB

v-stepanov: @adyach Why you expect that if you know that Nakadi validates event size as well?

src/main/java/org/zalando/nakadi/domain/EventPublishingStep.java
lmontrieux: https://github.com/zalando/nakadi/pull/515/commits/a06139221c7c0cd45bc7c4798b8b1ae7bd083551 and https://github.com/zalando/nakadi/commit/0a791da5ab7ee644797d845445df532d21516361 and https://github.com/zalando/nakadi/commit/df5b675d645fc83bfdba29751d29b8f85e8160e1

src/main/java/org/zalando/nakadi/service/EventPublisher.java
```diff
@@ -114,14 +123,18 @@ private void partition(final List<BatchItem> batch, final EventType eventType) t
...
         for (final BatchItem item : batch) {
             item.setStep(EventPublishingStep.VALIDATING);
             try {
                 validateSchema(item.getEvent(), eventType);
+                validateEventSize(item);
             } catch (final EventValidationException e) {
                 item.updateStatusAndDetail(EventPublishingStatus.FAILED, e.getMessage());
                 throw e;
+            } catch (final EventSizeValidationException e) {
+                item.updateStatusAndDetail(EventPublishingStatus.ABORTED, e.getMessage());
```
v-stepanov: I think here it should be `FAILED` like in a case when there was a schema validation problem. Because this is exactly the item that failed.

lmontrieux: https://github.com/zalando/nakadi/commit/957938b9bb4fa56ad95daeff1ea4d9c5d847e56f and https://github.com/zalando/nakadi/commit/49c184652e46d3748b1a8067d55297ec07c48fec

src/main/java/org/zalando/nakadi/service/EventPublisher.java
```diff
@@ -145,6 +151,11 @@ private void validateSchema(final JSONObject event, final EventType eventType) t
         }
     }
 
+    private void validateEventSize(final BatchItem item) throws EventValidationException {
+        if (item.getEvent().toString().getBytes().length > nakadiSettings.getEventMaxBytes())
```
valgog: Never use `getBytes()` without setting the default locale. Please always pass https://docs.oracle.com/javase/7/docs/api/java/nio/charset/StandardCharsets.html#UTF_8 object into that function.

Actually we have to check that we do not use cases where this stupid `getByte()` is done.

valgog: Another problem is to actually to do that conversion. There is the byte array representation that you can use. Or you should store the size into the item to avoid that conversion.

src/main/java/org/zalando/nakadi/service/EventPublisher.java
```diff
@@ -145,6 +151,11 @@ private void validateSchema(final JSONObject event, final EventType eventType) t
...
+            throw new EventValidationException("Event too large");
```
valgog: Please provide a proper error message that includes the size and the limit.

src/test/java/org/zalando/nakadi/service/EventPublisherTest.java
```diff
@@ -128,6 +139,147 @@ public void whenValidationFailsThenSubsequentItemsAreAborted() throws Exception
     }
 
     @Test
+    public void whenEventIsTooLargeThenResultIsAborted() throws Exception {
```
valgog: Please provide a test with multibyte chars in the message body.

lmontrieux: There are two tests with multibyte characters already, lines 252-280. The first one is an event that is just 1 byte over the limit; the second one is exactly at the limit. Multibyte events are created on lines 502-512

valgog:
@lmontrieux 

After thinking about this more, it looks like we do not have an well defined idea how to limit the size of the messages. 

The message itself has the size (in bytes).

The message is being enriched by information, that is also increasing the size of the message, written to Kafka. 

This means that actual limitation on the customer message size is influenced by Kafka message limits and the enrichments, that we provide.

To be able to correctly show the reason for applying some limits on the customer message size, we have to also be able to extract and show information about added sizes by every provided enrichment. Such numbers would be also nice to have for monitoring anyway.

My suggestion currently would be: 
1. extend Item object to keep both byte[] and String representation of the original message (check that conversions to and from String are done minimal amount of times)
2. extend Item object to keep statistics per enrichment (currently only added size in bytes per enrichment)
3. include information about the original and enriched sizes per enrichment into error message

What do you think?

v-stepanov:
@valgog 
I just wanted to mention that this was supposed to be a simple bugfix, not a spaceship-size feature.
Current situation is that message size is not checked anywhere and when somebody pushes message bigger than 1MB (kafka limit) - we respond with 207 (which is error situation for us) and have hole in the published batch.

The idea was very simple - to limit the message size to 1MB in Nakadi and 2MB in Kafka (so that 1MB + enrichment + kafka metadata will always fit to 2MB)

Also I have doubts about point 1 you suggested. That will in fact double Nakadi memory usage.

v-stepanov:
sorry, I accidentally closed PR :)

lmontrieux:
@valgog @v-stepanov I've had a short discussion offline with @valgog about this, but with Hack week I hadn't updated this PR yet. The way things work is:
* producer sends an event of size x;
* nakadi enriches the event, whose size is now x + e;
* the kafka client serialises the event somehow, and add a bit of metadata. Size is now x + e + k.
* kafka has 4 different configuration options for the max size of a message: the max size that a broker will accept (default 1MB); the max size of a log segment (must be higher than the max size of a message, default 1GB); the max size of data that a broker can replicate (default 1MB); the max size a consumer can read (default 1MB). Currently we use the defaults.

The idea behind this fix was to limit the size of an event to 1MB *before* enrichment (limit x only), but bump up all max values in Kafka to 2MB, to be safe (e+k are just a few bytes long). This way we will accept slightly larger messages than we do now. This only works assuming that increasing max values in Kafka will not increase its memory usage. I haven't looked into it but I'd be surprised in Kafka were to always allocate the max message size for each message.

As I recall, the outcome of this discussion was:
1/ limit size in validation, as originally planned;
2/ next week, as a different ticket, work on a more accurate way to limit the size of events, and keep statistics about the size of enrichments.

lmontrieux:
A quick look at the kafka code doesn't seem to indicate any allocation of max message size for each message.

src/main/java/org/zalando/nakadi/service/EventPublisher.java
```diff
@@ -145,6 +151,11 @@ private void validateSchema(final JSONObject event, final EventType eventType) t
         }
     }
 
+    private void validateEventSize(final BatchItem item) throws EventValidationException {
+        if (item.getEventSize() > nakadiSettings.getEventMaxBytes())
```
rcillo: We better add a rule in the code style for enforcing braces usage. I think it's the preferred style over the code base. While we don't do it, could you please add braces here?

src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -58,4 +59,8 @@ public void updateStatusAndDetail(final EventPublishingStatus publishingStatus,
         response.setPublishingStatus(publishingStatus);
         response.setDetail(detail);
     }
+
+    public int getEventSize() {
+        return event.toString().getBytes(StandardCharsets.UTF_8).length;
```
rcillo: I think we should keep the original string `eventsAsString` in order to avoid calling `toString` which is too costly.

rcillo: But we can try this way and see the impact. I think we can keep it as it is and test the impact on staging.

rcillo:
@lmontrieux for me it looks good 👍 I think we should make this limit explicit in the swagger file as well. Could you please add this information to the api file and make it clear how many bytes are accepted?

src/main/java/org/zalando/nakadi/domain/BatchItem.java
rcillo: @lmontrieux is it guaranteed that `toString` method won't add any spaces and eventually add some bytes to the message sent?

lmontrieux: @rcillo yes: https://developer.android.com/reference/org/json/JSONObject.html#toString() (I know this is the Android documentation, but the code on github says the same).

src/main/java/org/zalando/nakadi/service/EventPublisher.java
lmontrieux: https://github.com/zalando/nakadi/pull/515/commits/f3158365fe0e670e8479eb5581103ae62e0b4439

v-stepanov: @rcillo I agree, I will add a point for voting to our Code Style Radar

lmontrieux:
@rcillo done in 8871dea

rcillo:
Maybe someone wants to give some more feedback, but I'll leave here my approval.

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/domain/BatchItem.java
antban: @lmontrieux For me this approach is not so good.
At first we should decide which data we are storing to event repository.
If we are storing object(JSON), than it makes no sense to check it on validation step, because size of object means nothing on this step.
Probably we should pass items with strings or byte array to TopicRepository, at least these two types have estimation on size, TopicRepository should store this representation (string or byte[]), and this representation should be used on layer below (KafkaTopicRepositry)
So for me it's better to first validate, then partition, then enrich, than prepareToPublish. On step `prepareToPublish` one should convert nakadi representation to storage representation and validate size. In this approach we will avoid several toString converters, and we will be sure which kind of data is stored in TopicRepository.

lmontrieux: @antban @rcillo We've discussed something similar on Jira ticket 511. What I've been doing here is basically telling users "You can't send any event larger than 1,000,000 bytes". Pro: it's easy for producers to know how large their event can be; Con: we have to take into account that enrichment and metadata will increase the size of the event, so we have to configure Kafka to allow for events that are larger than 1,000,000 bytes.

The solution @antban mentions is different, I think: "We will not store events larger than 1,000,000 bytes". This means that the actual max size for a producer will be a bit lower, depending on the enrichment strategies selected. Pro: We know exactly how large the largest events will be in Kafka, and we can configure its maximum size very easily; Con: it's hard for producers to tell in advance whether an event will be accepted or not (if it's close to the limit).

Whatever solution we choose, I agree with @antban that we should improve BatchItem to reduce the number of toString() operations, but this can perhaps be done with another ticket.

rcillo: @antban I agree that calling `toString` might have an impact in performance, so I suggest to try load test on staging. But in terms of behaviour from the producer's point of view, we better provide a clear contract. If we had a variable contract, sometimes accepting 999,990 bytes, sometimes 999,985 bytes, depending on enrichment and so on, users might get frustrated by the lack of consistency.

For example, we are going to add schema version which enlarges the metadata even further. It would possibly break publishers with event close to the limit.

How we are going to achieve this is another problem.

PetrGlad:
I am sure that no one really interested in exact message sizes. The real concern is that unreasonably large ones are rejected. If someone _needs_ to send message that is on the verge then limit should be substantially larger.

rcillo:
@PetrGlad how do you answer the question: "What's the biggest message size I can send to Nakadi?"

PetrGlad:
@rcillo There was discussion how to measure message sizes, and the precision of that measurement WRT serialization and enrichment. I'd say that precision does not really matter. For me as a consumer of events more important what to expect rather whether it slightly less or slightly more than 1MiB.

lmontrieux:
After further discussion this morning, we agreed to check the event size as sent by the producer, i.e. before the events are transformed into an array of `JSONObject`s. If the producer includes unnecessary spaces in his events, they will count towards the maximum event size.

src/main/java/org/zalando/nakadi/domain/BatchFactory.java
```diff
@@ -1,16 +1,62 @@
...
 
 public class BatchFactory {
 
-    public static List<BatchItem> from(final JSONArray events) {
-        final List<BatchItem> batch = new ArrayList<>(events.length());
-        for (int i = 0; i < events.length(); i++) {
-            batch.add(new BatchItem(events.getJSONObject(i)));
+    public static List<BatchItem> from(final String events) {
+        final List<BatchItem> batch = new ArrayList<>();
+        StringBuilder sb = new StringBuilder();
```
rcillo: I was wondering if isn't there a way to build this substring with zero copy. I couldn't find, because since Java 7 String.substring creates a copy of the original char array.

I think it would be valuable to look at `JSONTokener` and try to figure out if isn't there a way to avoid extra copies of this string. This is the most executed code path in the project. I would expect that this change has an impact on garbage collection and overall performance.

rcillo: Maybe if it wasn't a string but an array of bytes, then we could just create strings using `String(bytes[] bytes, int offset, int length)`. I'm not sure, I think we should investigate more and reduce memory footprint.

src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -3,19 +3,22 @@
...
     private final BatchItemResponse response;
     private final JSONObject event;
     private String partition;
     private String brokerId;
+    private int eventSize;
 
-    public BatchItem(final JSONObject event) {
+    public BatchItem(final String event) {
+        this.event = new JSONObject(event);
+        this.eventSize = event.getBytes(StandardCharsets.UTF_8).length;
```
rcillo: I think this method also generates a copy of the original array.

src/main/java/org/zalando/nakadi/domain/BatchFactory.java
lmontrieux: Good point. I wonder if we could use `CharBuffer.wrap(..)`.

lmontrieux: I looked into this, and short of using another JSON parser, or doing some ugly size validation inside the controller, I don't see how we can improve this. We should discuss this in more detail.

rcillo: Ok. So let's move forward with this solution and we can test the performance impact on staging.

src/acceptance-test/java/org/zalando/nakadi/webservice/EventPublishingAT.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.http.HttpStatus;
+import org.junit.Test;
+import org.zalando.nakadi.domain.EventType;
+import org.zalando.nakadi.webservice.utils.NakadiTestUtils;
+
+import java.text.MessageFormat;
+
+import static com.jayway.restassured.RestAssured.given;
+
+public class EventPublishingAT extends BaseAT {
```
lmontrieux: Removed in https://github.com/zalando/nakadi/pull/515/commits/b9c36f89d2afabe292c43f0a284b6b4f9c4eb7a7

src/main/java/org/zalando/nakadi/service/EventPublisher.java
rcillo: @lmontrieux it seems that this message wasn't changed.

lmontrieux: @rcillo Pretty sure I removed the whole method :). The size computation is now in the `BatchItem` constructor

lmontrieux: @rcillo my mistake, I was wrong. Fixing it now

rcillo:
:shipit:

api/nakadi-event-bus-api.yaml
```diff
@@ -316,7 +316,9 @@ paths:
         incoming Events. Validation rules are evaluated in the order they are defined and the Event
         is **rejected** in the first case of failure. If the offending validation rule provides
         information about the violation it will be included in the `BatchItemResponse`.  If the
-        `EventType` defines schema validation it will be performed at this moment.
+        `EventType` defines schema validation it will be performed at this moment. The size of each
+        Event will also be validated. The maximum size per Event is 1,000,000 bytes, before
```
antban: @lmontrieux As far as I know, max.message.bytes for topic is 1,000,000 bytes. That means, that if we will enrich something, then it is guaranteed, that this message won't be written

src/main/java/org/zalando/nakadi/domain/BatchFactory.java
```diff
@@ -1,16 +1,61 @@
...
+                    if (sb.length() > 0) {
+                        batch.add(new BatchItem(sb.toString()));
+                    }
+                    sb = new StringBuilder();
+                }
+            }
+        }
+
+        if (sb.length() != 0) {
+            batch.add(new BatchItem(sb.toString()));
```
antban: Well, I completely disagree with this (I mean the fact that we are measuring user input), but there is a bug here: 
It seems that message `[          ]` will create one batch item (but there is nothing there).

api/nakadi-event-bus-api.yaml
lmontrieux: @antban we had agreed to set it to something a bit lower, but I was waiting for today's standup to agree on a value. I've suggested 999,000 bytes on hipchat.

lmontrieux: @antban https://github.com/zalando/nakadi/pull/515/commits/0330145b3c2d38afa0cfdc6a6b452179ed2474ee

src/main/java/org/zalando/nakadi/domain/BatchFactory.java
lmontrieux: @antban You were right, there was a bug with `[ ]`. No batch would be created, but a JSONException would be thrown. I fixed it in https://github.com/zalando/nakadi/pull/515/commits/0330145b3c2d38afa0cfdc6a6b452179ed2474ee

lmontrieux: @antban As for measuring user input, I thought it was what we had agreed on earlier this week (or was it last week?). Happy to discuss it offline if I'm wrong.

rcillo:
@lmontrieux is enrichment always using less than 100 bytes?

lmontrieux:
@rcillo Well, we have (just under) 1,000 bytes of wiggle room for enrichment. Currently we only have one enrichment strategy, which sets: `received_at`, `event_type`, `flow_id`, `partition`, and `version`. Looks fairly stable in its size.

Of course, if we implement more enrichment strategies, we'll have to revise this.

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/domain/BatchFactory.java
```diff
@@ -13,10 +13,35 @@
         int brackets = 0;
         boolean insideQuote = false;
         boolean escaped = false;
-        if ((!events.startsWith("[")) || (!events.endsWith("]"))) {
-            throw new JSONException("Array must be surrounded with square brackets");
+        int start = 0;
+        final int length = events.length();
+        int end = length - 1;
+
+        while ((events.charAt(start) == ' '
```
rcillo: I think this condition could be extracted to it's own method, something like `isBlankCharacter`.

src/main/java/org/zalando/nakadi/domain/BatchFactory.java
```diff
@@ -13,10 +13,35 @@
...
+                || events.charAt(start) == '\t'
+                || events.charAt(start) == '\n'
+                || events.charAt(start) == '\r')
+                && start < end) {
+            start++;
+        }
+        while ((events.charAt(end) == ' '
```
rcillo: If this condition is extracted to its own method, then this should be refactored to reuse it.

src/main/java/org/zalando/nakadi/domain/BatchFactory.java
rcillo: or as per the test name `isEmptyCharacter`

lmontrieux: Good point, will extract

lmontrieux:
If a batch had "empty" (carriage return, tab, space,...) characters around it, it was rejected with a 400 for malformed data. When I made the comment I thought the issue was with large events, but investigating further, I found that it wasn't a size issue, but a trailing carriage return after the batch.

src/main/java/org/zalando/nakadi/domain/BatchFactory.java
lmontrieux: https://github.com/zalando/nakadi/pull/515/commits/ffd729a245964c34656fc43e8e3044465f93e7ff

src/main/java/org/zalando/nakadi/domain/BatchFactory.java
lmontrieux: https://github.com/zalando/nakadi/pull/515/commits/ffd729a245964c34656fc43e8e3044465f93e7ff

rcillo:
:shipit:

src/main/resources/application.yml
```diff
@@ -59,7 +59,6 @@ nakadi:
     send.timeoutMs: 5000
     batch.size: 5242880
     linger.ms: 0
-    fetch.message.max.bytes: 2000000
```
rcillo: Why is this needed?

lmontrieux: It is not needed anymore, since we decided to stick to the default value for Kafka. I forgot to remove it, and now I see in Scalyr that it was never correct in the first place, so this line is just ignored.

rcillo:
:shipit:

lmontrieux:
@rcillo I broke the build with my previous commit. This removes all code related to the update in the consumer's max fetch size (does not affect producer's max event size btw)

src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -99,9 +99,7 @@ private void updateBrokers() {
     }
 
     public Properties getKafkaConsumerProperties() {
-        final Properties consumerProps = (Properties) kafkaProperties.clone();
-        consumerProps.put("fetch.message.max.bytes", kafkaSettings.getFetchMessageMaxBytes());
```
adyach: @lmontrieux I think we need to migrate to bigger message size

lmontrieux: @adyach yes, but this is not working. I get this in the logs: `[org.apache.kafka.clients.consumer.ConsumerConfig] --- The configuration fetch.message.max.bytes = 2000000 was supplied but isn't a known config.`

I think we could do this as a separate ticket, where we increase the max event size from 999,000 bytes to 1,000,000 bytes

rcillo:
:shipit:

v-stepanov:
:shipit:

PR #516, opened 2016-12-20
Introducing lombok

youssefmamdouh:
Introduce lombok to Nakadi. This PR only applies lombok for domain objects, if the idea is Ok, will apply it to more code.

### DIFF
--- a/src/main/java/org/zalando/nakadi/domain/EventType.java
+++ b/src/main/java/org/zalando/nakadi/domain/EventType.java
@@ -1,26 +1,31 @@
 package org.zalando.nakadi.domain;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import org.zalando.nakadi.partitioning.PartitionStrategy;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
 
 import javax.annotation.Nullable;
+
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 import javax.validation.constraints.Pattern;
 import javax.validation.constraints.Size;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Set;
 
-import static java.util.Collections.unmodifiableList;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
 public class EventType {
 
     public static final List<String> EMPTY_STRING_LIST = new ArrayList<>(0);
 
     @NotNull
-    @Pattern(regexp = "[a-zA-Z][-0-9a-zA-Z_]*(\\.[a-zA-Z][-0-9a-zA-Z_]*)*", message = "format not allowed" )
+    @Pattern(regexp = "[a-zA-Z][-0-9a-zA-Z_]*(\\.[a-zA-Z][-0-9a-zA-Z_]*)*", message = "format not allowed")
     @Size(min = 1, max = 255, message = "the length of the name must be >= 1 and <= 255")
     private String name;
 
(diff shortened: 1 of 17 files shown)


src/main/java/org/zalando/nakadi/domain/SubscriptionEventTypeStats.java
```diff
@@ -27,38 +26,13 @@ public String getEventType() {
...
-
-        public String getPartition() {
-            return partition;
-        }
-
-        public String getState() {
-            return state;
-        }
-
-        @Nullable
```
rcillo: @youssefmamdouh we try to follow a rule for null handling in this project (I guess it's not written anywhere, unfortunately).

Every method that could return a `null` is marked with `@Nullable` and the code that uses this method should be aware of that and treat it properly when necessary. Otherwise you can just consider that all other methods never return null.

This way we have been able to avoid null pointer problems so far.

Replacing all this code by a `getter/setter` annotation would be fine if we did not loose the `nullable` annotation. Is it possible to fix this?

rcillo:
I have no experience with `lombok` I did a quick research and found some different opinions on it. Are you aware of these concerns http://stackoverflow.com/a/37596394 ?

My first impression is that it makes code cleaner but at the expense of a lot of magic. The code it removed is mostly IDE generated code which is easy to maintain. I'm skeptical about new dependencies if they could compromise the long term maintainability of the project.

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -58,137 +63,4 @@
...
 
     private Set<String> readScopes;
-
-    public EventType() {
-        this.validationStrategies = Collections.emptyList();
-        this.enrichmentStrategies = Collections.emptyList();
-        this.partitionStrategy = PartitionStrategy.RANDOM_STRATEGY;
-        this.options = new EventTypeOptions();
-        this.writeScopes = Collections.emptySet();
-        this.readScopes = Collections.emptySet();
```
v-stepanov: All this initialization is lost now

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -58,137 +63,4 @@
...
-    public EventTypeStatistics getDefaultStatistic() {
-        return defaultStatistic;
-    }
-
-    public void setDefaultStatistic(final EventTypeStatistics defaultStatistic) {
-        this.defaultStatistic = defaultStatistic;
-    }
-
-    public List<String> getPartitionKeyFields() {
-        return unmodifiableList(partitionKeyFields != null ? partitionKeyFields : EMPTY_STRING_LIST);
```
v-stepanov: this logic is also lost now.

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -58,137 +63,4 @@
...
-    public EventTypeOptions getOptions() {
-        return options;
-    }
-
-    public void setOptions(final EventTypeOptions options) {
-        this.options = options;
-    }
-
-    public Set<String> getWriteScopes() {
-        return Collections.unmodifiableSet(writeScopes);
```
v-stepanov: this behavior is lost now

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -58,137 +63,4 @@
...
-    }
-
-    public void setWriteScopes(final Set<String> writeScopes) {
-        this.writeScopes = writeScopes == null ? Collections.emptySet() : writeScopes;
```
v-stepanov: this behavior is lost now

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -58,137 +63,4 @@
...
-    }
-
-    public Set<String> getReadScopes() {
-        return Collections.unmodifiableSet(readScopes);
```
v-stepanov: this behavior is lost now

src/main/java/org/zalando/nakadi/domain/EventType.java
```diff
@@ -58,137 +63,4 @@
...
-    }
-
-    public void setReadScopes(final Set<String> readScopes) {
-        this.readScopes = readScopes == null ? Collections.emptySet() : readScopes;
```
v-stepanov: this behavior is lost now

src/main/java/org/zalando/nakadi/domain/ItemsWrapper.java
```diff
@@ -20,8 +22,4 @@
     public ItemsWrapper(@JsonProperty("items") final List<T> items) {
         this.items = items;
     }
-
-    public List<T> getItems() {
-        return Collections.unmodifiableList(items);
```
v-stepanov: this behavior is lost now

src/main/java/org/zalando/nakadi/domain/SubscriptionEventTypeStats.java
youssefmamdouh: @rcillo if this rule is also strictly followed for Getters and Setters, then i do not think this can be done using lombok. The framework just assumes those methods are to be generated in a conventional way. Think i will close this PR, thanks for mentioning this rule.

v-stepanov:
@youssefmamdouh Hi. I marked several places that lost the original behavior.
I think that unfortunately lombok doesn't fully fit here.

youssefmamdouh:
@v-stepanov you're right, i can see this also now, thanks

PR #517, opened 2016-12-21
aruha-509 fix bug with $ref

rcillo:
$ref have been accidentally blocked.

This bug actually reveals a core schema meta-schema definition
problem. I've extended the meta-schema definition to accept $ref.

### DIFF
--- a/src/test/resources/org/zalando/nakadi/validation/invalid-json-schema-examples.json
+++ b/src/test/resources/org/zalando/nakadi/validation/invalid-json-schema-examples.json
@@ -147,9 +147,7 @@
         }
       }
     },
-    "errors": ["Invalid schema found in [#/definitions/paradox]: extraneous key [not] is not permitted",
-      "Invalid schema found in [#/items]: extraneous key [$ref] is not permitted",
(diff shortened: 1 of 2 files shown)


src/test/resources/org/zalando/nakadi/validation/invalid-json-schema-examples.json
```diff
@@ -147,9 +147,7 @@
         }
       }
     },
-    "errors": ["Invalid schema found in [#/definitions/paradox]: extraneous key [not] is not permitted",
-      "Invalid schema found in [#/items]: extraneous key [$ref] is not permitted",
```
v-stepanov: It's interesting that you actually had a test which checks that $ref is forbidden 😄

v-stepanov:
:shipit:

src/test/resources/org/zalando/nakadi/validation/invalid-json-schema-examples.json
rcillo: 😅 yes... but I maybe should open a PR to the json-schema meta-schema repository with this same fix. It's a bug in the original file from the specs repo as well.

PR #518, opened 2016-12-21
aruha-510: kafka streams in nakadi

adyach:
This PR brings possibility to apply expressions on the events stream using KafkaStreams in order to filter them.
Streaming of filtered events is possible in 2 ways:
- directly in the output stream in a form of json objects
- to provided topic

It requires Kafka 0.10.1.0


LappleApple:
Hi @adyach, what's the status of this PR, now that it has conflicts? Will you resolve them or find some other solution? cc @v-stepanov @rcillo @antban

adyach:
@LappleApple it was a hackweek project :) it is not possible to merge it right now, because we do not have the required Kafka version on live yet. Probably, it is better to close it for now.

PR #519, opened 2016-12-22
Experimental schema generation

rcillo:
This is an experimental endpoint to generate a schema from an event. It's a very simple solution, just to show that it would not be so hard to do it in a proper way.


src/main/java/org/zalando/nakadi/validation/schema/SchemaGeneration.java
```diff
@@ -0,0 +1,66 @@
...
+        if (event instanceof String) {
+            return new JSONObject().put("type", "string");
+        } else if (event instanceof Integer) {
+            return new JSONObject().put("type", "number");
+        } else if (event instanceof Boolean) {
+            return new JSONObject().put("type", "boolean");
+        } else if (event == JSONObject.NULL) {
+            return new JSONObject().put("type", "null");
+        } else if (event instanceof JSONArray) {
+            return schemaFor((JSONArray) event);
```
valgog: Will this code ever be called? 😊

rcillo: Yes 👍 if I removed this line of code, it would break this test case here https://github.com/zalando/nakadi/pull/519/files#diff-e0ec26c222e2d4ccfcfd792d8838ef00R96 for example.

rcillo:
Since this PR is not going to be merged any time soon, I'll close it and keep the remote branch if one day we resume this task.

PR #520, opened 2016-12-28
ARUHA-517: made braces required by checkstyle

v-stepanov:
@lmontrieux @antban @rcillo @adyach please review

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventStreamConfig.java
+++ b/src/main/java/org/zalando/nakadi/service/EventStreamConfig.java
@@ -85,19 +85,18 @@ public class EventStreamConfig {
 
     @Override
     public boolean equals(final Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
 
         final EventStreamConfig that = (EventStreamConfig) o;
 
-        if (batchLimit != that.batchLimit) return false;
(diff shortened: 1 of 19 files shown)


src/main/java/org/zalando/nakadi/service/EventStreamConfig.java
```diff
@@ -85,19 +85,18 @@ public String toString() {
...
-        if (streamLimit != that.streamLimit) return false;
-        if (batchTimeout != that.batchTimeout) return false;
-        if (streamTimeout != that.streamTimeout) return false;
-        if (streamKeepAliveLimit != that.streamKeepAliveLimit) return false;
-        if (!topic.equals(that.topic)) return false;
-        return cursors.equals(that.cursors);
-
+        return batchLimit == that.batchLimit && streamLimit == that.streamLimit && batchTimeout == that.batchTimeout
+                && streamTimeout == that.streamTimeout && streamKeepAliveLimit == that.streamKeepAliveLimit
+                && topic.equals(that.topic) && cursors.equals(that.cursors);
```
lmontrieux: Would it be more readable to align the `&&`?:
```
return batchLimit == that.batchLimit
        && streamLimit == that.streamLimit
        && batchTimeout == that.batchTimeout
```

v-stepanov: Well, we don't have a rule regarding that :) 
So I like it this way because it allows to use the whole width of line.
But feel free to put it on the vote in our code formatting radar.

antban:
:+1:

lmontrieux:
:shipit:

PR #522, opened 2016-12-29
ARUHA-520 use getBytes(..) to report actual batch size

lmontrieux:
This PR fixes a bug in batch size reporting for batches with multi-byte characters. I don't think a unit test is necessary for this.

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
@@ -108,7 +108,7 @@ public class EventPublishingController {
                                final String eventsAsString, final int eventCount) {
         if (result.getStatus() == EventPublishingStatus.SUBMITTED) {
             eventTypeMetrics.reportSizing(eventCount,
-                    eventsAsString.getBytes(StandardCharsets.UTF_8).length - eventCount + 1);
+                    eventsAsString.getBytes(StandardCharsets.UTF_8).length - eventCount - 1);
         } else if (result.getStatus() == EventPublishingStatus.FAILED && eventCount != 0) {
             final int successfulEvents= result.getResponses()
                     .stream()
--- a/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
@@ -107,7 +107,8 @@ public class EventPublishingController {
     private void reportMetrics(final EventTypeMetrics eventTypeMetrics, final EventPublishResult result,
                                final String eventsAsString, final int eventCount) {
         if (result.getStatus() == EventPublishingStatus.SUBMITTED) {
-            eventTypeMetrics.reportSizing(eventCount, eventsAsString.getBytes(StandardCharsets.UTF_8).length - 2);
+            eventTypeMetrics.reportSizing(eventCount,
+                    eventsAsString.getBytes(StandardCharsets.UTF_8).length - eventCount + 1);
         } else if (result.getStatus() == EventPublishingStatus.FAILED && eventCount != 0) {
             final int successfulEvents= result.getResponses()
                     .stream()
--- a/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
@@ -114,8 +114,8 @@ public class EventPublishingController {
                     .filter(r -> r.getPublishingStatus() == EventPublishingStatus.SUBMITTED)
                     .collect(Collectors.toList())
                     .size();
-            final double avgEventSize = eventsAsString.getBytes(StandardCharsets.UTF_8).length / eventCount;
-            eventTypeMetrics.reportSizing(successfulEvents,(int)Math.round(avgEventSize * successfulEvents));
+            final double avgEventSize = eventsAsString.getBytes(StandardCharsets.UTF_8).length / (double)eventCount;
+            eventTypeMetrics.reportSizing(successfulEvents, (int)Math.round(avgEventSize * successfulEvents));
         }
     }
 
--- a/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
@@ -1,5 +1,6 @@
 package org.zalando.nakadi.controller;
 
+import com.google.common.base.Charsets;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.slf4j.Logger;
@@ -82,7 +83,7 @@ public class EventPublishingController {
             final JSONArray eventsAsJsonObjects = new JSONArray(eventsAsString);
 
             final int eventCount = eventsAsJsonObjects.length();
-            eventTypeMetrics.reportSizing(eventCount, eventsAsString.length());
+            eventTypeMetrics.reportSizing(eventCount, eventsAsString.getBytes(Charsets.UTF_8).length);
 
             return response(publisher.publish(eventsAsJsonObjects, eventTypeName, client));
         } catch (final JSONException e) {
--- a/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
@@ -86,17 +86,7 @@ public class EventPublishingController {
 
             final int eventCount = eventsAsJsonObjects.length();
             final EventPublishResult result = publisher.publish(eventsAsJsonObjects, eventTypeName, client);
-            if (result.getStatus() == EventPublishingStatus.SUBMITTED) {
-                eventTypeMetrics.reportSizing(eventCount, eventsAsString.getBytes(StandardCharsets.UTF_8).length);
-            } else if (result.getStatus() == EventPublishingStatus.FAILED && eventCount != 0) {
-                final int successfulBatches = result.getResponses()
-                        .stream()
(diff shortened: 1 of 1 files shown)


v-stepanov:
👍

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -82,7 +83,7 @@ private ResponseEntity postEventInternal(final String eventTypeName,
             final JSONArray eventsAsJsonObjects = new JSONArray(eventsAsString);
 
             final int eventCount = eventsAsJsonObjects.length();
-            eventTypeMetrics.reportSizing(eventCount, eventsAsString.length());
+            eventTypeMetrics.reportSizing(eventCount, eventsAsString.getBytes(Charsets.UTF_8).length);
```
v-stepanov: @lmontrieux 
During validation I found one problem. It's not related to your fix, your fix works pretty good.
The thing is that we update statistics before publishing of events. So it can be that events are rejected because of validation problem but we anyway include event sizes to our metrics.
So for me it makes sense to include to statistics only the events that were actually published. What do you think?

lmontrieux: I agree, actually I was also surprised by this. It's not as trivial a fix as I initially thought however, and I think it would be better to include it in the discussion for ARUHA-473. What choices we make there will influence how we can solve that issue.

v-stepanov: But I actually think that to fix it will be quite easy - we just need to move this line down below "publisher.publish...."
And I think this fix will fit pretty good to this ticket.

lmontrieux: I think it's a bit more complicated than that: if we return a 207 then we need to re-compute the sum of the sizes of the events published successfully, as well as their number. However, on second thought, I agree that it makes more sense to fix it here.

lmontrieux: https://github.com/zalando/nakadi/pull/522/commits/faba3b9b8d4eb27f064cf5086e4d6c89e290a532

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -82,9 +85,21 @@ private ResponseEntity postEventInternal(final String eventTypeName,
             final JSONArray eventsAsJsonObjects = new JSONArray(eventsAsString);
 
             final int eventCount = eventsAsJsonObjects.length();
-            eventTypeMetrics.reportSizing(eventCount, eventsAsString.length());
+            final EventPublishResult result = publisher.publish(eventsAsJsonObjects, eventTypeName, client);
+            if (result.getStatus() == EventPublishingStatus.SUBMITTED) {
```
v-stepanov: I think it makes sense now to move this code to separate method (I mean the code that calculates metrics)

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -82,9 +85,21 @@ private ResponseEntity postEventInternal(final String eventTypeName,
...
+                eventTypeMetrics.reportSizing(eventCount, eventsAsString.getBytes(StandardCharsets.UTF_8).length);
+            } else if (result.getStatus() == EventPublishingStatus.FAILED && eventCount != 0) {
+                final int successfulBatches = result.getResponses()
```
v-stepanov: successful events, not batches.

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -82,9 +85,21 @@ private ResponseEntity postEventInternal(final String eventTypeName,
...
+                        .stream()
+                        .filter(r -> r.getPublishingStatus() == EventPublishingStatus.SUBMITTED)
+                        .collect(Collectors.toList())
+                        .size();
+                eventTypeMetrics.reportSizing(successfulBatches,
+                        (eventsAsString.getBytes(StandardCharsets.UTF_8).length / eventCount) * successfulBatches);
```
v-stepanov: I'm not sure about the precision here.
AFAIR if you divide int by int you will get int.

lmontrieux: https://github.com/zalando/nakadi/pull/522/commits/f4f0ca4ece0b1733e6d5da00cf35c4d9556f3ad9

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
lmontrieux: https://github.com/zalando/nakadi/pull/522/commits/f4f0ca4ece0b1733e6d5da00cf35c4d9556f3ad9

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
lmontrieux: https://github.com/zalando/nakadi/pull/522/commits/f4f0ca4ece0b1733e6d5da00cf35c4d9556f3ad9

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -114,6 +104,21 @@ private ResponseEntity postEventInternal(final String eventTypeName,
...
+                               final String eventsAsString, final int eventCount) {
+        if (result.getStatus() == EventPublishingStatus.SUBMITTED) {
+            eventTypeMetrics.reportSizing(eventCount, eventsAsString.getBytes(StandardCharsets.UTF_8).length - 2);
+        } else if (result.getStatus() == EventPublishingStatus.FAILED && eventCount != 0) {
+            final int successfulEvents= result.getResponses()
+                    .stream()
+                    .filter(r -> r.getPublishingStatus() == EventPublishingStatus.SUBMITTED)
+                    .collect(Collectors.toList())
+                    .size();
+            final double avgEventSize = eventsAsString.getBytes(StandardCharsets.UTF_8).length / eventCount;
```
v-stepanov: I think it will not help. It will divide int by int and the result will be int and then it will be converted to double.
So if length is 10 and eventCount is 3, the avgEventSize will be 3.0

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -114,6 +104,21 @@ private ResponseEntity postEventInternal(final String eventTypeName,
...
+            eventTypeMetrics.reportSizing(successfulEvents,(int)Math.round(avgEventSize * successfulEvents));
```
v-stepanov: space missing after `,`

lmontrieux: https://github.com/zalando/nakadi/pull/522/commits/dd961414c7c2bc4fd0709547dafd56cb284b362b

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
lmontrieux: Ooops. https://github.com/zalando/nakadi/pull/522/commits/dd961414c7c2bc4fd0709547dafd56cb284b362b

v-stepanov:
👍

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -99,6 +104,21 @@ private ResponseEntity postEventInternal(final String eventTypeName,
         }
     }
 
+    private void reportMetrics(final EventTypeMetrics eventTypeMetrics, final EventPublishResult result,
+                               final String eventsAsString, final int eventCount) {
+        if (result.getStatus() == EventPublishingStatus.SUBMITTED) {
+            eventTypeMetrics.reportSizing(eventCount, eventsAsString.getBytes(StandardCharsets.UTF_8).length - 2);
```
rcillo: @lmontrieux why `-2`?

lmontrieux: We're looking for the average size of an event. They are sent as an array, so we should ignore the opening and closing square brackets.

Come to think of it, we should also ignore the commas between events. I'll fix that.

rcillo: @lmontrieux I don't think it's going to be that easy. You need to be aware that there are commas in the events themselves. So you would be required to actually "parse" the string to identify these comma separators.

lmontrieux: @rcillo  There are one less comma than the number of JSONObjects in the JSONArray created a few lines earlier, so we can rely on that, I think. The problem is that white spaces, tabs, carriage returns, etc., between objects, will be counted, and they shouldn't.

rcillo: @lmontrieux I see, you're right. But it's already an improvement over what we have and i think it's not worth the effort of solving it with 100% correctness.

lmontrieux: https://github.com/zalando/nakadi/pull/522/commits/c73faaff36245e1521ead8818f7b7ad17ea9e7a7

rcillo:
If it's not too cumbersome (too many mocks) it would be good to have one test that illustrates an input and the expected values reported.

antban:
:+1:

PR #523, opened 2016-12-29
aruha-515 do not enforce additionalProperties for composed schemas

rcillo:
Additional properties should not be enforced when using composed
schemas since it would cause the rejection of all events different from
an empty json object.

### DIFF
--- a/src/main/java/org/zalando/nakadi/validation/JsonSchemaEnrichment.java
+++ b/src/main/java/org/zalando/nakadi/validation/JsonSchemaEnrichment.java
@@ -104,6 +104,7 @@ public class JsonSchemaEnrichment {
             OBJECT_SCHEMA_KEYWORDS.stream().anyMatch(schema::has) ||
             ARRAY_SCHEMA_KEYWORDS.stream().anyMatch(schema::has) ||
             COMPOSED_SCHEMA_KEYWORDS.stream().anyMatch(schema::has) ||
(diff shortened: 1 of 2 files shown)


rcillo:
@v-stepanov @lmontrieux @antban this is ready for review.

v-stepanov:
@rcillo I'm afraid I can't match the description of your PR with the actual code changes. Maybe you can explain it a little bit more detailed.

rcillo:
@v-stepanov in the fiel `strict-validation.json` I added one scenario that would be broken without this PR. Previous to this PR, it would add `additionalProperties=false` in reference schemas (those with `$ref`).

For example, give the user defines a schema

```
{
      "definitions": {
        "foo": {
          "type": "string"
        }
      },
      "properties": {
        "foo": { "$ref": "#/definitions/foo" }
      }
}
```

It would generate

```
{
      "definitions": {
        "foo": {
          "type": "string"
        }
      },
      "properties": {
        "foo": { 
          "$ref": "#/definitions/foo",
          "additionalProperties": false
        }
      },
      "additionalProperties": false
}
```

After this PR it generates

```
{
      "definitions": {
        "foo": {
          "type": "string"
        }
      },
      "properties": {
        "foo": { 
          "$ref": "#/definitions/foo"
        }
      },
      "additionalProperties": false
}
```

lmontrieux:
:shipit:

PR #524, opened 2016-12-29
aruha-507 remove support for allOf

rcillo:
Description of the problem https://techjira.zalando.net/browse/ARUHA-507

### DIFF
--- a/src/main/resources/schema.json
+++ b/src/main/resources/schema.json
@@ -125,7 +125,6 @@
             ]
         },
         "format": { "type": "string" },
-        "allOf": { "$ref": "#/definitions/schemaArray" },
         "anyOf": { "$ref": "#/definitions/schemaArray" },
         "oneOf": { "$ref": "#/definitions/schemaArray" }
     },
--- a/src/test/resources/org/zalando/nakadi/validation/invalid-json-schema-examples.json
+++ b/src/test/resources/org/zalando/nakadi/validation/invalid-json-schema-examples.json
@@ -161,5 +161,16 @@
       ]
     },
     "errors": ["Invalid schema found in [#/anyOf/0]: extraneous key [not] is not permitted"]
+  },
+  {
+    "description": "Do not support allOf",
+    "schema": {
+      "allOf": [
+        {
+          "type": "string"
+        }
+      ]
+    },
+    "errors": ["Invalid schema found in [#]: extraneous key [allOf] is not permitted"]
   }
 ]


lmontrieux:
:shipit:

lmontrieux:
There are style issues, which apparently stem from the new rule that `if` must always use curly brackets. Other than that, it look good :)

rcillo:
@lmontrieux @v-stepanov @antban I thought more about this problem and I have another proposal, instead of deprecating support for it.

Ideally, we should validate `allOf` as usual, without strict validation. But after it passes, we should be sure that no properties were left outside during validation. What I mean is that we should have 100% coverage of the event during validation, which is different from being strict.

Imagine that every time an event property is validated against a sub-schema we "mark" it as covered. Then, the problem is to check if the entire event was covered.

So, we could merge this and reintroduce support in the future once this has been implemented.

Since we have a new compatibility mode `forward` this `allOf` restriction won't affect our users.

rcillo:
There is also another possibility under discussion, which would involve the data-lake dropping properties that were not covered. So we would accept events without strict mode and it would be handled afterwards by the data-lake.

v-stepanov:
@rcillo 
> Ideally, we should validate allOf as usual, without strict validation. But after it passes, we should be sure that no properties were left outside during validation.
I don't really know. Does it change the default behavior of JSON-schema?

rcillo:
@v-stepanov I describe [here](https://docs.google.com/document/d/1FmGHxI9oklxN16LKZe1UG_U21jtQFVCtLtRiC2W1pAk/edit#heading=h.7hwuy21sl20) what I tried to achieve with strict validation.

The problem is that it's not trivial to implement and it's very specific to the problem we have in hands - supporting data-lake requirements for backward compatibility.

The way I chose to implement it, by forcing additionalProperties to false, had the side-effect of breaking `allOf`. We should either choose to implement another approach, or deprecate the support for `allOf`.

lmontrieux:
@rcillo @v-stepanov are we going ahead with this solution, or are we still discussing alternatives?

rcillo:
@lmontrieux let's discuss it on our weekly tech alignment. I'll further explain the problem, the options and we can decide what to do.

rcillo:
As discussed in our alignment meeting. https://techjira.zalando.net/browse/ARUHA-507?focusedCommentId=1685455&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1685455

PR #525, opened 2016-12-30
aruha-542 forward compatibility mode

rcillo:
This compatibility mode replaces the proposed fixed mode. It allows for
compatible schema changes, accepts the entire set of json-schema
features (as previously) and doesn't change the validation
behaviour. E.g. it's fully bacward compatible with the existing version
of Nakadi.


api/nakadi-event-bus-api.yaml
```diff
@@ -1591,18 +1591,18 @@ definitions:
...
-
           - 'compatible': Consumers can reliably parse events produced under different versions. Every event published
             since the first version is still valid based on the newest schema. When in compatible mode, it's allowed to
             add new optional properties and definitions to an existing schema, but no other changes are allowed.
             Under this mode, the following json-schema attributes are not supported: `not`, `patternProperties`,
             `additionalProperties` and `additionalItems`. When validating events, additional properties is `false`.
+
+          - 'forward': Compatible schema changes are allowed. It's possible to use the full json schema specification
+          for defining schemas. Consumers of forward compatible event types can safely read events tagged with the
+          latest schema version as long as they follow the robustness principle.
```
v-stepanov: formatting is broken here (indent is too small)

api/nakadi-event-bus-api.yaml
```diff
@@ -1591,18 +1591,18 @@ definitions:
...
-          - 'none': Any schema modification is accepted, even if it might break existing producers or consumers. When
-            validating events, no additional properties are accepted unless explicitly stated in the schema.
-
           - 'compatible': Consumers can reliably parse events produced under different versions. Every event published
             since the first version is still valid based on the newest schema. When in compatible mode, it's allowed to
             add new optional properties and definitions to an existing schema, but no other changes are allowed.
             Under this mode, the following json-schema attributes are not supported: `not`, `patternProperties`,
             `additionalProperties` and `additionalItems`. When validating events, additional properties is `false`.
+
+          - 'forward': Compatible schema changes are allowed. It's possible to use the full json schema specification
```
v-stepanov: as you removed the `fixed` now - will all old event-types be migrated to this one?

v-stepanov:
I think I can't find the new code that actually defines the logic of `forward` mode: that it's only allowed to add new fields, not to change existing ones. Or am I missing something?

api/nakadi-event-bus-api.yaml
rcillo: Exactly. Also, I changed the default value to be `forward` and not `compatible`. This way we can release it in a backward compatible way.

rcillo:
@v-stepanov it's a very subtle change. Since it behaves exactly the same way as the `compatible` mode when validating schema changes, no code was necessary to implement `forward`. The only change is in `validateCompatibleChanges` method, which now also validates changes for `forward` mode and not only `compatible` mode.

v-stepanov:
:shipit:

PR #526, opened 2017-01-03
aruha-538 disable subscription creation

rcillo:
In order to migrate the database, we need to disable features that write
to the database, so that we can dump & restore consistently, without
loosing any data.

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
+++ b/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
@@ -15,6 +15,9 @@ import org.springframework.web.util.UriComponents;
 import org.zalando.nakadi.domain.Subscription;
 import org.zalando.nakadi.domain.SubscriptionBase;
 import org.zalando.nakadi.exceptions.DuplicatedSubscriptionException;
+import org.zalando.nakadi.exceptions.NakadiException;
+import org.zalando.nakadi.exceptions.NoSuchSubscriptionException;
+import org.zalando.nakadi.exceptions.ServiceUnavailableException;
 import org.zalando.nakadi.plugin.api.ApplicationService;
 import org.zalando.nakadi.problem.ValidationProblem;
 import org.zalando.nakadi.security.Client;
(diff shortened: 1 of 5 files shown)


src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -105,6 +111,14 @@ public SubscriptionService(final SubscriptionDbRepository subscriptionRepository
                 .map(Optional::get)
                 .forEach(eventType -> client.checkScopes(eventType.getReadScopes()));
 
+        if (featureToggleService.isFeatureEnabled(DISABLE_SUBSCRIPTION_CREATION)) {
+            try {
+                return Result.ok(getExistingSubscription(subscriptionBase));
```
v-stepanov: if you just return existing subscription here - the user will get a wrong response code 201, like it was really created.

rcillo: oh I see. This is ackward. If I wanted to implement the toggle in the service level I would have to "throw" an exception so that it would be caught in the controller and `processDuplicatedSubscription` would be called.

Problably I should implement this in the controller layer.

rcillo:
@v-stepanov I changed the implementation to check the toggle on controller and avoid launching exceptions to control flow.

src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -72,22 +76,36 @@ public SubscriptionController(final FeatureToggleService featureToggleService,
...
 
         try {
+            if (featureToggleService.isFeatureEnabled(DISABLE_SUBSCRIPTION_CREATION)) {
+                try {
+                    return ok(subscriptionService.getExistingSubscription(subscriptionBase));
+                } catch (final NoSuchSubscriptionException e) {
+                    return Responses.create(new ServiceUnavailableException(
+                            "Subscription creation is temporarily unavailable", e).asProblem(), request);
+                } catch (final NakadiException e) {
+                    Responses.create(e.asProblem(), request);
```
antban: return missing

rcillo: @antban fixed in 220bdc9

antban:
:+1:

PR #529, opened 2017-01-04
ARUHA-519 Locking for timelines.

antban:


src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,365 @@
...
+            this.lockedEventTypes = new HashSet<>(lockedEventTypes);
+        }
+
+        @Override
+        public String toString() {
+            return "{version=" + version + ", lockedEventTypes=" + lockedEventTypes + '}';
+        }
+    }
+
+    private static final String ROOT_PATH = "/timelines";
```
v-stepanov: why it doesn't start with "/nakadi"?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,365 @@
...
+    private static final Charset CHARSET = Charset.forName("UTF-8");
+
+    private final ZooKeeperHolder zooKeeperHolder;
+    private final String thisId;
+    private final Set<String> lockedEventTypes = new HashSet<>();
+    private final Map<String, Integer> eventsBeingPublished = new HashMap<>();
+    private final Object localLock = new Object();
+    private final HashMap<String, List<Consumer<String>>> consumerListeners = new HashMap<>();
```
v-stepanov: Can be declared as `Map` instead of `HashMap`

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,365 @@
...
+        }
+
+        @Override
+        public String toString() {
+            return "{version=" + version + ", lockedEventTypes=" + lockedEventTypes + '}';
+        }
+    }
+
+    private static final String ROOT_PATH = "/timelines";
+    private static final Charset CHARSET = Charset.forName("UTF-8");
```
v-stepanov: In other places we use what is already there in guava
`com.google.common.base.Charsets.UTF_8`

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,365 @@
...
+    private final Map<String, Integer> eventsBeingPublished = new HashMap<>();
+    private final Object localLock = new Object();
+    private final HashMap<String, List<Consumer<String>>> consumerListeners = new HashMap<>();
+    private final BlockingQueue<DelayedChange> queuedChanges = new LinkedBlockingQueue<>();
+    private final Logger log;
+
+    @Autowired
+    public TimelineSyncImpl(final ZooKeeperHolder zooKeeperHolder, final UUIDGenerator uuidGenerator) {
+        this.thisId = uuidGenerator.randomUUID().toString();
+        this.log = LoggerFactory.getLogger("timelines.sync." + thisId);
```
v-stepanov: What is the idea of `thisId` and including it in the log?
As I understand there will be only one TimelineSyncImpl per Nakadi instance.
Maybe it's better to use instance IP as id.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,365 @@
...
+        } catch (final RuntimeException ex) {
+            throw ex;
+        } catch (final Exception ex) {
+            throw new RuntimeException(ex);
+        }
+    }
+
+    @Override
+    public Closeable workWithEventType(final String eventType) throws InterruptedException {
+        synchronized (localLock) {
```
v-stepanov: This localLock object is only one per instance? Or do I miss something?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @v-stepanov good point. Forgot about this.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: Ok, will use it here

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @v-stepanov 
Actually, i needed this thing only in unit tests. Now when everything works fine I will just remove it, because it is not needed (it's a singleton object)

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: yes, only one per instance

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,365 @@
...
+                return;
+            }
+            queuedChanges.poll();
+            log.info("Delayed change {} successfully processed", change);
+        }
+    }
+
+    /**
+     * MUST be executed under zk lock. Called when version is changed
+     */
```
antban: Oh dear... Didn't removed comment

src/main/java/org/zalando/nakadi/service/timeline/TimelineSync.java
```diff
@@ -0,0 +1,63 @@
...
+ * </pre>
+ */
+public interface TimelineSync {
+    /**
+     * Call while publishing to event type. When publishing complete - call {@link Closeable#close()}
+     *
+     * @param eventType Event type to publish to
+     * @return Closeable object, that should be closed when publishing complete
+     */
+    Closeable workWithEventType(String eventType) throws InterruptedException;
```
antban: One should provide timeout for this operation, forgot about this.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
v-stepanov: Don't you afraid the performance impact? As I understand, only one publishing thread can be inside this block at a time.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,365 @@
...
+    public Closeable workWithEventType(final String eventType) throws InterruptedException {
+        synchronized (localLock) {
+            while (lockedEventTypes.contains(eventType)) {
+                localLock.wait();
+            }
+            if (!eventsBeingPublished.containsKey(eventType)) {
+                eventsBeingPublished.put(eventType, 1);
+            } else {
+                eventsBeingPublished.put(eventType, 1 + eventsBeingPublished.get(eventType));
+            }
```
v-stepanov:
```java
eventsBeingPublished.put(eventType, eventsBeingPublished.getOrDefault(eventType, 0) + 1);
```

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,365 @@
...
+        }
+        return () -> {
+            synchronized (localLock) {
+                final int currentCount = eventsBeingPublished.get(eventType);
+                if (1 == currentCount) {
+                    eventsBeingPublished.remove(eventType);
+                    localLock.notifyAll();
```
v-stepanov: I think I don't really get it. Will it release all waits? Not only for current event-type?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @v-stepanov And actually I use thisId in order to distinguish different nodes one from each other

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @v-stepanov ok

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: Yes, but the publishing in this case looks like this: 
1. lock
2. unlock
3. publish
4. lock 
5. unlock

I hope that it will be enough. I can include locking right now to publishing and we can release it to be sure that it's fast enough during very first release.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: yes, all. It's simplified locking scheme, because I do not want to build very difficult locking scheme for operation that is executed from time to time (wait starts working only in case of timeline update)

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: ok

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
v-stepanov: Does it mean that changing Timeline for several event-types at the same time is not possible?

v-stepanov: And I still don't understand how it works.
As I see it: when the count of events being currently published is 0 for ANY event type - that will allow timeline switch for ANY OTHER event type. It seems that I miss somth

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,379 @@
...
+            Thread.sleep(TimeUnit.SECONDS.toMillis(1));
+            runLocked(() -> {
+                try {
+                    final List<String> nodes = zooKeeperHolder.get().getChildren().forPath(toZkPath("/nodes"));
+                    boolean allOk = true;
+                    for (final String node : nodes) {
+                        final int nodeVersion = readData("/nodes/" + node, null, Integer::valueOf);
+                        final boolean nodeOk = nodeVersion >= versionToWait.get();
+                        if (!nodeOk) {
+                            allOk = false;
```
v-stepanov: break?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
v-stepanov: I'm not against having id, but I would like to have an easy way to match it to actual instance. I suggest at least to log id right after generation so that in the logs we will be able to find which nakadi instance has that it.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,379 @@
...
+            final InterProcessLock lock =
+                    new InterProcessSemaphoreMutex(zooKeeperHolder.get(), ROOT_PATH + "/lock");
+            lock.acquire();
+            try {
+                action.run();
+            } finally {
+                try {
+                    lock.release();
+                } catch (final Exception ex) {
+                    LOG.warn("Failed to release lock", ex);
```
lmontrieux: Shouldn't this be give a higher severity? Presumably, if we can't release the lock, it's still there, and processes waiting for it will be stuck.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,379 @@
...
+        final String zkPath = toZkPath(relativeName);
+        if (null == watcher) {
+            data = zooKeeperHolder.get().getData().forPath(zkPath);
+        } else {
+            data = zooKeeperHolder.get().getData().usingWatcher(watcher).forPath(zkPath);
+        }
+        return converter.apply(new String(data, Charsets.UTF_8));
+    }
+
+    @Scheduled(fixedDelay = 500)
```
lmontrieux: Would it make sense to move this 500 to a configuration item?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
lmontrieux: I'm also a bit confused here

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: I think not

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: ok

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,379 @@
...
+        final long finishAt = System.currentTimeMillis() + timeoutMillis;
+        synchronized (localLock) {
+            long now = System.currentTimeMillis();
+            while (now < finishAt && lockedEventTypes.contains(eventType)) {
+                localLock.wait(finishAt - now);
+                now = System.currentTimeMillis();
+            }
+            if (lockedEventTypes.contains(eventType)) {
+                throw new TimeoutException("Timed out while waiting for event type " + eventType +
+                        " to unlock within " + timeoutMillis);
```
antban: ok

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @v-stepanov @lmontrieux 
There are two collections, access to which is synchronized using 1 lock (`localLock`): `eventsBeingPublished` and `lockedEventTypes`. When any of this collections is modified i notify about that using this lock.
Lock is not used for particular event type. Instead it is used for collection of event types.
@v-stepanov It is possible to have several event type timeline updates for different event types and for the same event type at the same moment.

Probably we should have offline review session for this.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @v-stepanov Initially (before tests) it was with break. But now it is like that because of additional logging (which makes sence while switching)

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,379 @@
...
+            });
+        }
+        LOG.info("Version update to {} complete", versionToWait.get());
+    }
+
+    @Override
+    public void startTimelineUpdate(final String eventType, final long timeoutMs) throws InterruptedException {
+        LOG.info("Starting timeline update for event type {} with timeout {} ms", eventType, timeoutMs);
+        final String etZkPath = toZkPath("/locked_et/" + eventType);
+        boolean successfull = false;
```
antban: ok

src/main/java/org/zalando/nakadi/service/timeline/TimelineSync.java
```diff
@@ -0,0 +1,66 @@
...
+public interface TimelineSync {
+    /**
+     * Call while publishing to event type. When publishing complete - call {@link Closeable#close()}
+     *
+     * @param eventType Event type to publish to
+     * @param timeoutMillis Timeout for operation
+     * @return Closeable object, that should be closed when publishing complete
+     * @throws TimeoutException In case when timeout passed and event type still wasn't unlocked
+     */
+    Closeable workWithEventType(String eventType, long timeoutMillis) throws InterruptedException, TimeoutException;
```
antban: ok

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
v-stepanov: @lmontrieux then it will be not possible to use it inside the tag

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
v-stepanov: wdyt about that?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
v-stepanov: ok, let's have offline discussion, because now for me it's not clear how this magic with a single lock object works :)

src/acceptance-test/java/org/zalando/nakadi/webservice/timelines/TimelineSyncAT.java
```diff
@@ -0,0 +1,201 @@
...
+                }
+            }, "timeline_update").start();
+            // Wait a little bit for thread to start timeline update
+            Thread.sleep(TimeUnit.SECONDS.toMillis(3));
+            Assert.assertEquals(false, updated.get());
+        } finally {
+            publishing.close();
+        }
+        int attemptsLeft = 5;
+        while (attemptsLeft-- > 0) {
```
rcillo: Have you considered using existent retry strategies like Retryer.executeWithRetry?

src/acceptance-test/java/org/zalando/nakadi/webservice/utils/ZookeeperTestUtils.java
```diff
@@ -13,6 +14,7 @@ public static CuratorFramework createCurator(final String zkUrl) {
         final CuratorFramework curator = CuratorFrameworkFactory.builder()
                 .connectString(zkUrl)
                 .retryPolicy(new RetryNTimes(5, 500))
+                .sessionTimeoutMs((int)TimeUnit.SECONDS.toMillis(10))
```
rcillo: What was the previous value and why is it a good idea to change it?

antban: Previous value was 40 seconds, and I was bored to wait for so much time between test runs (because old zk sessions were still present and therefore were still holding ephemeral nodes).
On production it is 6 seconds.

antban:
@v-stepanov @lmontrieux Added commit https://github.com/zalando/nakadi/pull/529/commits/6a5f604bd1433631b94977c0c8a6d596915bfa67 for code simplification (separate local lock to separate class)

src/main/java/org/zalando/nakadi/service/timeline/TimelineSync.java
```diff
@@ -0,0 +1,66 @@
...
+     * available in db and each node knows about it.
+     *
+     * @param eventType Event type to register listener for
+     * @param listener  Listener that will accept event type.
+     * @return Registration for listener. one should call cancel once registration is not needed anymore.
+     */
+    ListenerRegistration registerTimelineChangeListener(String eventType, Consumer<String> listener);
+
+    /**
+     * Returns current node identification string (used only in tests)
```
rcillo: In some places of the project we use a special annotation for this `@VisibleForTesting`. Maybe you want to use it.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSync.java
```diff
@@ -0,0 +1,66 @@
...
+ *  + - lock                    lock for timeline versions synchronization
+ *  + - version: {version}      monotonically incremented long value (version of timelines configuration)
+ *  + - locked_et:
+ *     | - et_1                 Event types that are paused right now (et_1)
+ *     | - et_2                 The same goes here (et_2)
+ *  + - nodes                   nakadi nodes
+ *    + - {node1}: {version}    Each nakadi node exposes version being used on this node
+ * </pre>
+ */
+public interface TimelineSync {
```
rcillo: Maybe we could change names to make it clear what's local what's remotely coordinated.

`LocalLocking` -> `TimelineLocalSync`
`TimelineSync` -> `TimelineRemoteSync`

What do you think?

v-stepanov:
@antban
Man, what do you do with your branch? :)
Each time I pull I get a merge conflict hell and I have to remove my local branch and checkout again.

src/acceptance-test/java/org/zalando/nakadi/webservice/timelines/TimelineSyncAT.java
antban: @rcillo I didn't thought about that, will fix it.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSync.java
antban: @rcillo No, I don't like it, because it doesn't reflect the actual meaning. TimelineSync - is a sync that contains local and remote synchronization parts for timelines. LocalLocking - is part of timelinesync, and may be hidden at all

src/main/java/org/zalando/nakadi/service/timeline/TimelineSync.java
antban: @rcillo I'll remove it completely.

src/acceptance-test/java/org/zalando/nakadi/webservice/timelines/TimelineSyncAT.java
```diff
@@ -0,0 +1,201 @@
...
+            new Thread(() -> {
+                try {
+                    t2.startTimelineUpdate(eventType, TimeUnit.SECONDS.toMillis(30));
+                    updated.set(true);
+                } catch (final InterruptedException ex) {
+                    throw new RuntimeException(ex);
+                }
+            }, "timeline_update").start();
+            // Wait a little bit for thread to start timeline update
+            Thread.sleep(TimeUnit.SECONDS.toMillis(3));
```
v-stepanov: I'm very greedy regarding seconds. Can we wait here for something specific? (e.g. some node created in ZK)
And use a retry library to wait till condition is met.

src/acceptance-test/java/org/zalando/nakadi/webservice/timelines/TimelineSyncAT.java
```diff
@@ -0,0 +1,201 @@
...
+            Assert.assertEquals(false, updated.get());
+        } finally {
+            publishing.close();
```
v-stepanov: be a hipster - use `try` with resource management :)

src/acceptance-test/java/org/zalando/nakadi/webservice/timelines/TimelineSyncAT.java
v-stepanov: No, it is even simpler.
You can just do `TestUtils.waitFor(() -> Assert.assertEquals(false, updated.get()));`

src/acceptance-test/java/org/zalando/nakadi/webservice/timelines/TimelineSyncAT.java
```diff
@@ -0,0 +1,201 @@
...
+        new Thread(() -> {
+            try (Closeable cl = t2.workWithEventType(eventType, TimeUnit.SECONDS.toMillis(5))) {
+                lockTaken.set(true);
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }, "publisher").start();
+
+        Thread.sleep(TimeUnit.SECONDS.toMillis(2)); // Wait for thread to start.
+        Assert.assertEquals(false, lockTaken.get());
```
v-stepanov: I suggest to use `TestUtils.waitFor(...)` here

src/acceptance-test/java/org/zalando/nakadi/webservice/timelines/TimelineSyncAT.java
```diff
@@ -0,0 +1,201 @@
...
+
+        t1.startTimelineUpdate(eventType1, TimeUnit.SECONDS.toMillis(3));
+        t1.startTimelineUpdate(eventType2, TimeUnit.SECONDS.toMillis(3));
+
+        Assert.assertEquals(0, l1.getData().size());
+        Assert.assertEquals(0, l2.getData().size());
+
+        t1.finishTimelineUpdate(eventType1);
+        // Wait for listeners
+        Thread.sleep(TimeUnit.SECONDS.toMillis(1));
```
v-stepanov: I suggest to use `TestUtils.waitFor(...)` here

rcillo: I would suggest replacing `sleep` call by retries whenever possible. From our experience it leads to a more stable test suite.

src/main/java/org/zalando/nakadi/service/timeline/LocalLocking.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Set;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LocalLocking {
+    private static final Logger LOG = LoggerFactory.getLogger(LocalLocking.class);
+    private final Set<String> lockedEventTypes = new HashSet<>();
+    private final Map<String, Integer> eventsBeingPublished = new HashMap<>();
```
rcillo: I did not wanted to suggest a bigger name, but the very fist time I read it I missed event**Type** part of it. Just `event` seems inaccurate. Maybe `eventTypesPublishing` ?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,333 @@
...
+            return "{version=" + version + ", lockedEventTypes=" + lockedEventTypes + '}';
+        }
+    }
+
+    private static final String ROOT_PATH = "/nakadi/timelines";
+    private static final Logger LOG = LoggerFactory.getLogger(TimelineSyncImpl.class);
+
+    private final ZooKeeperHolder zooKeeperHolder;
+    private final String thisId;
+    private final LocalLocking localLocking = new LocalLocking();
```
rcillo: Why isn't it inject in the constructor as well?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,333 @@
...
+        public String toString() {
+            return "{version=" + version + ", lockedEventTypes=" + lockedEventTypes + '}';
+        }
+    }
+
+    private static final String ROOT_PATH = "/nakadi/timelines";
+    private static final Logger LOG = LoggerFactory.getLogger(TimelineSyncImpl.class);
+
+    private final ZooKeeperHolder zooKeeperHolder;
+    private final String thisId;
```
adyach: `nodeId` is suggested to use

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,333 @@
...
+        } else {
+            data = zooKeeperHolder.get().getData().usingWatcher(watcher).forPath(zkPath);
+        }
+        return converter.apply(new String(data, Charsets.UTF_8));
+    }
+
+    @Scheduled(fixedDelay = 500)
+    public void reactOnEventTypesChange() throws InterruptedException {
+        checkForNewChange();
+        while (null != queuedChanges.peek()) {
```
adyach: You do peek twice (next line), I'd either move to a variable or use isEmpty

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,333 @@
...
+            final DelayedChange change = queuedChanges.peek();
+            LOG.info("Reacting on delayed change {}", change);
+            final Set<String> unlockedEventTypes = localLocking.lockedEventTypesChanged(change.lockedEventTypes);
+            // Notify consumers that they should refresh timeline information
+            for (final String unlocked : unlockedEventTypes) {
+                LOG.info("Notifying about unlock of {}", unlocked);
+                final List<Consumer<String>> toNotify;
+                synchronized (consumerListeners) {
+                    toNotify = consumerListeners.containsKey(unlocked) ?
```
adyach: Do you have reasons for not using ConcurrentHashMap for consumerListeners? It will simplify this code and avoid creation of additional list.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
rcillo: I could not find in the existing code what happens when this condition is met. The loop is over and I could not find where it subscribes for new changes.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,333 @@
...
+        LOG.info("Version update to {} complete", versionToWait.get());
+    }
+
+    @Override
+    public void startTimelineUpdate(final String eventType, final long timeoutMs) throws InterruptedException {
+        LOG.info("Starting timeline update for event type {} with timeout {} ms", eventType, timeoutMs);
+        final String etZkPath = toZkPath("/locked_et/" + eventType);
+        boolean successful = false;
+        try {
+            zooKeeperHolder.get().create().withMode(CreateMode.EPHEMERAL)
```
rcillo: Do you want to catch and rethrow already existent node exception, in order to change the status code from 500 to 409 or something else?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -0,0 +1,333 @@
...
+                } finally {
+                    if (!success) {
+                        newVersionPresent.set(true);
+                    }
+                }
+            });
+        }
+    }
+
+    private void versionChanged(final WatchedEvent ignore) {
```
adyach: What about different statuses of zookeeper?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
rcillo: Forget about it... it's scheduled up there. It got lost in the middle of comments.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
rcillo: Why not set a watcher on version and react when it changes?

src/main/java/org/zalando/nakadi/service/timeline/LocalLocking.java
antban: ok

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: ok

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: Why it is better?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @rcillo Because changing version in zk may fail, and if we won't retry, than timeline update will hang up for 30 seconds.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @adyach ok

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
rcillo: If you wanted to make some unit tests, it could help mocking.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @adyach Because of the way it is managed while adding listener. If I will use any other approach it will lead to possiblity of incorrect map update.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @adyach Well, yes. It is very good point. 
But here it doesn't matter what kind of change there is, because data will be re read and notification will be set up again.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @rcillo I had in mind that timeline update is an administrative task, and therefore doesn't have to be highly documented. The only thing is very important - exception should be correctly thrown or logged. I didn't wanted to introduce a lot of different exception types that will be mapped later.
But if the team insist on clear approach - of course I'll modify the code.

rcillo: Well, since it's administrative, I would say that good logging is enough.

src/acceptance-test/java/org/zalando/nakadi/webservice/timelines/TimelineSyncAT.java
antban: At this position it is not possible. I check, that after 2 seconds publishing is still not started. I don't have to wait for anything else.

src/acceptance-test/java/org/zalando/nakadi/webservice/timelines/TimelineSyncAT.java
antban: The way test works is completely different.
Here I check that after 2 seconds from thread start timeline update still not started.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
antban: @rcillo Anyway, I added additional exception to treat this information separately

rcillo: 👍

adyach:
Looks good for me 👍

v-stepanov:
👍

PR #531, opened 2017-01-06
ARUHA-522 Prepare database for timelines

antban:
Covered things: 
 - Migration scripts
 - CRD Storage
 - CRUD Timeline

### DIFF
--- a/database/migration/aruha-522-timelines.sql
+++ b/database/migration/aruha-522-timelines.sql
@@ -1,0 +1,21 @@
+CREATE TABLE zn_data.storage (
+  st_id            VARCHAR(36) NOT NULL PRIMARY KEY,
+  st_type          VARCHAR(32) NOT NULL,
+  st_configuration JSONB       NOT NULL
+);
+
+
+CREATE TABLE zn_data.timeline (
+  t_id              UUID                     NOT NULL PRIMARY KEY,
+  et_name           VARCHAR(255)             NOT NULL REFERENCES zn_data.event_type (et_name),
+  t_order           INT                      NOT NULL,
+  st_id             VARCHAR(36)              NOT NULL REFERENCES zn_data.storage (st_id),
+  t_topic           VARCHAR(255)             NOT NULL,
+  t_created_at      TIMESTAMP WITH TIME ZONE NOT NULL,
+  t_switched_at     TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+  t_cleanup_at      TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+  t_latest_position JSONB                    DEFAULT NULL,
+  UNIQUE (et_name, t_order)
+);
+
+CREATE INDEX ON zn_data.timeline(et_name);
--- a/database/migration/aruha-522-timelines.sql
+++ b/database/migration/aruha-522-timelines.sql
@@ -1,3 +1,5 @@
+SET ROLE zalando_nakadi_data_owner;
+
 CREATE TABLE zn_data.storage (
   st_id            VARCHAR(36) NOT NULL PRIMARY KEY,
   st_type          VARCHAR(32) NOT NULL,
@@ -6,16 +8,16 @@ CREATE TABLE zn_data.storage (
 
 
 CREATE TABLE zn_data.timeline (
-  t_id              UUID                     NOT NULL PRIMARY KEY,
-  et_name           VARCHAR(255)             NOT NULL REFERENCES zn_data.event_type (et_name),
-  t_order           INT                      NOT NULL,
-  st_id             VARCHAR(36)              NOT NULL REFERENCES zn_data.storage (st_id),
-  t_topic           VARCHAR(255)             NOT NULL,
-  t_created_at      TIMESTAMP WITH TIME ZONE NOT NULL,
-  t_switched_at     TIMESTAMP WITH TIME ZONE DEFAULT NULL,
-  t_cleanup_at      TIMESTAMP WITH TIME ZONE DEFAULT NULL,
-  t_latest_position JSONB                    DEFAULT NULL,
-  UNIQUE (et_name, t_order)
+  tl_id              UUID                     NOT NULL PRIMARY KEY,
+  et_name            VARCHAR(255)             NOT NULL REFERENCES zn_data.event_type (et_name),
+  tl_order           INT                      NOT NULL,
+  st_id              VARCHAR(36)              NOT NULL REFERENCES zn_data.storage (st_id),
+  tl_topic           VARCHAR(255)             NOT NULL,
+  tl_created_at      TIMESTAMP WITH TIME ZONE NOT NULL,
+  tl_switched_at     TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+  tl_cleanup_at      TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+  tl_latest_position JSONB                    DEFAULT NULL,
+  UNIQUE (et_name, tl_order)
 );
 
-CREATE INDEX ON zn_data.timeline(et_name);
+CREATE INDEX ON zn_data.timeline (et_name);
--- a/database/nakadi/10_data/04_tables/05_timeline.sql
+++ b/database/nakadi/10_data/04_tables/05_timeline.sql
@@ -1,14 +1,14 @@
 CREATE TABLE zn_data.timeline (
-  t_id              UUID                     NOT NULL PRIMARY KEY,
-  et_name           VARCHAR(255)             NOT NULL REFERENCES zn_data.event_type (et_name),
-  t_order           INT                      NOT NULL,
-  st_id             VARCHAR(36)              NOT NULL REFERENCES zn_data.storage (st_id),
-  t_topic           VARCHAR(255)             NOT NULL,
-  t_created_at      TIMESTAMP WITH TIME ZONE NOT NULL,
-  t_switched_at     TIMESTAMP WITH TIME ZONE DEFAULT NULL,
-  t_cleanup_at      TIMESTAMP WITH TIME ZONE DEFAULT NULL,
-  t_latest_position JSONB                    DEFAULT NULL,
-  UNIQUE (et_name, t_order)
+  tl_id              UUID                     NOT NULL PRIMARY KEY,
+  et_name            VARCHAR(255)             NOT NULL REFERENCES zn_data.event_type (et_name),
+  tl_order           INT                      NOT NULL,
+  st_id              VARCHAR(36)              NOT NULL REFERENCES zn_data.storage (st_id),
+  tl_topic           VARCHAR(255)             NOT NULL,
+  tl_created_at      TIMESTAMP WITH TIME ZONE NOT NULL,
+  tl_switched_at     TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+  tl_cleanup_at      TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+  tl_latest_position JSONB                    DEFAULT NULL,
+  UNIQUE (et_name, tl_order)
 );
 
-CREATE INDEX ON zn_data.timeline(et_name);
+CREATE INDEX ON zn_data.timeline (et_name);
--- a/database/nakadi/10_data/04_tables/05_timeline.sql
+++ b/database/nakadi/10_data/04_tables/05_timeline.sql
@@ -1,10 +1,3 @@
-CREATE TABLE zn_data.storage (
-  st_id            VARCHAR(36) NOT NULL PRIMARY KEY,
-  st_type          VARCHAR(32) NOT NULL,
-  st_configuration JSONB       NOT NULL
-);
-
-
 CREATE TABLE zn_data.timeline (
   t_id              UUID                     NOT NULL PRIMARY KEY,
   et_name           VARCHAR(255)             NOT NULL REFERENCES zn_data.event_type (et_name),
@@ -17,3 +10,5 @@ CREATE TABLE zn_data.timeline (
   t_latest_position JSONB                    DEFAULT NULL,
   UNIQUE (et_name, t_order)
 );
+
+CREATE INDEX ON zn_data.timeline(et_name);
--- a/src/acceptance-test/java/org/zalando/nakadi/repository/db/StorageDbRepositoryTest.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/repository/db/StorageDbRepositoryTest.java
@@ -1,0 +1,68 @@
+package org.zalando.nakadi.repository.db;
+
+import java.util.List;
+import java.util.Optional;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.zalando.nakadi.domain.Storage;
+
+public class StorageDbRepositoryTest extends AbstractDbRepositoryTest {
+    private StorageDbRepository repository;
+
+    public StorageDbRepositoryTest() {
+        super("zn_data.storage");
+    }
+
+    @Before
+    public void setUp() {
+        super.setUp();
+        repository = new StorageDbRepository(template, mapper);
+    }
+
+    static Storage createStorage(final String name, final String zkAddress, final String zkPath) {
+        final Storage.KafkaConfiguration config = new Storage.KafkaConfiguration();
+        config.setZkAddress(zkAddress);
+        config.setZkPath(zkPath);
+        final Storage storage = new Storage();
+        storage.setId(name);
+        storage.setType(Storage.Type.KAFKA);
+        storage.setConfiguration(config);
+        return storage;
+    }
+
+    @Test
+    public void testStorageCreated() {
+        final Storage storage = createStorage("default", "address", "path");
+
+        repository.createStorage(storage);
+
+        final Optional<Storage> createdCopy = repository.getStorage(storage.getId());
+        Assert.assertTrue(createdCopy.isPresent());
+        Assert.assertFalse(createdCopy.get() == storage);
+
+        Assert.assertEquals(storage, createdCopy.get());
+    }
+
+    @Test
+    public void testStorageOrdered() {
+        final Storage storage2 = repository.createStorage(createStorage("2", "address1", "path3"));
+        final Storage storage1 = repository.createStorage(createStorage("1", "address2", "path2"));
+        final Storage storage3 = repository.createStorage(createStorage("3", "address3", "path1"));
+
+        final List<Storage> storages = repository.listStorages();
+        Assert.assertEquals(3, storages.size());
+        Assert.assertEquals(storage1, storages.get(0));
+        Assert.assertEquals(storage2, storages.get(1));
+        Assert.assertEquals(storage3, storages.get(2));
+    }
+
+    @Test
+    public void testStorageDeleted() {
+        final Storage storage = repository.createStorage(createStorage("1", "address2", "path2"));
+        Assert.assertEquals(storage, repository.getStorage(storage.getId()).get());
+        repository.deleteStorage(storage.getId());
+        Assert.assertFalse(repository.getStorage(storage.getId()).isPresent());
+        Assert.assertEquals(0, repository.listStorages().size());
+    }
+}
--- a/src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
@@ -133,13 +133,8 @@ public class TimelineDbRepositoryTest extends AbstractDbRepositoryTest {
             final Date switchedAt,
             final Date cleanupAt,
             final Timeline.StoragePosition latestPosition) {
-        final Timeline timeline = new Timeline();
+        final Timeline timeline = new Timeline(eventType, order, storage, topic, createdAt);
         timeline.setId(id);
-        timeline.setStorage(storage);
-        timeline.setOrder(order);
-        timeline.setTopic(topic);
-        timeline.setEventType(eventType);
-        timeline.setCreatedAt(createdAt);
         timeline.setSwitchedAt(switchedAt);
         timeline.setCleanupAt(cleanupAt);
         timeline.setLatestPosition(latestPosition);
--- a/src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
@@ -1,0 +1,148 @@
+package org.zalando.nakadi.repository.db;
+
+import java.util.Date;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.LongStream;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.springframework.dao.DuplicateKeyException;
+import org.zalando.nakadi.domain.EventType;
+import org.zalando.nakadi.domain.Storage;
+import org.zalando.nakadi.domain.Timeline;
+import org.zalando.nakadi.exceptions.DuplicatedEventTypeNameException;
+import org.zalando.nakadi.exceptions.InternalNakadiException;
+import org.zalando.nakadi.utils.TestUtils;
+
+public class TimelineDbRepositoryTest extends AbstractDbRepositoryTest {
+
+    private TimelineDbRepository tRepository;
+    private StorageDbRepository sRepository;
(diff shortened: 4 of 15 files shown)


database/nakadi/10_data/04_tables/05_timeline.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE zn_data.timeline (
+  t_id              UUID                     NOT NULL PRIMARY KEY,
```
v-stepanov: why timeline id is UUID and storage id is varchar?

database/migration/aruha-522-timelines.sql
```diff
@@ -0,0 +1,21 @@
+CREATE TABLE zn_data.storage (
```
rcillo: It's important to switch roles before running this migration. In order not to forget about it, I suggest to add this line first:

`SET ROLE zalando_nakadi_data_owner;`

Otherwise, tables are created with wrong permissions and Nakadi is unable to access it.

database/migration/aruha-522-timelines.sql
```diff
@@ -0,0 +1,21 @@
...
+  st_id            VARCHAR(36) NOT NULL PRIMARY KEY,
+  st_type          VARCHAR(32) NOT NULL,
+  st_configuration JSONB       NOT NULL
+);
+
+
+CREATE TABLE zn_data.timeline (
+  t_id              UUID                     NOT NULL PRIMARY KEY,
```
rcillo: Why not `ti_id`?

src/acceptance-test/java/org/zalando/nakadi/repository/db/StorageDbRepositoryTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.zalando.nakadi.domain.Storage;
+
+public class StorageDbRepositoryTest extends AbstractDbRepositoryTest {
+    private StorageDbRepository repository;
```
v-stepanov: I think you forgot to commit it, I don't see this class in PR.

src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
```diff
@@ -0,0 +1,148 @@
...
+import org.zalando.nakadi.domain.EventType;
+import org.zalando.nakadi.domain.Storage;
+import org.zalando.nakadi.domain.Timeline;
+import org.zalando.nakadi.exceptions.DuplicatedEventTypeNameException;
+import org.zalando.nakadi.exceptions.InternalNakadiException;
+import org.zalando.nakadi.utils.TestUtils;
+
+public class TimelineDbRepositoryTest extends AbstractDbRepositoryTest {
+
+    private TimelineDbRepository tRepository;
```
v-stepanov: This class is also missing.

database/migration/aruha-522-timelines.sql
```diff
@@ -0,0 +1,21 @@
...
+  st_id             VARCHAR(36)              NOT NULL REFERENCES zn_data.storage (st_id),
+  t_topic           VARCHAR(255)             NOT NULL,
+  t_created_at      TIMESTAMP WITH TIME ZONE NOT NULL,
+  t_switched_at     TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+  t_cleanup_at      TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+  t_latest_position JSONB                    DEFAULT NULL,
+  UNIQUE (et_name, t_order)
+);
+
+CREATE INDEX ON zn_data.timeline(et_name);
```
rcillo: Maybe we also want to have a few more indexes on fields used by queries. Good candidates would be `t_order` and `t_created_at`.

src/acceptance-test/java/org/zalando/nakadi/repository/db/StorageDbRepositoryTest.java
v-stepanov: oops, sorry. It is there.

src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
v-stepanov: sorry, it is there actually.

src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
```diff
@@ -0,0 +1,148 @@
...
+        final EventType testEt = eRepository.saveEventType(TestUtils.buildDefaultEventType());
+        final Timeline t1 = tRepository.createTimeline(createTimeline(
+                storage, UUID.randomUUID(), 1, "test_topic", testEt.getName(),
+                new Date(), null, null, null));
+        Assert.assertEquals(1, tRepository.listTimelines(testEt.getName()).size());
+        tRepository.deleteTimeline(t1.getId());
+        Assert.assertEquals(0, tRepository.listTimelines(testEt.getName()).size());
+    }
+
+    private static Timeline createTimeline(
```
rcillo: What do you think about the builder pattern used for event types?

src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
```diff
@@ -0,0 +1,74 @@
...
+
+@DB
+@Repository
+public class StorageDbRepository extends AbstractDbRepository {
+    @Autowired
+    public StorageDbRepository(final JdbcTemplate template, final ObjectMapper mapper) {
+        super(template, mapper);
+    }
+
+    public List<Storage> listStorages() {
```
rcillo: Do you plan to add some pagination support? (offset + limit)

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
```diff
@@ -0,0 +1,129 @@
...
+    public List<Timeline> listTimelines(final String eventType) {
+        return jdbcTemplate.query(
+                BASE_TIMELINE_QUERY + " WHERE t.et_name=? order by t.t_order",
+                new Object[]{eventType},
+                timelineRowMapper);
+    }
+
+    public Optional<Timeline> getTimeline(final UUID id) {
+        final List<Timeline> timelines = jdbcTemplate.query(
+                BASE_TIMELINE_QUERY + " WHERE t.t_id=?",
```
rcillo: Maybe add `LIMIT 1`? But maybe it's not necessary since it's the primary key.

database/migration/aruha-522-timelines.sql
antban: Ok

database/migration/aruha-522-timelines.sql
antban: you meant use `tl_` (`ti_` doesn't say anything about timeline word)?

database/migration/aruha-522-timelines.sql
antban: Well, there are several things why I do not create indexes now. 
1. They are mostly not used, and sequential scan performs fast enough to deal with it. 
2. We will use timelines almost always from cache. That means that the only operation that probably should be optimized - getting all timelines for event type. 
3. I can not guarantee on this stage what kinds of queries should be optimized, therefore I made only index, that is used now (but again, for 1000 records we can go without indexes at all)

src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
antban: I thought about this, but decided not to do that. In fact, if we will have more than 10 active kafka clusters, I will be really surprised. If the cluster is not used, then we have to delete it to avoid mistakes while timelines creation

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
antban: @rcillo Yes, that is why i didn't included this limit. It is useless.

database/nakadi/10_data/04_tables/05_timeline.sql
antban: @v-stepanov Timeline id is machine generated, and it is better to have something unique while searching thru logs. 
Storage id - is something, that we will use as humans to perform switching, that is why it should be created by human and readable by human(values like default, kafka_10_test or whatever, but human readable)

database/migration/aruha-522-timelines.sql
rcillo: `tl` is good 👍 and it's consistent with the 2 chars rule (the exception `ets` is due to conflict).

rcillo:
@antban for me it looks good. Maybe you want to wait for @v-stepanov approval as well.

v-stepanov:
@antban 
Man, what do you do with your branch? :) 
Each time I pull I get a merge conflict hell and I have to remove my local branch and checkout again.

src/main/java/org/zalando/nakadi/domain/Timeline.java
```diff
@@ -0,0 +1,153 @@
+package org.zalando.nakadi.domain;
+
+import java.util.Date;
+import java.util.List;
+import java.util.Objects;
+import java.util.UUID;
+
+public class Timeline {
```
v-stepanov: I don't understand your nullability policy here. When I look at this class it seems that everything can be null as you check everything for null. But then in repository you check for null only the latest position.

src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.dao.DataAccessException;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.core.RowMapper;
+import org.springframework.stereotype.Repository;
+import org.zalando.nakadi.annotations.DB;
+import org.zalando.nakadi.domain.Storage;
+
+@DB
+@Repository
+public class StorageDbRepository extends AbstractDbRepository {
```
v-stepanov: I'm not sure about you policy regarding uppercase for SQL reserved words. It seems that you follow it but not in all places.

v-stepanov: the same applies to other storage as well.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
```diff
@@ -0,0 +1,129 @@
...
+@Repository
+public class TimelineDbRepository extends AbstractDbRepository {
+
+    public static final String BASE_TIMELINE_QUERY = "SELECT " +
+            "   t_id, et_name, t_order, t.st_id, t_topic, " +
+            "   t_created_at, t_switched_at, t_cleanup_at, t_latest_position, " +
+            "   st_type, st_configuration " +
+            " FROM " +
+            "   zn_data.timeline t " +
+            "   JOIN zn_data.storage st USING (st_id) ";
```
v-stepanov: Oh, Join! 
We are like adults now :)

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
```diff
@@ -0,0 +1,129 @@
...
+    @Autowired
+    public TimelineDbRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper jsonMapper) {
+        super(jdbcTemplate, jsonMapper);
+    }
+
+    public List<Timeline> listTimelines(final String eventType) {
+        return jdbcTemplate.query(
+                BASE_TIMELINE_QUERY + " WHERE t.et_name=? order by t.t_order",
+                new Object[]{eventType},
+                timelineRowMapper);
```
v-stepanov: a very small impovement, but you can use the method notation when the mapper is on second place and params vararg is on 3rd place. That will allow to pass eventType without creation of array.

src/main/java/org/zalando/nakadi/domain/Timeline.java
antban: OMG, i *forgot* about Nullable  policy! Shame on me.. Will fix it.

src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
antban: I didn't thought that we have this policy. will fix it.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
antban: ok

src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
antban: I don't think that using builder in this case is necessary. I will create static method for timeline creation as a part of `@Nullable` policy.

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
```diff
@@ -0,0 +1,128 @@
...
+                    timeline.getEventType(),
+                    timeline.getOrder(),
+                    timeline.getStorage().getId(),
+                    timeline.getTopic(),
+                    timeline.getCreatedAt(),
+                    timeline.getSwitchedAt(),
+                    timeline.getCleanupAt(),
+                    timeline.getLatestPosition() == null ? null :
+                            jsonMapper.writer().writeValueAsString(timeline.getLatestPosition()),
+                    timeline.getId());
```
v-stepanov: What will happen if id is null?

src/main/java/org/zalando/nakadi/domain/Timeline.java
```diff
@@ -0,0 +1,175 @@
...
+    private String eventType;
+    private int order;
+    private Storage storage;
+    private String topic;
+    private Date createdAt;
+    @Nullable
+    private Date switchedAt;
+    @Nullable
+    private Date cleanupAt;
+    @Nullable
```
v-stepanov: there's no need to mark private properties/methods as nullable

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    public Timeline createTimeline(final Timeline timeline) {
+        try {
+            jdbcTemplate.update(
+                    "INSERT INTO zn_data.timeline(" +
+                            " tl_id, et_name, tl_order, st_id, tl_topic, " +
+                            " tl_created_at, tl_switched_at, tl_cleanup_at, tl_latest_position) " +
+                            " VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?::jsonb)",
+                    timeline.getId(),
```
v-stepanov: What will happen if id is null?

src/main/java/org/zalando/nakadi/domain/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    public <T> T getConfiguration(final Class<T> clazz) {
+        if (configuration == null) {
+            return null;
+        }
+        if (!clazz.isAssignableFrom(configuration.getClass())) {
+            throw new IllegalStateException("Can not cast configuration " + configuration + " to class " + clazz);
+        }
+        return (T) configuration;
```
v-stepanov: It's a very strange use of generics. Why `configuration` property is not defined as T ?

src/main/java/org/zalando/nakadi/domain/Storage.java
```diff
@@ -0,0 +1,135 @@
...
+        public boolean equals(final Object o) {
+            if (this == o) {
+                return true;
+            }
+            if (!(o instanceof KafkaConfiguration)) {
+                return false;
+            }
+
+            final KafkaConfiguration that = (KafkaConfiguration) o;
+            return Objects.equals(zkAddress, that.zkAddress) && Objects.equals(zkPath, that.zkPath);
```
v-stepanov: Why do you use `Objects.equals()` for properties that are not nullable?

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
```diff
@@ -0,0 +1,128 @@
...
+                    timeline.getId());
+        } catch (final JsonProcessingException ex) {
+            throw new IllegalArgumentException("Can not serialize latest position to json", ex);
+        }
+    }
+
+    public void deleteTimeline(final UUID id) {
+        jdbcTemplate.update(
+                "DELETE FROM zn_data.timeline WHERE tl_id=?",
+                new Object[]{id});
```
v-stepanov: there's a vararg version of method, so you can skip array creation

src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
```diff
@@ -0,0 +1,143 @@
...
+import org.junit.Test;
+import org.springframework.dao.DuplicateKeyException;
+import org.zalando.nakadi.domain.EventType;
+import org.zalando.nakadi.domain.Storage;
+import org.zalando.nakadi.domain.Timeline;
+import org.zalando.nakadi.exceptions.DuplicatedEventTypeNameException;
+import org.zalando.nakadi.exceptions.InternalNakadiException;
+import org.zalando.nakadi.utils.TestUtils;
+
+public class TimelineDbRepositoryTest extends AbstractDbRepositoryTest {
```
v-stepanov: I'm always against this kind of tests. You test Blah.saveX() using Blah.loadX() and vise versa.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
antban: @v-stepanov Exception will be thrown. I use `@Nullable` here to support fake timelines, which do not have timeline id (will be introduced later). And of course it is incorrect to use this kind of timeline for database updates.

antban: @v-stepanov At least i hope that it will be like that.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
antban: @v-stepanov Here exception will be thrown, definitely.

src/main/java/org/zalando/nakadi/domain/Storage.java
antban: Because of unification

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
antban: missed this line.. will fix it.

src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
antban: I totally fine with this test, because it helped me to fix errors with sql in this home-made ORM (I renamed fields, and tests showed me 2 places where sql was not correct). 
So they already saved time for me.

src/main/java/org/zalando/nakadi/domain/Storage.java
antban: And I do not want to get NPE in equals or toString method because someone added annotation `@Nullable` or build object incorrectly. For me these methods should work correctly at any point of time.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
v-stepanov: But that ruins the purpose of Nullable annotation.
You suggest the following approach: in places A and B the property can be null, in places C and D the property cannot be null. This is a perfect scenario to have a lot of problems in future when the code will be written by someone who is not aware of these unannounced rules.

src/main/java/org/zalando/nakadi/domain/Storage.java
v-stepanov: That's why I'm a bit unhappy with the way you handle nulls. With such approach we will come to a point that we will check everything for null "just in case".

src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
v-stepanov: I'm not saying that it doesn't allow to spot problems. I'm saying that it involves a lot of Nakadi code for testing other Nakadi code. I'm not a fan of such approach.

v-stepanov:
@antban I wrote a lot of comments, but in general everything is fine for me :) I only would like to know regarding the generics question.

src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
antban: @rcillo Actually I created constructor

src/main/java/org/zalando/nakadi/domain/Storage.java
antban: @v-stepanov 
The `@Nullable` annotation doesn't guarantee anything. And I had several situations where method toString() was throwing NPE while printing debug information because of this approach. As a result - no one was able to see what actually happend. 
Using `@Nullable` in business code is probably a good thing, but I insist to think that in `equals`, `toString`, `hashCode` we must be prepared to nulls.

src/main/java/org/zalando/nakadi/domain/Storage.java
antban: Lets discuss it offline.

src/main/java/org/zalando/nakadi/domain/Timeline.java
antban: ok

v-stepanov:
:shipit:

database/migration/aruha-522-timelines.sql
```diff
@@ -0,0 +1,23 @@
+SET ROLE zalando_nakadi_data_owner;
+
+CREATE TABLE zn_data.storage (
+  st_id            VARCHAR(36) NOT NULL PRIMARY KEY,
+  st_type          VARCHAR(32) NOT NULL,
+  st_configuration JSONB       NOT NULL
+);
+
+
```
bocytko: You slightly alter the already used good old Zalando DB naming conventions.

Fields for `zn_data.storage` should have prefix `s_` (i.e. `s_type`). Fields for `zn_data.event_type` have prefix `et_`.

Following the same convention in `zn_data.timeline`, you have:
- `t_id`,
- `t_event_type_name`,
- `t_storage_id`,
- `t_topic`.

rcillo: @bocytko we didn't work with legacy databases at Zalando so this rule is not clear for us. Since in our codebase all tables had 2 letters, I thought the rule was to have 2 letters. So I asked to change from `t_id` to `tl_id` 😓

antban: @bocytko If I will rename columns in a way you propose I won't be able to use handy keyword `USING`. I know, that for old zalando db it is important to use the approach you say, but is it mandatory to use exactly the same approach among all microservices?

bocytko: Yes, you will never have matching column names across tables. Since the old rules are not forced anymore, you can adopt own ones as a team.

I just noted some inconsistency in your naming schema and wanted to make sure you know about the former conventions.

valgog: I tried to convince Dmitry that we had a reason to use the conventions we used and that it saved us a lot of trouble in the time of the years. But some advantages of using USING instead of joining on names probably are there. I personally do not see them, but what should I say.

I guess one learns on ones own mistakes and all the experiences are different.

antban: @bocytko @valgog 
- Ticket passed thru review of all the team members, so I believe that team adopted. 
- We do not use any automation now that relies on the rule you are saying about.
So I think that conversation should be closed, because it stops us from releasing this ticket.

valgog: I was absolutely sure that this review was closed long ago. Just want to express my point of view. 

Please try to close any reviews in shorter then a day or two. I suggest to do the reviews in person in a separate room even.

adyach:
:shipit:

PR #532, opened 2017-01-09
aruha-556 allow required extensions

rcillo:
This commit allows the extension of required properties when using forward
mode and also when migrating to compatible mode.

Since this only impatates producers, it's preserves the forward property
of not breaking consumers. For the consumer, it doesn't matter if a new
property is now required.

It's useful for both producers and consumers since it avoids the
necessity to do defensive programming when reading events.


v-stepanov:
For me it looks good

v-stepanov:
:shipit:

PR #533, opened 2017-01-11
aruha-485: added shutdown hook to free subscription resources

adyach:
This PR brings ability to free subscription resources when JVM is terminated (e.g. aws terminated the instance)


src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
```diff
@@ -92,9 +92,20 @@ public long getKafkaPollTimeout() {
 
     @Override
     public void stream() throws InterruptedException {
+        // bugfix ARUHA-485
+        Runtime.getRuntime().addShutdownHook(new Thread(() -> onShutdown()));
```
rcillo: I think we could add a unit test here by wrapping runtime in an object and inject it here. Then we could mock it and test that when we execute addShutdownHook it changes state in the expected manner.

rcillo: Actually, the target of the test is the function `onShutdown`. I would like to test it to be sure that if something changes in the future, it has a test to verify it.

adyach: @rcillo I made a little bit different fix eae593e. could you check this out?

rcillo:
Do you think there is no need to add any tests for this?

adyach:
@rcillo yes, here it is b669194
This test is not to check the logic of the clean up state but to fix it for future changes.

rcillo:
:shipit:

antban:
:+1:

PR #536, opened 2017-01-17
aruha-576 hotfix feature toggle

rcillo:
This feature toggle has been removed in
fc99630023d4b04e866348624e726aa299fa1d32 after being left abandoned for
several months. It ends up causing undesirable side effects and for that
reason I'm restoring it as a hotfix.

The definitive solution is yet to be presented. This is just hotfix.


lmontrieux:
:shipit:

PR #537, opened 2017-01-18
ARUHA-575 Change read and write parallelism values in readme

lmontrieux:
Readme example creates only 1 partition.


v-stepanov:
👍

PR #538, opened 2017-01-18
aruha-526: Include Timelines into cache

adyach:
This commit will bring possibility to request timelines using event typecache. Reaction on timelines changes is made using TimelineSync.ListenerRegistration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.opensource.zalan.do/stups/openjdk:8-42
+FROM registry.opensource.zalan.do/stups/openjdk:8-54
 
 MAINTAINER Team Aruha, team-aruha@zalando.de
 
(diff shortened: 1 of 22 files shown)


api/nakadi-event-bus-api.yaml
```diff
@@ -807,9 +807,13 @@ paths:
...
         - oauth2: ['nakadi.event_stream.read']
       description: |
-        Endpoint for committing offsets of the subscription. The client must commit at least once
-        every 60 seconds, otherwise Nakadi will consider the client to be gone and will close the
-        connection.
+        Endpoint for committing offsets of the subscription. If there is uncommited data, and no commits happen
+        for 60 seconds, then Nakadi will consider the client to be gone, and will close the connection. As long
+        as no events are sent, the client does not need to commit.
+
+        If the connection is closed, the client has 60 seconds to commit the events it received, from the moment
```
rcillo: I don't know if it's necessary, but this is true if the connection "is closed by the user" only. If the connection is closed by the server, it becomes immediately impossible to commit to that stream-id, as tested in https://techjira.zalando.net/browse/ARUHA-485

adyach: @rcillo sorry, I closed this pr due to a lot of unrelated changes. I will create a new one.

PR #540, opened 2017-01-18
aruha-526: Include Timelines into cache

adyach:


src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -20,75 +13,111 @@
...
+        this.timelineSync = timelineSync;
+        this.timelineRegistrations = new ConcurrentHashMap<>();
         if (null != cacheSync) {
             this.cacheSync.getListenable().addListener((curator, event) -> this.onZkEvent(event));
         }
-        preloadEventTypes(eventTypeRepository);
+        preloadEventTypes(eventTypeRepository, timelineRepository);
+    }
+
+    @PreDestroy
```
v-stepanov: I thought `EventTypeCache` is never destroyed?

adyach: @v-stepanov it is called when application context is going to be destroyed.
http://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-spring-application.html#boot-features-application-exit

v-stepanov: But is this a real case for us? Why do we need it?

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
```diff
@@ -31,6 +37,10 @@ public TimelineDbRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper
         super(jdbcTemplate, jsonMapper);
     }
 
+    public List<Timeline> list() {
```
v-stepanov: I think it's very inconsistent that we have methods named `list` and `listTimelines` here.

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
adyach: If you acquire resource you have to clean it up when you will finish using it. i do not think there is much sense in doing it here. I will remove it if it is mess for you.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
adyach: What is your proposal? Name both listTimelines?

v-stepanov: Yes, i think it's a good solution.

adyach: b385578

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
adyach: b385578

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -146,23 +172,19 @@ public EventTypeValidator getValidator(final String name) throws InternalNakadiE
...
-        }
-
-        final PathChildrenCache cacheSync = new PathChildrenCache(zkClient, ZKNODE_PATH, false);
-
-        // It is important to preload all data before specifying callback for updates, because otherwise preload won't
-        // give any effect - all changes will be removed.
-        cacheSync.start(PathChildrenCache.StartMode.BUILD_INITIAL_CACHE);
+    public Optional<Timeline> getActiveTimeline(final String name) throws InternalNakadiException,
+            NoSuchEventTypeException {
+        return getCached(name).map(CachedValue::getTimelines)
```
rcillo: Wouldn't it be better to store the active timeline directly so not to map+filter every time this method is called?

rcillo:
@adyach as we discussed offline, all cases are covered here. It seems good 👍

rcillo:
:shipit:

src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeCacheTestAT.java
```diff
@@ -129,13 +145,106 @@ public void invalidateCacheOnUpdate() throws Exception {
...
+
+    private List<Timeline> getMockedTimelines(final String etName) {
+        final Timeline t1 = new Timeline(etName, 0, new Storage(), "topic", new Date());
+        final Timeline t2 = new Timeline(etName, 1, new Storage(), "topic", new Date(System.currentTimeMillis() + 200));
+        t2.setSwitchedAt(new Date(System.currentTimeMillis() + 300));
+        final Timeline t3 = new Timeline(etName, 2, new Storage(), "topic", new Date(System.currentTimeMillis() + 500));
+        return ImmutableList.of(t1, t2, t3);
+    }
+
+    private class TestTimelineSync implements TimelineSync {
```
antban: Isn't it simpler to create mock?

adyach: This stub is used only for specific case when I want to call subscribes in order to invalidate cache.

adyach:
:shipit:

PR #541, opened 2017-01-18
aruha-578 Add changelog

rcillo:
In order to better communicate changes we are going to maintain a change
log file following a well defined standard.

Since this standard required semver adoption, we are using this
opportunity to bump version from 0.8.0 to 1.0.0, which better reflects
the maturity of the project's API definition.

Read more about the standard at http://keepachangelog.com

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Change Log
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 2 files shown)


rcillo:
After offline discussion, we decided not to maintain a versioned changelog.

For purposes of release notification we are going to summarise changes weekly, manually.

PR #542, opened 2017-01-23
ARUHA-521 Separate cursor model and view

antban:
NakadiCursor renamed to TopicPosition (it is more clear what it contains)

1. Cursor, SubscriptionCursor, TopicPartition moved to view package
2. TopicPosition, PartitionStatistics introduced to hold domain data for offsets and stuff
3. Topic name generation moved to TopicRepository


rcillo:
@antban `TopicPosition` has the word `Topic` in it, which is a word used only by Kafka in this project. What's the need for it? We have `NakadiCursor`, `KafkaCursor`, `TopicPosition` and `Cursor`. I don't understand the need for `TopicPosition`.

rcillo:
@antban just saw you edited the description. Got it.

rcillo:
@antban I don't like `TopicPosition` because `Topic` is a word used by Kafka and it makes me think that it's Kafka related. I preferred something like `NakadiCursor` because then it's clear that it's not Kafka related.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -131,45 +185,21 @@ public StreamingResponseBody streamEvents(
...
-                }
+                        .withConsumingAppId(client.getClientId())
+                        .withCursors(getStreamingStart(topic, cursorsStr))
+                        .build();
 
                 // acquire connection slots to limit the number of simultaneous connections from one client
                 if (featureToggleService.isFeatureEnabled(LIMIT_CONSUMERS_NUMBER)) {
-                    final List<String> partitions = cursors.stream()
-                            .map(Cursor::getPartition)
+                    final List<String> partitions = streamConfig.getCursors().stream()
```
antban: ok

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -178,26 +208,21 @@ public StreamingResponseBody streamEvents(
...
+                eventStream = eventStreamFactory.createEventStream(
+                        outputStream, eventConsumer, streamConfig, blacklistService);
 
                 outputStream.flush(); // Flush status code to client
 
                 eventStream.streamEvents(connectionReady);
+            } catch (final UnparseableCursorException e) {
+                LOG.debug("Incorrect syntax of X-nakadi-cursors header: {}. Respond with BAD_REQUEST.",
+                        e.getCursors(), e);
+                writeProblemResponse(response, outputStream, BAD_REQUEST,
```
antban: @rcillo On line 109 it was an exception. Probably I should get message from there.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaCursor.java
```diff
@@ -24,63 +29,78 @@ public long getOffset() {
...
+        return String.valueOf(partition);
     }
 
     public static int toKafkaPartition(final String partition) {
         return Integer.parseInt(partition);
     }
 
-    public static String toNakadiPartition(final int partition) {
-        return Integer.toString(partition);
+    public static KafkaCursor fromNakadiPosition(final TopicPosition tp) throws InvalidCursorException {
```
antban: @rcillo It's inside KafkaTopicRepository

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
antban: @rcillo And now it returns NakadiCursor, so I will not rename it

src/main/java/org/zalando/nakadi/domain/NakadiCursor.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Objects;
+
+public class NakadiCursor {
+    private final String topic;
+    private final String partition;
+    // NO BEGIN OR END HERE!
+    private final String offset;
+
+    public NakadiCursor(final String topic, final String partition, final String offset) {
+        assert null != topic && null != partition && null != offset;
```
v-stepanov: Why do you check for nulls here? If it's possible that nulls are passed here - please define params as Nullable. If no nulls are expected - then I think it's better to remove checks.

src/main/java/org/zalando/nakadi/exceptions/InvalidCursorException.java
```diff
@@ -1,44 +1,79 @@
...
 
     public InvalidCursorException(final CursorError error, final Cursor cursor) {
         super();
         this.error = error;
         this.cursor = cursor;
+        this.position = null;
+    }
+
+    public InvalidCursorException(final CursorError error, final NakadiCursor position) {
+        super();
```
v-stepanov: why do you call this? It will be called anyway by default.

src/main/java/org/zalando/nakadi/exceptions/InvalidCursorException.java
```diff
@@ -1,44 +1,79 @@
...
+        this.error = error;
+        this.cursor = null;
+        this.position = position;
+    }
+
+    public InvalidCursorException(final CursorError error) {
+        super();
```
v-stepanov: and here

src/main/java/org/zalando/nakadi/repository/TopicRepository.java
```diff
@@ -15,44 +16,28 @@
...
     List<String> listPartitionNames(final String topicId);
 
-    TopicPartition getPartition(String topicId, String partition) throws NakadiException;
-
-    EventConsumer createEventConsumer(String topic, List<Cursor> cursors) throws NakadiException,
+    EventConsumer createEventConsumer(List<NakadiCursor> positions) throws NakadiException,
             InvalidCursorException;
 
-    int compareOffsets(String firstOffset, String secondOffset) throws InternalNakadiException;
+    int compareOffsets(final NakadiCursor first, final NakadiCursor second) throws InternalNakadiException;
```
v-stepanov: no need in finals in interface

src/main/java/org/zalando/nakadi/repository/kafka/KafkaCursor.java
```diff
@@ -24,63 +29,78 @@ public long getOffset() {
...
-        return partition == that.partition && offset == that.offset;
+        return Objects.equals(topic, that.topic)
+                && partition == that.partition
+                && offset == that.offset;
     }
 
     @Override
     public int hashCode() {
-        int result = partition;
+        int result = topic != null ? topic.hashCode() : 0;
```
v-stepanov: can it be null?

antban: We already had conversation about nulls in equals/hashCode/toString(), I do not want to start it again

src/main/java/org/zalando/nakadi/exceptions/InvalidCursorException.java
antban: removed

src/main/java/org/zalando/nakadi/exceptions/InvalidCursorException.java
antban: removed

src/main/java/org/zalando/nakadi/repository/TopicRepository.java
antban: removed

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -398,142 +340,78 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
...
-                    .stream()
-                    .filter(tp -> tp.getPartitionId().equals(cursor.getPartition()))
-                    .findFirst()
-                    .orElseThrow(() -> new InvalidCursorException(PARTITION_NOT_FOUND, cursor));
-
-            if (Cursor.BEFORE_OLDEST_OFFSET.equals(cursor.getOffset())) {
-                continue;
-            } else if (Cursor.BEFORE_OLDEST_OFFSET.equals(topicPartition.getNewestAvailableOffset())) {
-                throw new InvalidCursorException(EMPTY_PARTITION, cursor);
+    private List<KafkaCursor> validateConsumerCursors(final List<NakadiCursor> cursors)
```
v-stepanov: Maybe it's better to rename this method as it actually transforms Nakadi cursors to Kafka cursors.

v-stepanov: Or it's even better to separate converting from nakadi cursors to kafka cursors and adding 1 to offset. Because adding 1 to offset doesn't have to do anything with convertion.

lmontrieux: @v-stepanov if I remember well, there is a difference of 1 between the nakadi offset and the kafka offset. In that case, it seems to make sense to add 1 to the offset during the conversion.

v-stepanov: @lmontrieux No, there's actually no difference between Nakadi and Kafka cursor. The event having offset 43248 in Kafka will have offset "43248" in Nakadi. The difference is in the reading semantics. In Kafka it is: "give me messages starting from offset X". In Nakadi it is "give me messages starting from event next after X". That's why we need to add 1 when we start reading in Kafka.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -398,142 +340,82 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
...
-        return kafkaTPs
-                .stream()
-                .collect(Collectors.toMap(
-                        org.apache.kafka.common.TopicPartition::partition,
-                        consumer::position
-                ));
-    }
-
-    @Override
-    public TopicPartition getPartition(final String topicId, final String partition)
```
v-stepanov: This was optimized to get offsets for a single partition because the seeking is a very time-consuming operation. Now you replaced it with getting offsets for all partitions and selecting a single one. That will lower the performance for GET /event-types/xxx/partitions/XXX

src/test/java/org/zalando/nakadi/controller/SubscriptionControllerTest.java
```diff
@@ -425,8 +424,9 @@ public void whenGetSubscriptionStatThenOk() throws Exception {
         when(zkSubscriptionClient.getOffset(partitionKey)).thenReturn(3L);
         when(eventTypeRepository.findByName("myET"))
                 .thenReturn(EventTypeTestBuilder.builder().name("myET").topic("topic").build());
-        when(topicRepository.listPartitions(Collections.singleton("topic")))
-                .thenReturn(Collections.singletonList(new TopicPartition("topic", "0", "3", "13")));
+        final List<PartitionStatistics> statistics = Collections.singletonList(
+                new KafkaPartitionStatistics("topic", 0,0, 13));
```
v-stepanov: you lost a space here.

v-stepanov:
I'm almost ready to approve this PR

v-stepanov:
:shipit:

antban:
@v-stepanov Fixed https://github.com/zalando/nakadi/pull/542/commits/d2323bcfc88aab82d6c35f7be6ca8d13f874deef

v-stepanov:
:shipit:

PR #543, opened 2017-01-23
aruha-580: Provide online access to OpenAPI reference definition

adyach:
This PR brings ability to view api swagger file using https://github.com/zalando-stups/twintip-spring-web


lmontrieux:
:shipit:

rcillo:
@adyach is there a way to keep a symbolic link from the old file to the new file. We have dozens of users that refer to the API directly through the link and this link will be broken due to renaming.

rcillo:
@adyach if there is no way to keep this link, we better search for everywhere that links to this file, like our techjira, readme, nakadi manual, nakadi libraries...

adyach:
@rcillo good point, i will check this.

src/main/resources/application.yml
```diff
@@ -78,6 +78,12 @@ nakadi:
   auth:
     plugin:
       factory: org.zalando.nakadi.plugin.auth.DefaultApplicationServiceFactory
+
+twintip:
+  mapping: /api
+  ui: /api/ui/
```
antban: What does it mean? I mean what is /api/ui and /api ?

adyach: http://stups.readthedocs.io/en/latest/components/twintip.html

rcillo:
:shipit:

rcillo:
:shipit:

antban:
:+1:

PR #544, opened 2017-01-23
aruha-584 consumer quota

rcillo:
In order to use Kafka quota on a fine grained level, we are going to
identify consumers using a combination of application name and event
type name. This will guarantee that an application can stream up to the
default quota from a single partition, since Kafka quotas are
implemented per client-id per broker.

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/EventStreamController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventStreamController.java
@@ -235,8 +235,10 @@ public class EventStreamController {
 
     /**
      * Every consumer identifies itself using a client-id to use its quota. The client id is a combination of
-     * application name and event type so that every application can consume up to the quota limit per partition.
-     * Leader partitions from a single event type are guaranteed to be located on different brokers.
+     * application name and event type so that every application can consume up to the quota limit per partition, given
+     * partitions from the same event type are located in different brokers. In case of broker failure, multiple
+     * partitions from the same event type could be served by the same broker due to Kafka fallback. In this case, the
+     * quota would be shared between partitions, reducing the overall throughput for that event type.
      **/
     private String getKafkaQuotaClientId(final String eventTypeName, final Client client) {
         return client.getClientId() + "-" + eventTypeName;
--- a/src/main/java/org/zalando/nakadi/controller/EventStreamController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventStreamController.java
@@ -178,7 +178,9 @@ public class EventStreamController {
                 consumerCounter = metricRegistry.counter(metricNameFor(eventTypeName, CONSUMERS_COUNT_METRIC_NAME));
                 consumerCounter.inc();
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -137,13 +139,23 @@ public void terminateProducer(final Producer<String, String> producer) {
         }
     }
 
+    public Consumer<String, String> getConsumer(final Properties properties) {
```
v-stepanov: can we maybe inline this method? Because it is used only in another getConsumer method.

adyach:
Good job 👍

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -231,6 +233,15 @@ public StreamingResponseBody streamEvents(
         };
     }
 
+    /**
+     * Every consumer identifies itself using a client-id to use its quota. The client id is a combination of
+     * application name and event type so that every application can consume up to the quota limit per partition.
+     * Leader partitions from a single event type are guaranteed to be located on different brokers.
```
lmontrieux: Is that always true? If some brokers are down, there could be less brokers than the number of partitions in a single event type, and this would not hold true anymore. Of course it would be temporary, until new brokers are up.

v-stepanov:
@rcillo what about default_statistic changes? Will you do it in separate PR?

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
rcillo: @lmontrieux you're right. I forgot to mention it here as well. I wasn't sure about how much I should document in the code and how much I should document in the concept or other means of documentation.

rcillo:
@v-stepanov yes, I think it's better to create a separate PR for default statistic. I'll create the ticket.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
lmontrieux: @rcillo I guess it's fine not to document it in the code, but in that case it would be better to remove this line.

rcillo: I made some changes in 0c78145 could you please take a look to see if better covers the behaviour?

lmontrieux: It's great 👍

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
rcillo: @v-stepanov my mistake. I should have reused it somewhere else, as well 280f533

rcillo:
@adyach @v-stepanov @lmontrieux do you have any additional comments?

lmontrieux:
:shipit: no more comments for me

v-stepanov:
:shipit:

PR #545, opened 2017-01-25
aruha-563: Nakadi allows creating event types schema with incorrect JSON format

adyach:


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -189,7 +189,15 @@ private void validateName(final String name, final EventTypeBase eventType) thro
 
     private void validateSchema(final EventTypeBase eventType) throws InvalidEventTypeException {
         try {
-            final JSONObject schemaAsJson = new JSONObject(eventType.getSchema().getSchema());
+            final String eventTypeSchema = eventType.getSchema().getSchema();
+            final JSONObject schemaAsJson = new JSONObject(eventTypeSchema);
+            final String parsedSchemaAsString = schemaAsJson.toString();
+
+            // bugfix ARUHA-563
+            // https://github.com/FasterXML/jackson-databind/issues/726
```
jhorstmann: How is this a jackson issue, if you use `org.json.JSONObject` to parse the schema?

Also, does this work if my schema is pretty-printed and contains whitespaces or newlines, and is it guaranteed to keep the order of properties?

adyach: @jhorstmann this is a mistake, i will remove the link, but the principe is the same for org.json.
If your schema has newlines and some additional space it will be rejected. The order could be a problem I have to reconsider this approach. Thanks!

PR #546, opened 2017-01-26
aruha-563: Nakadi allows creating event types schema with incorrect JSON format

adyach:
In order to validate event types schemas I have added library. It will check event type schema during validation and shows particular place of error if any.


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -211,6 +213,28 @@ private void validateSchema(final EventTypeBase eventType) throws InvalidEventTy
...
+    private void checkSchemaValidity(final String eventTypeSchema) throws InvalidEventTypeException {
+        final char[] chars = eventTypeSchema.toCharArray();
+        int bracketsCounter = 0;
+        for (int i = 0; i < chars.length; i++) {
+            if (chars[i] == '{' || chars[i] == '[') {
+                bracketsCounter++;
+            } else if (chars[i] == '}' || chars[i] == ']') {
+                bracketsCounter--;
+            }
+            // if all brackets were closed but we still have symbols in schema then something wrong with the schema
```
SergKam: Any symbols except white-spaces

lmontrieux: ... and \t, \n, etc., I suppose?

adyach: @SergKam it will look like:
`"{"additionalProperties": true}        "`
@lmontrieux 
```
"{"additionalProperties": true}

"
```

I do not think we should allow this.

lmontrieux: @adyach I think we should either allow both, or neither. Otherwise it makes even less sense to me.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -211,6 +213,28 @@ private void validateSchema(final EventTypeBase eventType) throws InvalidEventTy
         }
     }
 
+    /**
+     * bugifx ARUHA-563
+     *
+     * @param eventTypeSchema
+     * @throws InvalidEventTypeException
+     */
+    private void checkSchemaValidity(final String eventTypeSchema) throws InvalidEventTypeException {
```
v-stepanov: I we really want to do it like that I suggest not to do that at all :(
I think the proper solution would be to find a proper validation library or to create a node.js microservice that will do that.

adyach: Okay

adyach: @v-stepanov I was not able to find the library which makes such validations.

rcillo:
@adyach what about that alternative we discussed previously, to parse the json and then "toString" it back and store the result? So that we would be guaranteed to be a valid json.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -189,8 +192,10 @@ private void validateName(final String name, final EventTypeBase eventType) thro
 
     private void validateSchema(final EventTypeBase eventType) throws InvalidEventTypeException {
         try {
-            final JSONObject schemaAsJson = new JSONObject(eventType.getSchema().getSchema());
+            final String eventTypeSchema = eventType.getSchema().getSchema();
+            checkJsonIsValid(eventTypeSchema);
 
+            final JSONObject schemaAsJson = new JSONObject(eventTypeSchema);
```
jhorstmann: Looking at https://github.com/everit-org/json-schema#quickstart you can create a `JSONObject` from a `JSONTokener`, you could then call `next` or `more` on the tokener to determine if it reached the end of the input.

build.gradle
```diff
@@ -148,6 +148,7 @@ dependencies {
         exclude module: "json"
     }
     compile 'com.fasterxml.jackson.datatype:jackson-datatype-joda:2.4.0'
+    compile 'com.grack:nanojson:1.1'
```
jhorstmann: How many different json parsers does nakadi now depend on?

adyach: If you want to be really involved in this, I will give some context:
The original problem was found during test phase of our Nakadi UI. It happens that we use json parser wihch is not fully json rfc comptible that was about multiple roots and arbitary symbols at the end of the json schema. The solution I provided for that was pretty simple just go through the ET schema and count the number of the open/close brackets if something left after num of brackets is 0 then it was considered as invalid schema (https://github.com/zalando/nakadi/pull/546/commits/45b8bad8b37852cedb78223f3f9e523452a58b2d#diff-0cc0d4406e77f42ca8a2ee3e3ce445e1R222). 
Suddenly, another case appeared where schema had trailing commas in the json arrays and objects which of course was not covered by the solution.

We found out several possible ways to solve it:
1. Change nothing, leave it as it is
	As we understand it is not critical for our application, because we still can work with this malformed json schema and we have higher priority things to accomplish. In this approach we do not care about users that much (they lived with that for more than one year), becasuse they will see unprocessable json schema in the UI, but concentrate on more important task.

2. Store only parsed json object
	Right now after parsing ET json schema we store provided json schema not the parsed one. This approache considered stroring schema in our database after it was parsed. The first downside is that not clear for user since he provided one ET schema, but at the end got other and the second there could be mistakes, when user provided json schema with multiple roots, but we stored only the one of them (it how parsing works now).
	
3. Our own implementation
	The approache is quite interesting, but very error prone in the way that json parsing has a lot of cases and probably we'll miss somethig, which will introduce real user affecting bugs.
	
4. Third party dependecy
	We have several third party libraries connected to our project, but non of them are support json parsing in a proper way. This approache is less error prone comparing to our own implementation and of course it can have their own bugs. We used http://seriot.ch/json/parsing.html to find out the parser which better suites for our needs and nanojson looks like it.
	
About number of dependencies:
Actually until this moment we have not had any direct dependecies on json parser they are depenencies of other libraries we use. As to my calculation it is about 3: jackson, JSON, json-smart.

To conclude, probably, there are other solutions to that problem. I choose this one to be more customer centric with less effort possible I agree it is not the best possible solution, but more pragmatic. In other case I would rather go for 1.

jhorstmann: Please don't take this comment about dependencies to serious, my point was that this validation is completely possible with the existing dependencies and without parsing the string multiple times.

Both jackson and org.json support reading multiple objects from the same stream, so don't consider trailing characters after the first object as invalid. As a user of those libraries it is easy to validate that the whole input has been consumed parsing from a `JSONParser` or `JSONTokener`. The following method passes nearly all unit tests you have below, the only exception being that org.json allows trailing commas after a key-value pair (noted as n_object_trailing_comma.json in http://seriot.ch/json/parsing.html#24).

    private void validate(String input) {
        final JSONTokener tokener = new JSONTokener(input);
        try {
            final JSONObject json = new JSONObject(tokener);
        } catch (JSONException e) {
            throw new InvalidEventTypeException("Invalid json", e);
        }
        if (tokener.nextClean() != 0) {
            throw new InvalidEventTypeException("Invalid content after json");
        }
    }

This simple validation could be inlined and you could create the schema from the `JSONObject` without having to parse the input again.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -211,6 +216,15 @@ private void validateSchema(final EventTypeBase eventType) throws InvalidEventTy
         }
     }
 
+    @VisibleForTesting
```
v-stepanov: For me this annotation always smells, because it means that you want to test something that is supposed to be private. I think a better solution can be to create a separate class that does only json validation and put this method there. And EventTypeService will use this class.

adyach: @v-stepanov I usually do it the way you described, but the reason was that JsonParser returns a pretty good described error and points to the position where the error happened. my thoughts were it is better to have this annotation and keep the InvalidEventTypeException out of the separate class. I have thought it over again and I think if I name the method properly it will be fine to have the exception in separate class.
Here is a fix 57fa20f

v-stepanov:
for me everything is good

v-stepanov:
:shipit:

v-stepanov:
:shipit:

PR #548, opened 2017-01-26
Aruha 528 timeline locking when publish

v-stepanov:
The idea of this timeout thing was to introduce a global timeout for publishing endpoint, so that we will guarantee to our users that publishing will not take longer than X. But I think this idea need a further discussion.


src/main/java/org/zalando/nakadi/service/EventPublisher.java
```diff
@@ -82,6 +95,17 @@ public EventPublishResult publish(final String events, final String eventTypeNam
         } catch (final EventPublishingException e) {
             LOG.error("error publishing event", e);
             return failed(batch);
+        } catch (InterruptedException | TimeoutException e) {
```
antban: After catching InterruptedException one must do one of the following:
 - stop the thread
 - rise InterruptedException again
 - Call Thread.interrupt(), so the thread owner will be able to detect thread interruption

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -203,10 +204,6 @@ public boolean partitionExists(final String topic, final String partition) throw
                 }
             }));
             return result;
-        } catch (final InterruptException e) {
-            circuitBreaker.markSuccessfully();
```
antban: Who will call circuitBreaker in this case?

antban:
finished review

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
v-stepanov: The next catch, all InterruptException's will be handled there.

src/main/java/org/zalando/nakadi/service/EventPublisher.java
v-stepanov: I would better change timelineSync.workWithEventType() to throw some custom exception instead of InterruptedException and TimeoutException.

rcillo:
Should we also check `workWithEventType` when updating an event type so that we don't have problems with concurrent update and timeline creation?

v-stepanov:
@rcillo do you see a case when it can create a conflict

rcillo:
@v-stepanov this is a corner case. Let's suppose a user is updating the retention time of an event type. What could happen:

1. User PUT change to retention_time
2. Admin rollout a new Timeline for that event type
3. change to retention is applied to old Timeline
4. new Timeline is switched based on outdated retention_time

v-stepanov:
@rcillo Ok, it makes sense. I forgot that it's theoretically possible to change retention time. Ok, I will add that.

v-stepanov:
@rcillo added that for updating/deleting event-type https://github.com/zalando/nakadi/pull/548/commits/bedc09f4694447a74bc906662ab6697f1fa0a415

rcillo:
:shipit:

src/test/java/org/zalando/nakadi/service/subscription/StreamingContextTest.java
```diff
@@ -130,27 +127,4 @@ public void onExit() {
         Assert.assertArrayEquals(new boolean[]{true, true}, onExitCalls);
     }
 
-    @Test
-    public void testOnNodeShutdown() throws Exception {
```
adyach: This is not good, since this confirms the sequence of the actions for subscription clean up in case of instance termination.

v-stepanov: I merged now the fixed version of test from master

adyach:
:shipit:

src/main/java/org/zalando/nakadi/service/EventPublisher.java
```diff
@@ -42,28 +47,34 @@
...
 
     public EventPublishResult publish(final String events, final String eventTypeName, final Client client)
-            throws NoSuchEventTypeException, InternalNakadiException {
-        final EventType eventType = eventTypeCache.getEventType(eventTypeName);
+            throws NoSuchEventTypeException, InternalNakadiException, EventTypeTimeoutException {
+
+        Closeable publishingCloser = null;
         final List<BatchItem> batch = BatchFactory.from(events);
+        try {
+            publishingCloser = timelineSync.workWithEventType(eventTypeName, nakadiSettings.getTimelineWaitTimeoutMs());
```
antban: Isn't it better to do try with resource?

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -123,17 +139,37 @@ public EventTypeService(final EventTypeRepository eventTypeRepository, final Top
...
+            } catch (final IOException e) {
+                LOG.error("Exception occurred when releasing usage of event-type", e);
+            }
         }
     }
 
     public Result<Void> update(final String eventTypeName, final EventTypeBase eventTypeBase, final Client client) {
+        Closeable updatingCloser = null;
         try {
+            updatingCloser = timelineSync.workWithEventType(eventTypeName, nakadiSettings.getTimelineWaitTimeoutMs());
```
antban: I don't think that we need to lock here. Timelines are not changed on this step

src/main/java/org/zalando/nakadi/service/EventPublisher.java
v-stepanov: I wanted to do it like that in the beginning but it doesn't work as `workWithEventType` throws exception :(

src/main/java/org/zalando/nakadi/service/EventTypeService.java
v-stepanov: please see this comment from Ricardo https://github.com/zalando/nakadi/pull/548#issuecomment-276060597

The possible case can be the change of retention time (which will work sooner or later)

v-stepanov:
@antban Do you still want to discuss something here? I got an approval, so I plan to put this to validation.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
antban: ok

antban:
:+1:

adyach:
:shipit:

PR #549, opened 2017-01-26
Aruha 523 CRUD for Storages

lmontrieux:
This PR adds CRD (no U) operations for storages. Endpoint requires admin credentials.


api/nakadi-event-bus-api.yaml
```diff
@@ -1142,7 +1142,119 @@ paths:
         '204':
           description: Feature was successfully accepted.
 
+  /storages:
+    get:
+      tags:
+        - storages-api
```
antban: @lmontrieux Maybe it's better to mark it as timelines-api?

lmontrieux: f01d20f

api/nakadi-event-bus-api.yaml
```diff
@@ -1151,6 +1263,36 @@ paths:
 # ################################### #
 
 definitions:
+  Storage:
+      type: object
+      description: |
+        This is a storage backend
+        A real one.
```
antban: Sorry, I do not get the meaning of this description.

api/nakadi-event-bus-api.yaml
```diff
@@ -1151,6 +1263,36 @@ paths:
...
+      properties:
+        id:
+          description: The ID of the storage backend
+          type: string
+        storage_type:
+          description: the type of storage (currently, only Kafka)
+          type: string
```
antban: Maybe it is better to set it to enum providing possible values

api/nakadi-event-bus-api.yaml
```diff
@@ -1151,6 +1263,36 @@ paths:
...
+        kafka_configuration:
+          description: configuration settings (currently, only Kafka)
```
antban: Probably it's better to say that it is filled up only when backend type is kafka. For kinesis one will have `kinesis_configuration`, as far as I understand.

src/main/java/org/zalando/nakadi/controller/StoragesController.java
```diff
@@ -0,0 +1,86 @@
...
+            return status(FORBIDDEN).build();
+        }
+        final Result<Void> result = storageService.deleteStorage(id);
+        if (result.isSuccessful()) {
+            return status(HttpStatus.OK).build();
+        }
+        return Responses.create(result.getProblem(), request);
+    }
+
+    private boolean isNotAdmin(final Client client) {
```
antban: Isn't it better to use `@ExceptionHandler` and throw it in service, not it controller?

src/main/java/org/zalando/nakadi/domain/Storage.java
```diff
@@ -70,6 +75,7 @@ public String toString() {
     }
 
     private String id;
+    @JsonProperty("storage_type")
```
antban: I don't really like the way we do that (but of course it's up to team)
I thought that approach with different View and Model objects is much better. For example, from current code I can not figure out why configuration objects should be mapped to `kafka_configuration` value.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
```diff
@@ -38,6 +38,10 @@ public TimelineDbRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper
                 eventType);
     }
 
+    public List<Timeline> listTimelines() {
+        return jdbcTemplate.query(BASE_TIMELINE_QUERY + " order by t.tl_order", timelineRowMapper);
+    }
```
antban: It's not a good idea to select all timelines (more than 2000 records) only to check if any storage is used. Maybe it's better to create method `isStorageUsed(Storage)`

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -0,0 +1,97 @@
...
+    public List<Storage> listStorages() {
+        return storageDbRepository.listStorages();
+    }
+
+    public Result<Storage> getStorage(final String id) {
+        final Optional<Storage> storage = storageDbRepository.getStorage(id);
+        if (storage.isPresent()) {
+            return Result.ok(storage.get());
+        }
+        else {
```
antban: I thought that we decided to use write `} else {` without newline

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -0,0 +1,97 @@
...
+            return Result.problem(Problem.valueOf(Response.Status.NOT_FOUND, "No storage with id " + id));
+        }
+    }
+
+    public Result<Storage> createStorage(final JSONObject storageDetails) {
```
antban: It looks like it's injection of view logic into service code

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -0,0 +1,97 @@
...
+        final String type;
+        final JSONObject configuration;
+
+        try {
+            type = storageDetails.getString("storage_type");
+            configuration = storageDetails.getJSONObject("configuration");
```
antban: Here it is (I meant view logic) on service layer, that wrongly treats field name - `configuration` instead of `kafka_configuration`

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -0,0 +1,97 @@
...
+        } catch (JSONException e) {
+            return Result.problem(Problem.valueOf(UNPROCESSABLE_ENTITY, e.getMessage()));
+        }
+
+        final Storage storage = new Storage();
+        storage.setId(UUID.randomUUID().toString());
```
antban: I thought that id is generated by user, not by service. Imagine that you have 3 kafka clusters:
- default
- 123e4567-e89b-12d3-a456-426655440000
- 123e4567-e89b-12d3-a456-426555440000

They are all different, but I prefer to use: 
- default
- kafka09
- kafka10

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+        try {
+            storageDbRepository.deleteStorage(id);
+        } catch (DataAccessException e) {
+            return Result.problem(Problem.valueOf(Response.Status.INTERNAL_SERVER_ERROR, e.getMessage()));
+        }
+        return Result.ok();
+    }
+
+    private boolean isInUse(final String id) {
```
antban: As I already said, it's better to ask db for this logic

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,22 @@
...
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.zalando.nakadi.domain.Timeline;
+import org.zalando.nakadi.repository.db.TimelineDbRepository;
+
+import java.util.List;
+
+@Service
+public class TimelineService {
```
antban: I think that it's possible to use different db repositoies from one service, hence this class is not needed now

antban:
I finished review

api/nakadi-event-bus-api.yaml
lmontrieux: I wrote a dummy description and, of course, I forgot to change it later.

src/main/java/org/zalando/nakadi/service/StorageService.java
lmontrieux: I didn't know, will fix that

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
lmontrieux: That is true, but there is a ticket called "CRUD for timelines", so I think this service will be needed anyway.

api/nakadi-event-bus-api.yaml
lmontrieux: I thought of using `one_of`, but currently we only have Kafka, so it's a bit useless. Except that it indicates that there may be more storage types in the future

src/main/java/org/zalando/nakadi/controller/StoragesController.java
lmontrieux: I went for the same approach as what's already in the Settings controller, but I agree that @ExceptionHandler is better.

src/main/java/org/zalando/nakadi/service/StorageService.java
lmontrieux: 64805d8

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
lmontrieux: 5d21ddd

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
lmontrieux: Well it's not used anymore, so I've removed it: eee2135

src/main/java/org/zalando/nakadi/service/StorageService.java
lmontrieux: 5d21ddd

api/nakadi-event-bus-api.yaml
lmontrieux: 92c6f33

src/main/java/org/zalando/nakadi/service/StorageService.java
lmontrieux: That makes sense. I will change to String, and make it clear in the API documentation that the character limit is 36 characters.

api/nakadi-event-bus-api.yaml
```diff
@@ -1142,7 +1142,104 @@ paths:
...
+          description: Storage description
+          schema:
+            $ref: '#/definitions/Storage'
+      responses:
+        '204':
+          description: Storage backend was successfully registered. Returns storage object that was created.
+          schema:
+            $ref: '#/definitions/Storage'
+        '403':
+                  description: Access forbidden
```
v-stepanov: formatting broken

api/nakadi-event-bus-api.yaml
```diff
@@ -1142,7 +1142,104 @@ paths:
...
+        The oauth resource owner username has to be equal to 'nakadi.oauth2.adminClientId' property
+        to be able to access this endpoint.
+      parameters:
+        - name: storage
+          in: body
+          description: Storage description
+          schema:
+            $ref: '#/definitions/Storage'
+      responses:
+        '204':
```
v-stepanov: Why don't you want to use 201 here? Even in the description you use the word "created".

api/nakadi-event-bus-api.yaml
```diff
@@ -1142,7 +1142,104 @@ paths:
...
+                  description: Access forbidden
+                  schema:
+                    $ref: '#/definitions/Problem'
 
+  /storages/{id}:
+    get:
+      tags:
+        - timelines-api
+      description: |
+        Retrieves a storage backend from its ID.
```
v-stepanov: maybe better "**by** its ID" ?

api/nakadi-event-bus-api.yaml
```diff
@@ -1142,7 +1142,104 @@ paths:
...
+        The oauth resource owner username has to be equal to 'nakadi.oauth2.adminClientId' property
+        to be able to access this endpoint.
+      parameters:
+        - name: id
+          in: path
+          description: storage backend ID
+          type: string
+          required: true
+      responses:
+        '200':
```
v-stepanov: What do you think about 204 here?

api/nakadi-event-bus-api.yaml
```diff
@@ -1151,6 +1248,38 @@ paths:
 # ################################### #
 
 definitions:
+  Storage:
+      type: object
```
v-stepanov: formatting broken here? I can see 4 spaces instead of 2.

api/nakadi-event-bus-api.yaml
```diff
@@ -1151,6 +1248,38 @@ paths:
...
+        A storage backend.
+      properties:
+        id:
+          description: The ID of the storage backend.
+          type: string
+          maxLength: 36
+        storage_type:
+          description: the type of storage
+          type: string
+          enum:
```
v-stepanov: Using enum in API makes it impossible to change in compatible way. The reason is following: client writes the code that calls GET on this resource and he expects to get enum here with exactly one possible value. If we will extend enum in the future - that will break a client. So it's better to make it a string and describe possible values.
Like here for example: https://github.com/zalando/nakadi/blob/master/api/nakadi-event-bus-api.yaml#L1773

The example of this problem you can see in BatchItemResponse definition that has several enum properties that we are not able to change any more (and we already wanted to do that several times)

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -0,0 +1,86 @@
...
+    public Result<Void> deleteStorage(final String id) {
+        if (!storageDbRepository.getStorage(id).isPresent()) {
+            return Result.notFound("No storage with ID " + id);
+        }
+        if (storageDbRepository.isStorageUsed(id)) {
+            return Result.forbidden("Storage " + id + " is in use");
+        }
+        try {
+            storageDbRepository.deleteStorage(id);
+        } catch (DataAccessException e) {
```
v-stepanov: For me this is a bit mixing of layers. I think DataAccessException should be caught at repository level and throw more business-ish exception.

src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
```diff
@@ -53,6 +53,10 @@ public void deleteStorage(final String id) {
         jdbcTemplate.update("DELETE FROM zn_data.storage WHERE st_id=?", id);
     }
 
+    public boolean isStorageUsed(final String id) {
```
v-stepanov: I think you can add a test method for this into StorageDbRepositoryTest.

src/main/java/org/zalando/nakadi/controller/StoragesController.java
lmontrieux: I'm going to solve this in another ticket, since we have opened ARUHA-589 for that

api/nakadi-event-bus-api.yaml
lmontrieux: 640959e

src/main/java/org/zalando/nakadi/service/StorageService.java
lmontrieux: ba26e5f

api/nakadi-event-bus-api.yaml
lmontrieux: 853545c

api/nakadi-event-bus-api.yaml
lmontrieux: a9ce38f

api/nakadi-event-bus-api.yaml
lmontrieux: 6008de5

api/nakadi-event-bus-api.yaml
lmontrieux: Good point. Fixed in fe7b7b9

api/nakadi-event-bus-api.yaml
lmontrieux: @v-stepanov made a good point below about enums not being as easily extensible, so I went back to String, with the permitted value in the description: fe7b7b9

api/nakadi-event-bus-api.yaml
lmontrieux: 640959e

src/main/java/org/zalando/nakadi/service/StorageService.java
lmontrieux: In other places we throw an `InternalNakadiException`. I'm on this.

v-stepanov:
@lmontrieux For me everything looks fine now. So I'm ready to approve if @antban thinks that his issues are fixed as well.

src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
lmontrieux: 68f8a95 (with a bug fix as well)

src/main/java/org/zalando/nakadi/domain/Storage.java
lmontrieux: I agree. I will do this in another branch, as to not block our progress

src/main/java/org/zalando/nakadi/service/StorageService.java
lmontrieux: Will be fixed in the branch mentioned 3 comments up

src/main/java/org/zalando/nakadi/service/StorageService.java
lmontrieux: Will be fixed in the branch mentioned 4 comments up

api/nakadi-event-bus-api.yaml
lmontrieux: b7e9f7d

api/nakadi-event-bus-api.yaml
rcillo: This comment is in a weird place. I expected it to be made on line 1233.

src/main/java/org/zalando/nakadi/domain/Storage.java
rcillo: I don't know. Creating a second `Storage` to preserve layering but actually having very similar objects (all attributes with the same type) sometimes is might not help so much. I prefer to make a 2k lines refactoring every 6 months than being strict on layering and end up with a lot of equal classes.

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -0,0 +1,118 @@
...
+                    break;
+                default:
+                    return Result.problem(Problem.valueOf(UNPROCESSABLE_ENTITY,
+                            "Type '" + type + "' is not a valid storage type"));
+            }
+        } catch (JSONException e) {
+            return Result.problem(Problem.valueOf(UNPROCESSABLE_ENTITY, e.getMessage()));
+        }
+
+        try {
```
rcillo: It's better to create a unique key constraint in the DB and have it make the uniqueness check than implement it in the application level. In the best case, this code is redundant since the the DB already validates uniqueness. In the worst case (unique constraint is missing) it contains a concurrency problem.

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/domain/Storage.java
lmontrieux: When @antban commented on this, the json sent by the user didn't have an ID (it was generated by nakadi as a random UUID). Now, they're pretty much identical, so the issue is less clear cut.

src/main/java/org/zalando/nakadi/service/StorageService.java
lmontrieux: `st_id` is a primary key, so this code is indeed redundant.

lmontrieux: cda190f and b7079c3

rcillo:
:shipit:

lmontrieux:
@antban Fixed the 204/200 issue when deleting storages

antban:
:+1:

src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
```diff
@@ -46,11 +62,29 @@ public Storage createStorage(final Storage storage) throws DataAccessException {
...
         }
     }
 
-    public void deleteStorage(final String id) {
-        jdbcTemplate.update("DELETE FROM zn_data.storage WHERE st_id=?", id);
+    public void deleteStorage(final String id) throws InternalNakadiException {
+        try {
+            jdbcTemplate.update("DELETE FROM zn_data.storage WHERE st_id=?", id);
+        } catch (DataAccessException e) {
+            throw new InternalNakadiException("Error occurred when deleting Storage " + id, e);
```
adyach: I haven't found it to be logged somewhere. Probably, the logging of the error would be nice in the future.

src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
```diff
@@ -46,11 +62,29 @@ public Storage createStorage(final Storage storage) throws DataAccessException {
         } catch (final JsonProcessingException ex) {
             throw new IllegalArgumentException("Storage configuration " + storage.getConfiguration(Object.class) +
                     " can't be mapped to json", ex);
+        } catch (DuplicateKeyException e) {
```
adyach: It looks quite handy to use this kind of exceptions here, but it is not good to use presentation layer entities from storage layer. I suppose it is better to propagate runtime exception of jdbc template and process them on the service layer.

src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
```diff
@@ -22,20 +25,33 @@ public StorageDbRepository(final JdbcTemplate template, final ObjectMapper mappe
...
+                    id);
+        } catch (DataAccessException e) {
+            throw new InternalNakadiException("Error occurred when fetching Storage " + id, e);
+        }
         return Optional.ofNullable(storages.isEmpty() ? null : storages.get(0));
     }
 
-    public Storage createStorage(final Storage storage) throws DataAccessException {
+    public Storage createStorage(final Storage storage)
+            throws DataAccessException, DuplicatedStorageIdException, InternalNakadiException {
```
adyach: Actually, you catch DataAccessException down there.

PR #551, opened 2017-02-02
ARUHA-633 Disable consumer offset auto commit

lmontrieux:
This PR sets enable.auto.commit to false

### DIFF
--- a/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
@@ -54,6 +54,7 @@ public class KafkaRepositoryAT extends BaseAT {
     private static final int KAFKA_BATCH_SIZE = 1048576;
     private static final long KAFKA_LINGER_MS = 0;
     private static final long NAKADI_EVENT_MAX_BYTES = 1000000L;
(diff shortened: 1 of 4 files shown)


rcillo:
@lmontrieux as we talked previously, I have some questions:

1. is it safe to remove `__consumer_offsets` partition after disabling this feature?
2. when creating a new cluster, if this feature is never used, does it create `__consumer_offsets` automatically? When is this partition created?
3. when validating the impact of this change, is it possible to observe any changes in "bytes/flow" for this specific topic?

rcillo:
@lmontrieux don't you have to manually call `seek` to update the position to read on the next `poll` operation?

v-stepanov:
@rcillo I think we can test that quite easy.

lmontrieux:
@rcillo 

About 2.
It is created at startup, but it stays empty (confirmed locally)

Will update about the other questions

lmontrieux:
@rcillo looks like we only need to call `seek` upon restarting the client:

```
On restart restore the position of the consumer using seek(TopicPartition, long).
```

lmontrieux:
@rcillo for 1., I don't know, but I'll check in a sandbox.
For 3., I suspect that it is the case, since data replication is counted. In any case we can also connect to the kafka node that's leader for the partition, and consume from it to make sure nothing is written there anymore.

v-stepanov:
👍

PR #552, opened 2017-02-03
Aruha 596 commit performance

v-stepanov:


adyach:
@v-stepanov could you, please, describe shortly what you did here to improve it?

v-stepanov:
@adyach 
In general I did two things here:
1) I removed calling subscriptionClient.getZkSubscriptionNodeLocked() which locks the whole subscription in ZK and grabs all subscription data from ZK. I replaced that with getting only the info we need to validate the streamId, that reduced the amount of calls to ZK and no locking.

2) When writing new cursors I removed the lock of offset node - instead I get the version of node and try to write to the same version of node, if that is already replaced by the newer version (which should not normally happen if you don't commit the same partition from different threads) - it retries up to 5 times.

Yeah, and I also replaced the test for the cursors service as it was becoming a mock-nightmare. Now the test uses real ZK.

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -91,30 +86,60 @@ public CursorsService(final ZooKeeperHolder zkHolder,
...
+                            throw new NakadiRuntimeException(e);
+                        }
+                    })
+                    .collect(Collectors.toList());
+
+            if (!invalidCursors.isEmpty()) {
+                throw new InvalidStreamIdException("Cursors " + invalidCursors + " cannot be committed with stream id "
+                        + streamId);
+            }
+        } catch (final NakadiRuntimeException e) {
```
antban: Maybe it makes sense to catch RuntimeException

antban:
:+1:

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: I wanted to catch here only what I throw inside lambda. As I understand our general approach is that we don't catch runtime exceptions everywhere (for me it's like to check everything for null).

v-stepanov:
@antban I merged with master and had to do minor changes for that. Please re-review.

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -143,18 +168,36 @@ private boolean commitCursor(final String subscriptionId, final String eventType
...
+                                    .forPath(offsetPath, cursor.getOffset().getBytes(Charsets.UTF_8));
+                            return true;
+                        } else {
+                            return false;
+                        }
+                    },
+                    new RetryForSpecifiedCountStrategy<Boolean>(COMMIT_CONFLICT_RETRY_TIMES)
+                            .withExceptionsThatForceRetry(KeeperException.BadVersionException.class));
+
+            return Optional.ofNullable(committed).orElse(false);
```
antban: commit can not be null

src/main/java/org/zalando/nakadi/service/CursorsService.java
antban: Sorry, didn't saw that it is thrown right here

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: executeWithRetry is `@Nullable` so I don't really know if it will return null or not.

antban:
:+1:

PR #554, opened 2017-02-06
ARUHA-646: added query params to logging;

v-stepanov:
Sorry for this random PR, I just was browsing the logs and realized that there are no query params logged. And I also did what I always wanted: I added "statusCode" prefix to make searching for specific status code easier.


adyach:
👍

PR #556, opened 2017-02-07
aruha-645: fixed underterministic test

adyach:


src/test/java/org/zalando/nakadi/service/subscription/StreamingContextTest.java
```diff
@@ -145,9 +145,13 @@ public void onEnter() {
...
         t.start();
-        t.join(2000);
+        t.join(1000);
+
+        t = new Thread(() -> ctxSpy.onNodeShutdown());
+        t.start();
+        t.join(1000);
 
-        ctxSpy.onNodeShutdown();
+        Thread.sleep(1000);
```
v-stepanov: Why do you need this sleep? You already waited each thread 1 second.

adyach: onNodeShutdown switches the state to CleanupState which is processed by the first thread in the test, but after that CleanupState itself changes the state to DummyState which also has be processed by the first thread. Probably, I do not need the join on the second thread, because Thread.sleep should be enough.

adyach: @v-stepanov f4199ee

adyach: I run it 5 times in the row and passed 5 times.

v-stepanov:
👍

PR #557, opened 2017-02-07
ARUHA-648: extended logging

v-stepanov:
Just to see for whom it happens next time.


lmontrieux:
:shipit:

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -122,7 +122,8 @@ public void streamEvents(final AtomicBoolean connectionReady) {
         } catch (final IllegalStateException e) {
             LOG.info("Error occurred when streaming events (possibly server closed connection)", e);
         } catch (final KafkaException e) {
-            LOG.error("Error occurred when polling events from kafka", e);
+            LOG.error("Error occurred when polling events from kafka; consumer: " + config.getConsumingAppId() +
```
adyach: We usually use `{}` for params

v-stepanov: But then I will not be able to provide exception in the log.

adyach: you can pass it as the latest param
https://www.slf4j.org/faq.html#paramException

antban: @v-stepanov I agree with @adyach

adyach:
👍

src/main/java/org/zalando/nakadi/service/EventStream.java
v-stepanov: @adyach cool, I didn't know about that. I changed it now.

adyach:
:shipit:

PR #562, opened 2017-02-15
aruha-527: CRUD timelines

adyach:
This PR brings ability to use timelines only with one default topic repository.


api/nakadi-event-bus-api.yaml
```diff
@@ -1269,15 +1269,20 @@ definitions:
...
           zk_address:
             description: the Zookeeper address
             type: string
           zk_path:
             description: the Zookeeper path
             type: string
         required:
-          - zk_address
-          - zk_path
+          - exhibitor_address
```
antban: I don't get it. I thought that it is important to have either (zk_address, zk_path) pair or (exhibitor_address, exhibitor_port, zk_path). In fact - no required fields at all. But having exhibitor_address required is actually incorrect. 
Am I correct?

src/main/java/org/zalando/nakadi/config/NakadiSettings.java
```diff
@@ -74,4 +74,8 @@ public long getEventMaxBytes() {
         return eventMaxBytes;
     }
 
+    public long getTimelineWaitTimeoutMs() {
```
antban: This probably will be replaced by https://github.com/zalando/nakadi/blob/master/src/main/java/org/zalando/nakadi/config/NakadiSettings.java#L80 ?

api/nakadi-event-bus-api.yaml
adyach: Explained it offline as a result we want to add description for KafkaConfiguration class.

src/main/java/org/zalando/nakadi/config/NakadiSettings.java
adyach: yes, I added it before it was in master.

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
```diff
@@ -0,0 +1,50 @@
...
+    public ResponseEntity<?> createStorage(@RequestBody final TimelineRequest timelineRequest, final Client client) {
+        timelineService.createTimeline(timelineRequest, client);
+        return ResponseEntity.status(HttpStatus.CREATED).build();
+    }
+
+    @ResponseStatus(value = HttpStatus.FORBIDDEN)
+    @ExceptionHandler(ForbiddenAccessException.class)
+    public void forbidden(final Exception ex) {
+        LOG.debug("Request is forbidden", ex);
+    }
```
antban: Problem description is missing here.

antban: I  meant it is not returned in any response field

src/main/java/org/zalando/nakadi/repository/TopicRepositoryHolder.java
```diff
@@ -0,0 +1,100 @@
...
+                    final ZooKeeperHolder zooKeeperHolder = new ZooKeeperHolder(
+                            kafkaConfiguration.getZkAddress(),
+                            kafkaConfiguration.getZkPath(),
+                            kafkaConfiguration.getExhibitorAddress(),
+                            kafkaConfiguration.getExhibitorPort());
+                    final KafkaFactory kafkaFactory =
+                            new KafkaFactory(new KafkaLocationManager(zooKeeperHolder, kafkaSettings), metricRegistry);
+                    final KafkaTopicRepository kafkaTopicRepository = new KafkaTopicRepository(zooKeeperHolder,
+                            kafkaFactory, nakadiSettings, kafkaSettings, zookeeperSettings, uuidGenerator);
+                    return kafkaTopicRepository;
```
antban: maybe it makes sense to check if it is possible to call at least something from this topic repository (like list topics)

adyach: do you want to check if it was properly created?

src/main/java/org/zalando/nakadi/repository/zookeeper/ZooKeeperHolder.java
```diff
@@ -1,37 +1,57 @@
...
+    }
+
+    private void initExhibitor() throws Exception {
+        final RetryPolicy retryPolicy = new ExponentialBackoffRetry(EXHIBITOR_RETRY_TIME, EXHIBITOR_RETRY_MAX);
+        final Collection<String> exhibitorHosts = Arrays.asList(exhibitorAddresses.split("\\s*,\\s*"));
+        final Exhibitors exhibitors = new Exhibitors(exhibitorHosts, exhibitorPort,
+                () -> zookeeperBrokers + zookeeperKafkaNamespace);
+        final ExhibitorRestClient exhibitorRestClient = new DefaultExhibitorRestClient();
+        final ExhibitorEnsembleProvider ensembleProvider = new ExhibitorEnsembleProvider(exhibitors,
+                exhibitorRestClient, "/exhibitor/v1/cluster/list", EXHIBITOR_POLLING_MS, retryPolicy);
```
antban: I don't get it. Is it now impossible to create with zookeeper only?
There was FixedEnsembleProvider used for situation when only zookeeper available. Now `Exhibitors` instance always created, and ensemble provider will take into account that it it's exhibitors, not zookeepers.

src/main/java/org/zalando/nakadi/repository/zookeeper/ZooKeeperHolder.java
```diff
@@ -1,37 +1,57 @@
...
+        ensembleProvider.pollForInitialEnsemble();
```
antban: As for me - this call should die (but of course it's netflix exhibitor, so probably it will return something instead of saying that something is wrong)

src/main/java/org/zalando/nakadi/repository/zookeeper/ZookeeperConfig.java
```diff
@@ -10,11 +10,11 @@
 public class ZookeeperConfig {
 
     @Bean
-    public ZooKeeperHolder zooKeeperHolder(final Environment environment) {
+    public ZooKeeperHolder zooKeeperHolder(final Environment environment) throws Exception {
```
antban: Why Exception is here?

src/main/java/org/zalando/nakadi/repository/TopicRepositoryHolder.java
antban: yep, properly created and can work

src/main/java/org/zalando/nakadi/repository/zookeeper/ZooKeeperHolder.java
adyach: If you do not provide exhibitor address it will use provided zk backup connection

src/main/java/org/zalando/nakadi/repository/zookeeper/ZookeeperConfig.java
adyach: It was in ZooKeeperHolder post construct method before, since I want to use ZooKeeperHolder outside spring context I moved post construct in constructor.

src/main/java/org/zalando/nakadi/repository/zookeeper/ZooKeeperHolder.java
adyach: @antban 2259a97

api/nakadi-event-bus-api.yaml
adyach: bb72caa

src/main/java/org/zalando/nakadi/repository/TopicRepositoryHolder.java
adyach: f4dcfe5

src/main/java/org/zalando/nakadi/domain/Timeline.java
```diff
@@ -168,4 +172,22 @@ public boolean equals(final Object o) {
...
+            final int order,
+            final Storage storage,
+            final String topic,
+            final Date createdAt) {
+        final Timeline timeline = new Timeline(eventType, order, storage, topic, createdAt);
+        timeline.setId(UUID.randomUUID());
+        return timeline;
+    }
+
+    public static Timeline createFakeTimeline(final EventType eventType, final Storage storage) {
```
v-stepanov: Why do we need this?

src/main/java/org/zalando/nakadi/domain/Timeline.java
```diff
@@ -142,6 +142,10 @@ public void setCleanupAt(@Nullable final Date cleanupAt) {
         this.cleanupAt = cleanupAt;
     }
 
+    public boolean isFake() {
```
v-stepanov: What is fake timeline?

src/main/java/org/zalando/nakadi/domain/Timeline.java
adyach: I suppose you meant the fake timeline. This is for ability to switch back from the first real timeline we create in case something goes wrong, because fake timeline won't be created in DB.

src/main/java/org/zalando/nakadi/domain/Timeline.java
adyach: The timeline which is not registered in DB and equals to the topic of the currently used storage.

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private final TimelineService timelineService;
+
+    @Autowired
+    public TimelinesController(final TimelineService timelineService) {
+        this.timelineService = timelineService;
+    }
+
+    @RequestMapping(method = RequestMethod.POST)
+    public ResponseEntity<?> createStorage(@RequestBody final TimelineRequest timelineRequest, final Client client) {
```
v-stepanov: I don't exactly get what this endpoint is supposed to do. Is it not finished currently?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,149 @@
...
+                LOG.info("Switching timelines. Topics: {} -> {}", activeTimeline.getTopic(), nextTimeline.getTopic());
+                switchTimeline(eventType, nextTimeline, () -> {
+                    final Timeline.StoragePosition storagePosition =
+                            StoragePositionFactory.createStoragePosition(activeTimeline, currentTopicRepo);
+                    activeTimeline.setLatestPosition(storagePosition);
+                    nextTimeline.setSwitchedAt(new Date());
+                });
+            }
+        } catch (final NakadiException ne) {
+            throw new TimelineException(ne.getMessage(), ne);
```
v-stepanov: It's strange that you wrap all Nakadi exceptions to TimelineException.
As I see from the code TimelineException will result in 503 response. But you can get here exceptions because event-type is not found or storage is not found - these all should be 422, not 503. That's why I don't understand this wrapping.

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
adyach: The naming is wrong I will change it. In general this endpoint is done. It creates new timeline and switches event type to that timeline.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,149 @@
...
+        this.eventTypeCache = eventTypeCache;
+        this.storageDbRepository = storageDbRepository;
+        this.timelineSync = timelineSync;
+        this.nakadiSettings = nakadiSettings;
+        this.timelineDbRepository = timelineDbRepository;
+        this.topicRepositoryHolder = topicRepositoryHolder;
+    }
+
+    public void createTimeline(final TimelineRequest timelineRequest, final Client client)
+            throws ForbiddenAccessException, TimelineException, TopicRepositoryException {
```
v-stepanov: Is TopicRepositoryException handled somewhere in controller/spring level?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: Actually, I did it because I do not work with Result class here, because I consider it to be on controller level. As you can see on the https://github.com/zalando/nakadi/pull/562/files/290cc28cc9b8536144084acf52062af393765d08#diff-b4541809d4c98d3235abe7cbfb293ea2R91 I added handling of this exception and it does processing of all 422, 404 etc. from other services.
In my point of view if we have a framework which supports this kind of handling of exceptions why not to use it.

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
```diff
@@ -0,0 +1,56 @@
...
+        return ResponseEntity.status(HttpStatus.CREATED).build();
+    }
+
+    @ExceptionHandler(ForbiddenAccessException.class)
+    public ResponseEntity<Problem> forbidden(final ForbiddenAccessException ex, final NativeWebRequest request) {
+        LOG.error(ex.getMessage(), ex);
+        return Responses.create(Response.Status.FORBIDDEN, ex.getMessage(), request);
+    }
+
+    @ExceptionHandler(NotFoundException.class)
```
v-stepanov: Why you have some ExceptionHandler's in this controller? I thought that we only have a global Handler.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: I will add in controller advise for the next PR

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
adyach: It is not used anywhere except that particular place, that's why I do not see reasons to put in global handler yet.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,149 @@
...
+        if (!client.getClientId().equals(securitySettings.getAdminClientId())) {
+            throw new ForbiddenAccessException("Request is forbidden for user " + client.getClientId());
+        }
+
+        try {
+            final String eventTypeName = timelineRequest.getEventType();
+            final EventType eventType = eventTypeCache.getEventType(eventTypeName);
+
+            final Storage storage = storageDbRepository.getStorage(timelineRequest.getStorageId())
+                    .orElseThrow(() -> new NotFoundException("No storage with id: " + timelineRequest.getStorageId()));
```
v-stepanov: Why have you created some global NotFoundException ?
I think for such situations we use NoSuch...Exception.
Also, with current code that will result in 404 while in other places we usually respond with 422 (e.g. when you create subscription for event-type that is not found)

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,149 @@
...
+        this.securitySettings = securitySettings;
+        this.eventTypeCache = eventTypeCache;
+        this.storageDbRepository = storageDbRepository;
+        this.timelineSync = timelineSync;
+        this.nakadiSettings = nakadiSettings;
+        this.timelineDbRepository = timelineDbRepository;
+        this.topicRepositoryHolder = topicRepositoryHolder;
+    }
+
+    public void createTimeline(final TimelineRequest timelineRequest, final Client client)
```
v-stepanov: Shouldn't this method take some sort of lock for the event-type? Let's imagine that two requests for creation of timeline for the same ET happen at the same time.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
v-stepanov: Ok, i see now where it's unwrapped. But still can't understand why you wrap NakadiException and then unwrap it, while you could just throw it and handle it with handler.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,149 @@
...
+            final TopicRepository currentTopicRepo =
+                    topicRepositoryHolder.getTopicRepository(activeTimeline.getStorage());
+            final TopicRepository nextTopicRepo = topicRepositoryHolder.getTopicRepository(storage);
+            final List<PartitionStatistics> partitionStatistics =
+                    currentTopicRepo.loadTopicStatistics(Collections.singleton(activeTimeline.getTopic()));
+
+            if (activeTimeline.isFake()) {
+                LOG.info("Switching from fake timeline to real one. Current topic {}", activeTimeline.getTopic());
+                final Timeline nextTimeline = Timeline.createTimeline(activeTimeline.getEventType(),
+                        activeTimeline.getOrder() + 1, storage, activeTimeline.getTopic(), new Date());
```
v-stepanov: why it uses the topic name of active timeline? Why it doesn't create a new topic?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,149 @@
...
+                                final Timeline nextTimeline,
+                                final Runnable switcher) throws InterruptedException {
+        try {
+            timelineDbRepository.createTimeline(nextTimeline);
+            timelineSync.startTimelineUpdate(eventType.getName(), nakadiSettings.getTimelineWaitTimeoutMs());
+            switcher.run();
+            timelineDbRepository.updateTimelime(nextTimeline);
+        } catch (final Throwable th) {
+            LOG.error("Failed to switch timeline for event type {}", eventType.getName(), th);
+            timelineDbRepository.deleteTimeline(nextTimeline.getId());
```
v-stepanov: are you sure this is the right thing to do? What will happen if ET is already switched to new TL and then TL is deleted from DB?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: It should be fine, I do not see any problem with locking here, shared resources should be thread safe. In addition to that this is the administrative resource which is used only by our team, I think we can figure out somehow not to switch timeline for the same event type simultaneously :)

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: NoSuch...Exception is checked exception that's why it is not that handy in such cases like lambda. I will make to result in 404. Thanks!

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: This is the case for the fake timeline, it means that there is no real timelines in database for this event type and we can only switch to the same topic with the same offsets.

adyach: In that case we are able to revert a real timeline back to the fake one.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: I do not think that users of that class can really handle checked exceptions and fix the situation somehow (that's why we use checked exceptions), but by throwing them I will force people to wrap them into try catch block, because they expect to recover their state somehow.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: The actual switching happens when we call updateTimelime, that's why it should work. Another thing is that I do not update active timeline and it looks like a bug :)

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
v-stepanov: Ok, so it will switch from fake timeline to real timeline but the actual topic will stay the same, right? Why do we need this step? Can we go from fake timeline directly to real one with different topic?

adyach: Yes, this is right. If we use the new topic for the real timeline then it won't be possible to switch back to fake timeline, because the offsets will be different.
Using fake timeline allows us to think about event type as it does not have any timelines at the start and everything works as before, but it can be switched to real timeline meaning start using timelines.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
v-stepanov: do you plan to fix this bug within this PR?

src/main/java/org/zalando/nakadi/domain/Timeline.java
rcillo: Wouldn't it be better to create timelines for all existent topics?

src/main/java/org/zalando/nakadi/repository/TopicRepositoryHolder.java
```diff
@@ -0,0 +1,102 @@
...
+            readWriteLock.readLock().unlock();
+        }
+
+        readWriteLock.writeLock().lock();
+        try {
+            TopicRepository topicRepository = storageTopicRepository.get(storage);
+            if (topicRepository != null) {
+                return topicRepository;
+            }
+            topicRepository = createTopicRepository(storage);
```
rcillo: Why does calling `getTopicRepository` creates a new repository? Shouldn't creation be done in a separated operation?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,149 @@
...
+    private final EventTypeCache eventTypeCache;
+    private final StorageDbRepository storageDbRepository;
+    private final TimelineSync timelineSync;
+    private final NakadiSettings nakadiSettings;
+    private final TimelineDbRepository timelineDbRepository;
+    private final TopicRepositoryHolder topicRepositoryHolder;
+
+    @Autowired
+    public TimelineService(final SecuritySettings securitySettings,
+                           final EventTypeCache eventTypeCache,
```
rcillo: Cache was designed to be used transparently by client code. There is a `CachingEventTypeRepository` class which encapsulates the cache. Since it's the `@Primary` bean, it's safe to inject `EventTypeRepository` instead of `EventTypeCache`.

api/nakadi-event-bus-api.yaml
```diff
@@ -1269,15 +1269,18 @@ definitions:
         description: configuration settings for kafka storage. Only necessary if the storage type is 'kafka'
         type: object
         properties:
+          exhibitor_address:
+            description: the Zookeeper address
+            type: string
+          exhibitor_port:
+            description: the Zookeeper path
+            type: string
           zk_address:
```
adyach: These are two different things in terms curator framework. Could you take a look at how ZooKeeperHolder.java creates EnsembleProvider? 
It creates different EnsembleProvider depending on what you provide for it.
In case of exhibitor it is hostname of one of the exhibitors, but in case of the zk it is the list of each zookeper. Alos zks are used as a backup connection strategy in case of wrong/unreachable exhibitor address.

src/main/java/org/zalando/nakadi/repository/TopicRepositoryHolder.java
adyach: Topic repository holder is not under review in this PR (you can see it in description of PR, sorry about that, but I wanted to separate the work in favour of lighter reviews).
In my opinion it is fine that topic repository holder (maybe not the best name) creates repositories, but in the future it will know nothing about how to create the repository. You can see related changes in here https://github.com/zalando/nakadi/pull/565/files#diff-a161f5862153864c412c49e86a824adeR25

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: EventTypeCache is used as reading cache only here. The extension of EventTypeRepository to read data will introduce usable only by timeline service proxy methods in CachingEventTypeRepository.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: 72cb0e1

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
adyach: 77415fb

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: 3005e4f

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: 77415fb

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
adyach: 3005e4f

src/main/java/org/zalando/nakadi/domain/Timeline.java
antban: @rcillo As the one who invented this, I will answer. There are several types of timelines (we are considering, that now there are no timelines at all)
1. No timeline for event type. We do not have database entity "Timeline", but in order to unify code - we will generate this fake object. The cursor format for fake timeline will be the same as current one (no version is added).
2. Very first timeline. No new topic is created for this timeline, (so no storage change actually is there), but database object is created, hence offset now includes version. And another thing - this timeline may be deleted, because underlying storage is still the same as in 1. 
3. Usual timeline. The only difference from 2 is that it can not be deleted, because different topic was used

rcillo: @antban @adyach it's clear now. Thanks for the explanation.

src/main/java/org/zalando/nakadi/domain/Timeline.java
```diff
@@ -168,4 +172,22 @@ public boolean equals(final Object o) {
...
     }
+
+    public static Timeline createTimeline(
+            final String eventType,
+            final int order,
+            final Storage storage,
+            final String topic,
+            final Date createdAt) {
+        final Timeline timeline = new Timeline(eventType, order, storage, topic, createdAt);
+        timeline.setId(UUID.randomUUID());
```
antban: We have uuidService

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -174,17 +174,21 @@ public EventTypeValidator getValidator(final String name) throws InternalNakadiE
 
     public Optional<Timeline> getActiveTimeline(final String name) throws InternalNakadiException,
             NoSuchEventTypeException {
-        return getCached(name).map(CachedValue::getTimelines)
-                .orElseThrow(() -> new NoSuchEventTypeException("Event type " + name + " does not exists"))
-                .stream()
+        final List<Timeline> timelines = getTimelines(name);
+        if (timelines == null) {
```
antban: Where is PR Police @v-stepanov ?
`getTimelines(String)` is not nullable!

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,156 @@
...
+
+    public TopicRepository getTopicRepository(final EventType eventType)
+            throws TopicRepositoryException, TimelineException {
+        final Timeline timeline = getTimeline(eventType);
+        return topicRepositoryHolder.getTopicRepository(timeline.getStorage());
+    }
+
+    private void switchTimeline(final EventType eventType,
+                                final Timeline nextTimeline,
+                                final Runnable switcher) {
```
antban: I thing that it's better to check `isFake` second time instead of having callback.

adyach: @antban This is not a callback. This is a function which is called during execution of switchTimeline method.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,156 @@
...
+        try {
+            final String eventTypeName = eventType.getName();
+            final Optional<Timeline> activeTimeline = eventTypeCache.getActiveTimeline(eventTypeName);
+            if (activeTimeline.isPresent()) {
+                return activeTimeline.get();
+            }
+
+            final Storage storage = storageDbRepository.getStorage(DEFAULT_STORAGE)
+                    .orElseThrow(() -> new UnableProcessException("Fake timeline creation failed for event type " +
+                            eventType.getName() + ".No default storage defined"));
```
antban: I don't see any place where DEFAULT_STORAGE is created in DB. We expect that someone will create it for us?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,156 @@
...
+        this.eventTypeCache = eventTypeCache;
+        this.storageDbRepository = storageDbRepository;
+        this.timelineSync = timelineSync;
+        this.nakadiSettings = nakadiSettings;
+        this.timelineDbRepository = timelineDbRepository;
+        this.topicRepositoryHolder = topicRepositoryHolder;
+        this.transactionTemplate = transactionTemplate;
+    }
+
+    public void createTimeline(final TimelineRequest timelineRequest, final Client client)
```
antban: I think that it's better to have eventType parameter and storageId parameter instead of view-level TimelineRequest

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: It has to be created depending on the environment we deploy. I have planned to add a separate task for that.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,156 @@
...
+    public TopicRepository getTopicRepository(final EventType eventType)
+            throws TopicRepositoryException, TimelineException {
+        final Timeline timeline = getTimeline(eventType);
+        return topicRepositoryHolder.getTopicRepository(timeline.getStorage());
+    }
+
+    private void switchTimeline(final EventType eventType,
+                                final Timeline nextTimeline,
+                                final Runnable switcher) {
+        transactionTemplate.execute(status -> {
```
v-stepanov: what this transaction template will do?

adyach: the behaviour is the same as for @Transactional, but the problem is that @Transactional does not work for the methods calls in the same class.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: 69aecb8

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
adyach: 69aecb8

src/main/java/org/zalando/nakadi/domain/Timeline.java
adyach: 69aecb8

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
v-stepanov: I also don't know what `@Transactional` would do here.
Can you please explain that a bit?

adyach: In general, it will just wrap your code in try-catch block and rollback transaction in catch block if exception happens.

v-stepanov: How will it do the rollback?

adyach: You can check out the details here: https://github.com/spring-projects/spring-framework/blob/21ae169653b9da60f9ec8bfe18f5d1bf9e801c94/spring-tx/src/main/java/org/springframework/transaction/support/TransactionTemplate.java#L125

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
antban: @adyach Yes, I agree, and I thought that it will be better not to create this function at all.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
v-stepanov: discussed offline. Now everything is clear.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: 6d68691

v-stepanov:
I'm good with your PR and almost ready to approve

antban:
I'm almost ready to approve PR

antban:
:+1:

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
rcillo: If I understood correctly, Nakadi no longer works without this `DEFAULT_STORAGE` in the database. That means that creating the `DEFAULT_STORAGE` in the database is an additional step for deploying Nakadi successfully.

Considering someone else uses Nakadi as opensource software, this should be documented in the release process otherwise it's just going to break in a unexpected manner.

Why not create this "default storage" directly from application configuration? Then Nakadi would have 2 modes of operation: single timeline (or default) and with multiple timelines.

It would make the release process simple and not force every Nakadi appliance user to run some "curl" during setup.

adyach: Honestly, I have not thought about Nakadi usage outside of the our stack. I will add it.

adyach: @rcillo 0214131

src/main/java/org/zalando/nakadi/repository/tool/DefaultStorage.java
```diff
@@ -0,0 +1,42 @@
...
+        if (!storageOpt.isPresent()) {
+            final Storage storage = new Storage();
+            storage.setId(DEFAULT_STORAGE);
+            storage.setType(Storage.Type.KAFKA);
+            storage.setConfiguration(new Storage.KafkaConfiguration(
+                    environment.getProperty("nakadi.zookeeper.exhibitor.brokers"),
+                    Integer.valueOf(environment.getProperty("nakadi.zookeeper.exhibitor.port", "0")),
+                    environment.getProperty("nakadi.zookeeper.brokers"),
+                    environment.getProperty("nakadi.zookeeper.kafkaNamespace", "")));
+            storageDbRepository.createStorage(storage);
```
rcillo: Is there any problem if an exception is launched here?

adyach: Application won't start, no other problems :)

adyach: Probably, it won't be good if several connections will request creation of the storage and only one will succeeded.

adyach: 8716b80

rcillo:
:shipit:

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,133 @@ paths:
...
+          description: Name of the EventType
+          type: string
+          required: true
+        - name: timeline_request
+          in: body
+          schema:
+            description: Storage id to be used for timeline creation
+            type: object
+            properties:
+              storage_id:
```
v-stepanov: shouldn't `storage_id` be a required property here?

adyach: 76a0039431946f408ec8f62895feb37172ea8700

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -164,14 +164,10 @@ private void switchTimelines(final Timeline activeTimeline, final Timeline nextT
                 Thread.currentThread().interrupt();
                 throw new TimelineException("Failed to switch timeline for:" + activeTimeline.getEventType());
             } catch (final Exception ex) {
+                LOG.error(ex.getMessage(), ex);
```
v-stepanov: I thought we decided to log exception in controller's exception-handler

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -215,12 +211,18 @@ private void switchToFakeTimeline(final UUID uuid, final Timeline activeTimeline
...
+    }
+
+    private void finishTimelineUpdate(final String eventTypeName) throws TimelineException {
+        try {
+            timelineSync.finishTimelineUpdate(eventTypeName);
+        } catch (final InterruptedException ie) {
+            Thread.currentThread().interrupt();
+            throw new TimelineException("Timeline update was interrupted for:" + eventTypeName);
+        } catch (final RuntimeException re) {
+            throw new TimelineException("Timeline update was interrupted for:" + eventTypeName, re);
```
v-stepanov: You use here the same message as for Interrupted exception. Is it ok?

src/main/java/org/zalando/nakadi/service/timeline/TimelineSync.java
```diff
@@ -36,14 +36,15 @@
...
+    void startTimelineUpdate(String eventType, long timeoutMs)
+            throws InterruptedException, IllegalStateException, RuntimeException;
 
     /**
      * Release publishing lock to event type
      *
      * @param eventType Event type to unlock publishing to.
      */
-    void finishTimelineUpdate(String eventType) throws InterruptedException;
+    void finishTimelineUpdate(String eventType) throws InterruptedException, RuntimeException;
```
v-stepanov: I don't think that it makes much sense to declare throwing of general RuntimeException which actually can be thrown by any method.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: Yes, you are right, but this is kind of exception case here: If exception is thrown here and after that another exception is thrown in finally block then we will lose this exception stack trace.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSync.java
adyach: Probably, you are right, RuntimeException has to be replaced with something more meaningful, but this method throws RuntimeException and it happens that I lost this exception, that's why I decided to catch it.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: 1bed903

src/main/java/org/zalando/nakadi/service/timeline/TimelineSync.java
v-stepanov: Ok, then it can make sense as we explicitly throw RuntimeException.
However I think we should throw something more specific, but this discussion is out of this PR scope.

v-stepanov:
👍

v-stepanov:
I approved but you also need green tests :)

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -146,11 +148,18 @@ public TopicRepository getDefaultTopicRepository() throws TopicRepositoryExcepti
...
         LOG.info("Switching timelines from {} to {}", activeTimeline, nextTimeline);
-        transactionTemplate.execute(status -> {
-            try {
+        try {
+            timelineSync.startTimelineUpdate(activeTimeline.getEventType(), nakadiSettings.getTimelineWaitTimeoutMs());
+        } catch (final InterruptedException ie) {
+            Thread.currentThread().interrupt();
+            throw new TimelineException("Failed to switch timeline for:" + activeTimeline.getEventType());
+        } catch (final IllegalStateException ie) {
+            throw new ConflictException("Timeline is being created for:" + activeTimeline.getEventType(), ie);
```
v-stepanov: This sounds like a success message.

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -279,19 +279,23 @@ public void startTimelineUpdate(final String eventType, final long timeoutMs)
...
+        boolean successful = false;
+        try {
+            updateVersionAndWaitForAllNodes(timeoutMs);
+            successful = true;
         } catch (final InterruptedException ex) {
             throw ex;
-        } catch (final Exception e) {
-            throw new RuntimeException(e);
-        } finally {
+        }  finally {
```
v-stepanov: two spaces

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
```diff
@@ -79,4 +80,10 @@ public TimelinesController(final TimelineService timelineService) {
         return Responses.create(Response.Status.NOT_FOUND, ex.getMessage(), request);
     }
 
+    @ExceptionHandler(ConflictException.class)
+    public ResponseEntity<Problem> conflict(final ConflictException ex, final NativeWebRequest request) {
+        LOG.error(ex.getMessage(), ex);
```
v-stepanov: I think it's not an error, it can be logged with "debug" level

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: This is a message details, which tells what is going on with the system right now (the process), but the actually error code tells that it was an error.

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
adyach: ebfd88b

src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
adyach: ebfd88b

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
v-stepanov: For me the message tells that Timeline is being created, which is good - because this is exactly what I wanted to do - to create a timeline :)
I think something like "Timeline is already being created for: " will make more sense;

Another very minor point - in your messages you miss a space after ":"

adyach: c12a1aa

v-stepanov:
👍

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,135 @@ paths:
...
+                  type: string
+                  format: date-time
+                switched_at:
+                  type: string
+                  format: date-time
+                cleaned_up_at:
+                  type: string
+                  format: date-time
+                latest_position:
+                  type: object
```
v-stepanov: why don't you want to specify structure of this object?

antban:
:+1:

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -0,0 +1,257 @@
...
+                nextTimeline = Timeline.createTimeline(activeTimeline.getEventType(),
+                        activeTimeline.getOrder() + 1, storage, activeTimeline.getTopic(), new Date());
+            } else {
+                final String newTopic = nextTopicRepo.createTopic(partitionStatistics.size(),
+                        eventType.getOptions().getRetentionTime());
+                nextTimeline = Timeline.createTimeline(activeTimeline.getEventType(),
+                        activeTimeline.getOrder() + 1, storage, newTopic, new Date());
+            }
+
+            switchTimelines(activeTimeline, nextTimeline);
```
v-stepanov: I'm currently working on another ticket about timelines. I can't understand how the cache is notified that new timeline is created? I could not find where we trigger that.

adyach: @v-stepanov actual notification is sent here
https://github.com/zalando/nakadi/blob/6b2e65b21dd46b414601f2ef14ebf832cd8b3d5d/src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java#L135
we subscribed for timeline changes here:
https://github.com/zalando/nakadi/blob/6b2e65b21dd46b414601f2ef14ebf832cd8b3d5d/src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java#L142
we notify about changes here by TimelineSync.startTimelineUpdate and TimelineSync.finishTimelineUpdate

PR #563, opened 2017-02-16
Aruha 596 cursors compaction

v-stepanov:
@lmontrieux @adyach @antban @rcillo please review
- Added cursors compaction when committing (when there are many cursors for the same partition, only the latest one will be committed)
- Added extra logging to detect the place that consumes a lot of time in a case of commit spikes


src/acceptance-test/java/org/zalando/nakadi/webservice/CursorsServiceAT.java
```diff
@@ -180,6 +174,44 @@ public void whenFirstCursorIsNotCommittedThenNextCursorsAreNotSkipped() throws E
     }
 
     @Test
+    public void whenMultipleCursorsForSamePartitionThenResultsAreCorrect() throws Exception {
+        CURATOR.setData().forPath(offsetPath(P1), "000100".getBytes(UTF_8));
```
adyach: why do not you want to use 18 digits offset, the offset we decided to use in Nakadi?

src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -83,6 +87,10 @@ public CursorsController(final CursorsService cursorsService,
                                            @NotNull @RequestHeader("X-Nakadi-StreamId") final String streamId,
                                            final NativeWebRequest request,
                                            final Client client) {
+
+        LOG.debug("[COMMIT_CURSORS] committing {} cursor(s) for subscription {}", cursors.getItems().size(),
```
adyach: maybe we can add `[COMMIT_CURSORS]` as a part of the logger definition?

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -80,11 +81,37 @@ public CursorsService(final ZooKeeperHolder zkHolder,
 
         validateStreamId(cursors, streamId, subscriptionId);
 
+        LOG.debug("[COMMIT_CURSORS] stream IDs validation finished");
+
+        final Iterator<EventTypePartition> distinctPartitions = cursors.stream()
+                .map(c -> new EventTypePartition(c.getEventType(), c.getPartition()))
+                .distinct()
```
adyach: This looks strange that we can not provide comparator for that and instead we have a separate class to hold the values. I see that EventTypePartition class is used only in CursorsService, do you think it will good to have as inner class?

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -80,11 +81,37 @@ public CursorsService(final ZooKeeperHolder zkHolder,
...
+                .iterator();
+
+        final HashMap<EventTypePartition, Iterator<Boolean>> partitionCommits = new HashMap<>();
+        while (distinctPartitions.hasNext()) {
```
adyach: Usually, iterators are used for changing iterating collection, while you iterate it. Probably it can be replaced with foreach loop.

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -80,11 +81,37 @@ public CursorsService(final ZooKeeperHolder zkHolder,
...
+
+            final EventTypePartition etPartition = distinctPartitions.next();
+            final List<SubscriptionCursor> partitionCursors = cursors.stream()
+                    .filter(etPartition::ownsCursor)
+                    .collect(Collectors.toList());
+
+            final Iterator<Boolean> commitResultIterator = processPartitionCursors(subscriptionId, partitionCursors,
```
adyach: It is not clear why you need iterator here. You get only next() value after all.

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -80,11 +81,37 @@ public CursorsService(final ZooKeeperHolder zkHolder,
 
         validateStreamId(cursors, streamId, subscriptionId);
 
+        LOG.debug("[COMMIT_CURSORS] stream IDs validation finished");
+
+        final Iterator<EventTypePartition> distinctPartitions = cursors.stream()
```
adyach: What do you think about this approach?
1. group by etname_partition -> list of subscriptions
2. group by etname_partition -> max offset for the list of subscriptions

src/acceptance-test/java/org/zalando/nakadi/webservice/CursorsServiceAT.java
v-stepanov: makes sense; fixed that.

src/main/java/org/zalando/nakadi/controller/CursorsController.java
v-stepanov: This controller does not only commits so I would not like to include that to logger itself. And anyway these logs are temporary, they will be removed after the problem with performance is resolved.

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: Why we need a comparator for that?
I think class EventTypePartition can be used in any other place where someone wants a singe entity for event-type/partition pair.

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: Yes, here iterator doesn't make much sense, I replaced with foreach now.

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: Sorry, I didn't get it. Can you please explain more detailed?

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: I need the iterator here because I need to fill the result list in the same order as user provided cursors. For that I need to iterate through cursors and and for each cursor get a result from appropriate list depending on EventTypePartition. Using iterators was the easiest way to implement that because iterators keep iteration pointer.

src/main/java/org/zalando/nakadi/service/CursorsService.java
adyach: I meant that Java does not have an option in distinct method to provide a comparator, it would be very handy to distinct by specific function.

src/main/java/org/zalando/nakadi/service/CursorsService.java
adyach: In this way you can get the map, etname_partition -> cursor with max offset per partition:

`        cursors.stream()
                .collect(Collectors.groupingBy((s -> s.getEventType() + "_" + s.getPartition()))).entrySet().stream()
                .map(entry -> cursors.stream().max(Comparator.comparingLong(sc -> Long.valueOf(sc.getOffset()))));`

After that you can just commit these cursors.

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: I see.

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: Yes I wanted to do it like that in the beginning , in this case the implementation would be much more simpler. But the tricky point here is that we need to return response for each cursor. So if I would just commit the max and will ignore the others I will not know which result to return for each cursor. For example if user provides these cursors for a partition: `24, 25, 27, 10, 35, 30` we should respond with `true, true, true, false, true, false`

In this test you can see which cases are possible https://github.com/zalando/nakadi/pull/563/files#diff-d92c44fce8107aa3413fffc22d08eed9R175

adyach: Okay, got it! 
It looks like you go through the whole list of cursors every time. It makes more sense then to pass the preprocessed map to next method calls.
cursors.stream().collect(Collectors.groupingBy((s -> s.getEventType() + "_" + s.getPartition())))
In that case you will just grab the list for particular event type partition and work with it: compare offsets and commit the largest one. It seems legitimate since the order should be correct per partition.

adyach: What do you think about it?

v-stepanov: Yes, thanks! I changed that, it simplified the code. I was not aware of this Collectors.groupingBy(..) this was actually what I needed here.

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -144,62 +165,86 @@ private boolean isActiveSession(final String subscriptionId, final String stream
...
-                        // Yep, here we are trying to hack a little. This code
-                        // should be removed during timelines implementation
-                        final NakadiCursor cursorToCommit = new NakadiCursor(eventType, cursor.getPartition(),
-                                cursor.getOffset());
-                        final NakadiCursor currentCursor = new NakadiCursor(eventType, cursor.getPartition(),
+                        NakadiCursor currentMaxCursor = new NakadiCursor(topic, etPartition.getPartition(),
                                 currentOffset);
-                        if (topicRepository.compareOffsets(cursorToCommit, currentCursor) > 0) {
+
+                        final List<Boolean> commits = Lists.newArrayList();
```
adyach: Having `committed` property in NakadiCursor can help to avoid creating list with bools. Instead of it, creating the map of cursor->NakadiCursor and returning in back to commitCursors will help to avoid creation of EventTypePartition for every cursor, when preparing CursorCommitResult.

src/main/java/org/zalando/nakadi/controller/CursorsController.java
antban: @v-stepanov I expected to see stream id here as well

src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -92,12 +99,16 @@ public CursorsController(final CursorsService cursorsService,
...
+            LOG.debug("[COMMIT_CURSORS] scopes validation finished");
+
+            final List<CursorCommitResult> items = cursorsService.commitCursors(streamId, subscriptionId,
                     cursors.getItems());
-            final List<CursorCommitResult> items = result.entrySet().stream()
-                    .map(entry -> new CursorCommitResult(entry.getKey(), entry.getValue()))
-                    .collect(Collectors.toList());
-            final boolean allCommited = result.values().stream().reduce(Boolean::logicalAnd).orElseGet(() -> true);
+
+            LOG.debug("[COMMIT_CURSORS] commit finished");
```
antban: I don't like it. We need to create something that will generate one line with timings, or we don't need it at all.

src/main/java/org/zalando/nakadi/domain/CursorCommitResult.java
```diff
@@ -0,0 +1,47 @@
...
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        final CursorCommitResult that = (CursorCommitResult) o;
+        return cursor.equals(that.cursor) && result.equals(that.result);
+    }
+
+    @Override
+    public int hashCode() {
+        int result1 = cursor.hashCode();
```
antban: Maybe result2 is a better fit? :)

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -144,62 +165,86 @@ private boolean isActiveSession(final String subscriptionId, final String stream
...
+                        if (Cursor.BEFORE_OLDEST_OFFSET.equals(cursor.getOffset())) {
+                            cursorToProcess = new SubscriptionCursor(cursor.getPartition(), "-1", cursor.getEventType(),
+                                    cursor.getCursorToken());
+                        }
+                        final NakadiCursor nakadiCursor = new NakadiCursor(
+                                eventType.getTopic(),
+                                cursorToProcess.getPartition(),
+                                cursorToProcess.getOffset());
+                        try {
+                            topicRepository.validateCommitCursor(nakadiCursor);
```
antban: Because this method is one that expects a lot of load, I think that it's better not to call kafka several times for list of partitions (here we call listPartitions() each time when validating nakadiCursor). And this call is actually not needed, because partition is correct (as it is present in zk node). So in the end - we are doing many calls that are not needed.

src/main/java/org/zalando/nakadi/controller/CursorsController.java
v-stepanov: Each log message already has a timestamp. 
Also, all these "[COMMIT_CURSORS] ..." logs are temporary and will be removed after the performance problem is resolved.

src/main/java/org/zalando/nakadi/domain/CursorCommitResult.java
v-stepanov: Fixed :)

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: I don't think we should have that in NakadiCursor. In most cases where NakadiCursor is used it's not about committing, so it will always have some additional field that is not used.

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: I changed cursors validation now. Now it only checks the format of offset and the null values. Partitions are not checked.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -429,11 +429,7 @@ public int compareOffsets(final NakadiCursor first, final NakadiCursor second) t
 
     @Override
     public void validateCommitCursor(final NakadiCursor position) throws InvalidCursorException {
```
adyach: Could we fix `position`? 
You call validateCursorForNulls, but after you call KafkaCursor.fromNakadiCursor, which basically does the same checks. From other side it is not clear why you call KafkaCursor.fromNakadiCursor and do not work with the KafkaCursor.

v-stepanov: Yes, true. There's no need to call validateCursorForNulls here. I removed it now.
 I call KafkaCursor.fromNakadiCursor for validation purposes (it will throw InvalidCursorException if something is wrong), so I don't really need Kafka cursors that it returns.

antban:
:+1:

antban:
Even if I don't like additinal logs

adyach:
:shipit:

PR #564, opened 2017-02-17
ARUHA-668 Enable auto commit to reproduce live incident

lmontrieux:
This PR is NOT to be merged. I just need to re-enable auto offset commit to try and reproduce the incident conditions.


v-stepanov:
👍

PR #565, opened 2017-02-20
aruha 527: topic repository refactoring

adyach:
This PR will bring refactoring of topic repository replacing it with the timeline service all over the project.


src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
```diff
@@ -358,6 +365,7 @@ public void whenExceedMaxConsumersNumThen429() throws IOException, InterruptedEx
                 break;
             }
         }
+        range(0, 15).forEach(value -> NakadiTestUtils.publishEvent(etName, "{\"foo\": \"please, close connection\"}"));
```
antban: @adyach ?

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -193,7 +201,7 @@ public StreamingResponseBody streamEvents(
                         .withStreamKeepAliveLimit(streamKeepAliveLimit)
                         .withEtName(eventTypeName)
                         .withConsumingAppId(client.getClientId())
-                        .withCursors(getStreamingStart(topic, cursorsStr))
+                        .withCursors(getStreamingStart(topicRepository, topic, cursorsStr))
```
antban: Cursors here might be in different timeline...

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -98,7 +103,9 @@ public EventStreamController(final EventTypeRepository eventTypeRepository, fina
     }
 
     @VisibleForTesting
-    List<NakadiCursor> getStreamingStart(final String topic, final String cursorsStr)
+    List<NakadiCursor> getStreamingStart(final TopicRepository topicRepository,
+                                         final String topic,
+                                         final String cursorsStr)
```
antban: In case of begin the very first timeline should be used, in case of cursors absence - the very last one

src/main/java/org/zalando/nakadi/controller/PartitionsController.java
```diff
@@ -46,12 +49,12 @@ public PartitionsController(final EventTypeRepository eventTypeRepository, final
         LOG.trace("Get partitions endpoint for event-type '{}' is called", eventTypeName);
         try {
             final EventType eventType = eventTypeRepository.findByName(eventTypeName);
-
+            final TopicRepository topicRepository = timelineService.getTopicRepository(eventType);
```
antban: And here the same - the begin may be in one timeline, and end - in another.

src/main/java/org/zalando/nakadi/repository/TopicRepositoryHolder.java
```diff
@@ -1,51 +1,41 @@
...
-        this.nakadiSettings = nakadiSettings;
-        this.kafkaSettings = kafkaSettings;
-        this.zookeeperSettings = zookeeperSettings;
-        this.uuidGenerator = uuidGenerator;
-        this.metricRegistry = metricRegistry;
+    public TopicRepositoryHolder(@Qualifier("kafka") final TopicRepositoryCreator kafkaRepository) {
         this.readWriteLock = new ReentrantReadWriteLock();
         this.storageTopicRepository = new HashMap<>();
+        this.repositoryCreators = new HashMap<>();
+        this.repositoryCreators.put(Storage.Type.KAFKA, kafkaRepository);
```
antban: I would prefer to see method getSupportedStorageType() in TopicRepositoryCreator.

antban:
@adyach In general I like PR, but can you also create tickets for places that are working with cursors so we can support multiple timelines in different tasks.
For example - `/event-types/{event_type}/partitions` may work with several timelines, and we need to explicitly support it.

src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
adyach: 0f35cd7

src/main/java/org/zalando/nakadi/repository/TopicRepositoryHolder.java
adyach: ffee833

adyach:
@antban added FIXME TIMELINE todo.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -181,6 +190,7 @@ public StreamingResponseBody streamEvents(
                 client.checkScopes(eventType.getReadScopes());
 
                 // validate parameters
+                final TopicRepository topicRepository = timelineService.getTopicRepository(eventType);
```
rcillo: Shouldn't topic be retrieved based on `event type + cursor`? I mean, even if you don't plan to implement it now, since this is a refactoring, maybe it would be a good idea to define an API closer to the final version, something like `topicRepositoryPool.get(eventType, cursor)` that returns something like a map from cursor to repository. Wdyt?

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -159,14 +159,17 @@ private boolean processCursor(final String subscriptionId, final SubscriptionCur
...
     }
 
-    private boolean commitCursor(final String subscriptionId, final String eventType, final SubscriptionCursor cursor)
+    private boolean commitCursor(final String subscriptionId,
+                                 final EventType eventType,
+                                 final SubscriptionCursor cursor)
             throws ServiceUnavailableException, NoSuchSubscriptionException, InvalidCursorException {
 
-        final String offsetPath = format(PATH_ZK_OFFSET, subscriptionId, eventType, cursor.getPartition());
+        final String topic = eventType.getTopic();
```
rcillo: This is something that maybe doesn't make sense anymore. Here we can see that `event type` is coupled to the `kafka topic name`. But since this relation is made through a timeline now, it seems that the proper way to get this topic is through `event type + cursors` from the timelines.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
adyach: I wanted to make the transition smoother and implement only the part for the changes in topic repository in that way we can test it on live not introducing too much changes. If I am going to create topic repository for each cursor it will touch consumption as well. From the other side it will be simpler to implement consumption later If I have the right api method now, but this is not so many changes.

adyach: In general, i think it makes sense, but I did not aim it in this task.

rcillo: @adyach ok. I think it's a good idea to deliver it part by part. Let's revisit this when implementing consumer.

src/main/java/org/zalando/nakadi/service/CursorsService.java
adyach: I fixed it here aae9086, but again this is not fully correct approach to do it. The timeline has to be retrieved according to offset. I checked that EventType.getTopic is used in a lot of places, for example EventTypeController.delete, there should be removed topics for all timelines. In my opinion it has to addressed in the following PRs.

src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
```diff
@@ -312,6 +318,7 @@ public void whenReadEventsForBlockedConsumerThen403() throws Exception {
                 .statusCode(HttpStatus.OK.value());
 
         SettingsControllerAT.blacklist(EVENT_TYPE_NAME, BlacklistService.Type.CONSUMER_ET);
+        Thread.sleep(100);
```
adyach: f8210a9

rcillo:
If I understood the scope correctly, this PR is fine. It's not introducing any new features. It's just retrieving repositories through timelines. In the future, it will be required to extend this interface with cursors.

rcillo:
👍

rcillo:
👍

PR #561, opened 2017-02-14
ARUHA-554 Introduce zero-padded offsets

antban:

PR #568, opened 2017-02-21
ARUHA-554 Introduce zero-padded offsets

antban:
Initially it was there: https://github.com/zalando/nakadi/pull/561


v-stepanov:
👍

PR #574, opened 2017-02-23
aruha-527: delete and list timelines

adyach:
aruha-527: delete and list timelines


api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,112 @@ paths:
...
+          schema:
+            description: Event type and storage id to be used for timeline creation.
+            type: object
+            properties:
+              event_type:
+                type: string
+              strorage_id:
+                type: string
+          required: true
+      responses:
```
lmontrieux: Should we add a response if the event type and/or storage id don't exist?

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,112 @@ paths:
...
+      parameters:
+        - name: timeline_request
+          in: body
+          schema:
+            description: Event type and storage id to be used for timeline creation.
+            type: object
+            properties:
+              event_type:
+                type: string
+              strorage_id:
```
lmontrieux: `strorage_id` -> `storage_id`

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,112 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+  /timelines:
+    post:
+      tags:
+        - timelines-api
+      description: |
+        Creates timeline and switches to it.
```
lmontrieux: I would say something like 'Creates a new timeline for an event type, and switches event production to the new timeline.'

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,112 @@ paths:
...
+          description: Access forbidden
+          schema:
+            $ref: '#/definitions/Problem'
+
+  /timelines/{id}#1:
+    delete:
+      tags:
+        - timelines-api
+      description: |
+        Deletes a timeline in case it is the first one.
```
lmontrieux: If I were someone who hasn't read the code, I think I would not really understand this.
* the first timeline of what?
* what happens if the timeline is not first?

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,112 @@ paths:
...
+        Method call is quite dangerous, because it will remove any timeline.
```
lmontrieux: This somehow contradicts the previous line. In the previous line you say the timeline is deleted in case it is the first one. Here you say it will be deleted no matter what

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,112 @@ paths:
...
+              type: object
+              properties:
+                id:
+                  type: string
+                  format: uuid
+                event_type:
+                  type: string
+                order:
+                  type: integer
+                strorage_id:
```
lmontrieux: `strorage_id` -> `storage_id`

src/main/java/org/zalando/nakadi/repository/TopicRepository.java
```diff
@@ -17,10 +12,14 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
 public interface TopicRepository {
 
-    String createTopic(int partitionCount, Long retentionTimeMs)
-            throws TopicCreationException, DuplicatedEventTypeNameException;
+    String createTopic(int partitionCount, Long retentionTimeMs) throws TopicCreationException;
```
lmontrieux: Of course this make sense, since the implementation of this interface uses `TopicCreationException` in case of duplicate topics, but I wonder why this is in a PR about timeline lists and deletion?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -165,4 +177,35 @@ private void switchTimelines(final EventType eventType,
         });
     }
 
+    public void delete(final String id, final Client client) {
```
lmontrieux: I think we agreed to explicitely declare unchecked exceptions that we throw?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -165,4 +177,35 @@ private void switchTimelines(final EventType eventType,
...
+        final UUID uuid = uuidGenerator.fromString(id);
+        final Optional<Timeline> timelineOpt = timelineDbRepository.getTimeline(uuid);
+        if (!timelineOpt.isPresent()) {
+            throw new UnableProcessException("Timeline with id: " + uuid + " not found");
+        }
+
+        timelineDbRepository.deleteTimeline(uuid);
+    }
+
+    public List<TimelineView> getTimelines(final String eventTypeName, final Client client) {
```
lmontrieux: Same comment about `throws`

src/main/java/org/zalando/nakadi/repository/TopicRepository.java
adyach: I found that it is wrong and fixed it.

adyach: It is all about one big task of timelines

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: 1db5405

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: 1db5405

api/nakadi-event-bus-api.yaml
adyach: 1db5405

api/nakadi-event-bus-api.yaml
adyach: 1db5405

api/nakadi-event-bus-api.yaml
adyach: 1db5405

api/nakadi-event-bus-api.yaml
adyach: 1db5405

api/nakadi-event-bus-api.yaml
adyach: 1db5405

src/main/java/org/zalando/nakadi/view/TimelineView.java
```diff
@@ -0,0 +1,67 @@
...
+public class TimelineView {
+
+    private final UUID id;
+    private final String eventType;
+    private final int order;
+    private final String storageId;
+    private final String topic;
+    private final Date createdAt;
+    private final Date switchedAt;
+    private final Date cleanupAt;
```
lmontrieux: You have `created` and `switched` (both verbs in past participle form), but here `cleanup`, which is a noun. Maybe use `cleanedUpAt`? (because it would be `cleaned up`)

lmontrieux: Of course this comment is also valid for `Timeline` and the API spec

adyach: 07fb77d

lmontrieux: 👍 . There are a few occurrences of `cleanupAt`, but they're only in tests, so I think this is fine.

lmontrieux:
:shipit:

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,115 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+  /timelines:
```
antban: I thought it will be better to have timelines under event type - like `/event-types/{event-type}/timelines`

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,115 @@ paths:
...
+          description: Access forbidden
+          schema:
+            $ref: '#/definitions/Problem'
+
+  /timelines/{id}#1:
+    delete:
+      tags:
+        - timelines-api
+      description: |
+        Deletes a timeline by its id. Method call is quite dangerous, because it will remove any timeline.
```
antban: I do not agree with any timeline. It should delete only first timeline if it is last at the same time.

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,115 @@ paths:
...
+        '422':
+          description: Unprocessable entity due to non existing timeline
+          schema:
+            $ref: '#/definitions/Problem'
+        '403':
+          description: Access forbidden
+          schema:
+            $ref: '#/definitions/Problem'
+
+  /timelines/{event_type}#2:
```
antban: I do not like `#1` and `#2`

api/nakadi-event-bus-api.yaml
adyach: I had only one point to make timeline api separate from event types: It is administrative api, which does not intersects with the event types api, that was an initial reason for that.

api/nakadi-event-bus-api.yaml
adyach: Good point 👍  In general, it was a workaround since swagger does not differentiate the same path with the different methods.

api/nakadi-event-bus-api.yaml
adyach: 83a370d

api/nakadi-event-bus-api.yaml
adyach: 83a370d

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,123 @@ paths:
...
+        '422':
+          description: Unprocessable entity due to non existing timeline
+          schema:
+            $ref: '#/definitions/Problem'
+        '403':
+          description: Access forbidden
+          schema:
+            $ref: '#/definitions/Problem'
+
+  /event-types/{name}/timelines#get:
```
antban: was ist '#get'?

adyach: 1fdbb9a

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -152,17 +162,76 @@ private void switchTimelines(final EventType eventType,
...
+            try {
+                timelineSync.finishTimelineUpdate(activeTimeline.getEventType());
+            } catch (final InterruptedException ie) {
+                Thread.currentThread().interrupt();
+                throw new TimelineException("Timeline update was interrupted for:" + activeTimeline.getEventType());
+            }
+        }
+    }
+
+    public List<TimelineView> getTimelines(final String eventTypeName, final Client client)
```
antban: Generating view on service level

adyach: 78edca0

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,121 @@ paths:
...
+              storage_id:
+                type: string
+          required: true
+      responses:
+        '201':
+          description: New timeline is created and in use.
+        '403':
+          description: Access forbidden
+          schema:
+            $ref: '#/definitions/Problem'
```
antban: There is no 422

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,121 @@ paths:
...
+                  type: string
+                created_at:
+                  type: string
+                  format: date-time
+                switched_at:
+                  type: string
+                  format: date-time
+                cleanedup_at:
+                  type: string
+                  format: date-time
```
antban: I think that this value should not be exposed now...

antban: It is not clear if it is more related to implementation, rather than to user view of it

api/nakadi-event-bus-api.yaml
```diff
@@ -1245,6 +1245,121 @@ paths:
...
+                latest_position:
+                  type: string
```
antban: Maybe list of cursor makes more sense?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -152,17 +160,74 @@ private void switchTimelines(final EventType eventType,
...
 
+    public void delete(final String eventTypeName, final String timelineId, final Client client)
+            throws ForbiddenAccessException, UnableProcessException, TimelineException {
+        if (!client.getClientId().equals(securitySettings.getAdminClientId())) {
+            throw new ForbiddenAccessException("Request is forbidden for user " + client.getClientId());
+        }
+
+        final UUID uuid = uuidGenerator.fromString(timelineId);
+        final List<Timeline> timelines = timelineDbRepository.listTimelines(eventTypeName);
+        if (timelines.size() != DELETE_ONLY_FIRST_TIMELINE) {
```
antban: I think that having constant for that makes a lot of sense.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -152,17 +160,74 @@ private void switchTimelines(final EventType eventType,
...
+            timelineDbRepository.deleteTimeline(uuid);
+        } catch (final InterruptedException ie) {
+            Thread.currentThread().interrupt();
+            throw new TimelineException("Failed to switch timeline for:" + activeTimeline.getEventType());
+        } finally {
+            try {
+                timelineSync.finishTimelineUpdate(activeTimeline.getEventType());
+            } catch (final InterruptedException ie) {
+                Thread.currentThread().interrupt();
+                throw new TimelineException("Timeline update was interrupted for:" + activeTimeline.getEventType());
```
antban: This may eat original exception

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -317,19 +317,6 @@ public void whenPostDuplicatedEventTypeReturn409() throws Exception {
...
-        final Problem expectedProblem = Problem.valueOf(Response.Status.CONFLICT, "dummy message");
-        final EventType et = buildDefaultEventType();
-        doReturn(et).when(eventTypeRepository).saveEventType(any(EventType.class));
-
-        Mockito.doThrow(new DuplicatedEventTypeNameException("dummy message")).when(topicRepository)
-                .createTopic(anyInt(), any());
-
-        postEventType(et).andExpect(status().isConflict()).andExpect(content().contentType("application/problem+json"))
-                .andExpect(content().string(matchesProblem(expectedProblem)));
-    }
```
antban: Why you removed this test?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
antban: I was wrong, in this case it makes no sense

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
adyach: There is no need in it anymore, since TopicRepository does not throw DuplicatedEventTypeNameException.

api/nakadi-event-bus-api.yaml
adyach: 6338b79

api/nakadi-event-bus-api.yaml
adyach: I think it is good to have it for debug purpose

api/nakadi-event-bus-api.yaml
adyach: This is a json object, fixed type 6338b79

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: 6338b79

antban:
:+1:

PR #576, opened 2017-02-24
ARUHA-657 Fix committing to subscription for closed stream

antban:
It is possible to have situation, when stream is closing, and at the same time there is a task for streaming data to client in subscription queue. In this case streaming session was closed without waiting for uncommitted events. This fix moves calculation of uncommitted events to the very end of StreamingState (so no further writes may occur after this calculation).


src/acceptance-test/java/org/zalando/nakadi/webservice/SettingsControllerAT.java
```diff
@@ -58,6 +59,8 @@ public void testGetFlooders() throws Exception {
         final EventType eventType = NakadiTestUtils.createEventType();
         blacklist(eventType.getName(), BlacklistService.Type.CONSUMER_ET);
 
+        Thread.sleep(TimeUnit.SECONDS.toMillis(1)); // give time for TreeCache refresh
```
rcillo: I thought @adyach had improved this test on another PR. He implemented some code that waits for a specific condition, instead of adding a generic `sleep`.

antban: But I want my tests to pass too.

rcillo: @antban maybe you could do like him https://github.com/zalando/nakadi/pull/562/files

adyach: https://github.com/zalando/nakadi/pull/562/files#diff-e66b120a1c73878f7df2b10f74871df0L61

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
```diff
@@ -73,6 +74,24 @@ public void before() throws IOException {
...
+    public void whenStreamTimeoutReachedPossibleToCommit() throws Exception {
+        publishEvent(eventType.getName(),"{\"foo\":\"bar\"}");
+        publishEvent(eventType.getName(),"{\"foo\":\"bar\"}");
+        final TestStreamingClient client = TestStreamingClient
+                .create(URL, subscription.getId(), "stream_limit=1&stream_timeout=1")
+                .start();
+        waitFor(() -> Assert.assertFalse(client.getBatches().isEmpty()), TimeUnit.SECONDS.toMillis(1), 100);
+        final SubscriptionCursor toCommit = client.getBatches().get(0).getCursor();
+        client.close(); // connection is closed, and stream as well
+
```
rcillo: Don't you want to wait a little bit to make sure the connection close is finished?

rcillo: Or maybe query for the metrics that keep track of the amount of consumer connected. When it reaches 0 then proceed.

adyach: Or even try to publish something, which should trigger IOException and close the connection from Nakadi side

src/main/java/org/zalando/nakadi/service/subscription/state/ClosingState.java
```diff
@@ -39,9 +43,10 @@ public void onExit() {
 
     @Override
     public void onEnter() {
-        final long timeToWaitMillis = getParameters().commitTimeoutMillis - (System.currentTimeMillis()
-                - lastCommitMillis);
-        if (timeToWaitMillis > 0) {
+        final long timeToWaitMillis = getParameters().commitTimeoutMillis -
+                (System.currentTimeMillis() - lastCommitSupplier.getAsLong());
+        uncommittedOffsets = uncommittedOffsetsSupplier.get();
+        if (!uncommittedOffsets.isEmpty() && timeToWaitMillis > 0) {
```
rcillo: Is this new check for `!uncommittedOffsets.isEmpty()` part of the bug fix or an optimization?

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -38,6 +38,8 @@
     private long committedEvents;
     private long sentEvents;
     private long batchesSent;
+    // Uncommitted offsets are calculated right on exiting from Streaming state.
+    private Map<Partition.PartitionKey, Long> uncommittedOffsets;
```
rcillo: Maybe you could initialise it with an empty `Map` so that there will be no need for null checks.

rcillo:
👍

src/main/java/org/zalando/nakadi/service/subscription/state/ClosingState.java
```diff
@@ -39,9 +43,10 @@ public void onExit() {
 
     @Override
     public void onEnter() {
-        final long timeToWaitMillis = getParameters().commitTimeoutMillis - (System.currentTimeMillis()
-                - lastCommitMillis);
-        if (timeToWaitMillis > 0) {
+        final long timeToWaitMillis = getParameters().commitTimeoutMillis -
```
adyach: It is not clear why use function here, but if it is a case then it is better to move this lines to constructor and make vars final.

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
antban: close() finishes when connection is closed. Connection is closed when subscription finished traveling thru all the steps. I do not see any reason to wait for a little bit more

src/acceptance-test/java/org/zalando/nakadi/webservice/SettingsControllerAT.java
antban: Ok, I will just remove this line and try to push "build" on travis several time.

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
rcillo: @antban client connection finishes but server connection could still be open, not?

src/main/java/org/zalando/nakadi/service/subscription/state/ClosingState.java
antban: Well, it is exactly the cause of the bug. State travels thru following steps:
1. ctor
1. onEnter
1. onExit

The one that makes sense for us is `onEnter` because all the preconditions for this state should be calculated in that moment. That is why it is wrong to calculate something that dynamically changed on `ctor` stage.

src/main/java/org/zalando/nakadi/service/subscription/state/ClosingState.java
antban: It is part of bugfix. Initially this check was in `StreamingState`, now it was moved to ClosingState.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
antban: I do not check it for nulls anywhere in this case.
Because it is stated, that every state will go thru `onExit` and `onEnter` calls. From the other hand, if I will initialize this field with something incorrect (for example empty), than someone who use it will be able to treat it incorrectly. It is better to return value that will lead to exception than.

rcillo: Please mark getter as `@Nullable` so.

rcillo: Have you considered raising null when calling get?

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
antban: @rcillo in this case - no.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
antban: After some conversation we decided to use Preconditions.notNull on getter.

antban: https://github.com/zalando/nakadi/pull/576/commits/e3fadc93e26ff513876e5021ceb4ee5ecbc65231

adyach:
:shipit:

v-stepanov:
👍

v-stepanov:
👍

adyach:
:shipit:

PR #577, opened 2017-02-24
aruha-592 collect metrics for all requests individually

rcillo:
This PR is part of the task https://techjira.zalando.net/browse/ARUHA-592

Here I'll collect metrics for all requests. So to have timers for every path, like:

- `GET/subscriptions/a7671c51-49d1-48e6-bb03-b50dcf14f3d3/events`
- `POST/subscriptions/a7671c51-49d1-48e6-bb03-b50dcf14f3d3/cursors`
- `GET/event-types/my-event-type/partitions`

After that, I'll build a dashboard called `Nakadi Police` where I'll rank these entries per rate usage per functionality.

This way we will be able to find very quickly which subscription is committing so much, which event type partitions is called like crazy, which event type is slow to publish, which subscription is slow to commit.

I thought about appending/prepending the name of the application to the metric. So we could very fast also spot who is abusing the service or is been impacted by degraded service quality. Wdyt? Something like:

- `application-name.GET/event-types/my-event-type/events`

### DIFF
--- a/src/main/java/org/zalando/nakadi/config/MetricsConfig.java
+++ b/src/main/java/org/zalando/nakadi/config/MetricsConfig.java
@@ -1,0 +1,67 @@
+package org.zalando.nakadi.config;
+
+import com.codahale.metrics.MetricRegistry;
+import com.codahale.metrics.jvm.BufferPoolMetricSet;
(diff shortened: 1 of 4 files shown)


rcillo:
I'm closing this PR in favor of a new one with even more metrics. Stay tuned.

PR #578, opened 2017-02-28
ARUHA-592: faceted metrics

rcillo:
https://techjira.zalando.net/browse/ARUHA-592

Aruha team used to have a hard time to quickly answer questions about specific application performance.

Questions like:

- Who consumes my event types and how much?
- How many active consumer a given application has at the moment?
- How fast are Kafka consumers streaming?

In order not to grow even more our existent `/metrics` endpoint, I'm created 2 new endpoints:

1. /requests-metrics
- request timer on a per application basis for every URI path:  application-id.http-method.http-path
- counter for number of consumers from a given subscription: nakadi.subscriptions.subscription-id.consumers
2. /stream-metrics
- it exposes Kafka consumer metrics for HiLA: hila.application-id.subscription-id.bytes-flushed
- it exposes Kafka consumer metrics for LoLA: lola.application-id.event-type.bytes-flushed

With all these metrics, we are going to be able to detect slowdowns specific to certain applications, subscriptions or event types. Also, we will be able to detect who's abusing the system, by calling a given endpoint too often or streaming too much.

All of this was only possible previously through log analyses.


src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -174,6 +182,12 @@ public StreamingResponseBody streamEvents(
             final AtomicBoolean connectionReady = closedConnectionsCrutch.listenForConnectionClose(request);
             Counter consumerCounter = null;
             EventStream eventStream = null;
+            EventConsumer eventConsumer = null;
+            final String kafkaClientMetricsName = MetricRegistry.name(
+                    "lola",
+                    client.getClientId(),
+                    eventTypeName,
```
rcillo: Maybe I should remove `.` from event type name?

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -124,12 +132,17 @@ public StreamingResponseBody streamEvents(
...
 
         return outputStream -> {
+            Counter consumerCounter = null;
 
             if (!featureToggleService.isFeatureEnabled(HIGH_LEVEL_API)) {
                 response.setStatus(HttpServletResponse.SC_NOT_IMPLEMENTED);
                 return;
             }
 
+            final String metricName = metricNameForSubscription(subscriptionId, CONSUMERS_COUNT_METRIC_NAME);
```
rcillo: Maybe I should also scope this metric per `application-id`?

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -254,6 +273,10 @@ public StreamingResponseBody streamEvents(
                 if (eventStream != null) {
                     eventStream.close();
                 }
+                if (eventConsumer != null) {
+                    this.kafkaClientMetrics.removeMatching((name, metric) -> {
+                        return name.startsWith(kafkaClientMetricsName); });
```
adyach: No need in `return`

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
adyach: Do you consider using MetricUtils to generate the name? It replaces `.` in event type name with `#`.

src/main/java/org/zalando/nakadi/metrics/KafkaClientMetrics.java
```diff
@@ -0,0 +1,30 @@
...
+    public KafkaClientMetrics(final Consumer<String, String> kafkaConsumer) {
+        this.kafkaConsumer = kafkaConsumer;
+    }
+
+    @Override
+    public Map<String, Metric> getMetrics() {
+        final Map<String, Metric> gauges = new HashMap<>();
+        for (final Object metricObject : kafkaConsumer.metrics().entrySet()) {
+            final Map.Entry<MetricName, KafkaMetric> metricEntry = (Map.Entry<MetricName, KafkaMetric>) metricObject;
+            gauges.put(metricEntry.getKey().name(), (Gauge) () -> { return metricEntry.getValue().value(); });
```
adyach: No need in `return`

src/main/java/org/zalando/nakadi/metrics/KafkaClientMetrics.java
```diff
@@ -0,0 +1,30 @@
...
+    private final Consumer<String, String> kafkaConsumer;
+
+    public KafkaClientMetrics(final Consumer<String, String> kafkaConsumer) {
+        this.kafkaConsumer = kafkaConsumer;
+    }
+
+    @Override
+    public Map<String, Metric> getMetrics() {
+        final Map<String, Metric> gauges = new HashMap<>();
+        for (final Object metricObject : kafkaConsumer.metrics().entrySet()) {
```
adyach: You could use Map.Entry<MetricName, ? extends org.apache.kafka.common.Metric>

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
adyach: I think we can add it in the feature, if there is a need.

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -124,12 +132,17 @@ public StreamingResponseBody streamEvents(
             throws IOException {
 
         return outputStream -> {
+            Counter consumerCounter = null;
```
adyach: You can move it to the metric construction (down there) and make final.

src/main/java/org/zalando/nakadi/metrics/MonitoringRequestFilter.java
```diff
@@ -34,11 +41,28 @@ public void doFilter(final ServletRequest request, final ServletResponse respons
         openHttpConnectionsCounter.inc();
         final Timer.Context timerContext = httpConnectionsTimer.time();
 
+        Timer.Context perPathTimerContext = null;
+        if (request instanceof HttpServletRequest) {
+            final HttpServletRequest httpRequest = (HttpServletRequest) request;
+            final String clientId = Optional.ofNullable(httpRequest.getUserPrincipal())
+                    .map(Principal::getName)
+                    .orElse("unauthenticated");
```
adyach: Is it useful to gather metrics for unauthenticated users? and what can we gather?

adyach:
It looks good 👍 except tests are not passing yet and some minor things I've noticed

src/main/java/org/zalando/nakadi/metrics/MonitoringRequestFilter.java
rcillo: for example, `/health` and `/metrics` endpoint. We might realise that they are called too often for example.

rcillo:
@adyach fixed comments and simplified metrics. Now there is going to be a `/stream-metrics` endpoint with stream metrics. It's not bound to Kafka anymore.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -219,8 +225,18 @@ public StreamingResponseBody streamEvents(
                 final EventConsumer eventConsumer = topicRepository.createEventConsumer(
                         kafkaQuotaClientId,
                         streamConfig.getCursors());
+
+                final String bytesFlushedMetricName = MetricRegistry.name(
```
adyach: Do you think it makes sense to use it as in subscriptions through MetricUtils?

rcillo: This name looks like "lola.stups_some-application-name.some#event-ype-name.bytes-flushed". The existent methods in `MetricUtils` looks like "nakadi.subscriptions.subscription-id.bytes-flushed". It has 2 problems:

1. it has no space for application-id
2. it has no space for lola/hila.

I could create new methods there, to have it all in a single class, though. Is this what you meant?

adyach:
👍

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
rcillo: @adyach done 92cc082

rcillo:
@adyach I need a new approval after I refactored `MetricUtils` a little bit.

adyach:
👍

rcillo:
@adyach I used spring filters and it seems to have solved. I tried locally with authentication turned on.

rcillo:
the thing about it is that it no longer registers the time taken to authenticate, obviously.

src/main/java/org/zalando/nakadi/metrics/MonitoringRequestFilter.java
```diff
@@ -3,61 +3,56 @@
...
-        }
+        final Timer.Context perPathTimerContext = perPathMetricRegistry.timer(perPathMetricKey).time();
 
         try {
 
-            chain.doFilter(request, response);
+            filterChain.doFilter(request, response);
 
         } finally {
             if (perPathTimerContext != null) {
```
adyach: Is it possible that it is `null` here?

rcillo: Not at all. I forgot to remove it after refactoring. Done https://github.com/zalando/nakadi/pull/578/commits/c06e4871d24dd3e67f4a6d41f22f5b01ab27b1b7

adyach:
👍

adyach:
:shipit:

PR #579, opened 2017-02-28
Aruha 591 start from cursors

v-stepanov:
https://techjira.zalando.net/browse/ARUHA-591

I also did a lot of refactoring, sorry :(


src/test/java/org/zalando/nakadi/service/SubscriptionValidationServiceTest.java
```diff
@@ -0,0 +1,218 @@
...
+                new SubscriptionCursorWithoutToken(ET2, P0, ""),
+                new SubscriptionCursorWithoutToken(ET3, P0, "a"),
+                new SubscriptionCursorWithoutToken(ET3, P0, "b")
+        ));
+        try {
+            subscriptionValidationService.validateSubscription(subscriptionBase, client);
+            fail("WrongInitialCursorsException expected");
+        } catch (final WrongInitialCursorsException e) {
+            assertThat(e.getMessage(),
+                    equalTo("there should be not more than 1 cursor for each partition in initial_cursors"));
```
rcillo: small suggestion `not` -> `no`.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -334,30 +340,77 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
...
-                    consumer::position));
-        } catch (final Exception e) {
-            throw new ServiceUnavailableException("Error occurred when fetching partitions offsets", e);
         }
+    }
 
+    private Map<String, Long> getSubscriptionTopicInitPositions(final String topic,
+                                                                final List<SubscriptionCursorWithoutToken> etCursors) {
+        final Map<String, Long> positions = new HashMap<>();
+        final List<SubscriptionCursorWithoutToken> cursorsWithBegin = etCursors.stream()
```
rcillo: Should you also consider cursors with `END`?

v-stepanov: It's not possible to have END in a cursor, we don't have such a value.
The only place we use "end" is a "read_from" property of subscription.

src/main/java/org/zalando/nakadi/controller/PostSubscriptionController.java
```diff
@@ -0,0 +1,119 @@
...
+
+import static org.springframework.http.HttpStatus.NOT_IMPLEMENTED;
+import static org.springframework.http.HttpStatus.OK;
+import static org.zalando.nakadi.util.FeatureToggleService.Feature.CHECK_OWNING_APPLICATION;
+import static org.zalando.nakadi.util.FeatureToggleService.Feature.DISABLE_SUBSCRIPTION_CREATION;
+import static org.zalando.nakadi.util.FeatureToggleService.Feature.HIGH_LEVEL_API;
+
+
+@RestController
+public class PostSubscriptionController {
```
v-stepanov: The controller became too big, and the unit test for controller was just huge that's why I decided to separate that. Also I wanted to start this new life with runtime exceptions - it's easier to do that in separate class.

src/main/java/org/zalando/nakadi/controller/ExceptionHandling.java
```diff
@@ -85,4 +87,18 @@ private String generateErrorTraceId() {
...
     }
+
+    @ExceptionHandler(RepositoryProblemException.class)
+    public ResponseEntity<Problem> handleRepositoryProblem(final RepositoryProblemException exception,
+                                                           final NativeWebRequest request) {
+        LOG.error("Repository problem occurred", exception);
+        return Responses.create(Response.Status.SERVICE_UNAVAILABLE, exception.getMessage(), request);
+    }
+
+    @ExceptionHandler(InconsistentStateException.class)
```
v-stepanov: I was considering two reasons to have this:
1) I want to log the exception explicitly.
2) I want to expose to user only the exception message, not the stacktrace.

src/main/java/org/zalando/nakadi/controller/PostSubscriptionController.java
```diff
@@ -0,0 +1,119 @@
...
+        this.subscriptionService = subscriptionService;
+    }
+
+    @RequestMapping(value = "/subscriptions", method = RequestMethod.POST)
+    public ResponseEntity<?> createOrGetSubscription(@Valid @RequestBody final SubscriptionBase subscriptionBase,
+                                                     final Errors errors,
+                                                     final NativeWebRequest request,
+                                                     final Client client) {
+        if (!featureToggleService.isFeatureEnabled(HIGH_LEVEL_API)) {
+            return new ResponseEntity<>(NOT_IMPLEMENTED);
```
v-stepanov: Yes, we still need that. We are considering temporary switching off HILA when migrating new new versions of cursors and when decoupling zookeepers.

src/test/java/org/zalando/nakadi/service/SubscriptionValidationServiceTest.java
v-stepanov: fixed https://github.com/zalando/nakadi/pull/579/commits/f5ca984cfcbf6088876b314afd1bd49a79db564f

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -334,30 +340,77 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
...
+                        .stream()
+                        .map(p -> new org.apache.kafka.common.TopicPartition(eventType.getTopic(), p.partition()))
+                        .toArray(org.apache.kafka.common.TopicPartition[]::new);
+                consumer.assign(Arrays.asList(kafkaTPs));
+                if (position == SubscriptionBase.InitialPosition.BEGIN) {
+                    consumer.seekToBeginning(kafkaTPs);
+                } else if (position == SubscriptionBase.InitialPosition.END) {
+                    consumer.seekToEnd(kafkaTPs);
+                } else {
+                    throw new IllegalArgumentException("Bad offset specification " + position + " for topic " +
```
v-stepanov: Yeah I don't really know why we have this check, I removed it now https://github.com/zalando/nakadi/pull/579/commits/9f92f45884426ac890b7128164a629d1828f3c81

src/main/java/org/zalando/nakadi/exceptions/runtime/MyNakadiRuntimeException1.java
```diff
@@ -0,0 +1,13 @@
+package org.zalando.nakadi.exceptions.runtime;
+
+public class MyNakadiRuntimeException1 extends RuntimeException {
```
v-stepanov: added comment :) https://github.com/zalando/nakadi/pull/579/commits/9f92f45884426ac890b7128164a629d1828f3c81

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: true 👍 I forgot about it.

src/main/java/org/zalando/nakadi/controller/PostSubscriptionController.java
v-stepanov: Oh yeah, and another reason was that the unit test for SubscriptionController is testing controller together with service class. I wanted to fix this at least for creation of subscription: to have a separate test for controller and separate test for service class.

src/main/java/org/zalando/nakadi/controller/ExceptionHandling.java
rcillo: Maybe it would be better to always log the stacktrace but to the user just send a generic error message: internal server error. To the user it doesn't matter so much if it's the database or a bug in the code. At the end, the only thing the user can do is to retry later.

rcillo: But, I'm fine with this approach.

rcillo:
:shipit:

src/main/java/org/zalando/nakadi/controller/ExceptionHandling.java
adyach: I agree that we should use more generic approach, otherwise we will have a lot of methods like that one. For example, RepositoryProblemException and InconsistentStateException are the same to methods to handle it.

v-stepanov: For me it's difficult to say right now. It was a first step to a "new life" so I wanted to make everything very explicit here. Maybe when we work in this way some more time we will see that it's better to have some more general handler for 500-resulting exceptions

v-stepanov: @adyach I think that we can just reuse this method for all other 500-resulting exceptions by just adding exception class to `@ExceptionHandler` annotation. (And of course we will change parameter type to MyNakadiRuntimeException1)

adyach: @v-stepanov if you want to have some specific exceptions processed in specific way you can have exception handler on the controller level. Another way is to define global exception handler for MyNakadiRuntimeException1. Wdyt?

v-stepanov: Ok I changed it now to have a global handler for MyNakadiRuntimeException1 returning 500. I think this is what we agreed to on our meeting.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -334,30 +340,74 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
     }
 
     @Override
-    public Map<String, Long> materializePositions(final String topicId, final SubscriptionBase.InitialPosition position)
```
adyach: Do you think that subscription entity belongs to the KafkaTopicRepository class?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -334,30 +340,74 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
...
+    public Map<String, Long> materializePositions(final EventType eventType, final Subscription subscription)
```
adyach: Do you think that subscription entity belongs to the KafkaTopicRepository class?

v-stepanov: No, it doesn't belong, but I see no problems to use Subscription on repository layer. Subscription belongs to domain layer and I think we can use it in repositories.

adyach: I agree that it belongs to the domain layer. In my opinion it has to be used in database repository layer and service layer which works with them, but topic repository slightly different thing which abstracts messaging backend system and probably, it should know nothing about Subscriptions.

adyach:
Good job integrating RuntimeExceptions 👍

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
v-stepanov: No, I think it works only in opposite direction. Service layer should not know anything about specific repository implementation. But repository should use domain objects, in other case it's not clear how we can use repositories at all.

v-stepanov: Well, as discussed offline @antban will anyway refactor this method. So I will leave it as it is.

v-stepanov:
@rcillo @adyach please re-approve. I fixed checkstyle problem.

adyach:
:shipit:

rcillo:
👍

rcillo:
👍

PR #581, opened 2017-03-03
ARUHA-530 Inject Timelines into NakadiCursor

antban:
https://techjira.zalando.net/browse/ARUHA-684
Refactored NakadiCursor to contain timeline. 
Use the same conversion method from and to NakadiCursor everywhere in the code.


src/main/java/org/zalando/nakadi/service/CursorConverter.java
```diff
@@ -1,46 +1,28 @@
...
-    public SubscriptionCursor convert(
-            final NakadiCursor position, final String eventType, final String token) {
-        final Cursor oldCursor = convert(position);
-        return new SubscriptionCursor(oldCursor.getPartition(), oldCursor.getOffset(), eventType, token);
-    }
+    // Convert from view to model
+    NakadiCursor convert(String eventTypeName, Cursor cursor) throws
+            InternalNakadiException, NoSuchEventTypeException, InvalidCursorException, ServiceUnavailableException;
 
+    NakadiCursor convert(SubscriptionCursorWithoutToken c) throws
```
v-stepanov: maybe it's better to rename it to "cursor"

src/main/java/org/zalando/nakadi/service/converter/CursorConverterImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * Method tries to get version of cursor. If version can not be restored, than {@link NakadiCursor.Version#ZERO}
+     * will be returned
+     *
+     * @param offset Offset to guess version from
+     * @return Version of offset.
+     */
+    @VisibleForTesting
```
v-stepanov: I think this is an indicator that something is not right. For example there could be a separate class that will guess versions.

src/main/java/org/zalando/nakadi/service/converter/VersionOneConverter.java
```diff
@@ -0,0 +1,72 @@
...
+    public NakadiCursor convert(final String eventTypeStr, final Cursor cursor)
+            throws InternalNakadiException, NoSuchEventTypeException, InvalidCursorException {
+        final String orderStr = cursor.getOffset().substring(
+                NakadiCursor.VERSION_LENGTH, NakadiCursor.VERSION_LENGTH + TIMELINE_ORDER_LENGTH);
+        final int order = Integer.parseInt(orderStr, TIMELINE_ORDER_BASE);
+        final List<Timeline> timelines = eventTypeCache.getTimelines(eventTypeStr);
+        final EventType eventType = eventTypeCache.getEventType(eventTypeStr);
+        final String realOffset = cursor.getOffset().substring(NakadiCursor.VERSION_LENGTH + TIMELINE_ORDER_LENGTH);
+        if (timelines.isEmpty()) {
+            // Here one have problems. Timeline probably was there some time ago, but it was rolled back.
```
v-stepanov: has

src/main/java/org/zalando/nakadi/service/converter/VersionedConverter.java
```diff
@@ -0,0 +1,24 @@
...
+
+    NakadiCursor.Version getVersion();
+    /**
+     * Converts cursor to nakadi cursor having in mind that preliminary checks for format were performed.
+     *
+     * @param eventTypeStr Name of event type
+     * @param cursor       cursor to convert.
+     * @return NakadiCursor instance.
+     */
+    NakadiCursor convert(final String eventTypeStr, final Cursor cursor) throws
```
v-stepanov: it doesn't make sense to define parameters final in an interface.

src/main/java/org/zalando/nakadi/service/converter/VersionedConverter.java
```diff
@@ -0,0 +1,24 @@
...
+            InternalNakadiException, NoSuchEventTypeException, ServiceUnavailableException, InvalidCursorException;
+
+    String formatOffset(final NakadiCursor nakadiCursor);
```
v-stepanov: and here too

src/main/java/org/zalando/nakadi/service/subscription/KafkaClient.java
```diff
@@ -1,57 +1,88 @@
...
-                       final TimelineService timelineService,
-                       final EventTypeRepository eventTypeRepository) {
+    public KafkaClient(final Subscription subscription, final TimelineService timelineService,
+                       final CursorConverter cursorConverter)
+            throws InternalNakadiException, NoSuchEventTypeException {
         this.subscription = subscription;
-        this.eventTypeRepository = eventTypeRepository;
+        this.timelineService = timelineService;
+        this.cursorConverter = cursorConverter;
         // FIXME TIMELINE: for refactoring purposes, has to be removed during timeline event consumption task
```
v-stepanov: is it fixed now? Can this comment be removed?

src/main/java/org/zalando/nakadi/service/subscription/KafkaClient.java
```diff
@@ -1,57 +1,88 @@
...
+                    }
+                    break;
+                case END:
+                    for (String eventType : subscription.getEventTypes()) {
+                        final Timeline timeline = timelineService.getActiveTimelinesOrdered(eventType).get(0);
+                        timelineService.getTopicRepository(timeline)
+                                .loadTopicStatistics(Collections.singletonList(timeline))
+                                .forEach(stat -> existingPositions.add(stat.getLast()));
+                    }
+                    break;
```
v-stepanov: branches for BEGIN and END are almost identical. I think the code can be improved to avoid duplication. For example, to check if it's BEGIN or END only when you do `stat.getBeforeFirst()` or `stat.getLast()`

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -80,7 +79,7 @@ public SubscriptionService(final SubscriptionDbRepository subscriptionRepository
     }
 
     public Subscription createSubscription(final SubscriptionBase subscriptionBase, final Client client)
-            throws DuplicatedSubscriptionException, TooManyPartitionsException, RepositoryProblemException,
+            throws TooManyPartitionsException, RepositoryProblemException, DuplicatedSubscriptionException,
```
v-stepanov: didn't you like the original order :)

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionValidationService.java
```diff
@@ -100,25 +102,16 @@ private void validateInitialCursors(final SubscriptionBase subscription,
...
-
         try {
-            // FIXME TIMELINES: has to be fixed during consumption task
-            final TopicRepository topicRepository = timelineService.getTopicRepository(eventTypes.get(0));
-            topicRepository.validateReadCursors(cursorsWithoutBegin);
-        } catch (InvalidCursorException e) {
-            throw new WrongInitialCursorsException(e.getMessage(), e);
-        } catch (ServiceUnavailableException e) {
-            throw new RepositoryProblemException("Topic repository problem occurred when validating cursors", e);
+            for (SubscriptionCursorWithoutToken cursor : subscription.getInitialCursors()) {
```
v-stepanov: final

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -129,12 +148,26 @@ public Timeline getTimeline(final EventTypeBase eventType) throws TimelineExcept
         }
     }
 
+    public Timeline getFakeTimeline(final EventType eventType) {
+        try {
+            final Storage storage = storageDbRepository.getStorage(DEFAULT_STORAGE)
+                    .orElseThrow(() -> new UnableProcessException("No default storage defined"));
```
v-stepanov: I think we agreed that method should define which exceptions it throws. 
Also, I'm not sure if UnableProcessException is a right exception to throw here, for me it's more like InconsistentStateException as we could not find default storage which should always be there.

v-stepanov:
@antban please insert link to ticket into PR description (currently zappr fails to approve PR sepcification in your PR)

src/acceptance-test/java/org/zalando/nakadi/webservice/CursorsServiceAT.java
```diff
@@ -173,14 +156,14 @@ public void whenETRepoExceptionThenException() throws Exception {
...
+        final NakadiCursor c2 = new NakadiCursor(timeline, P2, NEW_OFFSET);
         testCursors = ImmutableList.of(c1, c2);
 
-        final List<CursorCommitResult> result = cursorsService.commitCursors(streamId, sid, testCursors);
+        final List<Boolean> result = cursorsService.commitCursors(streamId, sid, testCursors);
 
-        assertThat(result.get(0).getResult(), equalTo(CursorCommitResult.OUTDATED));
-        assertThat(result.get(1).getResult(), equalTo(CursorCommitResult.COMMITTED));
+        assertEquals(Boolean.FALSE, result.get(0));
+        assertEquals(Boolean.TRUE, result.get(1));
```
v-stepanov: if you don't use hamcrest-style anyway, you can just use here assertTrue and assertFalse

src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -73,7 +83,10 @@ public CursorsController(final CursorsService cursorsService,
         }
         try {
             validateSubscriptionReadScopes(client, subscriptionId);
-            final List<SubscriptionCursor> cursors = cursorsService.getSubscriptionCursors(subscriptionId);
+            final List<SubscriptionCursor> cursors = cursorsService.getSubscriptionCursors(subscriptionId)
+                    .stream()
+                    .map(cursor -> cursorConverter.convert(cursor, cursorTokenService.generateToken()))
```
v-stepanov: Do you think it's ok to generate these tokens on controller level?

src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -99,25 +112,31 @@ public CursorsController(final CursorsService cursorsService,
...
-                    .allMatch(cursor -> CursorCommitResult.COMMITTED.equals(cursor.getResult()));
-            final ItemsWrapper<CursorCommitResult> body = new ItemsWrapper<>(items);
-            return allCommited ? noContent().build() : ok(body);
-        } catch (final NoSuchEventTypeException e) {
+            final boolean allCommited = items.stream().allMatch(item -> item);
+            if (allCommited) {
+                return noContent().build();
+            } else {
+                final List<CursorCommitResult> body = IntStream.range(0, cursors.getItems().size())
+                        .mapToObj( idx -> new CursorCommitResult(cursors.getItems().get(idx), items.get(idx)))
```
v-stepanov: I think it's possible to do that more elegant by streaming one list and using iterator of another list in that stream.

src/main/java/org/zalando/nakadi/domain/NakadiCursor.java
```diff
@@ -1,21 +1,43 @@
...
 
+import com.google.common.base.Preconditions;
 import java.util.Objects;
 
 public class NakadiCursor {
-    private final String topic;
+    public static final int VERSION_LENGTH = 3;
+    public enum Version {
+        ZERO("000"),
+        ONE("C00"),;
```
v-stepanov: I'm really still not sure if we should start it with "C00". Comparing BEGIN to other cursors is something wrong. Also, BEGIN is not always the smallest offset. For example when you start stream from BEGIN for some partition - this BEGIN is bigger than all offsets that are expired for this partition.

src/main/java/org/zalando/nakadi/domain/NakadiCursor.java
```diff
@@ -1,21 +1,43 @@
...
+            Preconditions.checkArgument(
+                    code.length() == VERSION_LENGTH,
+                    "Version field length should be equal to " + VERSION_LENGTH);
+            this.code = code;
+        }
+    }
+
+    private final Timeline timeline;
     private final String partition;
     // NO BEGIN OR END HERE!
```
v-stepanov: We actually don't have any END for offset. I think it's better to remove it from this comment not to confuse people.

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -97,9 +97,13 @@ private void preloadEventTypes(final EventTypeRepository eventTypeRepository,
         try {
             final Map<String, List<Timeline>> eventTypeTimelines = timelineRepository.listTimelines().stream()
                     .collect(Collectors.groupingBy(Timeline::getEventType));
+            final List<Timeline> emptyList = Collections.unmodifiableList(new ArrayList<>());
```
v-stepanov: `ImmutableList.of();`

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -185,7 +189,6 @@ public EventTypeValidator getValidator(final String name) throws InternalNakadiE
                 .max(Comparator.comparing(Timeline::getOrder));
     }
 
-    @Nullable
```
v-stepanov: very good, very good

src/main/java/org/zalando/nakadi/repository/kafka/KafkaCursor.java
```diff
@@ -33,12 +34,14 @@ public KafkaCursor addOffset(final long toAdd) {
...
-        return new NakadiCursor(topic, toNakadiPartition(partition), toNakadiOffset(offset));
+    public NakadiCursor toNakadiCursor(final Timeline timeline) {
+        return new NakadiCursor(timeline,
+                toNakadiPartition(partition),
+                toNakadiOffset(offset));
     }
 
     public static String toNakadiOffset(final long offset) {
-        return String.valueOf(offset);
+        return offset >= 0 ? String.format("%018d", offset) : String.valueOf(offset);
```
v-stepanov: You do this `String.format("%018d", offset)` in many places. Maybe it will make sense to make some helper static method that does it with respect to the length constant we have for cursors with leading zeroes.

src/main/java/org/zalando/nakadi/service/CursorConverter.java
```diff
@@ -1,46 +1,28 @@
...
-    }
-
-    public SubscriptionCursor convert(
-            final NakadiCursor position, final String eventType, final String token) {
-        final Cursor oldCursor = convert(position);
-        return new SubscriptionCursor(oldCursor.getPartition(), oldCursor.getOffset(), eventType, token);
-    }
+    // Convert from view to model
+    NakadiCursor convert(String eventTypeName, Cursor cursor) throws
+            InternalNakadiException, NoSuchEventTypeException, InvalidCursorException, ServiceUnavailableException;
```
v-stepanov: I'm a little bit worried that we still use checked exceptions. But I understand that you already did a lot of changes with this PR, so maybe it's better to leave it as it is for now.

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -167,52 +159,42 @@ private boolean isActiveSession(final String subscriptionId, final String stream
...
                         } catch (final InvalidCursorException e) {
                             throw new NakadiRuntimeException(e);
                         }
                     })
                     .collect(Collectors.toList());
 
-            LOG.debug("[COMMIT_CURSORS] finished validation of {} cursor(s) for partition {}", cursors.size(),
-                    eventTypePartition);
+            LOG.debug("[COMMIT_CURSORS] finished validation of {} cursor(s) for partition {} {}", cursors.size(),
+                    cursors.get(0).getTimeline().getEventType(), cursors.get(0).getTopic());
```
v-stepanov: `cursors` can be empty (at least we don't check anywhere that it's not empty)

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -167,52 +159,42 @@ private boolean isActiveSession(final String subscriptionId, final String stream
...
-    private List<Boolean> commitPartitionCursors(final String subscriptionId,
-                                                 final EventType eventType,
-                                                 final List<NakadiCursor> cursors,
-                                                 final EventTypePartition etPartition)
+    private List<Boolean> commitPartitionCursors(final String subscriptionId, final List<NakadiCursor> cursors)
             throws ServiceUnavailableException {
-        final String topic = eventType.getTopic();
-        final String offsetPath = format(PATH_ZK_OFFSET, subscriptionId, topic, etPartition.getPartition());
-        final TopicRepository topicRepository = timelineService.getTopicRepository(eventType);
+        final NakadiCursor first = cursors.get(0);
```
v-stepanov: `cursors` can be empty (at least we don't check anywhere that it's not empty)

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -167,52 +159,42 @@ private boolean isActiveSession(final String subscriptionId, final String stream
...
+        final String offsetPath = format(
+                PATH_ZK_OFFSET, subscriptionId, first.getTopic(), first.getPartition());
+        final TopicRepository topicRepository = timelineService.getTopicRepository(first.getTimeline());
```
v-stepanov: what if offsets of different cursors will belong to different timelines?

src/test/java/org/zalando/nakadi/service/converter/VersionOneConverterTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+        final NakadiCursor nakadiCursor = converter.convert(eventTypeName, cursor);
+        Assert.assertEquals(fakeTimeline, nakadiCursor.getTimeline());
+        Assert.assertEquals("1", nakadiCursor.getPartition());
+        Assert.assertEquals(String.format("%018d", 12345), nakadiCursor.getOffset());
+    }
+
+    @Test
+    public void testInvalidCursorExceptionOnNotExistentTimeline() throws Exception {
+        final Cursor cursor = new Cursor("1", "C000002012345");
```
v-stepanov: I thought we will have fancy dashes like "C00-0002-012345" :(

src/main/java/org/zalando/nakadi/service/converter/VersionOneConverter.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @Override
+    public NakadiCursor.Version getVersion() {
+        return NakadiCursor.Version.ONE;
+    }
+
+    public NakadiCursor convert(final String eventTypeStr, final Cursor cursor)
+            throws InternalNakadiException, NoSuchEventTypeException, InvalidCursorException {
+        final String orderStr = cursor.getOffset().substring(
+                NakadiCursor.VERSION_LENGTH, NakadiCursor.VERSION_LENGTH + TIMELINE_ORDER_LENGTH);
```
v-stepanov: how do you handle "BEGIN" here?

src/main/java/org/zalando/nakadi/service/converter/VersionOneConverter.java
```diff
@@ -0,0 +1,72 @@
...
+        final int order = Integer.parseInt(orderStr, TIMELINE_ORDER_BASE);
+        final List<Timeline> timelines = eventTypeCache.getTimelines(eventTypeStr);
+        final EventType eventType = eventTypeCache.getEventType(eventTypeStr);
+        final String realOffset = cursor.getOffset().substring(NakadiCursor.VERSION_LENGTH + TIMELINE_ORDER_LENGTH);
```
v-stepanov: I think that will make sense to catch StringIndexOutOfBoundsException and throw InvalidCursorException here

src/test/java/org/zalando/nakadi/service/converter/VersionOneConverterTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.zalando.nakadi.domain.Timeline;
+import org.zalando.nakadi.exceptions.InvalidCursorException;
+import org.zalando.nakadi.repository.db.EventTypeCache;
+import org.zalando.nakadi.service.timeline.TimelineService;
+import org.zalando.nakadi.view.Cursor;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class VersionOneConverterTest {
```
v-stepanov: no tests for incorrect format like "C000043" or "C00fjur48rre64545"

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -111,6 +113,23 @@ public void createTimeline(final String eventTypeName, final String storageId, f
...
     }
 
+    /**
+     * Returns list of ACTIVE timelines for event type.
+     * @param eventType
+     * @return list of active timelines. List is always NOT empty! At least fake timeline present there.
+     * @throws InternalNakadiException everything can happen
+     * @throws NoSuchEventTypeException No such event type
+     */
+    public List<Timeline> getActiveTimelinesOrdered(final String eventType)
```
v-stepanov: I don't see a test for this new method

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -129,12 +148,26 @@ public Timeline getTimeline(final EventTypeBase eventType) throws TimelineExcept
         }
     }
 
+    public Timeline getFakeTimeline(final EventType eventType) {
```
v-stepanov: no test for this method

src/test/java/org/zalando/nakadi/service/SubscriptionValidationServiceTest.java
```diff
@@ -179,6 +181,8 @@ public void whenInvalidCursorThenException() throws Exception {
                 new SubscriptionCursorWithoutToken(ET2, P0, ""),
                 new SubscriptionCursorWithoutToken(ET3, P0, "")
         ));
+        final NakadiCursor tmp = mock(NakadiCursor.class);
```
v-stepanov: maybe it's better to give it a nicer name, I believe `tmp` is a very bad name.

src/test/java/org/zalando/nakadi/service/SubscriptionValidationServiceTest.java
```diff
@@ -191,25 +195,12 @@ public void whenServiceUnavailableThenException() throws Exception {
...
```
v-stepanov: and here too

src/main/java/org/zalando/nakadi/service/subscription/KafkaClient.java
```diff
@@ -1,57 +1,88 @@
...
-            throw new NakadiRuntimeException(e);
-        }
+        this.topicRepository = timelineService.getTopicRepository(
+                timelineService.getActiveTimelinesOrdered(subscription.getEventTypes().iterator().next()).get(0));
         // FIXME TIMELINE: for refactoring purposes, has to be removed during timeline event consumption task
     }
 
-    public Map<Partition.PartitionKey, Long> getSubscriptionOffsets() {
-        final Map<Partition.PartitionKey, Long> offsets = new HashMap<>();
+    public Map<Partition.PartitionKey, NakadiCursor> getSubscriptionOffsets() {
```
v-stepanov: I'm missing test for this method

v-stepanov:
Great refactoring 👍  This PR simplifies a lot.

src/main/java/org/zalando/nakadi/controller/CursorsController.java
antban: @v-stepanov Unless it is needed on service level - it will be generated there

src/main/java/org/zalando/nakadi/controller/CursorsController.java
antban: @v-stepanov I don't think that having separate iterator without checking hasNext() is more elegant way...

src/main/java/org/zalando/nakadi/domain/NakadiCursor.java
antban: @v-stepanov Let's discuss it offline

src/main/java/org/zalando/nakadi/domain/NakadiCursor.java
antban: @v-stepanov ok

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
antban: @v-stepanov ok

src/main/java/org/zalando/nakadi/repository/kafka/KafkaCursor.java
antban: @v-stepanov ok, will do that

src/main/java/org/zalando/nakadi/service/CursorConverter.java
antban: @v-stepanov That is exactly what I thought

src/main/java/org/zalando/nakadi/service/CursorConverter.java
antban: ok

src/main/java/org/zalando/nakadi/service/CursorsService.java
antban: @v-stepanov According to the way they are collected - it can not be empty.

src/main/java/org/zalando/nakadi/service/CursorsService.java
antban: @v-stepanov According to the way they are collected - it can not be empty.

src/main/java/org/zalando/nakadi/service/CursorsService.java
antban: @v-stepanov This PR is part of LOLA timelines part. When one will get to Hila part, then we will support it

src/main/java/org/zalando/nakadi/service/converter/CursorConverterImpl.java
antban: @v-stepanov In this case instead of annotation `@VisibleForTesting` one will hack and introduce public class used only to test only one method. I do not want to separate logic because of unit tests

src/main/java/org/zalando/nakadi/service/converter/VersionOneConverter.java
antban: @v-stepanov It is version one converter, here it should not be handled.

src/main/java/org/zalando/nakadi/service/converter/VersionedConverter.java
antban: ok

src/main/java/org/zalando/nakadi/service/converter/VersionedConverter.java
antban: ok

src/main/java/org/zalando/nakadi/service/subscription/KafkaClient.java
antban: No, it is still there

src/main/java/org/zalando/nakadi/service/subscription/KafkaClient.java
antban: @v-stepanov Yep, and there is a bug there...

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
antban: @v-stepanov rebase problems

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionValidationService.java
antban: @v-stepanov probably one should reconfigure our code checking...

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
antban: @v-stepanov I will check how this method was created and used. Maybe it is some artifact from rebase and can be safely removed

src/test/java/org/zalando/nakadi/service/converter/VersionOneConverterTest.java
antban: We still can use it, but it won't help in any way.

src/main/java/org/zalando/nakadi/service/subscription/KafkaClient.java
antban: @v-stepanov Fixed one bug, so they are more different now.

src/main/java/org/zalando/nakadi/service/converter/CursorConverterImpl.java
v-stepanov: Decoupling is not a hack, it's a good approach that will allow you to have good code that can be easily tested.
Having package-level method instead of private method just to be able to test it -> this is a hack :)

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
antban: Wil change it to IllegalStateException

src/acceptance-test/java/org/zalando/nakadi/webservice/CursorsServiceAT.java
antban: ok

src/main/java/org/zalando/nakadi/controller/CursorsController.java
v-stepanov: You anyway do the same by calling items.get(idx), so the code "safeness" will not change if using solution with iterator, but the code will be more laconic.
Also, you have extra space in that line so you will have to change it anyway :)

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: agree

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: agree

src/main/java/org/zalando/nakadi/service/CursorsService.java
v-stepanov: maybe then it's better to put some sort of TODO here

src/main/java/org/zalando/nakadi/service/converter/VersionOneConverter.java
v-stepanov: where is it handled then?

src/test/java/org/zalando/nakadi/service/converter/VersionOneConverterTest.java
antban: added tests

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
antban: I don't see any reason to make test for one if cause and one sort method.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
antban: The same here - test only for orElseThrow condition ?

src/main/java/org/zalando/nakadi/service/converter/CursorConverterImpl.java
antban: @v-stepanov decoupling is something to loose coupling. I do not see it here at all. 
Having class with name `CursorConverterVisibleForTesting` (and the name is correct, because it is class purpose) it a bad approach.

src/main/java/org/zalando/nakadi/service/CursorsService.java
antban: ok

src/main/java/org/zalando/nakadi/service/converter/CursorConverterImpl.java
v-stepanov: Well, I was thinking about a class named `CursorVersionGuesser`.
But anyway, we already discussed this offline, so I can live with current code.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -282,7 +282,7 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
     @Override
     public Optional<PartitionStatistics> loadPartitionStatistics(final Timeline timeline, final String partition)
             throws ServiceUnavailableException {
-        try (final Consumer<String, String> consumer = kafkaFactory.getConsumer()) {
+        try (Consumer<String, String> consumer = kafkaFactory.getConsumer()) {
```
v-stepanov: why you removed final?

antban:
The same reason as public and final in interfaces - resource in try with
resource is final by default, and it is not possible to change it.

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -106,15 +99,14 @@ public CursorsService(final ZooKeeperHolder zkHolder,
...
                             cursor.getPartition());
-                    final Boolean committed = partitionCommits.get(etPartition).next();
-                    return new CursorCommitResult(cursor, committed);
+                    return partitionCommits.get(etPartition).next();
                 })
                 .collect(Collectors.toList());
     }
 
-    private void validateStreamId(final List<SubscriptionCursor> cursors, final String streamId,
+    private void validateStreamId(final List<NakadiCursor> cursors, final String streamId,
```
adyach: On line 133:
partitionSession = getPartitionSession(subscriptionId, eventType.getTopic(), cursor);
 you take topic from event type. I think it does not make much sense.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -111,6 +112,24 @@ public void createTimeline(final String eventTypeName, final String storageId, f
...
+     * @throws InternalNakadiException  everything can happen
+     * @throws NoSuchEventTypeException No such event type
+     */
+    public List<Timeline> getActiveTimelinesOrdered(final String eventType)
+            throws InternalNakadiException, NoSuchEventTypeException {
+        final List<Timeline> timelines = eventTypeCache.getTimelines(eventType);
+        if (timelines.isEmpty()) {
+            return Collections.singletonList(getFakeTimeline(eventTypeCache.getEventType(eventType)));
+        } else {
+            return timelines.stream().sorted(Comparator.comparing(Timeline::getOrder)).collect(Collectors.toList());
```
adyach: It should be already ordered for you, since it requested from DB by t.tl_order.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaCursor.java
antban: @v-stepanov Actually now it is the only place

src/main/java/org/zalando/nakadi/service/CursorsService.java
antban: ok, will fix

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
antban: I will rename method for getting timelines

src/main/java/org/zalando/nakadi/domain/NakadiCursor.java
```diff
@@ -7,7 +7,7 @@
     public static final int VERSION_LENGTH = 3;
     public enum Version {
         ZERO("000"),
-        ONE("C00"),;
+        ONE("001"),;
```
v-stepanov: My problem now is that there's no space between versions ZERO and ONE. But you remember we have a "Plan B" how to rollout timelines without changing offsets for users. If we decide to apply that plan we will probably need some version between ZERO and ONE.
Wdyt?

src/main/java/org/zalando/nakadi/service/converter/VersionOneConverter.java
```diff
@@ -33,20 +33,30 @@
 
     public NakadiCursor convert(final String eventTypeStr, final Cursor cursor)
             throws InternalNakadiException, NoSuchEventTypeException, InvalidCursorException {
-        if (cursor.getOffset().length() <= NakadiCursor.VERSION_LENGTH + TIMELINE_ORDER_LENGTH) {
+        final String[] splitted = cursor.getOffset().split("-", 3);
```
v-stepanov: splitted is a wrong word as split is an irregular verb.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -122,11 +121,11 @@ public void createTimeline(final String eventTypeName, final String storageId, f
      */
     public List<Timeline> getActiveTimelinesOrdered(final String eventType)
             throws InternalNakadiException, NoSuchEventTypeException {
-        final List<Timeline> timelines = eventTypeCache.getTimelines(eventType);
+        final List<Timeline> timelines = eventTypeCache.getTimelinesOrdered(eventType);
         if (timelines.isEmpty()) {
             return Collections.singletonList(getFakeTimeline(eventTypeCache.getEventType(eventType)));
         } else {
-            return timelines.stream().sorted(Comparator.comparing(Timeline::getOrder)).collect(Collectors.toList());
+            return timelines.stream().filter(t -> t.getSwitchedAt() != null).collect(Collectors.toList());
```
v-stepanov: This is a very good answer why you need a test for this method.
Currently this method is not tested and anyone can change it in any way and nobody will notice.

src/main/java/org/zalando/nakadi/domain/NakadiCursor.java
antban: I will leave this comment unanswered

v-stepanov:
I will put real approve when the tests and zappr spec. check are greeen :-D

antban:
I squashed everything to one commit and rebased on master

v-stepanov:
👍

src/main/java/org/zalando/nakadi/controller/CursorsController.java
v-stepanov: @antban oh no, you have not addressed this comment, I want to take my approval back :)
At least you should remove extra space.

PR #583, opened 2017-03-06
aruha-683: cache default storage

adyach:
Currently, there is a performance problem with the default storage, becasue we go to db for every request of default topic repository. In this PR the caching of default storage was introduced.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
+++ b/src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
@@ -37,6 +37,8 @@ import java.util.Date;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 @Service
 public class TimelineService {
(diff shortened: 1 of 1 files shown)


rcillo:
@adyach wouldn't it be better to have a generic cache for storages instead of something specific to the default storage?

adyach:
@rcillo caching of storages is done as part of timeline caching. Each timeline has its storage which is picked up during event type cache filling.

rcillo:
I'm fine with locks implementation even if personally I find guava cache easier to use.

rcillo:
👍

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -136,13 +139,33 @@ public TopicRepository getTopicRepository(final EventTypeBase eventType)
...
+        readWriteLock.readLock().lock();
+        try {
+            if (defaultStorage != null) {
+                return defaultStorage;
+            }
+        } finally {
+            readWriteLock.readLock().unlock();
+        }
+
+        readWriteLock.writeLock().lock();
```
v-stepanov: I don't exactly get why we need a lock here at all?

adyach: @v-stepanov the lock is needed to properly initialise the storage, but I will change this impl, because there is simpler way to do it.

PR #584, opened 2017-03-06
aruha-683: cache default storage

adyach:
Currently, there is a performance problem with the default storage, becasue we go to db for every request of default topic repository. In this PR the caching of default storage was introduced.


src/main/java/org/zalando/nakadi/config/NakadiConfig.java
```diff
@@ -60,4 +66,24 @@ public ApplicationService applicationService(@Value("${nakadi.auth.plugin.factor
...
+        final Storage storage = new Storage();
+        storage.setId("default");
+        storage.setType(Storage.Type.KAFKA);
+        storage.setConfiguration(new Storage.KafkaConfiguration(
+                environment.getProperty("nakadi.zookeeper.exhibitor.brokers"),
+                Integer.valueOf(environment.getProperty("nakadi.zookeeper.exhibitor.port", "0")),
+                environment.getProperty("nakadi.zookeeper.brokers"),
+                environment.getProperty("nakadi.zookeeper.kafkaNamespace", "")));
+        try {
+            storageDbRepository.createStorage(storage);
```
rcillo: Why do we need to have the default storage in the database if the application only uses it's in memory representation?

rcillo: Maybe because it's needed when switching timelines?

rcillo:
This approach is better. 👍

rcillo:
:shipit:

PR #587, opened 2017-03-07
issue-585 fix major version bump

### LINKED ISSUE #585: Major version bump does not reset minor version
When doing a major version bump from, say, 1.1.0, it results in version 2.1.0 instead of 2.0.0.

I believe the culprit is here: https://github.com/zalando/nakadi/blob/master/src/main/java/org/zalando/nakadi/domain/Version.java#L63.

rcillo:
Fix #585


lmontrieux:
👍

PR #589, opened 2017-03-07
Reduce allocations writing batch events for event consumers.

dehora:
If `SEND_BATCH_VIA_OUTPUT_STREAM` is enabled, send the batch and its events directly to the consumer's OutputStream as bytes instead of building up an intermediate batch String via `createStreamEvent`. The standard parts of the batch are allocated statically as bytes instead of per call.

This is done to reduce the number of allocations generated for a batch and is based on some performance tuning investigations.

`EventStreamTest` is cloned as is to `EventStreamTestSendBatchViaOutputStream`, to exercise the current tests without changing them (to avoid adding test bugs), with some extra tests add to `EventStreamTest` that check the new `writeStreamEvent` method directly. The clone can be removed once the flagged path becomes the default and the flag is removed.


src/test/java/org/zalando/nakadi/service/EventStreamTest.java
```diff
@@ -343,4 +358,88 @@ public void whenReadFromMultiplePartitionsThenGroupedInBatchesAccordingToPartiti
...
+            verify(meter, times(1)).mark(anyInt());
+
+        } catch (IOException e) {
+            fail(e.getMessage());
+        }
+    }
+
+    @SuppressWarnings("unchecked")
+    @Test
+    public void testWriteStreamEventEmpty() {
```
rcillo: @dehora I'm not 100% sure, but I think that when no events are provided (keep alive scenario) we don't even write "events:" to the output stream. I'm worried that a "null" could break consumers that are not used to receive a null value for this field. Am I right?

dehora: @rcillo there's a test for not emitting `events` here https://github.com/zalando/nakadi/pull/589/files#diff-bd375e9e6fae6aed0a5c3d08446a9378R434. If that's not the right test, I'll change it. If it's the right test, I can change the test name, which might be misleading.

rcillo: @dehora I mean, the `key` events should not be present. This test should launch an exception when doing `batch.get("events");`. @v-stepanov could you confirm this behaviour?

v-stepanov: @rcillo Yes, when the batch is empty we don't even send the "events" field. But I think Bill's test tests exactly that behavior.

dehora: @rcillo @v-stepanov updated to make the check clearer in 2360d36

rcillo: @dehora my fault. I was confused about `get` behaviour. It would be clearer for me if the test checked `containsKey("events")` false. But it's fine like this.

v-stepanov:
@dehora have you considered reading bytes from Kafka instead of String? Because you actually anyway convert string event to bytes. You could configure Kafka consumer to use org.apache.kafka.common.serialization.ByteArrayDeserializer (currently StringDeserializer is used)

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -159,13 +183,53 @@ public static String createStreamEvent(final Cursor cursor, final List<String> e
...
+            final String streamEvent = createStreamEvent(cursorConverter.convert(topicPosition), currentBatch);
+            final byte[] batchBytes = streamEvent.getBytes(UTF8);
+            outputStream.write(batchBytes);
+            bytesFlushedMeter.mark(batchBytes.length);
+            outputStream.flush();
+        }
+    }
+
+    @VisibleForTesting
+    void writeStreamEvent(final OutputStream os, final Cursor cursor, final List<String> events)
```
v-stepanov: Also I think it will make sense to move the low-level writing together with bytes constants into a separate class.

v-stepanov: Then you will be able to use it in Subscription streaming

dehora:
> have you considered reading bytes from Kafka instead of String

@v-stepanov 👍  it was also part of the performance tuning exercise. I was going to do it in a different pr to reduce the amount of changes.

v-stepanov:
@dehora for me everything looks good 👍  The only thing is that I think it's better to extract the low-level stuff (writing to stream + bytes constants) into a separate class, because in other case EventStream class becomes a little bit mixed class: it contains both logic of streaming as well as low-level writing stuff.

dehora:
@v-stepanov ok! What I'd like to do is start prs for byte based consumers and a refactor like you said. I'd like to sketch those out on top of #589/#590 just to see what they'd look like. I'm going to mark this as do not merge for a day or so and sanity check the work here doesn't those follow on changes harder..

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -159,13 +183,53 @@ public static String createStreamEvent(final Cursor cursor, final List<String> e
...
+        byteCount += partition.length;
+        os.write(B_OFFSET_BEGIN);
+        final byte[] offset = cursor.getOffset().getBytes(StandardCharsets.UTF_8);
+        os.write(offset);
+        byteCount += offset.length;
+        os.write(B_CURSOR_PARTITION_END);
+        if (!events.isEmpty()) {
+            os.write(B_EVENTS_ARRAY_BEGIN);
+            for (int i = 0; i < events.size(); i++) {
+                final byte[] event = events.get(i).getBytes(StandardCharsets.UTF_8);
```
lhotari: To get rid of these byte array allocations, it would help to use `java.io.OutputStreamWriter`. Using `java.io.OutputStreamWriter` is slightly more efficient that creating byte arrays from Strings and writing those separately.
Using `com.google.common.io.CountingOutputStream` would simplify the byte counting code.

lhotari: Furthermore, the String->byte[] conversion problem originates from the Kafka consumer config. Using `org.apache.kafka.common.serialization.ByteArrayDeserializer` could help to get rid of the unnecessary byte[]->String and String->byte[] conversions and garbage that it causes.

lhotari: `StringDeserializer` configured at least in https://github.com/zalando/nakadi/blob/e5ee95c848aee5c4c59af9cfe7647969436f0666/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java#L80

lhotari: I started an experiment to switch to consume byte[] instead of String. Example of using CountingOutputStream: https://github.com/lhotari/nakadi/commit/d78a1ddefc1fac3110504e534f7d26f086a65ee4#diff-3456695fa478cd40e115f5abba5b87ddR149 . I didn't have time yet to complete the experiment. Perhaps consuming byte[] instead of String is something that would be worth doing?

lhotari: I finished my experiment and created PR #605 with my changes.

dehora: @lhotari 

> Perhaps consuming byte[] instead of String is something that would be worth doing?

It would have been better to have coordinated first before spinning up #605; see https://github.com/zalando/nakadi/pull/589#issuecomment-285317114 (relative to internal work this is part of a ongoing tuning exercise). I'll starting commenting on

I'm aware of CountingOutputStream; imvho it would be better to do that as part of a separate pr to keep the changes isolated (count bytes isn't ideal but it follows what's already here).

dehora:
I've done some local testing against the output with the flag enabled and allowing for some flakiness across runs, the ATs look ok. Lifted the do not merge flag.

v-stepanov:
@dehora can you please update with the status of this one?
1) Do you want to merge it as a first version before implementing the version with byte kafka-consumer?
2) If yes - have you considered doing the refactoring to separate low level byte stuff?

dehora:
> Do you want to merge it as a first version before implementing the version with byte kafka-consumer?

I do; it needs to be rebased again first. Then https://github.com/zalando/nakadi/pull/590 can follow

> If yes - have you considered doing the refactoring to separate low level byte stuff?

Yes. There's a pr here that can help with https://github.com/zalando/nakadi/pull/605. It'll need to be rebased once this and 590 are merged

src/main/java/org/zalando/nakadi/service/EventStream.java
dehora: Ok. That can be done in #590

dehora:
>> @dehora can you please update with the status of this one?
> it needs to be rebased again first.

rebased and latest is 7756324

antban:
:+1:

src/main/java/org/zalando/nakadi/service/EventStreamWriterString.java
```diff
@@ -0,0 +1,36 @@
...
+        final byte[] batchBytes = eventsString.getBytes(UTF_8);
+        os.write(batchBytes);
+        return batchBytes.length;
+    }
+
+    private EventStreamWriterString() {
+
+    }
+
+    static final EventStreamWriterString INSTANCE = new EventStreamWriterString();
```
v-stepanov: @antban 
Why do you make a singleton from it instead just making it a Spring bean which is a singleton by default?

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
```diff
@@ -0,0 +1,65 @@
...
+            }
+        }
+        os.write(B_CLOSE_CURLY_BRACKET);
+        os.write(B_BATCH_SEPARATOR);
+        return byteCount;
+    }
+
+    private EventStreamWriterBinary() {
+    }
+    static final EventStreamWriterBinary INSTANCE = new EventStreamWriterBinary();
```
v-stepanov: ... and here as well

https://github.com/zalando/nakadi/pull/589/commits/e97a12fc0aee38da1acd09475fc112806ba8c1ad#r115971321

src/main/java/org/zalando/nakadi/service/EventStreamWriterString.java
antban: https://github.com/zalando/nakadi/pull/589/commits/5bd154dc7fe5d989b56a7c779a94d474cf2bceee

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
antban: https://github.com/zalando/nakadi/pull/589/commits/5bd154dc7fe5d989b56a7c779a94d474cf2bceee

src/main/java/org/zalando/nakadi/service/EventStreamWriterString.java
antban: Because of `11 changed files  with 108 additions and 505 deletions`

antban:
:+1:

v-stepanov:
👍

antban:
This pr is closed in favor of #659

PR #590, opened 2017-03-07
Reduce allocations writing batch events for subscription consumers. 

dehora:
**do not merge: needs to be rebased against #589 when that merges** 

Write batch subscription events directly to the output stream.

If `SEND_SUBSCRIPTION_BATCH_VIA_OUTPUT_STREAM` is enabled, send the batch and its events directly to the `SubscriptionOutput`'s underlying OutputStream as bytes instead of building up an intermediate batch String via `serializeBatch`. The standard parts of the batch are allocated statically as bytes instead of per call. This is done to reduce the number of allocations generated for a batch and is based on some performance tuning investigations.

Each call to `SubscriptionOutput.streamData` in the production path calls flush on its underlying OutputStream. This might be a problem and if it is, `SubscriptionOutput` will probably have to allow StreamState to access to its OutputStream.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
@@ -26,6 +26,7 @@ import org.zalando.nakadi.service.subscription.state.StartingState;
 import org.zalando.nakadi.service.subscription.state.State;
 import org.zalando.nakadi.service.subscription.zk.ZKSubscription;
 import org.zalando.nakadi.service.subscription.zk.ZkSubscriptionClient;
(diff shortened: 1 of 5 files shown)


dehora:
I've done some local testing against the output with the flag enabled and allowing for some flakiness across runs, the ATs look ok. Pending a rebase against #598 when that merges

adyach:
@dehora hi Bill, what do you think about rebasing on top of ARUHA-753 branch and completing it? Let me know if you have enough time for that. 
Thanks!

rcillo:
@dehora I hope you don't mind. I'm rebasing it.

rcillo:
Closed in favor of https://github.com/zalando/nakadi/pull/673

PR #591, opened 2017-03-07
ARUHA-541 consumer lag monitoring endpoint

rcillo:
Ticket link https://techjira.zalando.net/browse/ARUHA-531

Previous to Timelines, cursor's offsets used to be sequential Long
numbers as Strings, for example "00000123". Because of that, several
consumers used to convert such strings back into numbers, our example
above would be converted to 123, and later use them as numbers for
several purposes.

One of the things consumers used to do was to query for `/partitions`
endpoint to get the newest available offset and then calculate the
distance between the newest and the latest consumed offset. This is how
many consumers from low level API would detect lag in consumer.

After introducing timelines, it will be no longer possible to easily
convert an offset to a number and perform such operations on the client
side. For that reason, we are introducing an extension to the
`/partitions` endpoint that will allow consumers to get the number of
unconsumed events without the need to perform any conversions or
calculations on their side.

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -632,7 +632,7 @@ paths:
             The flow id of the request, which is written into the logs and passed to called services. Helpful
             for operational troubleshooting and log analysis.
           type: string
-        - name: offset
+        - name: consumed_offset
           in: parameter
           description: |
             Offset to query for unconsumed events. Depends on `partition` parameter. When present adds the property
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -574,6 +574,25 @@ paths:
             The flow id of the request, which is written into the logs and passed to called services. Helpful
             for operational troubleshooting and log analysis.
           type: string
+        - name: X-nakadi-cursors
+          in: header
+          description: |
+            Cursors indicating the partitions and offsets consumed by the client.
+
+            When this header is provided, Nakadi calculates the consumer lag, e.g. the number of events between the
+            provided offset and the most recent published event.
+
+            In this situation, the response will contain the `unconsumed_events` property for every partition.
+
+            The lag calculation is non inclusive, e.g. if the provided offset is the offset of the latest published
+            event, the number of `unconsumed_events` will be zero.
+
+            If this header is not present, partitions won't display `unconsumed_events`.
+          required: false
+          type: array
+          items:
+            type: string
+            format: '#/definitions/Cursor'
       responses:
         '200':
           description: OK
@@ -613,6 +632,15 @@ paths:
             The flow id of the request, which is written into the logs and passed to called services. Helpful
             for operational troubleshooting and log analysis.
           type: string
+        - name: partition
+          in: parameter
+          description: |
+            Partition to query for unconsumed events. Depends on `offset` parameter.
+        - name: offset
+          in: parameter
+          description: |
+            Offset to query for unconsumed events. Depends on `partition` parameter. When present adds the property
+            `unconsumed_events` to the response partition object.
       responses:
         '200':
           description: OK
@@ -1626,6 +1654,11 @@ definitions:
           Might assume the special name BEGIN, meaning a pointer to the offset of the oldest
           available event in the partition.
         type: string
+      unconsumed_events:
+        description: |
+          Number of events unconsumed by the client. This is also known as consumer lag and is used for monitoring
+          purposes by consumers interested in keeping an eye on the number of unconsumed events.
+        type: number
 
   StreamInfo:
     type: object
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -548,6 +548,64 @@ paths:
           description: |
             Schema object.
 
+  '/event-types/{name}/cursors-lag':
+    post:
+      tags:
+        - unmanaged-api
+        - monitoring
+        - management-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        This endpoint is mostly interesting for monitoring purposes. Used when a consumer wants to know how far behing
+        in the stream its application is lagging.
+
+        It provides the number of unconsummed events for each cursor's partition.
+      parameters:
+        - name: name
+          in: path
+          description: EventType name
+          type: string
+          required: true
+        - name: X-Flow-Id
+          in: header
+          description: |
+            The flow id of the request, which is written into the logs and passed to called services. Helpful
+            for operational troubleshooting and log analysis.
+          type: string
+        - name: cursors
+          in: body
+          description: |
+            Each cursor indicates the partition and offset consumed by the client. When a cursor is provided,
+            Nakadi calculates the consumer lag, e.g. the number of events between the provided offset and the most
+            recent published event. This lag will be present in the response as `unconsumed_events` property.
+
+            It's not mandatory to specify cursors for every partition.
+
+            The lag calculation is non inclusive, e.g. if the provided offset is the offset of the latest published
+            event, the number of `unconsumed_events` will be zero.
+
+            All provided cursors must be valid, i.e. a non expired cursor of an event in the stream.
+          required: true
+          schema:
+            type: array
+            items:
+              - $ref: '#/definitions/Cursor'
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
+            description: |
+              List of partitions with the number of unconsummed events.
+            items:
+              $ref: '#/definitions/Partition'
+        '401':
+          description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+
+
   '/event-types/{name}/partitions':
     get:
       tags:
@@ -1654,6 +1712,7 @@ definitions:
           Approximate number of events unconsumed by the client. This is also known as consumer lag and is used for
           monitoring purposes by consumers interested in keeping an eye on the number of unconsumed events.
         type: number
+        format: int64
 
   StreamInfo:
     type: object
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1651,8 +1651,8 @@ definitions:
         type: string
       unconsumed_events:
         description: |
-          Number of events unconsumed by the client. This is also known as consumer lag and is used for monitoring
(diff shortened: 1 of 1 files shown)


api/nakadi-event-bus-api.yaml
```diff
@@ -574,6 +574,25 @@ paths:
...
           type: string
+        - name: X-nakadi-cursors
+          in: header
+          description: |
+            Cursors indicating the partitions and offsets consumed by the client.
+
+            When this header is provided, Nakadi calculates the consumer lag, e.g. the number of events between the
+            provided offset and the most recent published event.
+
+            In this situation, the response will contain `unconsumed_events` property for every partition.
```
lmontrieux: add `the` after `contain`

api/nakadi-event-bus-api.yaml
```diff
@@ -1626,6 +1654,11 @@ definitions:
           Might assume the special name BEGIN, meaning a pointer to the offset of the oldest
           available event in the partition.
         type: string
+      unconsummed_events:
+        description: |
+          Number of events unconsummed by the client. This is also known as consumer lag and is used for monitoring
```
lmontrieux: `unconsummed` -> `unconsumed` (only one `m`)

api/nakadi-event-bus-api.yaml
```diff
@@ -1626,6 +1654,11 @@ definitions:
...
+          purposes by consumers interested in detecting the trend in growth of the number of unconsumed events.
```
lmontrieux: `detecting the trend in growth of the number of unconsumed events.` -> `keeping an eye on the number of unconsumed events.` or `monitoring the evolution of the number of unconsumed events.`

api/nakadi-event-bus-api.yaml
```diff
@@ -574,6 +574,25 @@ paths:
             The flow id of the request, which is written into the logs and passed to called services. Helpful
             for operational troubleshooting and log analysis.
           type: string
+        - name: X-nakadi-cursors
```
adyach: It is not mandatory (http://zalando.github.io/restful-api-guidelines/naming/Naming.html), but I think it is pretty common to use hyphenated pascal case for headers.

api/nakadi-event-bus-api.yaml
```diff
@@ -613,6 +632,15 @@ paths:
             The flow id of the request, which is written into the logs and passed to called services. Helpful
             for operational troubleshooting and log analysis.
           type: string
+        - name: partition
```
adyach: could you elaborate on that, please? it is not clear why we want partition in params if we have it in path already.

api/nakadi-event-bus-api.yaml
rcillo: I used the same name that is used for opening streams. I prefer homogeneity.

api/nakadi-event-bus-api.yaml
rcillo: Removed redundant parameter b91e4f8

adyach:
👍

api/nakadi-event-bus-api.yaml
```diff
@@ -574,6 +574,25 @@ paths:
...
+
+            When this header is provided, Nakadi calculates the consumer lag, e.g. the number of events between the
+            provided offset and the most recent published event.
+
+            In this situation, the response will contain the `unconsumed_events` property for every partition.
+
+            The lag calculation is non inclusive, e.g. if the provided offset is the offset of the latest published
+            event, the number of `unconsumed_events` will be zero.
+
+            If this header is not present, partitions won't display `unconsumed_events`.
```
antban: Is it mandatory to specify cursors for all partitions at the same time or it is possible to do on per partition basis?

api/nakadi-event-bus-api.yaml
```diff
@@ -613,6 +632,11 @@ paths:
             The flow id of the request, which is written into the logs and passed to called services. Helpful
             for operational troubleshooting and log analysis.
           type: string
+        - name: offset
+          in: parameter
+          description: |
+            Offset to query for unconsumed events. Depends on `partition` parameter. When present adds the property
+            `unconsumed_events` to the response partition object.
```
antban: Maybe it is better to change name to `consumed_offset`, so it will hold some information about which offset to use.

api/nakadi-event-bus-api.yaml
rcillo: I would say it should be possible to pass only the cursors of interest. What do you think? Is it clear from the description?

api/nakadi-event-bus-api.yaml
rcillo: 👍 done c32cd34

api/nakadi-event-bus-api.yaml
```diff
@@ -574,6 +574,25 @@ paths:
             The flow id of the request, which is written into the logs and passed to called services. Helpful
             for operational troubleshooting and log analysis.
           type: string
+        - name: X-nakadi-cursors
+          in: header
```

api/nakadi-event-bus-api.yaml
```diff
@@ -1626,6 +1650,11 @@ definitions:
           Might assume the special name BEGIN, meaning a pointer to the offset of the oldest
           available event in the partition.
         type: string
+      unconsumed_events:
+        description: |
+          Number of events unconsumed by the client. This is also known as consumer lag and is used for monitoring
```
dehora: I'd suggest documenting this as approximate if it's approximate, or mentioning that it might be out of step with what's exactly in the stream at the time the count is examined. The "latency physics" suggest by the time i read it, it's possibly gotten out of date for a high volume stream. SQS is an example here which goes as far as naming their metadata with "approximate" in some case, eg http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html, but a sentence might be enough here.

api/nakadi-event-bus-api.yaml
```diff
@@ -613,6 +632,11 @@ paths:
             The flow id of the request, which is written into the logs and passed to called services. Helpful
             for operational troubleshooting and log analysis.
           type: string
+        - name: consumed_offset
```
dehora: There might be systems reasons for not doing so, but would it be reasonable to just send back `unconsumed_events` in the data as an additive field? As a client dev `unconsumed_events` sounds like something I'd going to want to have as default information to help with instrumentation on the consumer side.

rcillo: @dehora I'm not sure I understood what you mean by additive field. This property `unconsumed_events` is only present when `consumed_offset` is provided as a parameter in the query.

api/nakadi-event-bus-api.yaml
rcillo: Fixed f026f5e

api/nakadi-event-bus-api.yaml
rcillo: Fixed query parameter f1c76e3

api/nakadi-event-bus-api.yaml
rcillo: c26e4a0

api/nakadi-event-bus-api.yaml
rcillo: Thank you for the detailed explanation. Here is our second proposal c26e4a0

api/nakadi-event-bus-api.yaml
rcillo: I've moved this parameter to the URL and changed the name to `cursors` c26e4a0

rcillo:
@dehora I see no way to return `unconsumed_events`  by default, since it depends on the user providing a cursor pointing to the latest consumed event. Since it's about low level API, the server has no information about which offsets have been successfully consumed by the client.

rcillo:
@tkrop @dehora @antban @v-stepanov @adyach @lmontrieux @SergKam I've acted on all comments. I chose to move from the `header` approach to encoding the json in a query parameter.

The other alternative would be to have a POST endpoint. I've received the feedback that POST should be used when changing state in the server and so I tried to keep using the GET approach.

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,64 @@ paths:
...
+
+            It's not mandatory to specify cursors for every partition.
+
+            The lag calculation is non inclusive, e.g. if the provided offset is the offset of the latest published
+            event, the number of `unconsumed_events` will be zero.
+
+            All provided cursors must be valid, i.e. a non expired cursor of an event in the stream.
+          required: true
+          schema:
+            type: array
```
tkrop: You may consider to wrap the array with an object as suggested by this guideline (http://zalando.github.io/restful-api-guidelines/data-formats/DataFormats.html#must-use-json-as-the-body-payload) to simplify future extensions.

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,64 @@ paths:
...
+            items:
+              - $ref: '#/definitions/Cursor'
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
```

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,64 @@ paths:
...
+        - unmanaged-api
+        - monitoring
+        - management-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        This endpoint is mostly interesting for monitoring purposes. Used when a consumer wants to know how far behing
+        in the stream its application is lagging.
+
+        It provides the number of unconsummed events for each cursor's partition.
```
GJL: Should be *unconsumed*

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,64 @@ paths:
...
+  '/event-types/{name}/cursors-lag':
+    post:
+      tags:
+        - unmanaged-api
+        - monitoring
+        - management-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        This endpoint is mostly interesting for monitoring purposes. Used when a consumer wants to know how far behing
```
GJL: Typo: "[...] know how far *behind* [...]"

rcillo:
👍

rcillo:
👍

GJL:
Will the API also accept the current offsets, i.e., numeric ones?

antban:
:+1:

v-stepanov:
👍

PR #592, opened 2017-03-07
aruha-691 Cursor navigation

rcillo:
Previous to Timelines it was possible to take one offset, for example
"000000000000000009", transform it into a number and later add or
subtract a given number from it, in order to craft another valid offset
that could be used for random access.

After timelines, cursors are no longer simple sequences and can no
longer be easily or safely manipulated by the client for random access.

For this reason we are providing this extension to the API, as an
alternative for randomly accessing the stream.


api/nakadi-event-bus-api.yaml
```diff
@@ -1647,6 +1695,17 @@ definitions:
         type: string
         description: |
           Offset of the event being pointed to.
+      shift:
+        type: number
+        description: |
+          This number is a modifier for the offset. It moves the cursor forward or backwards by the amount provided.
```
lmontrieux: `amount` -> `number of events`

api/nakadi-event-bus-api.yaml
```diff
@@ -1647,6 +1695,17 @@ definitions:
...
+          For example, suppose a user wants to read events starting 100 positions before "001-000D-0000000000000009A8",
+          it's possible to specify shift with -100 and Nakadi will make the necessary calculations to move the cursor
+          relatively to the given offset.
```
lmontrieux: Should we indicate here what happens if the user provides a shift that's either too far in the past (before BEGIN), or too far in the future (not enough events have been published yet)?

rcillo: I don't think we need to provide any special status code since we already have 422 for this scenario. In practice, the error message will be slightly different, to make it clearer that the requested "shifted offset" is not available. But this is not new, since we already return 422 for offsets "out of bounds".

api/nakadi-event-bus-api.yaml
```diff
@@ -1647,6 +1695,18 @@ definitions:
         type: string
         description: |
           Offset of the event being pointed to.
+      shift:
```
v-stepanov: Do you think it makes sense to include it in original cursor definition? I'm just afraid that 95% of users who don't need this functionality at all will be confused with this "shift".
Or do you plan that it will be possible to pass such a cursor directly to streaming endpoint? If yes - then I don't get why we need a separate endpoint that will shift cursors.

api/nakadi-event-bus-api.yaml
```diff
@@ -623,6 +623,54 @@ paths:
...
+              format: '#/definitions/Cursor'
+        responses:
+          '200':
+            description: OK
+            schema:
+              type: array
+              description: An array of `Cursor`s
+              items:
+                $ref: '#/definitions/Cursor'
+          '401':
```
v-stepanov: Can be skipped.

api/nakadi-event-bus-api.yaml
rcillo: I considered adding it to the stream endpoint to avoid one more round trip. Nowadays clients can directly craft a shifted offset on their side, so I though about what the performance impact would be if we forced everyone to make one more call to "materialize" a shifted offset into a non shifted offset.

I considered not publishing the extra endpoint since I also think it's redundant. But I did anyway since someone would ask for it.

rcillo: @v-stepanov maybe we should actually make it be costly so that people won't replace new-offset by "BEGIN" + "old-offset" as shift.

v-stepanov: @rcillo Yeah, I think it's better to have that only in a separate endpoint. So that users can do [ cursors + shift = shifted-cursors ]
That will also allow not to include `shift` into original cursor definition.
That will allow not to expose that shift stuff to all usual users.

v-stepanov: I mean my point is that shifting cursors is a very specific use-case and we actually don't want this to be widely used.

rcillo: @v-stepanov done a366665

v-stepanov:
👍

api/nakadi-event-bus-api.yaml
```diff
@@ -623,6 +623,50 @@ paths:
...
+            type: array
+            items:
+              type: string
+              format: '#/definitions/ShiftedCursor'
+        responses:
+          '200':
+            description: OK
+            schema:
+              type: array
+              description: An array of `Cursor`s with shift applied to the offset.
```
antban: Are there any guaranties about order of cursors?

antban: Are there any special cases when it is not possible to calculate this information?

api/nakadi-event-bus-api.yaml
```diff
@@ -623,6 +623,50 @@ paths:
...
+            description: |
+              The flow id of the request, which is written into the logs and passed to called services. Helpful
+              for operational troubleshooting and log analysis.
+            type: string
+          - name: X-nakadi-cursors
+            in: header
+            description: |
+              Cursors indicating the positions to be calculated. If a cursors contains `shift` property, this property
+              will be used to calculate the actual offset for the given position and in the response the cursor will
+              no longer contain the `shift`.
```
antban: Is it possible to calculate this information for every possible cursor (for example if data related to new cursor was deleted)
If cursor will contain some invalid offset (for example less then -1 in case of kafka or value that is greater then latest offset), will it report some kind of error for this cursor?

rcillo: 👍 sure, just added the missing status code ea62a5a

rcillo: But I won't detail every possibility in the API since the `Problem` object returned will contain more details about what went wrong.

api/nakadi-event-bus-api.yaml
rcillo: I don't think we need to promise any order in the API. Since the returned object contains all the information about the query (I mean, `ShiftedCursor` has `partition` and `offset`) so that the requester don't need to rely on order.

I added 422 to the response.

api/nakadi-event-bus-api.yaml
```diff
@@ -623,6 +623,50 @@ paths:
...
+            type: string
+            required: true
+          - name: X-Flow-Id
+            in: header
+            description: |
+              The flow id of the request, which is written into the logs and passed to called services. Helpful
+              for operational troubleshooting and log analysis.
+            type: string
+          - name: X-nakadi-cursors
+            in: header
```

dehora: @tkrop 

> We will shortly improve our API guidelines, to give better guidance on 
> the usage of headers and get parameters with respect to semantic and 
> size constraints.

@tkrop Understood, the challenge is in the absence of clear guidance, the API owners have to guess what the guidelines might be in advance of them existing (this pattern has gotten through a guild review in the past, so there's no way to have known it's now non-preferred in advance).

With that it mind I see the following options:

- Proceed with the pattern that's here. Add the guild's preferred option in the future on both resources when it exists and deprecate the header pattern (assuming the team is happy to support both options).

- Wait for the pattern to be in place from the guild

- Design a non-header option, hope it lines up with the guild model.

- Have the guild work directly with the team now, and document the outcome.


/+ @ePaul @fmueller as they are often stewards of the guild process :)

api/nakadi-event-bus-api.yaml
```diff
@@ -1648,6 +1698,25 @@ definitions:
         description: |
           Offset of the event being pointed to.
 
+  ShiftedCursor:
```
dehora: `"OffsetCursor"` ? That would make the field name `"offset"`. Just a thought, to reduce the number of terms in the API's design.

api/nakadi-event-bus-api.yaml
```diff
@@ -1648,6 +1698,25 @@ definitions:
...
+    allOf:
+      - $ref: "#/definitions/Cursor"
+      - type: object
+        properties:
+          shift:
```
dehora: it's probably worth documenting (or clarifying) the following - :

- what happens when shift is 0

- what happens when shift moves back beyond the stream horizon. Clients can't reason about this, they'll just have to make a call. If the server "rounds down" to the stream's horizon, that's possibly going to break someone being creative with using the shift to count on the consumer (we have real world experience people will count things if they're given anything that remotely looks like a number :) If the server rejects the request, the client's going to have to play a game of hi-lo.

- what happens when shift moves ahead of now, the head of the stream. Again, clients can't reason about this.

- when the number exceeds int32 plus/minus. If it's meant to be a long, that's worth saying. I've got no problem imagining a Nakadi stream with more than 2^32 events - volumes only ever go up :)

- if large numbers are allowed whether JSON libs can mess them up (eg some systems protect against this by drilling very large numbers through strings - as ugly as that is, it stops bugs caused by rounding)

api/nakadi-event-bus-api.yaml
dehora: > Are there any guaranties about order of cursors?
>I don't think we need to promise any order in the API

Good spot. If there's no ordering, I'd add a short sentence saying that.

api/nakadi-event-bus-api.yaml
tkrop: @dehora we are working on providing a clear guidance. We would be please to work out the guidance together with the team.

@dehora @rcillo @codecov-io @v-stepanov @lmontrieux Please contribute to the discussion in https://github.com/zalando/restful-api-guidelines/issues/207 how the guidance should look like.

api/nakadi-event-bus-api.yaml
tlossen: but `Cursor` has an `offset` field already.

api/nakadi-event-bus-api.yaml
tlossen: json schema defines `number` as 
> An arbitrary-precision, base-10 decimal number value, from the JSON "number" production

so this should not be limited to 2^32 or even 2^64. 

in case there is any limit from nakadi side, it is worth pointing out of course.

api/nakadi-event-bus-api.yaml
ePaul: I'll try to concentrate on this specific API here instead of the general guideline issue.

The API in the current form takes a list of "shifted cursor" objects, where each is a combination of partition, offset and (optional) shift, and produces a list of cursor objects (partition/offset).
(I guess the idea is that each input array entry corresponds to one result array entry, in the same order?.)

One problem here is that OpenAPI in its 2.0 incarnation doesn't allow putting complex objects (or arrays thereof) into a header, which manifests in using the reference to the schema in a "format" property, because there is no proper place for that. It is not clear to me how this array of objects would be encoded in the header.

The other issue is that the _main input_ of an operation, not just a context information like language, accepted media type, request content type etc., is hidden in a header, instead of being expressed in a URL (or possibly a POST body).

I would suggest putting the three values – partition, offset and shift – into query parameters. Of course, this makes it complicated to have a bulk request as it is currently, but is this bulk request really needed? If so, you could possibly put the list of objects into a POST body.

api/nakadi-event-bus-api.yaml
ePaul: Yeah, I would say `offset` doesn't work here, as the object would have two `offset` fields with different meaning (one opaque string, one a numeric difference).

api/nakadi-event-bus-api.yaml
```diff
@@ -623,6 +623,68 @@ paths:
...
+            required: true
+            schema:
+              type: array
+              items:
+                - $ref: '#/definitions/ShiftedCursor'
+        responses:
+          '200':
+            description: OK
+            schema:
+              type: array
```

rcillo:
👍

antban:
:+1:

v-stepanov:
👍

rcillo:
👍

antban:
:+1:

PR #593, opened 2017-03-07
ARUHA-692 Cursor distance calculation

rcillo:
Previous to Timelines it used to be possible to calculate how many
events here is between two offsets. For example, between offsets
"000000000000000001" and "000000000000000005" there are obviously 4
events.

This kind of comparison is used extensively by consumers in order to
verify that their implementation is correctly processing the same number
of events that Nakadi sent to them.

This endpoint provides an alternative way of calculating this distance
after Timelines was introduced.


api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,43 @@ paths:
...
+    get:
+      tags:
+        - unmanaged-api
+        - monitoring
+        - management-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        Calculate the distance between two offsets. This is useful for performing checks for data completeness, when
+        a client had read some batches and wants to be sure that all delivered events have been correctly processed.
```
lmontrieux: `had` -> `has`

api/nakadi-event-bus-api.yaml
```diff
@@ -1634,6 +1671,32 @@ definitions:
...
+  CursorDistanceResult:
+    required:
+      - partition
+      - distance
+    properties:
+      distance:
+        type: number
+        description: |
+          Number of events between two offsets. Initial offset is exclusive. It's only zero when both provided offsets
+          are equal.
```
lmontrieux: What happens if the older offset is in fact newer than the newer offset? Do we reply with an error, or do we give a negative value for distance?

rcillo: Good. I'll make it explicit. The idea is to reply with error.

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,43 @@ paths:
...
+            format: "#/definitions/CursorDistanceQuery"
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
+            description: An array of `CursorDistanceResult`s with the distance between two offsets.
+            items:
+              $ref: '#/definitions/CursorDistanceResult'
+        '401':
```
v-stepanov: you can skip "default" status codes in swagger file until they have some special meaning/use-case

api/nakadi-event-bus-api.yaml
```diff
@@ -1634,6 +1671,32 @@ definitions:
...
+      final_offset:
+        type: string
+        description: |
+          Offset of the newer event in the stream.
+
+  CursorDistanceResult:
+    required:
+      - partition
+      - distance
+    properties:
```
v-stepanov: `partition` property is missing

api/nakadi-event-bus-api.yaml
```diff
@@ -1634,6 +1671,32 @@ definitions:
       purposes. We recommend logging this object in order to solve connection issues. Clients
       should not parse this structure.
 
+  CursorDistanceQuery:
+    required:
+      - partition
+      - initial_offset
+      - final_offset
+    properties:
```
v-stepanov: `partition` property is missing

api/nakadi-event-bus-api.yaml
```diff
@@ -1634,6 +1666,40 @@ definitions:
...
+      partition:
+        type: string
+        description: |
+          Partition for which the offset distance is measured.
+      initial_offset:
```
v-stepanov: will `BEGIN` be supported here?

rcillo: @v-stepanov yes.

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,38 @@ paths:
...
+      description: |
+        Calculate the distance between two offsets. This is useful for performing checks for data completeness, when
+        a client has read some batches and wants to be sure that all delivered events have been correctly processed.
+      parameters:
+        - name: X-nakadi-cursor-distances-query
+          in: header
+          description: |
+            List of cursors with two special offsets: `initial_offset` and `final_offset`. Used as input by Nakadi to
+            calculate how many events there are between two offsets. The distance doesn't include the `initial_offset`
+            but includes the `final_offset`. So they are equal the result is zero. If the `initial_offset` is the one
```
v-stepanov: So **when** they are equal the result is zero

antban: Is it possible to use offsets from future or from past?

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,38 @@ paths:
...
+          type: array
+          items:
+            type: string
+            format: "#/definitions/CursorDistanceQuery"
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
+            description: An array of `CursorDistanceResult`s with the distance between two offsets.
```
antban: count of elements and their positions is the same as in requested parameter?

api/nakadi-event-bus-api.yaml
```diff
@@ -1634,6 +1666,40 @@ definitions:
...
+      initial_offset:
+        type: string
+        description: |
+          Offset of the older event in the stream.
+      final_offset:
+        type: string
+        description: |
+          Offset of the newer event in the stream.
+
+  CursorDistanceResult:
```
antban: Does it make sense to extend `CursorDistanceQuery` and include some kind of additional field like `error` in case of error?

rcillo: In case of error I thought about using the default approach: return 422 with `Problem`.

rcillo: But extending seems to be a good idea. I'll do it.

rcillo: Done caedf39

api/nakadi-event-bus-api.yaml
rcillo: After I made `Result` and extension of the `Query` it became unambiguous. It'll be very clear that the distance is the number of positions in between initial and final offsets.

api/nakadi-event-bus-api.yaml
rcillo: Will result in 422 1548a2a

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,38 @@ paths:
...
+        - monitoring
+        - management-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        Calculate the distance between two offsets. This is useful for performing checks for data completeness, when
+        a client has read some batches and wants to be sure that all delivered events have been correctly processed.
+      parameters:
+        - name: X-nakadi-cursor-distances-query
+          in: header
```

api/nakadi-event-bus-api.yaml
```diff
@@ -1634,6 +1670,37 @@ definitions:
       purposes. We recommend logging this object in order to solve connection issues. Clients
       should not parse this structure.
 
+  CursorDistanceQuery:
+    required:
+      - partition
+      - initial_offset
+      - final_offset
```
tlossen: another option would be for `CursorDistanceQuery` to consist of `initial_cursor` and `final_cursor`, both of type `Cursor`. this would introduce a tiny amount of duplication (same `partition`), but might be more user friendly?

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,42 @@ paths:
...
+      tags:
+        - unmanaged-api
+        - monitoring
+        - management-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        Calculate the distance between two offsets. This is useful for performing checks for data completeness, when
+        a client has read some batches and wants to be sure that all delivered events have been correctly processed.
+      parameters:
```
ePaul: The `name` path parameter is missing here ... I guess this is just a single oversight?

api/nakadi-event-bus-api.yaml
ePaul: I would suggest three query parameters for a GET request:

  - partition
  - initial_offset
  - final_offset

This doesn't allow bulk requests, though I dare to question the need for those.
If a bulk request is really needed, put the list of objects into the request body with a POST.

tkrop: I'm not that sure about that, because a subscription listener in best case wants to apply this operation on all partitions he is listening to at once. Surely it is an option to provide two end point for this request, but I would understand the decision to only support one bulk API, as it is already complex and big enough.

api/nakadi-event-bus-api.yaml
tkrop: Nice idea. From programmer perspective this allows to directly use the stored cursors.

api/nakadi-event-bus-api.yaml
rcillo: 👍 fixed

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,47 @@ paths:
...
+          type: string
+          required: true
+        - name: cursors-distances-query
+          in: body
+          description: |
+            List of pairs of cursors: `initial_cursor` and `final_cursor`. Used as input by Nakadi to
+            calculate how many events there are between two cursors. The distance doesn't include the `initial_cursor`
+            but includes the `final_cursor`. So if they are equal the result is zero.
+          schema:
+            type: array
```
tkrop: You may consider to wrap the array with an object as suggested by this guideline (http://zalando.github.io/restful-api-guidelines/data-formats/DataFormats.html#must-use-json-as-the-body-payload) to simplify future extensions.

api/nakadi-event-bus-api.yaml
```diff
@@ -548,6 +548,47 @@ paths:
...
+            items:
+              $ref: "#/definitions/CursorDistanceQuery"
+          required: true
+      responses:
+        '200':
+          description: OK
+          schema:
+            type: array
```

rcillo:
👍

v-stepanov:
👍

antban:
:+1:

rcillo:
👍

v-stepanov:
👍

PR #599, opened 2017-03-08
aruha-695 publishing metrics

rcillo:
This PR exposes bytes sent per application per event type.

The name of the metric is "publishing.application-name.event-type-name.bytes-sent" and it's a meter.

It can be found in the `/stream-metrics`. These metrics are been introduced for administration purposes at first, but later we are going to document them so that users can also rely on them.

Previous to this PR, the only way to monitor publishing was by topic,
which is Kafka internal information, which we don't want to expose to
users.

This way users can monitor based on Nakadi only.

### DIFF
--- a/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
@@ -1,10 +1,7 @@
 package org.zalando.nakadi.repository.kafka;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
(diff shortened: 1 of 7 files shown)


adyach:
@rcillo I understood your point about integrating metrics into the topic repository since it gives the number of sent bytes to repository without iterating over all events and checking their status. Nevertheless, If you look at the EventTypeController.reportMetrics method, it already does some job related to the event sizing(number of events, translating string to bytes etc.) and actually iterates over all all events while checking publishing status. 
According to that maybe it will good if we integrate calculation of sent bytes into the controller level as well?

rcillo:
@adyach I'll finally get back to you. Yes, you are right. Some of the metrics are already collected in the right layer. I'll close this one and re write it in the correct layer.

PR #600, opened 2017-03-09
aruha-694: HIlaAT.whenStreamTimeoutReachedPossibleToCommit fix

adyach:
This PR bring fix for failing from time to time test, the problem was in the implementation of the streaming client, which was hanging while closing the stream, but the connection was closed from Nakadi side and commit timeout expired thus it was not possible to commit cursors.
image

### DIFF
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/utils/TestStreamingClient.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/utils/TestStreamingClient.java
@@ -2,38 +2,42 @@ package org.zalando.nakadi.webservice.utils;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.Lists;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.zalando.nakadi.config.JsonConfig;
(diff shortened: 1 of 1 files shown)


src/acceptance-test/java/org/zalando/nakadi/webservice/utils/TestStreamingClient.java
```diff
@@ -117,11 +115,8 @@ public TestStreamingClient start() {
 
     public boolean close() {
         if (running) {
-            try {
-                connection.disconnect();
-            } finally {
-                running = false;
-            }
+            running = false;
+            connection.disconnect();
```
v-stepanov: @adyach have you discussed that with @antban? I know that in his branch he changed the way of closing connection by closing connection of URL object.

adyach: I think he already pushed it to master, but it did not help. The `connection` is the instance of the `HttpUrlConnection`. Is that what you meant?

v-stepanov: Ah ok, I see now. I took a look at his branch - yes he actually closes HttpUrlConnection. So everything is ok.

antban:
:+1:

PR #602, opened 2017-03-10
ARUHA-530 Allow to stream from multiple timelines at the same time

antban:
This is preliminary pull requests that shows how it will look like. 
Tests will be added as additional commit


v-stepanov:
@antban only 20 files. You are getting old :trollface:

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,265 @@
...
+    @Override
+    public Set<TopicPartition> getAssignment() {
+        return latestOffsets.values().stream().map(NakadiCursor::getTopicPartition).collect(Collectors.toSet());
+    }
+
+
+    @Override
+    public Optional<ConsumedEvent> readEvent() {
+        Runnable task;
+        while ((task = queuedTasks.poll()) != null) {
```
v-stepanov: you really like task queues :)

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,265 @@
...
+            task.run();
+        }
+        Optional<ConsumedEvent> result = Optional.ofNullable(eventsQueue.poll());
+        if (!result.isPresent()) {
+            poll();
+            result = Optional.ofNullable(eventsQueue.poll());
+        }
```
v-stepanov: maybe above 5 lines can be inside poll method? So that poll will simply return Optional<ConsumedEvent>

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,265 @@
...
+            final NakadiCursor position = result.get().getPosition();
+            final EventTypePartition etp = position.getEventTypePartition();
+            latestOffsets.put(etp, position);
+            final String borderOffset = borderOffsets.get(etp);
+            if (null != borderOffset && borderOffset.compareTo(position.getOffset()) <= 0) {
+                queuedTasks.add(() -> {
+                    try {
+                        reelectTopicRepositories();
+                    } catch (final NakadiException | InvalidCursorException e) {
+                        throw new NakadiRuntimeException(e);
```
v-stepanov: Come on, why you still use this wrapper?

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,265 @@
...
+            poll();
+            result = Optional.ofNullable(eventsQueue.poll());
+        }
+        if (result.isPresent()) {
+            final NakadiCursor position = result.get().getPosition();
+            final EventTypePartition etp = position.getEventTypePartition();
+            latestOffsets.put(etp, position);
+            final String borderOffset = borderOffsets.get(etp);
+            if (null != borderOffset && borderOffset.compareTo(position.getOffset()) <= 0) {
+                queuedTasks.add(() -> {
```
v-stepanov: why you don't want to do this synchronously? I think that will simplify the code.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,265 @@
...
+import org.zalando.nakadi.exceptions.NakadiRuntimeException;
+import org.zalando.nakadi.exceptions.ServiceUnavailableException;
+import org.zalando.nakadi.service.timeline.TimelineService;
+import org.zalando.nakadi.service.timeline.TimelineSync;
+import org.zalando.nakadi.util.NakadiUtils;
+
+public class MultiTimelineEventConsumer implements EventConsumer {
+    private final String clientId;
+    private final Map<EventTypePartition, NakadiCursor> latestOffsets = new HashMap<>();
+    private final Map<String, List<Timeline>> consumedEventTypes = new HashMap<>();
```
v-stepanov: hm... Is it supposed to read from multiple event-type?
Do you plan to reuse it for HILA?

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,265 @@
...
+        final List<String> toRemove = consumedEventTypes.keySet().stream()
+                .filter(et -> !eventTypes.contains(et))
+                .collect(Collectors.toList());
+        for (final String item : toRemove) {
+            consumedEventTypes.remove(item);
+        }
+        // Add refreshed data.
+        for (final String eventType : eventTypes) {
+            consumedEventTypes.put(eventType, timelineService.getActiveTimelinesOrdered(eventType));
+        }
```
v-stepanov: does this code really changes the set of event-types? Or does it only "updates" event-types?

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    @Override
+    public Set<TopicPartition> getAssignment() {
+        return latestOffsets.values().stream().map(NakadiCursor::getTopicPartition).collect(Collectors.toSet());
+    }
+
+
+    @Override
+    public Optional<ConsumedEvent> readEvent() {
```
v-stepanov: what is the expected time of execution of this method? Can it me more than 1 second? I ask this because EventStream class which uses it, has some expectations. 
For example if batch_flush_timeout = 1, then EventStream want to send batch (or empty batch) at least every second. And if this method will run more than that - that can produce behavior which user doesn't expect.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,265 @@
...
+        Optional<ConsumedEvent> result = Optional.ofNullable(eventsQueue.poll());
+        if (!result.isPresent()) {
+            poll();
+            result = Optional.ofNullable(eventsQueue.poll());
+        }
+        if (result.isPresent()) {
+            final NakadiCursor position = result.get().getPosition();
+            final EventTypePartition etp = position.getEventTypePartition();
+            latestOffsets.put(etp, position);
+            final String borderOffset = borderOffsets.get(etp);
```
v-stepanov: How this `borderOffsets` map filled initially? I could not find any initialization.

v-stepanov: Ok, I see now. It will be triggered by calling `reassign`, right?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -165,6 +169,13 @@ public TopicRepository getDefaultTopicRepository() throws TopicRepositoryExcepti
         return topicRepositoryHolder.getTopicRepository(defaultStorage);
     }
 
+    public EventConsumer createEventConsumer(final String clientId, final List<NakadiCursor> positions)
+            throws NakadiException, InvalidCursorException {
+        final MultiTimelineEventConsumer result = new MultiTimelineEventConsumer(clientId, this, timelineSync);
+        result.reassign(positions);
```
v-stepanov: what is the purpose of making it a separate call? Can this be passed through constructor? 
Because I don't see any other place where 'reassign' is called.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @v-stepanov  Yes, I do

v-stepanov:
@antban first iteration of review is finished

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @v-stepanov Basically this method won't introduce more than usual NakadiEventConsumer.

v-stepanov: What about the place where it needs first to execute all tasks from queue to proceed? Theoretically that can result in creating of several event consumers which can be not so fast.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: In this case it is very important thing - all changes are applied only on streaming path

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: ok

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: yes, on reelectTopicRepositories - when timelines or consumed event types changed

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: Yes, this should be made without queuedTasks, missed it.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: I'll try to use different one.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: This code will do that while streaming from subscription

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
antban: I do not want to have it inside constructor, because there will be 2 ways of initialization. And reinitialization I will need in any case for subscriptions

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
v-stepanov: Ah, ok I see. Because of subscription rebalance the set of streamed event-types can be changed, right?

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: Yes, it may happen, but I think that this limitation we can not overcome.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: But in general there is nothing to add there - it will be logged in `EventStreamController#streamEvents` call

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @v-stepanov yes

src/main/java/org/zalando/nakadi/util/NakadiUtils.java
```diff
@@ -0,0 +1,26 @@
...
+        public final Set<T> added;
+        public final Set<T> removed;
+
+        Diff(final Set<T> added, final Set<T> removed) {
+            this.added = added;
+            this.removed = removed;
+        }
+    }
+
+    public static <T> Diff<T> difference(final Set<T> oldCollection, final Set<T> newCollection) {
```
v-stepanov: would love to see a test for this

v-stepanov: Also, why do define params as Sets? I think those can be just Collections

src/main/java/org/zalando/nakadi/util/NakadiUtils.java
```diff
@@ -0,0 +1,26 @@
+package org.zalando.nakadi.util;
+
+import java.util.Set;
+import java.util.stream.Collectors;
+
+public class NakadiUtils {
```
v-stepanov: Nakadi**Collection**Utils? :)
Or maybe the method name can be improved to reflect that it's about collections.

v-stepanov:
@antban only one questions left - regarding acceptance test. Do you plan that we make a big Timeline AT when all timeline stuff is implemented?

src/main/java/org/zalando/nakadi/util/NakadiUtils.java
antban: @v-stepanov In collections values may be not unique, and that is why the way it will work won't be so straightforward.

src/main/java/org/zalando/nakadi/util/NakadiUtils.java
antban: Ok

antban:
@v-stepanov Regarding acceptance test I think that they may be written in separate PR, cause actually they will test everything - from creating timeline to consuming timeline, and they will be really big.

src/main/java/org/zalando/nakadi/util/NakadiUtils.java
v-stepanov: Ok, I see now. Then it maybe should be renamed to `setsDifference`
And params can be renamed to something like `oldSet`, `newSet`

antban: Ok

src/main/java/org/zalando/nakadi/util/NakadiCollectionUtils.java
```diff
@@ -3,7 +3,7 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 
-public class NakadiUtils {
+public class NakadiCollectionUtils {
     public static class Diff<T> {
         public final Set<T> added;
         public final Set<T> removed;
```
v-stepanov: I just noticed that these fields are public. What is the reason for that?

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,264 @@
...
+import org.zalando.nakadi.util.NakadiCollectionUtils;
+
+public class MultiTimelineEventConsumer implements EventConsumer {
+    private final String clientId;
+    private final Map<EventTypePartition, NakadiCursor> latestOffsets = new HashMap<>();
+    private final Map<String, List<Timeline>> consumedEventTypes = new HashMap<>();
+    private final Map<String, TimelineSync.ListenerRegistration> timelineRefreshListeners = new HashMap<>();
+    private final Map<TopicRepository, EventConsumer> eventConsumers = new HashMap<>();
+    private final LinkedList<ConsumedEvent> eventsQueue = new LinkedList<>();
+    private final Map<EventTypePartition, String> borderOffsets = new HashMap<>();
```
rcillo: There are some new abstractions in this class which are not trivial. Could you clarify what `borderOffsets` is about? Does it hold the offsets pointing to the newest or oldest event in this partition?

If it's the latest available offset for that partition, I would suggest renaming to `latestAvailableOffset` to make it more clear or adding some comments. I find comments very useful when reading code, because I don't need to read the entire class to understand what is the meaning of something.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,264 @@
...
+        while ((task = queuedTasks.poll()) != null) {
+            task.run();
+        }
+        final Optional<ConsumedEvent> result = poll();
+        if (result.isPresent()) {
+            final NakadiCursor position = result.get().getPosition();
+            final EventTypePartition etp = position.getEventTypePartition();
+            latestOffsets.put(etp, position);
+            final String borderOffset = borderOffsets.get(etp);
+            if (null != borderOffset && borderOffset.compareTo(position.getOffset()) <= 0) {
```
rcillo: Does it mean that when reading, the consumer reached the limit of available events? I find it useful to extract the result of the expression to a variable with a good explanatory name or add a comment like: "switch to a new timeline when all events from the previous timeline have been consumed".

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,264 @@
...
+                try {
+                    reelectTopicRepositories();
```
rcillo: The usage of the verb "election" alludes to something non deterministic and probabilistic. I don't think you mean that.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,264 @@
...
+                }
+            }
+        }
+        return result;
+    }
+
+    private Optional<ConsumedEvent> poll() {
+        Optional<ConsumedEvent> result = Optional.ofNullable(eventsQueue.poll());
+        if (!result.isPresent()) {
+            eventConsumers.values().forEach(consumer -> consumer.readEvent().ifPresent(eventsQueue::add));
```
rcillo: what about reading in parallel? I don't know if that would optimise things since kafka consumer already pre-fetches data, so that when calling `poll` the data is already available. What do you think?

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -0,0 +1,264 @@
...
+                .orElseThrow(() -> new ServiceUnavailableException(
+                        "It is expected that partition statistics exists for timeline " + electedTimeline +
+                                " and partition " + partition + ", but it wasn't found")).getBeforeFirst();
+    }
+
+    private void reelectTopicRepositories() throws NakadiException, InvalidCursorException {
+        final Map<TopicRepository, List<NakadiCursor>> newAssignment = new HashMap<>();
+        borderOffsets.clear();
+        // load new topic repositories and possibly replace cursors to newer timelines.
+        for (final NakadiCursor cursor : latestOffsets.values()) {
```
rcillo: Are these `latestOffsets`  the `latestConsumedOffsets`? What do you think about renaming it?

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @rcillo 
There is already latestOffsets in this class - which points to latest offsets that were sent to user of this class. 
borderOffsets - is a list of offsets that should trigger timeline switch while reaching them, for example, if I'm streaming partiton `test:0`, and I know, that there are 3 timelines there with next configuraion:
```
timeline | start | latest
   t0    |   0   |  120
   t1    |   0   |  100
   t2    |   0   |   - 
```
then in case if i'm on `t0`, border will be 120. Once border is reached, active timeline changed to next one (t1), border will become 100, then again after reaching border - switch. after that - there is no border anymore.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
rcillo: As far as I could see, the only task added here is `timelinesChanged`. That means that in fact, this task abstraction serves a single purposed: if there is a timeline change, react accordingly.

Do you plan to add any other kind of tasks? Is it possible to queue multiple timeline changes? In case there are multiple timeline changes queued, isn't it enough to react on a single task, maybe the latest one?

Overall, I think that abstractions like this make it harder to understand the code and they allow for some flexibility (adding several tasks or different types of tasks) which is not used so far.

It would be clearer for me if it did something more straighforward like `hasTimelineChanged` and then call something `switchTimeline`. It informs me more about what's going on than reading `queuedTask.poll` and `task.run()` which are very generic.

rcillo: Or maybe just document what are the possible tasks so that the reader can understand what's going on without the need to read the entire class.

v-stepanov:
👍

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
rcillo: maybe rename it to `eventTypesTimelines` so to make it clear that it contains a map from event types names to timelines.

src/test/java/org/zalando/nakadi/repository/MultiTimelineEventConsumerTest.java
```diff
@@ -0,0 +1,193 @@
...
+import org.zalando.nakadi.service.timeline.TimelineService;
+import org.zalando.nakadi.service.timeline.TimelineSync;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+public class MultiTimelineEventConsumerTest {
```
rcillo: I can feel the pain of writing this test with dozens of mocks. Wouldn't it be better to have single acceptance test that stream from a topic with 2 timelines and after reading everything creates a third timeline to observe that it reacts to the creation of a new timeline on-the-fly?

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
rcillo: @antban  Thank you for the explanation. I get it now. But the next person to read this code will have the same question and will have to go through the entire class to figure out what `borderOffset` is about. That's why I suggest to add some comment in the same line of your answer "borderOffsets - is a list of offsets that should trigger timeline switch while reaching them".

src/test/java/org/zalando/nakadi/repository/MultiTimelineEventConsumerTest.java
rcillo: @antban I just saw you comment mentioning that a full acceptance test in planned for the future. So no need to do it here https://github.com/zalando/nakadi/pull/602#issuecomment-287804799

rcillo:
@antban I made several comments, all of them regarding clarity. I had questions when reading the code for the fist time and I think this first impression feedback is very valuable to improve readability and easy of understanding. But anyway, they do not change the behaviour so I have nothing against merging this and refactoring in the future.

rcillo:
👍

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @rcillo good point

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @rcillo 
For me this line looks like:
`if (border is set and border <= currentOffset) then do something`
and for me it is was obvious not to wrap this simple check to variable. I'll do that

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @rcillo 
For me election `is a formal decision-making process by which a population chooses an individual` (from wikipedia), and it is exactly what it does right now. The only thing I can do is to remove `re`.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @rcillo I thought about that. But in fact in almost all the cases there will be only one TopicRepository. I don't see any reason to optimize it even more.

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @rcillo In fact there would be no such problem if one just get rid of queue and Optional<ConsumedEvent> at all. I think I will do it, because it really simplifies development.

src/test/java/org/zalando/nakadi/repository/MultiTimelineEventConsumerTest.java
antban: @rcillo But before that I wanted to check two situations in which this thing should work. I'll remove test once Acceptance test is ready

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @rcillo ok

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
antban: @rcillo I will add comment

antban:
@rcillo @v-stepanov 
I didn't want to do that now, but for me right now it is definitely clear, that one should change the overall behavior of EventConsumer.
Approach with receiving one event at a time and having queues of events or each step just exhausted itself.
As a consequence of that approach I believe we will have CPU usage drop and overall simplification across all the code.

rcillo:
@antban it looks good 👍 but if you want to further improve it, I'm not against it =)

rcillo:
👍

antban:
:+1:

antban:
:+1:

lmontrieux:
👍

antban:
:+1:

lmontrieux:
👍

antban:
:+1:

src/main/java/org/zalando/nakadi/repository/MultiTimelineEventConsumer.java
```diff
@@ -134,13 +133,23 @@ private TopicRepository selectCorrectTopicRepo(
...
+            } else if (latest.compareTo(cursor) > 0) {
+                // There is a border case - latest is equal to begin (that means that there are no available events
+                // there), and one should position on timeline that have something inside.
+                final NakadiCursor firstItem = timelineService.getTopicRepository(toCheck)
+                        .loadPartitionStatistics(toCheck, cursor.getPartition())
+                        .get().getFirst();
+                if (latest.compareTo(firstItem) > 0) {
+                    electedTimeline = toCheck;
+                } else {
+                    LOG.info("Timeline {} doesn't have any data inside, skipping", toCheck);
```
lmontrieux: `doesn't have any data inside` -> `is empty`

lmontrieux:
Apart from the error message, this looks fine to me, so I'll approve when the message is fixed.

lmontrieux:
👍

lmontrieux:
👍

antban:
:+1:

PR #603, opened 2017-03-10
aruha-600: api change for reseting subscription cursor

adyach:
In order to be able to reset cursor for given subscription we need api to be adjusted

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -871,6 +871,35 @@ paths:
           description: Unprocessable Entity
           schema:
             $ref: '#/definitions/Problem'
+    patch:
+      tags:
+        - subscription-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        Endpoint for reseting subscription committed offsets to provided.
+      parameters:
+        - name: subscription_id
+          in: path
+          type: string
+          description: Id of subscription
+          required: true
+        - name: cursors
+          in: body
+          schema:
+            type: object
+            properties:
+              items:
+                description: |
+                  List of cursors that the consumer acknowledges to have successfully processed.
+                type: array
+                items:
+                  $ref: '#/definitions/SubscriptionCursor'
+            required:
+              - items
+      responses:
+        '200':
+          description: Offsets were reseted
 
   /subscriptions/{subscription_id}/events:
     get:
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -877,7 +877,10 @@ paths:
       security:
         - oauth2: ['nakadi.event_stream.read']
       description: |
-        Endpoint for reseting subscription offsets.
+        Endpoint for resetting subscription offsets.
+        The subscription consumers will be disconnected during offset reset. The request can hang up until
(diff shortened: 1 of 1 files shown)


api/nakadi-event-bus-api.yaml
```diff
@@ -871,6 +871,35 @@ paths:
           description: Unprocessable Entity
           schema:
             $ref: '#/definitions/Problem'
+    patch:
+      tags:
+        - subscription-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        Endpoint for reseting subscription committed offsets to provided.
```
antban: It's too hard to understand what it means...

api/nakadi-event-bus-api.yaml
```diff
@@ -871,6 +871,35 @@ paths:
...
+                description: |
+                  List of cursors that the consumer acknowledges to have successfully processed.
+                type: array
+                items:
+                  $ref: '#/definitions/SubscriptionCursor'
+            required:
+              - items
+      responses:
+        '200':
+          description: Offsets were reseted
```
antban: I think that at least 2 possible errors should be described:
 - Subscription not found
 - cursor list is not complete

And why 200 but not 204? There is no content there in response.

adyach: I think these are common response codes, that's why I have not mentioned them here to not litter.

adyach: @antban 56ee35a

api/nakadi-event-bus-api.yaml
```diff
@@ -871,6 +871,35 @@ paths:
...
+          in: body
+          schema:
+            type: object
+            properties:
+              items:
+                description: |
+                  List of cursors to reset to.
+                type: array
+                items:
+                  $ref: '#/definitions/SubscriptionCursor'
```
v-stepanov: I think it makes sense to use `SubscriptionCursorWithoutToken` here instead, because user will not really have a token when resetting cursors.

api/nakadi-event-bus-api.yaml
```diff
@@ -871,6 +871,35 @@ paths:
...
+            required:
+              - items
+      responses:
+        '204':
+          description: Offsets were reseted
```
v-stepanov: reset~~ed~~

api/nakadi-event-bus-api.yaml
v-stepanov: I also think that it makes sense to describe these two cases in api (subscription not found and incomplete list of cursors) - because here it is part of behavior of our endpoint.

adyach:
@antban @v-stepanov requested fix, please, take a look

antban:
:+1:

api/nakadi-event-bus-api.yaml
```diff
@@ -871,6 +871,46 @@ paths:
...
+          description: Id of subscription
+          required: true
+        - name: cursors
+          in: body
+          schema:
+            type: object
+            properties:
+              items:
+                description: |
+                  List of cursors to reset subscription to.
```
v-stepanov: will we allow to reset only some of the cursors and leave others unchanged?

adyach: Yes, I think so

adyach:
:shipit:

PR #604, opened 2017-03-10
aruha-701 make commit timeout configurable by client

### LINKED ISSUE #594: Make commit timeout configurable
If a subscription API consumer doesn't commit a batch within 60s, Nakadi considers the connection as broken.

*As* a consumer of the subscription API
*I want to* be able to configure this setting to another value (e.g. 5min)
*Such that* I don't break if I need longer and can e.g. implement back pressure.

##### Explanation for the back pressure:
We consume events and write them into a datastore. If our datastore becomes slow, we might come into a situation where we cannot write events in the same pace to the datastore in which we consume them from Nakadi. Hence, we throttle Nakadi consumption by delaying the commit.
Is there a better way to do that?

rcillo:
Fix #594

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
+++ b/src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
@@ -16,6 +16,7 @@ import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.servlet.mvc.method.annotation.StreamingResponseBody;
 import org.zalando.nakadi.config.NakadiSettings;
 import org.zalando.nakadi.exceptions.NakadiException;
+import org.zalando.nakadi.exceptions.UnprocessableEntityException;
 import org.zalando.nakadi.security.Client;
 import org.zalando.nakadi.service.BlacklistService;
 import org.zalando.nakadi.service.ClosedConnectionsCrutch;
@@ -128,6 +129,7 @@ public class SubscriptionStreamController {
             @Nullable @RequestParam(value = "stream_timeout", required = false) final Long streamTimeout,
             @Nullable
             @RequestParam(value = "stream_keep_alive_limit", required = false) final Integer streamKeepAliveLimit,
+            @RequestParam(value = "commit_timeout", required = false, defaultValue = "60") final int commitTimeout,
             final HttpServletRequest request, final HttpServletResponse response, final Client client)
             throws IOException {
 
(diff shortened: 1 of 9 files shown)


src/main/resources/application.yml
```diff
@@ -49,7 +49,9 @@ nakadi:
       rotationMs: 86400000 # 1 day
   stream:
     timeoutMs: 31536000000 # 1 year :-P
-    default.commitTimeout: 60 # 1 minute
+    default:
+      minCommitTimeout: 60 # 1 minute
```
antban: I think it makes sense to use slightly smaller value - 5 seconds for example

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -152,9 +154,18 @@ public StreamingResponseBody streamEvents(
                     return;
                 }
 
+                if (commitTimeout < nakadiSettings.getMinCommitTimeoutSeconds()
+                        || commitTimeout > nakadiSettings.getMaxCommitTimeoutSeconds()) {
+                    final String errorMessage = "commit_timeout must be bigger than "
```
antban: bitter or equal (or not smaller), the same for `smaller than`

src/main/resources/application.yml
```diff
@@ -49,7 +49,9 @@ nakadi:
       rotationMs: 86400000 # 1 day
   stream:
     timeoutMs: 31536000000 # 1 year :-P
-    default.commitTimeout: 60 # 1 minute
+    default:
```
v-stepanov: I see only min and max but what is the default?

rcillo: The default is now only visible in the controller, where I used `defaultValue` for the query parameter.

antban: @rcillo Maybe it's better to make it configurable?

rcillo: Well, we already use this approach for other default values. But I agree that having all configurations in a single file would be ideal. If I'm doing it here, maybe we should consider no longer use `defaultValue` feature from Spring.

rcillo:
We could not come to an agreement on what to do with this PR. There are strong cons to enabling this feature:

1. users could inadvertently block themselves from consuming for 5 minutes (max value).

So, this feature alone is more dangerous than valuable. It should be made available in case there was a complementary feature in place, to forcefully disconnect clients, so to be able to free slots in some emergency situations.

For the time being, I'll close this PR and keep the issue that motivated it open, so that we can provide a better (more complete) solution in the future.

PR #605, opened 2017-03-11
Consume payload as byte[] instead of java.lang.String to reduce GC overhead and to improve performance

lhotari:
- currently consuming deserializes the payload to a `java.lang.String` instance. a `String` instance has to be converted back to a `byte[]` when it's sent to REST clients. 
- This PR uses `org.apache.kafka.common.serialization.ByteArrayDeserializer` as the `value.deserializer` so that the value returned by the consumer is a `byte[]` instead of a `java.lang.String`.
- Besides this change, it makes the writing of events to clients use a streaming approach in `org.zalando.nakadi.service.EventStream` and `org.zalando.nakadi.service.subscription.state.StreamingState`. 
  - Apache Commons IO's  `org.apache.commons.io.output.CountingOutputStream` is used to calculate the number of bytes written to the stream. `java.io.OutputStreamWriter` is used for writing String instances to the output stream.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventStream.java
+++ b/src/main/java/org/zalando/nakadi/service/EventStream.java
@@ -2,6 +2,7 @@ package org.zalando.nakadi.service;
 
 import com.codahale.metrics.Meter;
 import com.google.common.collect.Lists;
+import org.apache.commons.io.output.CountingOutputStream;
 import org.apache.kafka.common.KafkaException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -11,6 +12,8 @@ import org.zalando.nakadi.repository.EventConsumer;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
 import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Map;
@@ -18,6 +21,7 @@ import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+
 import org.zalando.nakadi.view.Cursor;
 
 import static java.lang.System.currentTimeMillis;
@@ -54,7 +58,7 @@ public class EventStream {
             int messagesRead = 0;
             final Map<String, Integer> keepAliveInARow = createMapWithPartitionKeys(partition -> 0);
 
-            final Map<String, List<String>> currentBatches =
+            final Map<String, List<byte[]>> currentBatches =
                     createMapWithPartitionKeys(partition -> Lists.newArrayList());
             // Partition to NakadiCursor.
             final Map<String, NakadiCursor> latestOffsets = config.getCursors().stream().collect(
@@ -142,30 +146,47 @@ public class EventStream {
                 .collect(Collectors.toMap(identity(), valueFunction));
     }
 
-    public static String createStreamEvent(final Cursor cursor, final List<String> events) {
-        final StringBuilder builder = new StringBuilder()
-                .append("{\"cursor\":{\"partition\":\"").append(cursor.getPartition())
-                .append("\",\"offset\":\"").append(cursor.getOffset()).append("\"}");
+    public static int writeStreamEvent(final OutputStream outputStream, final Cursor cursor, final List<byte[]> events)
+            throws IOException {
+        final CountingOutputStream countingOutputStream = new CountingOutputStream(outputStream) {
+            @Override
+            public void flush() throws IOException {
+                // block flushing
+            }
+        };
+        final Writer writer = new OutputStreamWriter(countingOutputStream);
+        writer.write("{\"cursor\":{\"partition\":\"");
+        writer.write(cursor.getPartition());
+        writer.write("\",\"offset\":\"");
+        writer.write(cursor.getOffset());
+        writer.write("\"}");
         if (!events.isEmpty()) {
-            builder.append(",\"events\":[");
-            events.forEach(event -> builder.append(event).append(","));
-            builder.deleteCharAt(builder.length() - 1).append("]");
+            writer.write(",\"events\":[");
+            writer.flush();
+            boolean first = true;
+            for (final byte[] event : events) {
+                if (!first) {
+                    countingOutputStream.write(',');
+                }
+                countingOutputStream.write(event);
+                first = false;
+            }
+            countingOutputStream.write(']');
         }
-
-        builder.append("}").append(BATCH_SEPARATOR);
-
-        return builder.toString();
+        writer.write("}");
+        writer.write(BATCH_SEPARATOR);
+        writer.flush();
+        outputStream.flush();
+        return countingOutputStream.getCount();
     }
 
-    private void sendBatch(final NakadiCursor topicPosition, final List<String> currentBatch)
+    private void sendBatch(final NakadiCursor topicPosition, final List<byte[]> currentBatch)
             throws IOException {
         // create stream event batch for current partition and send it; if there were
         // no events, it will be just a keep-alive
-        final String streamEvent = createStreamEvent(cursorConverter.convert(topicPosition), currentBatch);
-        final byte[] batchBytes = streamEvent.getBytes(UTF8);
-        outputStream.write(batchBytes);
-        bytesFlushedMeter.mark(batchBytes.length);
-        outputStream.flush();
+        final int batchBytesCount = writeStreamEvent(outputStream, cursorConverter.convert(topicPosition),
+                currentBatch);
+        bytesFlushedMeter.mark(batchBytesCount);
     }
 
     public void close() throws IOException {
--- a/src/main/java/org/zalando/nakadi/service/EventStream.java
+++ b/src/main/java/org/zalando/nakadi/service/EventStream.java
@@ -14,7 +14,6 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
-import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -25,13 +24,13 @@ import java.util.stream.Collectors;
 import org.zalando.nakadi.view.Cursor;
 
 import static java.lang.System.currentTimeMillis;
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.function.Function.identity;
 
 public class EventStream {
 
     private static final Logger LOG = LoggerFactory.getLogger(EventStream.class);
     public static final String BATCH_SEPARATOR = "\n";
-    public static final Charset UTF8 = Charset.forName("UTF-8");
 
     private final OutputStream outputStream;
     private final EventConsumer eventConsumer;
@@ -154,7 +153,7 @@ public class EventStream {
                 // block flushing
             }
         };
-        final Writer writer = new OutputStreamWriter(countingOutputStream);
+        final Writer writer = new OutputStreamWriter(countingOutputStream, UTF_8);
         writer.write("{\"cursor\":{\"partition\":\"");
         writer.write(cursor.getPartition());
         writer.write("\",\"offset\":\"");
--- a/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
@@ -32,6 +32,8 @@ import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 
 class StreamingState extends State {
     private final Map<Partition.PartitionKey, PartitionData> offsets = new HashMap<>();
@@ -230,7 +232,7 @@ class StreamingState extends State {
                 // block flushing
             }
         };
-        final Writer writer = new OutputStreamWriter(countingOutputStream);
+        final Writer writer = new OutputStreamWriter(countingOutputStream, UTF_8);
 
         writer.write("{\"cursor\":");
         writer.write(cursorSerialized);
--- a/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
@@ -1,20 +1,23 @@
 package org.zalando.nakadi.service.subscription.state;
 
 import com.codahale.metrics.Meter;
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.google.common.base.Preconditions;
+import org.apache.commons.io.output.CountingOutputStream;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecords;
 import org.apache.kafka.common.TopicPartition;
 import org.slf4j.LoggerFactory;
+import org.zalando.nakadi.domain.Timeline;
 import org.zalando.nakadi.metrics.MetricUtils;
 import org.zalando.nakadi.service.EventStream;
 import org.zalando.nakadi.service.subscription.model.Partition;
 import org.zalando.nakadi.service.subscription.zk.ZKSubscription;
 import org.zalando.nakadi.view.SubscriptionCursor;
-import org.zalando.nakadi.domain.Timeline;
 
 import java.io.IOException;
+import java.io.OutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -37,7 +40,7 @@ class StreamingState extends State {
     // correctly, and p0 is not receiving any updates - reassignment won't complete.
     private final Map<Partition.PartitionKey, Long> releasingPartitions = new HashMap<>();
     private ZKSubscription topologyChangeSubscription;
-    private Consumer<String, String> kafkaConsumer;
+    private Consumer<String, byte[]> kafkaConsumer;
     private boolean pollPaused;
     private long lastCommitMillis;
     private long committedEvents;
@@ -135,7 +138,7 @@ class StreamingState extends State {
             scheduleTask(this::pollDataFromKafka, getKafkaPollTimeout(), TimeUnit.MILLISECONDS);
             return;
         }
-        final ConsumerRecords<String, String> records = kafkaConsumer.poll(getKafkaPollTimeout());
+        final ConsumerRecords<String, byte[]> records = kafkaConsumer.poll(getKafkaPollTimeout());
         if (!records.isEmpty()) {
             for (final TopicPartition tp : records.partitions()) {
                 final Partition.PartitionKey pk = new Partition.PartitionKey(tp.topic(),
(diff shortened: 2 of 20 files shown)


lhotari:
I did some profiling after applying this PR. It looks pretty good. No allocation hotspots seem to show up from the EventStream class. I used Java Flight Recorder for profiling.

One way to enable JFR is to add these JVM args to the bootRun task in build.gradle:
```
bootRun {
    addResources = false
    main = 'org.zalando.nakadi.Application'
    environment 'NAKADI_OAUTH2_MODE', 'OFF'
    environment 'NAKADI_FEATURETOGGLE_DEFAULT', 'true'

    jvmArgs "-XX:+UnlockCommercialFeatures", "-XX:+FlightRecorder", "-XX:+UnlockDiagnosticVMOptions", "-XX:+DebugNonSafepoints",
      "-XX:FlightRecorderOptions=stackdepth=1024,defaultrecording=true,settings=profile,disk=true,maxsize=500M,dumponexit=true"
}
```

Some bottlenecks I noticed in profiling:
```
org.zalando.nakadi.service.BlacklistService.isBlocked(BlacklistService$Type, String) - Calls to Curator are slow
org.zalando.nakadi.repository.kafka.KafkaCursor.toNakadiOffset(long) - String.format is slow
org.zalando.nakadi.service.ClosedConnectionsCrutch.getCurrentConnections(InputStream) - creates a large number of IOExceptions
org.apache.kafka.common.utils.Crc32.update(byte[], int, int) - https://issues.apache.org/jira/browse/KAFKA-3174
```

lhotari:
I created PRs #606 , #607 and #608 to optimize the hotspots I found in Java Flight Recorder profiling.

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -142,30 +146,47 @@ public void streamEvents(final AtomicBoolean connectionReady) {
...
-                .append("\",\"offset\":\"").append(cursor.getOffset()).append("\"}");
+    public static int writeStreamEvent(final OutputStream outputStream, final Cursor cursor, final List<byte[]> events)
+            throws IOException {
+        final CountingOutputStream countingOutputStream = new CountingOutputStream(outputStream) {
+            @Override
+            public void flush() throws IOException {
+                // block flushing
+            }
+        };
+        final Writer writer = new OutputStreamWriter(countingOutputStream);
```
valgog: The UTF8 `Charset` should be used here as a second parameter.

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -142,30 +146,47 @@ public void streamEvents(final AtomicBoolean connectionReady) {
...
         }
-
-        builder.append("}").append(BATCH_SEPARATOR);
-
-        return builder.toString();
+        writer.write("}");
+        writer.write(BATCH_SEPARATOR);
+        writer.flush();
+        outputStream.flush();
+        return countingOutputStream.getCount();
```
valgog: I am wondering if the correct way would be to use `getByteCount()` and `long` here and in the counters later.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -223,18 +224,39 @@ private String serializeBatch(final Partition.PartitionKey partitionKey, final l
...
-
-        builder.append("}").append(EventStream.BATCH_SEPARATOR);
-        return builder.toString();
+        if (metadata.isPresent()) {
+            writer.write(",\"info\":{\"debug\":\"");
+            writer.write(metadata.get());
+            writer.write("\"}");
+            writer.flush();
+        }
+        writer.append("}").append(EventStream.BATCH_SEPARATOR);
```
valgog: Why `append` and not `write`?

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -223,18 +224,39 @@ private String serializeBatch(final Partition.PartitionKey partitionKey, final l
...
         }
-        metadata.ifPresent(s -> builder.append(",\"info\":{\"debug\":\"").append(s).append("\"}"));
-
-        builder.append("}").append(EventStream.BATCH_SEPARATOR);
-        return builder.toString();
+        if (metadata.isPresent()) {
+            writer.write(",\"info\":{\"debug\":\"");
+            writer.write(metadata.get());
+            writer.write("\"}");
+            writer.flush();
```
valgog: Why to flush here?

src/test/java/org/zalando/nakadi/service/subscription/StreamingContextTest.java
```diff
@@ -29,8 +30,8 @@ public void onException(final Exception ex) {
             }
 
             @Override
-            public void streamData(final byte[] data) throws IOException {
-
+            public OutputStream getOutputStream() {
+                return null;
```
dehora: Based on this, is it right to say there's no test coverage on the new outputstream use?

dehora:
> I created PRs #606 , #607 and #608 to optimize the hotspots I found in Java Flight Recorder profiling.

@lhotari can you open issues for each of those so we can track them, thanks!

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -142,30 +146,47 @@ public void streamEvents(final AtomicBoolean connectionReady) {
...
-    public static String createStreamEvent(final Cursor cursor, final List<String> events) {
-        final StringBuilder builder = new StringBuilder()
-                .append("{\"cursor\":{\"partition\":\"").append(cursor.getPartition())
-                .append("\",\"offset\":\"").append(cursor.getOffset()).append("\"}");
+    public static int writeStreamEvent(final OutputStream outputStream, final Cursor cursor, final List<byte[]> events)
+            throws IOException {
+        final CountingOutputStream countingOutputStream = new CountingOutputStream(outputStream) {
+            @Override
+            public void flush() throws IOException {
+                // block flushing
```
lhotari: Since this is used as a wrapper, it will just block the flush call from the wrapped OutputStream to propagate to the "upstream". It would be cleaner code to have a separate wrapper for blocking the flushes. 

The reason to do this is that when the OutputStreamWriter is flushed so that that flush call doesn't propagate and cause unwanted side effects. OutputStreamWriter has to be flushed after it's use is over since it will internally buffer bytes. 

I guess that extracting more classes and adding more unit tests would be necessary to make the code and logic understandable.

src/main/java/org/zalando/nakadi/service/EventStream.java
lhotari: Yes. I'll pass `java.nio.charset.StandardCharsets.UTF_8` as the 2nd parameter.

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -142,30 +146,47 @@ public void streamEvents(final AtomicBoolean connectionReady) {
...
         if (!events.isEmpty()) {
-            builder.append(",\"events\":[");
-            events.forEach(event -> builder.append(event).append(","));
-            builder.deleteCharAt(builder.length() - 1).append("]");
+            writer.write(",\"events\":[");
+            writer.flush();
+            boolean first = true;
+            for (final byte[] event : events) {
+                if (!first) {
+                    countingOutputStream.write(',');
```
lhotari: the `writer` writes to `countingOutputStream` so it's the whole size. The `writer.flush()` is called before switching to use `countingOutputStream` directly so that the internal buffers of the `writer` get flushed before that.

src/main/java/org/zalando/nakadi/service/EventStream.java
lhotari: If we'd like to support over 2GB... :) I guess it would be good practise to use a `long` for a byte counter. However it might be irrelevant for the byte counter of an event so I'm not yet acting on this. I think this event writing logic should be extracted to a separate class so that it can be shared also in `StreamingState` which currently has duplicate logic.

src/test/java/org/zalando/nakadi/repository/kafka/NakadiKafkaConsumerTest.java
```diff
@@ -98,18 +98,18 @@ public void whenCreateConsumerThenKafkaConsumerConfiguredCorrectly() {
     public void whenReadEventsThenGetRightEvents() {
 
         // ARRANGE //
-        final String event1 = randomString();
-        final String event2 = randomString();
+        final byte[] event1 = randomString().getBytes();
```
lhotari: I made this consistent now. Ran a check for `getBytes()` in the code base.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
lhotari: That's explained in an earlier comment on this PR. The reason is that the possibly buffered bytes in the `OutputStreamWriter` have to be flushed before switching to write to the `OutputStream` which by-passes the `OutputStreamWriter`.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
lhotari: I guess that just slipped there. I guess it would be consistent to use `write`. I'll extract the stream writing logic from `EventStream` and `StreamingState` and make it consistent.

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -11,13 +12,16 @@
...
+import java.io.OutputStreamWriter;
+import java.io.Writer;
 import java.nio.charset.Charset;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+
```
lhotari: I'm using IntelliJ and the current formatting settings add it and that's why it slips easily in the changes. Do you have an advice how to setup IntelliJ styles properly so that it doesn't mess the imports?

src/test/java/org/zalando/nakadi/repository/kafka/NakadiKafkaConsumerTest.java
```diff
@@ -98,18 +98,18 @@ public void whenCreateConsumerThenKafkaConsumerConfiguredCorrectly() {
...
+        final byte[] event2 = randomString().getBytes();
```
lhotari: Done.

src/test/java/org/zalando/nakadi/service/EventStreamTest.java
```diff
@@ -46,7 +49,7 @@
 public class EventStreamTest {
 
     private static final String TOPIC = randomString();
-    private static final String DUMMY = "DUMMY";
+    private static final byte[] DUMMY = "DUMMY".getBytes();
```
lhotari: Done.

src/test/java/org/zalando/nakadi/service/EventStreamTest.java
```diff
@@ -226,7 +229,8 @@ public void whenReadingEventsTheOrderIsCorrect() throws NakadiException, IOExcep
                 .range(0, eventNum)
                 .boxed()
                 .map(index -> new ConsumedEvent(
-                        "event" + index, new NakadiCursor(TIMELINE, "0", KafkaCursor.toNakadiOffset(index))))
+                        ("event" + index).getBytes(), new NakadiCursor(TIMELINE, "0",
```
lhotari: Done.

src/test/java/org/zalando/nakadi/service/EventStreamTest.java
```diff
@@ -327,16 +332,16 @@ private static NakadiKafkaConsumer endlessDummyConsumer() throws NakadiException
...
     }
 
     private static String jsonBatch(final String partition, final String offset,
-                                    final Optional<List<String>> eventsOrNone, final Optional<String> metadata) {
+                                    final Optional<List<byte[]>> eventsOrNone, final Optional<String> metadata) {
         final String eventsStr = eventsOrNone
                 .map(events -> {
                     final StringBuilder builder = new StringBuilder(",\"events\":[");
-                    events.forEach(event -> builder.append("\"").append(event).append("\","));
+                    events.forEach(event -> builder.append("\"").append(new String(event)).append("\","));
```
lhotari: I passed UTF-8 Charset to the constructor for now. What kind of test would you like to add?

src/test/java/org/zalando/nakadi/service/subscription/state/PartitionDataTest.java
```diff
@@ -38,7 +40,7 @@ public void onNewOffsetsShouldSupportCommitInFuture() {
     public void normalOperationShouldNotReconfigureKafkaConsumer() {
         final PartitionData pd = new PartitionData(null, 100L);
         for (long i = 0; i < 100; ++i) {
-            pd.addEventFromKafka(100L + i + 1, "test_" + i);
+            pd.addEventFromKafka(100L + i + 1, ("test_" + i).getBytes());
```
lhotari: Done.

src/test/java/org/zalando/nakadi/service/subscription/state/PartitionDataTest.java
```diff
@@ -55,24 +57,24 @@ public void normalOperationShouldNotReconfigureKafkaConsumer() {
     public void eventsMustBeReturnedInGuaranteedOrder() {
         final PartitionData pd = new PartitionData(null, 100L);
         for (long i = 0; i < 100; ++i) {
-            pd.addEventFromKafka(200L - i, "test_" + (200L - i));
+            pd.addEventFromKafka(200L - i, ("test_" + (200L - i)).getBytes());
```
lhotari: Done.

src/test/java/org/zalando/nakadi/service/subscription/state/PartitionDataTest.java
```diff
@@ -55,24 +57,24 @@ public void normalOperationShouldNotReconfigureKafkaConsumer() {
...
-            final SortedMap<Long, String> data = pd.takeEventsToStream(currentTimeMillis(), 10, 0L);
+            final SortedMap<Long, byte[]> data = pd.takeEventsToStream(currentTimeMillis(), 10, 0L);
             assertNotNull(data);
             assertEquals(10, data.size());
             assertEquals((i + 1) * 10, pd.getUnconfirmed());
             assertEquals(0, pd.getKeepAliveInARow());
             assertEquals(100L + i * 10L + 1L, data.firstKey().longValue());
             assertEquals(100L + i * 10L + 10L, data.lastKey().longValue());
-            data.forEach((k, v) -> assertEquals("test_" + k, v));
+            data.forEach((k, v) -> assertArrayEquals(("test_" + k).getBytes(), v));
```
lhotari: Done.

src/test/java/org/zalando/nakadi/service/subscription/state/PartitionDataTest.java
```diff
@@ -55,24 +57,24 @@ public void normalOperationShouldNotReconfigureKafkaConsumer() {
...
         }
-        final SortedMap<Long, String> data = pd.takeEventsToStream(currentTimeMillis(), 10, 0L);
+        final SortedMap<Long, byte[]> data = pd.takeEventsToStream(currentTimeMillis(), 10, 0L);
         assertNotNull(data);
         assertEquals(1, data.size());
         assertEquals(201L, data.firstKey().longValue());
-        assertEquals("fake", data.get(data.firstKey()));
+        assertArrayEquals("fake".getBytes(), data.get(data.firstKey()));
```
lhotari: Done.

src/test/java/org/zalando/nakadi/service/subscription/state/PartitionDataTest.java
```diff
@@ -96,9 +98,9 @@ public void eventsShouldBeStreamedOnTimeout() throws InterruptedException {
         final long timeout = TimeUnit.SECONDS.toMillis(1);
         final PartitionData pd = new PartitionData(null, 100L);
         for (int i = 0; i < 100; ++i) {
-            pd.addEventFromKafka(i + 100L + 1, "test");
+            pd.addEventFromKafka(i + 100L + 1, "test".getBytes());
```
lhotari: Done.

src/test/java/org/zalando/nakadi/service/subscription/state/PartitionDataTest.java
```diff
@@ -108,7 +110,7 @@ public void eventsShouldBeStreamedOnTimeout() throws InterruptedException {
         assertEquals(100, data.size());
 
         for (int i = 100; i < 200; ++i) {
-            pd.addEventFromKafka(i + 100L + 1, "test");
+            pd.addEventFromKafka(i + 100L + 1, "test".getBytes());
```
lhotari: Done.

src/test/java/org/zalando/nakadi/service/subscription/state/PartitionDataTest.java
```diff
@@ -125,10 +127,10 @@ public void eventsShouldBeStreamedOnBatchSize() {
         final long timeout = TimeUnit.SECONDS.toMillis(1);
         final PartitionData pd = new PartitionData(null, 100L);
         for (int i = 0; i < 100; ++i) {
-            pd.addEventFromKafka(i + 100L + 1, "test");
+            pd.addEventFromKafka(i + 100L + 1, "test".getBytes());
```
lhotari: Done.

src/test/java/org/zalando/nakadi/service/subscription/StreamingContextTest.java
lhotari: Just saying that there were no unit tests for the previous `streamData` method either. :) Adding more test coverage makes sense.

lhotari:
>> I created PRs #606 , #607 and #608 to optimize the hotspots I found in Java Flight Recorder profiling.
>
> @lhotari can you open issues for each of those so we can track them, thanks!

This is a bit confusing to me since PRs are a type of issue in Github, for example the numbering is shared between issues and PRs. I'm just wondering if a 1-to-1 mapping between an issue and a PR makes sense. 
What do you recommend in this case?

src/main/java/org/zalando/nakadi/service/EventStream.java
dehora: @lhotari I have the same setup (using [square's style guide](https://github.com/square/java-code-styles)).  I'm not sure the project has a formatting guideline, so I just try to remove/reduce incidental changes.

src/test/java/org/zalando/nakadi/repository/kafka/NakadiKafkaConsumerTest.java
dehora: @lhotari did you push the getBytes() changes? Just in case my view is stale, I'm looking at 5fd783da9b32a18693d052a4062d870f59baa007 as the latest and it's still showing bare getBytes()

lhotari: @dehora I pushed it now. I had mistakenly pushed it to a different branch in my fork.

src/test/java/org/zalando/nakadi/service/subscription/StreamingContextTest.java
dehora: > Just saying that there were no unit tests for the previous streamData method either. :) 

Fair enough! I can do some local testing on this branch and we can take it from there 😄

lmontrieux:
@lhotari @dehora am I correct that this has been implemented in #673 ? Can we close this PR?

rcillo:
@lmontrieux actually, this implementation is a little bit cleaner, since it doesn't count bytes manually, which is error prone. But the change about byte deserialiser has been merged already in the mentioned PR. So I think we could close this one.

PR #607, opened 2017-03-13
Replace slow String.format call with alternative implementation using StringUtils.leftPad

lhotari:
Profiling showed that `org.zalando.nakadi.repository.kafka.KafkaCursor.toNakadiOffset` is one of the hotspots. This can be seen in JFR/JMC's allocation profiling. For example:
```
java.util.regex.Pattern.matcher(CharSequence)
   java.util.Formatter.parse(String)
      java.util.Formatter.format(Locale, String, Object[])
         java.util.Formatter.format(String, Object[])
            java.lang.String.format(String, Object[])
               org.zalando.nakadi.repository.kafka.KafkaCursor.toNakadiOffset(long)
```
`java.lang.String.format` is very unefficient and should be avoided in hotspots. This PR replaces the usage of `String.format` with `StringUtils.leftPad` in `KafkaCursor.toNakadiOffset`.

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaCursor.java
+++ b/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaCursor.java
@@ -1,5 +1,6 @@
 package org.zalando.nakadi.repository.kafka;
 
+import org.apache.commons.lang3.StringUtils;
 import org.zalando.nakadi.domain.NakadiCursor;
 import org.zalando.nakadi.domain.Timeline;
 import org.zalando.nakadi.exceptions.runtime.InvalidCursorException;
@@ -10,6 +11,7 @@ import static org.zalando.nakadi.domain.CursorError.INVALID_FORMAT;
 import static org.zalando.nakadi.domain.CursorError.PARTITION_NOT_FOUND;
 
 public class KafkaCursor implements Comparable<KafkaCursor> {
+    private static final int CURSOR_OFFSET_LENGTH = 18;
     private final String topic;
     private final int partition;
     private final long offset;
@@ -43,7 +45,8 @@ public class KafkaCursor implements Comparable<KafkaCursor> {
     }
 
     public static String toNakadiOffset(final long offset) {
-        return offset >= 0 ? String.format("%018d", offset) : String.valueOf(offset);
+        return offset >= 0 ? StringUtils.leftPad(String.valueOf(offset), CURSOR_OFFSET_LENGTH, '0')
+                : String.valueOf(offset);
     }
 
     public static String toNakadiPartition(final int partition) {


rcillo:
Could you please share with us your profiling results (before and after)?

lhotari:
> Could you please share with us your profiling results (before and after)?

What type of comparison are you looking for? I simply made this change since this method showed up in hotspots while profiling with JFR. After making this change, this hotspot obviously went away. 
Do you have a performance test setup to demonstrate performance improvements?

valgog:
@lhotari this is a good question. Performance tests should be part of the project as some point. Do you have some best practice examples of setting such tests up?

Have you checked that the unit tests are covering this part of the code good enough?

lhotari:
> @lhotari this is a good question. Performance tests should be part of the project as some point. Do you have some best practice examples of setting such tests up?
@valgog I can recommend the performance testing approach there is at Gradle since I have personal experience of it. My previous colleage has written [a blog post covering some of the background](https://blog.gradle.org/performance-is-a-feature#performance-test-suite). Gradle's solution is opensource and [available in the Gradle codebase](https://github.com/gradle/gradle/tree/master/subprojects/performance). However it's not a generic solution for all and therefore I think it's a lot of work and effort to build a proper performance testing solution for Nakadi. Perhaps there are better examples of getting up to speed quickly, but I don't have first hand experience of such solutions. 
The general principle of the Gradle performance testing solution is that performance tests are part of the CD pipeline. Performance regressions are catched early when performance tests are automated. It's very useful to have an established way to create performance test suites for important use cases so that it's possible to measure how small changes affect the results of important use cases. 
Before getting to a fully automated performance test suite that's in the CD pipeline, I'd suggest to first start with a local development environment setup for performance tests. That would include the test data and solution to create and run performance tests and to do measurements. It's useful to work to the direction that can be automated later on. Therefore I wouldn't invest much in solutions that are very complex to setup. For a lot of use cases, it might be certainly ok to simplify the test data and setup since I'd assume that the target of performance testing is mainly the parts of the solution that can be affected by the development team. 

> Have you checked that the unit tests are covering this part of the code good enough?
There doesn't seem to be any unit tests for the org.zalando.nakadi.repository.kafka.KafkaCursor.toNakadiOffset method. What would you recommend for that?

lmontrieux:
@lhotari @rcillo @valgog There has been no progress or discussion on this PR for almost a year. However, I see that @adyach is doing some profiling work, so perhaps some of the contributions in this PR would be useful?

valgog:
@lmontrieux to say the truth, it is a common sense that a leftpad function will be faster then format (one can compare the source code). 

So changing the formatting of the offsets to a more efficient version, should be simple and safe (I just suppose, that such an important part is covered well by unit tests).

From the point of view of the profiling, another example to check, would be pgjdbc project, that has profiling set up and required for all changes.

adyach:
@lhotari could we resolve conflicts (if you are still interested) and merge it ?

lhotari:
> @lhotari could we resolve conflicts (if you are still interested) and merge it ?

@adyach I have rebased the commit and resolved the conflict.

adyach:
👍

lmontrieux:
👍

PR #608, opened 2017-03-13
Optimize BlacklistService by adding fast first level cache

lhotari:
- Zk / Curator's TreeCache is relatively slow
- first level has short expiration time (1000ms)

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/BlacklistService.java
+++ b/src/main/java/org/zalando/nakadi/service/BlacklistService.java
@@ -1,5 +1,8 @@
 package org.zalando.nakadi.service;
 
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ImmutableMap;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.recipes.cache.ChildData;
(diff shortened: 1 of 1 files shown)


adyach:
@lhotari thaks for contributing! I saw you created a couple of PRs. In general we do not have any super strick guidlines for contributers, but in my opinion it should be disscussed with the project team first, since some of the features/bugfixes could be on the way (as it is for working with byte arrays)

> Zk / Curator's TreeCache is relatively slow

I do not see any reason to tune it now. Could you elaborate on that?

lhotari:
> I do not see any reason to tune it now. Could you elaborate on that? 

I made some profiling locally with Java Flight Recorder and I noticed BlacklistService.isBlocked showing up in some hotspots. I'd assume that this is a minor hotspot. 
Do you have a way to performance test Nakadi and verify if a single change improves the performance of a certain use case?

adyach:
We have load testing tool in our armoury, but it can be done only in real env like staging. It should be in our plans to do that, but currently it is not, because we have other things to accomplish. 
Usually, before doing this kind of tasks we write concepts and do performance tests which shows us that it will signifacantly change something([example](https://docs.google.com/document/d/13O2XcU7--gWzHQB_F5d8xtBvQeQNlxG7BRl6ExawOuo/edit#heading=h.11dq0cs752ml)). From your PR it is still not clear how it will impact the performance, why at all we should it and support after all.

lhotari:
@adyach Feel free to close this PR if this is useless for you. I'm ok with that.

adyach:
As we disscussed with @lhotari in the chat I will close it for now and we will get back to it, when we have performance tests or issues with it (depends on what will come first).

PR #611, opened 2017-03-14
Aruha 535 cleanup timelines

v-stepanov:
https://techjira.zalando.net/browse/ARUHA-535


database/nakadi/10_data/04_tables/05_timeline.sql
```diff
@@ -8,6 +8,7 @@ CREATE TABLE zn_data.timeline (
   tl_switched_at     TIMESTAMP WITH TIME ZONE DEFAULT NULL,
   tl_cleanup_at      TIMESTAMP WITH TIME ZONE DEFAULT NULL,
   tl_latest_position JSONB                    DEFAULT NULL,
+  tl_deleted         BOOLEAN                  DEFAULT FALSE,
```
v-stepanov: Well, right now when event-type is deleted - it is removed from DB, so it will make no sense to keep timelines for ET that was deleted. 
But when ET still exists I think it's better to keep all timelines for the history (at least for now when timeline feature is not really tested well)

src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
```diff
@@ -116,20 +88,46 @@ public void testListTimelinesOrdered()
...
+
+        insertTimeline(tomorrow.toDate(), true, 0);
+        insertTimeline(tomorrow.toDate(), false, 1);
+        insertTimeline(yesterday.toDate(), true, 2);
+        final Timeline t4 = insertTimeline(yesterday.toDate(), false, 3);
+        final Timeline t5 = insertTimeline(twoDaysAgo.toDate(), false, 4);
+
+        final List<Timeline> expiredTimelines = tRepository.getExpiredTimelines();
+
+        Assert.assertEquals(2, expiredTimelines.size());
```
v-stepanov: Good point 👍 
I now added explicit check for NULL and extended test https://github.com/zalando/nakadi/pull/611/commits/b0f8b6b49c16ad5b4f6a33e9599757faaa6d0172

src/acceptance-test/java/org/zalando/nakadi/repository/db/TimelineDbRepositoryTest.java
```diff
@@ -116,20 +88,46 @@ public void testListTimelinesOrdered()
...
+        Assert.assertEquals(ImmutableList.of(t4, t5), expiredTimelines);
+    }
+
+    private Timeline insertTimelineWithOrder(final int order) {
+        return tRepository.createTimeline(createTimeline(
+                storage, UUID.randomUUID(), order, "test_topic", testEt.getName(),
+                new Date(), null, null, null));
+    }
+
+    private Timeline insertTimeline(final Date cleanupDate, final boolean deleted, final int order) {
```
v-stepanov: I could not find a good name for this method, so I renamed the other one :)

src/acceptance-test/java/org/zalando/nakadi/service/job/ExclusiveJobWrapperAT.java
```diff
@@ -0,0 +1,99 @@
...
+
+    @Test
+    public void whenExecuteJobThenOk() throws Exception {
+        jobWrapper.runJobLocked(dummyJob);
+        assertThat(jobExecuted.get(), is(true));
+
+        assertThat("lock node should be removed", CURATOR.checkExists().forPath(lockPath), is(nullValue()));
+
+        final byte[] data = CURATOR.getData().forPath(latestPath);
+        final DateTime lastCleaned = objectMapper.readValue(new String(data, Charsets.UTF_8), DateTime.class);
```
v-stepanov: I think the current "standard" is to use guava Charsets class.
Changed https://github.com/zalando/nakadi/pull/611/commits/c869f991250d95cd892d4e79300f3ca724b9466d

src/acceptance-test/java/org/zalando/nakadi/webservice/TimelinesControllerAT.java
```diff
@@ -6,28 +6,36 @@
...
+    private static final long RETENTION_TIME_MS = Duration.of(2, DAYS).toMillis();
 
-    @BeforeClass
-    public static void setUp() throws JsonProcessingException {
-        NakadiTestUtils.createEventTypeInNakadi(eventType);
-    }
+    private final EventType eventType;
 
-    @AfterClass
-    public static void tearDown() {
```
v-stepanov: I don't really see a reason to cleanup. In out ATs we usually don't cleanup because it will just add extra time to execution and it is not really needed as these tests are never run on real DBs. 
The only exception is UserJourneyAT, we run it on real environments and there we do cleanup after test.

src/main/java/org/zalando/nakadi/service/job/ExclusiveJobWrapper.java
```diff
@@ -0,0 +1,132 @@
...
+
+        if (!timeToRunAction()) {
+            return;
+        }
+        if (!createJobLockNode()) {
+            return;
+        }
+        // we need to check it again after we acquired the lock
+        if (!timeToRunAction()) {
+            return;
```
v-stepanov: Holy crap! Good that you noticed that 👍 
I moved this now under the `try` so that lock is released by `finally` block. 
https://github.com/zalando/nakadi/pull/611/commits/abbee68f6446b6e7484852109861eca885c0331f

src/main/java/org/zalando/nakadi/service/job/ExclusiveJobWrapper.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+        try {
+            log.info("Execution of job will be performed on this instance");
```
v-stepanov: Fixed https://github.com/zalando/nakadi/pull/611/commits/ac02cc6e1304e32d75a080522a1ad581a2bc5096

database/nakadi/10_data/04_tables/05_timeline.sql
lmontrieux: That's a good idea, but unfortunately the column `st_id` in table `timeline` is a foreign key to `st_id` in table `storage`. So we wouldn't be able to delete a storage, if there is one deleted timeline left on the storage.

v-stepanov: Ok, this is an interesting point. But I think that it should be fixed from other side. When the  storage is deleted it should also delete all timelines of this storage. And it should also check if at least one timeline is still used - then it should forbid deletion of storage.
I will add these changes to storage deletion.

lmontrieux:
👍

v-stepanov:
@lmontrieux I reworked deletion of storage. So now when storage is deleted - it also removes all timelines with deleted=true.

lmontrieux:
@v-stepanov would you like to add an acceptance test that deletes a storage containing exclusively deleted timelines?

lmontrieux:
👎  I'm blocking this for now, after this morning's discussion with @antban during SU.

We need to discuss what the best strategy is wrt deletion of timelines, storages, and event types. If we decide to mark timelines as deleted instead of actually deleting them, then perhaps it makes sense to do the same with storages and/or event types.

lmontrieux:
👎

lmontrieux:
@v-stepanov So, we discussed this on Friday, and decided that we will only allow a storage to be deleted if there are no timelines. Even if all timelines in the storage are marked as deleted, we do not  delete the storage.

src/acceptance-test/java/org/zalando/nakadi/webservice/TimelinesControllerAT.java
adyach: I don not agree with this, If test initialises resource it has to free them. It will avoid situations when tests work with the event type, which was modified in other test.

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -87,24 +105,33 @@ public StorageService(final ObjectMapper objectMapper,
...
-            if (storageDbRepository.isStorageUsed(id)) {
-                return Result.forbidden("Storage " + id + " is in use");
-            }
-            storageDbRepository.deleteStorage(id);
-        } catch (InternalNakadiException e) {
+            transactionTemplate.execute(action -> {
+                storageDbRepository.deleteStorage(id);
+                return null;
+            });
+        } catch (final NoStorageException e) {
```
adyach: Do you think we can continue with the exceptions refactoring you started in repository level and throw these exceptions to controller level catching them with exception handling mechanism?

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -87,24 +105,33 @@ public StorageService(final ObjectMapper objectMapper,
...
         try {
-            if (!storageDbRepository.getStorage(id).isPresent()) {
-                return Result.notFound("No storage with ID " + id);
-            }
-            if (storageDbRepository.isStorageUsed(id)) {
-                return Result.forbidden("Storage " + id + " is in use");
-            }
-            storageDbRepository.deleteStorage(id);
-        } catch (InternalNakadiException e) {
+            transactionTemplate.execute(action -> {
```
adyach: It is not clear why you do one operation in transaction. Could you elaborate on that?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -184,6 +188,7 @@ private void switchTimelines(final Timeline activeTimeline, final Timeline nextT
                     final Timeline.StoragePosition storagePosition =
                             topicRepositoryHolder.createStoragePosition(activeTimeline);
                     activeTimeline.setLatestPosition(storagePosition);
+                    scheduleTimelineCleanup(activeTimeline);
```
adyach: Shouldn't it be done when we create topic?

src/main/java/org/zalando/nakadi/service/job/TimelineCleanupJob.java
```diff
@@ -0,0 +1,81 @@
...
+
+    @Scheduled(
+            fixedDelayString = "${nakadi.jobs.checkRunMs}",
+            initialDelayString = "${random.int(${nakadi.jobs.checkRunMs})}")
+    public void cleanupTimelines() {
+        try {
+            jobWrapper.runJobLocked(() ->
+                    timelineDbRepository.getExpiredTimelines().stream()
+                            .forEach(timeline -> {
+                                deleteTimelineTopic(timeline);
```
adyach: In my opinion we should stop cleaning up if deletion of the topic is failed and also if we deleted the topic, but the timeline deletion is failed, what then?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -197,6 +202,20 @@ private void switchTimelines(final Timeline activeTimeline, final Timeline nextT
...
     }
 
+    private void scheduleTimelineCleanup(final Timeline timeline) throws InconsistentStateException {
+        try {
+            final EventType eventType = eventTypeCache.getEventType(timeline.getEventType());
+            final Long retentionTime = eventType.getOptions().getRetentionTime();
+            if (retentionTime == null) {
+                throw new InconsistentStateException("Event type should has information about its retention time");
+            }
+            final Date cleanupDate = new Date(System.currentTimeMillis() + retentionTime);
```
adyach: There is a possibility to change retention time of the event type. CleanupUpAt has to be changed accordingly.

src/main/java/org/zalando/nakadi/service/job/ExclusiveJobWrapper.java
```diff
@@ -0,0 +1,132 @@
...
+        this.objectMapper = objectMapper;
+        this.jobPeriodMs = jobPeriodMs;
+        this.zkPath = ZKPaths.makePath(NAKADI_JOBS_PATH, jobName);
+        this.log = LoggerFactory.getLogger("nakadi-job." + jobName);
+    }
+
+    public void runJobLocked(final Runnable action) {
+        log.info("Trying to run job...");
+
+        if (!timeToRunAction()) {
```
adyach: Could you please elaborate on the approach to run job at certain time as well as checking that it is time to run the action? I am thinking of something like, run jobs at certain time and see if there is a lock, then act or not.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
```diff
@@ -2,26 +2,31 @@
...
+
+import java.io.IOException;
+import java.sql.SQLException;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
 
 @DB
 @Repository
 public class TimelineDbRepository extends AbstractDbRepository {
```
adyach: Shouldn't this change affect methods?
- listTimelinesOrdered
- listTimelinesOrdered(final String eventType)

src/main/java/org/zalando/nakadi/service/StorageService.java
v-stepanov: I would love to do that but this PR already grew too much. But later we of course should remove this Result.

src/main/java/org/zalando/nakadi/service/StorageService.java
v-stepanov: Before there used to be 2 actions inside. As now there was only one call to DB - no need in transaction any more, so I now removed this wrapper.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
v-stepanov: yes, I know. But the plan was that it will be implemented within the ticket for updating retention time :)
In this one https://github.com/zalando/nakadi/pull/621

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
v-stepanov: Why?
When we are at this point the topic is already created.
So I thought that it will be a good place to add this when we update the previous timeline.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: interesting, at least I was not aware about it, when implementing #621 :) then I will file another ticket to tie it together in different pr. wdyt?

v-stepanov: Ok, makes sense. 
We can do it as a separate PR when both of current PRs are merged.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: It will be closer to the date, when topic was really created, so the clean up time will be more precise, honestly, so much time passed since I wrote it I do not even remember why :)

src/main/java/org/zalando/nakadi/service/job/TimelineCleanupJob.java
v-stepanov: 1) If the topic deletion failed and timeline is marked as deleted - the topic will "hang in air"
2) If the topic is deleted and the timeline is failed to be marked as deleted - then it will try again the next time Job runs.

It seems that #1 can be improved but I don't really see how. In general we can change it so that if topic failed to be removed - we don't mark timeline as deleted - so that it will try the next time. But the problem can be that the topic failed to be removed because it doesn't exist (e.g. was removed by hand or whatever), then in this case the timeline will try to be cleaned each Job run (so it will get into this infinite loop).

Do you see any ways to improve that?

src/main/java/org/zalando/nakadi/service/job/TimelineCleanupJob.java
```diff
@@ -0,0 +1,81 @@
...
+            topicRepository.deleteTopic(timeline.getTopic());
+        } catch (final TopicDeletionException e) {
+            LOG.error("Failed to delete topic {} for expired timeline {}", timeline.getTopic(), timeline.getId(), e);
+        }
+    }
+
+    private void markTimelineDeleted(final Timeline timeline) {
+        try {
+            timeline.setDeleted(true);
+            timelineDbRepository.updateTimelime(timeline);
```
v-stepanov: So here we have two failure cases:
1) Timeline failed to be updated in DB - the exception will be thrown and the cache will not try to be updated.
2) Timeline was updated in DB but the cache failed to be updated. Well there is not much to do in this case. Do you suggest to revert timeline update in this case.

src/main/java/org/zalando/nakadi/service/job/ExclusiveJobWrapper.java
v-stepanov: It is done like that because the job can be executed on any of Nakadi instances and Nakadi instances have different start time so they have different timers for this Job.
So if I would implement it just to run each N minutes - then the job can be executed just few seconds after it was executed on another Nakadi instance - so the running of Job will be quite random. That's why I need this check if it's time to run job (it checks the timestamp when the Job was executed the last time)

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
v-stepanov: Such a precision will not make much sense here as the Job that cleans timelines is run once per hour.

src/acceptance-test/java/org/zalando/nakadi/webservice/TimelinesControllerAT.java
v-stepanov: @adyach In our acceptance tests we almost never clean (except user journey).

My opinion is that cleaning-after + tests-that-assume-the-DB-is-empty is a bad approach. It will always lead to some tests that forget to clean and some other tests fail because of that.

I think the better approach is not to clean after tests + all tests assume that DB is dirty + the tests that need clean-DB clean it in `@Before`. This is actually the approach we have now in Nakadi; I think for us it works fine.

src/main/java/org/zalando/nakadi/service/job/TimelineCleanupJob.java
adyach: My point of not understanding that is `topicRepository.deleteTopic` does not delete topic form Kafka, it just schedule it for deletion (maybe we should reconsider naming). Only exception it throws is UnknownTopicOrPartitionException (according to Kafka client code), so I think it is safe to not handle this. Methods renaming will be good.

src/main/java/org/zalando/nakadi/service/job/TimelineCleanupJob.java
adyach: Yes, if the update of the cache is failed, It is better to remove rollback timeline update.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
v-stepanov: This is true. It seems that inactive timelines are not filtered out currently. I will add that.

v-stepanov: I added filtering of inactive timelines.
https://github.com/zalando/nakadi/pull/611/commits/8fcce0bda1dba198b3bd93a01bacf58bbd320661

src/main/java/org/zalando/nakadi/service/job/TimelineCleanupJob.java
v-stepanov: Ok, I added this functionality https://github.com/zalando/nakadi/pull/611/commits/69d1e9ed5a3589d65e93ff00e1a6253ba65a1adc

src/main/java/org/zalando/nakadi/service/job/TimelineCleanupJob.java
```diff
@@ -65,16 +65,31 @@ private void deleteTimelineTopic(final Timeline timeline) {
     }
 
     private void markTimelineDeleted(final Timeline timeline) {
+        boolean timelineUpdatedInDB = false;
+        boolean cacheUpdated = false;
         try {
             timeline.setDeleted(true);
             timelineDbRepository.updateTimelime(timeline);
+            timelineUpdatedInDB = true;
```
adyach: Do you really need this? In my opinion only cacheUpdated is enough.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
adyach: Maybe we also should change listTimelinesOrdered, because it is used in the cache? In that case maybe it is better to do it on db level.

src/main/java/org/zalando/nakadi/service/job/TimelineCleanupJob.java
v-stepanov: I need this to know if timeline was updated in DB. Because if it wasn't updated - then there is nothing to revert.

src/main/java/org/zalando/nakadi/repository/db/TimelineDbRepository.java
v-stepanov: After I merge with new consuming code in master I will check again all use-cases. But for now it seems for me that it's better to do the filtering only on Service layer.

v-stepanov: I merged with latest master and checked all the use-cases, the only place where we need to filter out inactive timelines is `TimelineService.getActiveTimelinesOrdered`. In all other places it is ok to have all timelines (including expired ones).

v-stepanov:
I think I now addressed all the comments.

v-stepanov:
👍

adyach:
👍

adyach:
👍

v-stepanov:
👍

adyach:
👍

PR #613, opened 2017-03-15
Aruha 531 cursor operations

rcillo:
https://techjira.zalando.net/browse/ARUHA-531

This service exposes 2 methods for operating on cursors:

- `calculateDistance`: used to calculate the distance between two cursors. Consumer lag can be calculated using this method. Just use as `initial_offset` the consumed offset and the latest available as `final_offset`.
- `unshiftCursor`: receives a cursor with a shift request and resolve it to a plain `NakadiCursor`. Allows users to move forward and backward given a reference cursor.


src/main/java/org/zalando/nakadi/domain/Timeline.java
```diff
@@ -151,6 +151,14 @@ public boolean isFake() {
         return fake;
     }
 
+    public boolean isOpen() {
+        return getSwitchedAt() == null;
+    }
+
+    public boolean isClosed() {
```
v-stepanov: I don't know why, but I feel strange having both of these methods. I think the more traditional way is to have only one of them (so that users can do `!` themselves)
But that is just my opinion.

src/main/java/org/zalando/nakadi/exceptions/runtime/InvalidCursorOperation.java
```diff
@@ -0,0 +1,22 @@
...
+    public enum Reason {
+        INVERTED_TIMELINE_ORDER,
+        TIMELINE_NOT_FOUND,
+        INVERTED_OFFSET_ORDER,
+        PARTITION_NOT_FOUND,
+        CURSORS_WITH_DIFFERENT_PARTITION
+    }
+
+    public InvalidCursorOperation(final Reason reason) {
+        super();
```
v-stepanov: I think you can skip this.

src/main/java/org/zalando/nakadi/exceptions/runtime/MyNakadiRuntimeException1.java
```diff
@@ -6,6 +6,10 @@
  */
 public class MyNakadiRuntimeException1 extends RuntimeException {
 
+    public MyNakadiRuntimeException1() {
+        super();
```
v-stepanov: I think this will be called automatically.

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
```diff
@@ -0,0 +1,169 @@
...
+        } else {
+            final long distance = getDistanceDifferentTimelines(query, initialCursor, finalCursor);
+            return new NakadiCursorDistanceResult(query, distance);
+        }
+    }
+
+    public NakadiCursor unshiftCursor(final ShiftedNakadiCursor cursor) throws InvalidCursorOperation {
+        if (cursor.getShift() < 0) {
+            return moveCursorBackwards(cursor.getTopic(), cursor.getTimeline(), cursor.getPartition(),
+                    parse(cursor.getOffset()), cursor.getShift());
```
v-stepanov: hmm, but this `parse` will work only for version-zero cursors. For version one that will fail as offset is not a number.

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.stream.Collectors;
+
+import static org.zalando.nakadi.exceptions.runtime.InvalidCursorOperation.Reason.CURSORS_WITH_DIFFERENT_PARTITION;
+import static org.zalando.nakadi.exceptions.runtime.InvalidCursorOperation.Reason.INVERTED_OFFSET_ORDER;
+import static org.zalando.nakadi.exceptions.runtime.InvalidCursorOperation.Reason.INVERTED_TIMELINE_ORDER;
+import static org.zalando.nakadi.exceptions.runtime.InvalidCursorOperation.Reason.PARTITION_NOT_FOUND;
+import static org.zalando.nakadi.exceptions.runtime.InvalidCursorOperation.Reason.TIMELINE_NOT_FOUND;
+
+public class CursorOperationsService {
+    public static final int CURSOR_OFFSET_LENGTH = 18;
```
v-stepanov: I think it duplicates `VersionZeroConverter.VERSION_ZERO_MIN_OFFSET_LENGTH`

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
v-stepanov: Ah, no. That will actually work fine. But I think that math with offsets should be done on Kafka/EventBroker level.

v-stepanov: I mean, from my point of view, only on Kafka/EventBroker level we know what to do with the offset. And on service layer offset is just a string for us.

src/main/java/org/zalando/nakadi/domain/Timeline.java
rcillo: 👍 improved in c8fbfc4

src/main/java/org/zalando/nakadi/exceptions/runtime/InvalidCursorOperation.java
rcillo: Fixed 9dc9d9d

src/main/java/org/zalando/nakadi/exceptions/runtime/MyNakadiRuntimeException1.java
rcillo: Fixed 9dc9d9d

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
```diff
@@ -0,0 +1,169 @@
...
+                final long finalOffset = offset + shift;
+                final String paddedOffset = getPaddedOffset(finalOffset);
+                return new NakadiCursor(timeline, partition, paddedOffset);
+            } else { // move right, next timeline
+                final Timeline nextTimeline = getTimeline(topic, timeline.getOrder() + 1);
+                final long totalEventsInTimeline = totalEventsInTimeline(nextTimeline, partition);
+                final long newShift = shift - (totalEventsInTimeline - offset);
+                return moveCursorForward(topic, nextTimeline, partition, 0, newShift);
+            }
+        } else { // Open timeline. Avoid checking for the latest available offset for performance reasons
```
rcillo: @v-stepanov I'm not sure about this decision. Maybe we should actually check for the latest available offset, cache it if necessary, and throw an error in case the shift exceeds it.

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
```diff
@@ -0,0 +1,169 @@
...
+        // get all intermediary timelines sizes
+        final String partitionString = initialCursor.getPartition();
+        for (int order = initialCursor.getTimeline().getOrder() + 1;
+             order < finalCursor.getTimeline().getOrder();
+             order++) {
+            final Timeline t = getTimeline(initialCursor.getTopic(), order);
+            distance += totalEventsInTimeline(t, partitionString);
+        }
+
+        // do not check if the offset exists because there is no reason to do it.
```
rcillo: Here as well. Maybe we should be more strict.

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
rcillo: @v-stepanov you're right. In the `parse` method I make the assumption that the offset is a decimal zero padded number. But for `ONE` it could be hexa, right?

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
v-stepanov: I don't know. For me it's fine as it is. In the worst case it will be the cursor that points so something that doesn't exist yet. Anyway when user will try to read from it - he will get the error.

rcillo:
@v-stepanov here is the refactoring I mentioned previously https://github.com/zalando/nakadi/pull/613/commits/d8b3710f9a37f2d24ed2fbcee58d8b8d1a913976

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -375,6 +380,33 @@ public int compareOffsets(final NakadiCursor first, final NakadiCursor second) t
...
+
+    public long numberOfEventsBeforeCursor(final NakadiCursor cursor) {
+        return kafkaOffset(cursor);
+    }
+
+    public String getOffsetForPosition(final long offset) {
+        return StringUtils.leftPad(String.valueOf(offset), VERSION_ZERO_MIN_OFFSET_LENGTH, '0');
+    }
+
+    private long kafkaOffset(final NakadiCursor cursor) {
```
v-stepanov: it actually duplicates functionality of [KafkaCursors.toKafkaOffset](https://github.com/zalando/nakadi/blob/3d551da20b2fa4b5ed99a77a8aad0e120c38e3e6/src/main/java/org/zalando/nakadi/repository/kafka/KafkaCursor.java#L69)

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -375,6 +380,33 @@ public int compareOffsets(final NakadiCursor first, final NakadiCursor second) t
         return KafkaCursor.fromNakadiCursor(first).compareTo(KafkaCursor.fromNakadiCursor(second));
     }
 
+    public long totalEventsInPartition(final Timeline timeline, final String partitionString) {
+        final Timeline.StoragePosition positions = timeline.getLatestPosition();
+        if (positions == null) {
+            throw new InvalidCursorOperation(InvalidCursorOperation.Reason.PARTITION_NOT_FOUND);
```
v-stepanov: Why do you throw this kind of exception?
As I understand the situation that `timeline.getLatestPosition() == null` means that current timeline is still "active".

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -375,6 +380,33 @@ public int compareOffsets(final NakadiCursor first, final NakadiCursor second) t
...
+        }
+        final int partition = Integer.valueOf(partitionString);
+        final Timeline.KafkaStoragePosition kafkaPositions = (Timeline.KafkaStoragePosition) positions;
+        final List<Long> offsets = kafkaPositions.getOffsets();
+        if (offsets.size() - 1 < partition) {
+            throw new InvalidCursorOperation(InvalidCursorOperation.Reason.PARTITION_NOT_FOUND);
```
v-stepanov: In general it seems that this method still "thinks" that it is inside `CursorOperationsService` class :)
It is strange that it throws InvalidCursorOperation because it doesn't have any "cursor" parameter

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -375,6 +381,37 @@ public int compareOffsets(final NakadiCursor first, final NakadiCursor second) t
...
+            throws InvalidCursorOperation {
+        try {
+            final Timeline.StoragePosition positions = timeline.getLatestPosition();
+            if (positions == null) {
+                final Optional<PartitionStatistics> statsO = loadPartitionStatistics(timeline, partitionString);
+                return statsO.map(stats -> numberOfEventsBeforeCursor(stats.getLast())).orElseThrow(
+                        MyNakadiRuntimeException1::new
+                );
+            }
+            final int partition = Integer.valueOf(partitionString);
```
antban: Maybe it is better to use `KafkaKursor.toKafkaPartition`

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -375,6 +381,37 @@ public int compareOffsets(final NakadiCursor first, final NakadiCursor second) t
         return KafkaCursor.fromNakadiCursor(first).compareTo(KafkaCursor.fromNakadiCursor(second));
     }
 
+    public long totalEventsInPartition(final Timeline timeline, final String partitionString)
+            throws InvalidCursorOperation {
+        try {
+            final Timeline.StoragePosition positions = timeline.getLatestPosition();
+            if (positions == null) {
+                final Optional<PartitionStatistics> statsO = loadPartitionStatistics(timeline, partitionString);
```
antban: If we expect that there will be a lot of calls to this method, then maybe it makes sense to use KafkaConsumer here, because it will eliminate the need of seekToBegining()

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
```diff
@@ -0,0 +1,209 @@
...
+    public NakadiCursorDistanceResult calculateDistance(final NakadiCursorDistanceQuery query)
+            throws InvalidCursorOperation {
+        final NakadiCursor initialCursor = query.getInitialCursor();
+        final NakadiCursor finalCursor = query.getFinalCursor();
+
+        // Validate query
+        if (!initialCursor.getPartition().equals(finalCursor.getPartition())) {
+            throw new InvalidCursorOperation(CURSORS_WITH_DIFFERENT_PARTITION);
+        } else if (initialCursor.getTimeline().getOrder() > finalCursor.getTimeline().getOrder()) {
+            throw new InvalidCursorOperation(INVERTED_TIMELINE_ORDER);
```
antban: For me it makes sense not to have this check. If we are performing mathematical operation, than if  `A-B=C` then `B-A=-C`

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
```diff
@@ -0,0 +1,209 @@
...
+
+    private long totalEventsInTimeline(final NakadiCursor cursor) {
+        final TopicRepository topicRepository = getTopicRepository(cursor.getTimeline());
+        return topicRepository.totalEventsInPartition(cursor.getTimeline(), cursor.getPartition());
+    }
+
+    private Timeline getTimeline(final String topic, final int order) {
+        final List<Timeline> timelines;
+        try {
+            timelines = timelineService.getActiveTimelinesOrdered(topic);
```
antban: As far as I remember, this method accepts event type, not topic

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -375,6 +381,37 @@ public int compareOffsets(final NakadiCursor first, final NakadiCursor second) t
...
+            throw new MyNakadiRuntimeException1("Problem calculating partition statistics", e);
+        }
+    }
+
+    public long numberOfEventsBeforeCursor(final NakadiCursor cursor) {
+        return KafkaCursor.toKafkaOffset(cursor.getOffset());
+    }
+
+    public String getOffsetForPosition(final long offset) {
+        return StringUtils.leftPad(String.valueOf(offset), VERSION_ZERO_MIN_OFFSET_LENGTH, '0');
```
antban: It is better to use `KafkaCursor.toNakadiOffset`

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
```diff
@@ -0,0 +1,202 @@
...
+import javax.validation.Valid;
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static org.springframework.http.HttpStatus.OK;
+import static org.springframework.http.ResponseEntity.status;
+
+@RestController
+public class CursorOperationsController {
```
antban: Are there any checks regarding security?

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
```diff
@@ -0,0 +1,202 @@
...
+
+        if (errors.hasErrors()) {
+            return Responses.create(new ValidationProblem(errors), request);
+        }
+
+        final List<NakadiCursorDistanceQuery> domainQueries = queries.stream()
+                .map(this.toCursorDistanceQuery(eventTypeName))
+                .collect(Collectors.toList());
+
+        final List<NakadiCursorDistanceResult> domainResult = cursorOperationsService.calculateDistance(domainQueries);
```
antban: For me it seems that there is a lot of unneeded complex things here.
What is the purpose of class NakadiCursorDistanceQuery?
What is the purpose of class NakadiCursorDistanceResult?

I thought that we can work within our domain model:
```
long distance = cursorOperationsService.calculateDistance(NakadiCursor, NakadiCursor)
```

src/main/java/org/zalando/nakadi/domain/NakadiCursorLag.java
```diff
@@ -0,0 +1,57 @@
+package org.zalando.nakadi.domain;
+
+public class NakadiCursorLag {
+    private String topic;
+    private String partition;
```
antban: What is the point of information duplication? I mean even if topic is needed here, then there are 2 cursors there, that should contain the same information.

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @RequestMapping(path = "/event-types/{eventTypeName}/cursors-lag", method = RequestMethod.POST)
+    public ResponseEntity<?> cursorsLag(@PathVariable("eventTypeName") final String eventTypeName,
+                                         @Valid @RequestBody final List<Cursor> cursors,
+                                         final Errors errors,
+                                         final NativeWebRequest request) {
+
+        if (errors.hasErrors()) {
+            return Responses.create(new ValidationProblem(errors), request);
+        }
```
antban: Isn't it better to add exception handler for `MethodArgumentNotValidException` instead of writing this check every time call made.
In this case it will be possible to get rid of `ResponseEntity<?>`, and it will clean the code.

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    private NakadiCursor moveCursorBackwards(final String topic, final Timeline timeline, final String partition,
+                                             final long offset, final long shift) {
+        final long shiftedOffset = offset + shift;
+        if (shiftedOffset >= 0) { // move left in the same timeline
+            final String paddedOffset = getOffsetForPosition(timeline, shiftedOffset);
+            return new NakadiCursor(timeline, partition, paddedOffset);
+        } else { // move to previous timeline
+            final Timeline previousTimeline = getTimeline(topic, timeline.getOrder() - 1);
```
antban: I don't feel comfortable with calling getActiveTimelines() several times without need for that

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: We don't expect too many calls to reach this branch. Actually, I'm afraid it might never get here. I just implemented this branch for the sake of completeness, in case this method would be used in another context in the future.

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
rcillo: but, in the other had, the word `distance` implies something that is always positive. So it would be the modulus of that? To avoid confusion when interpreting results, I thought about making it more strict, with less options. If `initial` is always older than `final` than `distance` is always positive and very easy to understand.

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
rcillo: @antban do you have any examples on how it would look like?

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
rcillo: It's supposed to be cached, right?

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
rcillo: Well reminded https://github.com/zalando/nakadi/pull/613/commits/fd53217febeecd4dd75116e57fbc460b743f0a47

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
antban: yes, it is cached

src/main/java/org/zalando/nakadi/view/CursorDistanceResult.java
```diff
@@ -0,0 +1,32 @@
+package org.zalando.nakadi.view;
+
+public class CursorDistanceResult {
```
antban: Class is not used

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
rcillo: Read scopes checked a4c4405

src/main/java/org/zalando/nakadi/domain/NakadiCursorLag.java
```diff
@@ -0,0 +1,36 @@
+package org.zalando.nakadi.domain;
+
+public class NakadiCursorLag {
+    private NakadiCursor firstCursor;
+    private NakadiCursor lastCursor;
```
antban: Does it make sense to make it final?

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
rcillo: Completely removed this code since it's already handled automatically by another exception handler. It even provides a very nicely formatted error message explaining the reason and the path for each error

`{"violations":[{"field":"valid_list_wrapper.list[0].offset","message":"may not be null"}],"status":422,"title":"Constraint Violation","type":"https://github.com/zalando/problem/wiki/constraint-violation"}`

rcillo: f3bd714

src/main/java/org/zalando/nakadi/domain/NakadiCursorLag.java
rcillo: done 6a8ef22

src/main/java/org/zalando/nakadi/view/CursorDistanceResult.java
rcillo: bf1d38f

rcillo:
@antban all comments addressed. Could you please take a final look.

src/main/java/org/zalando/nakadi/domain/NakadiCursorLag.java
rcillo: 965aeb9

antban:
:+1:

rcillo:
👍

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
```diff
@@ -0,0 +1,196 @@
...
+            case INVERTED_OFFSET_ORDER: return "Inverted offsets. Final cursor offsets must be newer than initial " +
+                    "cursor offsets";
+            case PARTITION_NOT_FOUND: return "Partition not found.";
+            case CURSORS_WITH_DIFFERENT_PARTITION: return "Cursors with different partition. Pairs of cursors should " +
+                    "have matching partitions.";
+            default: return unexpectedErrorReason(reason);
+        }
+    }
+
+    private void accessControl(final String eventTypeName, final Client client) {
```
antban: Maybe it's better to select different name (one of verbs because noun is not really describing what is going to happen), I think that `checkReadScopes` or `verifyCanRead` matches better.

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
```diff
@@ -0,0 +1,196 @@
...
+            eventType = eventTypeRepository.findByName(eventTypeName);
+            client.checkScopes(eventType.getReadScopes());
+        } catch (final InternalNakadiException e) {
+            throw new MyNakadiRuntimeException1("failed to get event type", e);
+        } catch (final NoSuchEventTypeException e) {
+            throw new NoEventTypeException(e.getMessage(), e);
+        }
+    }
+
+    private String unexpectedErrorReason(final InvalidCursorOperation.Reason reason) throws MyNakadiRuntimeException1 {
```
antban: I think it is better to embed this method, cause it breaks usual code read flow

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
```diff
@@ -0,0 +1,196 @@
...
+                nakadiCursorLag.getLag()
+        );
+    }
+
+    private Function<Cursor, NakadiCursor> toNakadiCursor(final String eventTypeName) {
+        return cursor -> {
+            try {
+                return cursorConverter.convert(eventTypeName, cursor);
+            } catch (final NakadiException | InvalidCursorException e) {
+                throw new CursorConversionException("problem converting cursors", e);
```
antban: What do you think of injecting this kind of method to `CursorConverter` (and later on - remove the old one)?
(Although maybe it's not the best ticket to do this modification)

src/main/java/org/zalando/nakadi/controller/ExceptionHandling.java
```diff
@@ -66,6 +70,12 @@ private String generateErrorTraceId() {
         return Responses.create(Response.Status.BAD_REQUEST, message, request);
     }
 
+    @ExceptionHandler(NoEventTypeException.class)
+    public ResponseEntity<Problem> noEventTypeException(final NoEventTypeException exception,
+                                                               final NativeWebRequest request) {
+        return Responses.create(UNPROCESSABLE_ENTITY, exception.getMessage(), request);
```
antban: As far as I see - this is class for default exception handlers. Maybe default one for `NoSuchEventType` is `NOT_FOUND`?

antban:
And I see, that this branch should be rebased...

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
rcillo: I just extracted it because in our code style it's not allowed to have nested blocks and extracting the method was the way I found to avoid it.

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
rcillo: If we see that it's repeating somewhere we can move it.

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
rcillo: 625dcbd

src/main/java/org/zalando/nakadi/controller/ExceptionHandling.java
rcillo: 33a98dd

rcillo:
@antban I made the changes that were clear for me. If there is anything missing, please let me know.

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
antban: @rcillo I meant that method says that it returns String, but actually it only generates exception.
That is why I think that line `return unexpectedErrorReason(reason);` leads to misunderstanding. 
After reading method and it's call My brain was broken.
I think that 
```
default:
    throw generateUnexpectedError(reason)
``` 
is much better. 
And from the other hand - why not to do
```
default:
    LOG.error("Unexpected invalid cursor operation reason " + reason);
    throw new MyNakadiRuntimeException1();
```
?

rcillo: @antban you're right. There is not need to have a nested block... I don't know what I was thinking about. Done in a9ad2f6

antban:
:+1:

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
```diff
@@ -0,0 +1,199 @@
...
+
+        // get distance from initialCursor to the end of the timeline
+        distance += totalEventsInTimeline(initialCursor) - numberOfEventsBeforeCursor(initialCursor);
+
+        // get all intermediary timelines sizes
+        final String partitionString = initialCursor.getPartition();
+        for (int order = initialCursor.getTimeline().getOrder() + 1;
+             order < finalCursor.getTimeline().getOrder();
+             order++) {
+            final Timeline timeline = getTimeline(initialCursor.getTopic(), order);
```
antban: @rcillo Just found during testing that it dies here (topic is used instead of event type)

rcillo: Fixed here https://github.com/zalando/nakadi/pull/613/commits/e8387f64909ae97d25ca697c78b218caede6fadb

rcillo:
@antban I fixed the issue with `s/topic/event-type/` and added an acceptance test for multiple timelines.

antban:
:+1:

antban:
:+1:

rcillo:
👍

rcillo:
@antban I fixed a refactoring problem in one of the tests. The build should pass now.

rcillo:
👍

antban:
:+1:

Issue #614, opened 2017-03-15
DuplicatedEventTypeNameException is not processed


adyach:
When creating event type which exists in database, topic is not deleted from Kafka

PR #615, opened 2017-03-15
issue-614: process duplicated event type exception

adyach:
#614

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -112,7 +112,7 @@ public class EventTypeService {
                 try {
                     topicRepository.deleteTopic(eventType.getTopic());
                 } catch (final TopicDeletionException ex) {
-                    LOG.warn("failed to delete topic for event type that failed to be created", ex);
+                    LOG.error("failed to delete topic for event type that failed to be created", ex);
                 }
             }
             return Result.problem(e.asProblem());
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -99,22 +99,30 @@ public class EventTypeService {
             validateSchema(eventType);
             enrichment.validate(eventType);
             partitionResolver.validate(eventType);
+
             final String topicName = topicRepository.createTopic(
                     partitionsCalculator.getBestPartitionsCount(eventType.getDefaultStatistic()),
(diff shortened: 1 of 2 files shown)


adyach:
:shipit:

v-stepanov:
👍

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -104,7 +105,8 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
             eventType.setTopic(topicName);
             eventTypeRepository.saveEventType(eventType);
             return Result.ok();
-        } catch (final InvalidEventTypeException | NoSuchPartitionStrategyException e) {
+        } catch (final InvalidEventTypeException | NoSuchPartitionStrategyException |
+                DuplicatedEventTypeNameException e) {
```
rcillo: Could you please change log in line "113/115" to `error` since this is leaving the system in an inconsistent state, i.e. orphan topic in Kafka without corresponding event type in DB.

v-stepanov:
:+1:

src/main/java/org/zalando/nakadi/service/EventTypeService.java
v-stepanov: I addressed your comment in https://github.com/zalando/nakadi/pull/615/commits/fb1025e1010868dbd93791071b24b73c3f0bdac3

antban: I would like to use different approach. The problem there is that it is not safe to do like that in case of runtime exceptions. Please consider the following approach
```
boolean eventTypeSaved = false;
try {
    eventType.setTopic(topicName);
    eventTypeSaved = true;
} finally {
    if (!eventTypeSaved) {
        try {
            topicRepository.deleteTopic(eventType.getTopic());
        } catch (final Exception ex) {
            LOG.error("failed to delete topic for event type that failed to be created", ex);
        }
    }
}
```

v-stepanov: @antban makes sense. I changed it now in a way you suggested: https://github.com/zalando/nakadi/pull/615/commits/349f721502c1aa19d1947cb02e1913326572104e

v-stepanov:
👍

antban:
:+1:

antban:
:+1:

v-stepanov:
👍

v-stepanov:
👍

adyach:
👍

PR #616, opened 2017-03-21
Aruha 681 remove event type with timelines

lmontrieux:
[ARUHA-681: Remove event type with timelines](https://techjira.zalando.net/browse/ARUHA-681)

Remove even type from Nakadi by removing its timelines and topics from repositories.


src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -161,6 +163,11 @@ public TopicRepository getTopicRepository(final Timeline timeline) {
         return topicRepositoryHolder.getTopicRepository(timeline.getStorage());
     }
 
+    public TopicRepository getTopicRepositoryForTimeline(final Timeline timeline)
+    throws TopicRepositoryException, TimelineException {
```
v-stepanov: indent problem here

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -161,6 +163,11 @@ public TopicRepository getTopicRepository(final Timeline timeline) {
         return topicRepositoryHolder.getTopicRepository(timeline.getStorage());
     }
 
+    public TopicRepository getTopicRepositoryForTimeline(final Timeline timeline)
```
v-stepanov: But why do you need this method? There's exactly the same one few lines above.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -227,6 +234,17 @@ public void delete(final String eventTypeName, final String timelineId, final Cl
         }
     }
 
+    public void deleteAllTimelinesForEventType(final String eventTypeName)
+            throws TopicDeletionException, TimelineException, NotFoundException {
+        LOG.info("Deleting all timelines for event type {}", eventTypeName);
+        final List<Timeline> timelines = getTimelines(eventTypeName);
+        for (final Timeline timeline:timelines) {
```
v-stepanov: spaces missing around ":"

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -227,6 +234,17 @@ public void delete(final String eventTypeName, final String timelineId, final Cl
...
+            final TopicRepository topicRepository = getTopicRepositoryForTimeline(timeline);
+            topicRepository.deleteTopic(timeline.getTopic());
```
v-stepanov: As you saw from my PR - topic can already be removed for old timeline. I think you should handle here the case that topic doesn't exist and continue with removal of next timeline (but not interrupt the whole process).

src/main/java/org/zalando/nakadi/controller/EventTypeController.java
```diff
@@ -139,4 +148,39 @@ private void setDefaultEventTypeOptions(final EventTypeBase eventType) {
         }
         return status(HttpStatus.OK).body(result.getValue());
     }
+
+    @ExceptionHandler(EventTypeDeletionException.class)
```
v-stepanov: I like how you use the new approach of runtime exceptions 👍

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
lmontrieux: I was wondering how I could have missed this, but git blame tells me that the identical function a few lines above appeared when I merged master into this branch, after having written this function. I'll fix this.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
lmontrieux: 4a3c356

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
lmontrieux: Maybe we need a style check for this c769fb5

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
lmontrieux: Fixed c769fb5

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
lmontrieux: 1f286f6, 139cda1, and 6d0436c

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -161,8 +162,8 @@ public void delete(final String eventTypeName, final Client client)
             LOG.error("Failed to wait for timeline switch", e);
             throw new EventTypeUnavailableException("Event type "+ eventTypeName
                     + " is currently in maintenance, please repeat request", e);
-        } catch (final TopicDeletionException e) {
-            LOG.error("Problem deleting kafka topic " + eventTypeName, e);
+        } catch (final TopicDeletionException|TopicRepositoryException e) {
```
v-stepanov: spaces missing around "|"

lmontrieux: 6050c71

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -231,12 +231,14 @@ public void delete(final String eventTypeName, final String timelineId, final Cl
...
     public void deleteAllTimelinesForEventType(final String eventTypeName)
-            throws TopicDeletionException, TimelineException, NotFoundException {
+            throws TopicDeletionException, TopicRepositoryException, TimelineException, NotFoundException {
         LOG.info("Deleting all timelines for event type {}", eventTypeName);
         final List<Timeline> timelines = getTimelines(eventTypeName);
         for (final Timeline timeline : timelines) {
             final TopicRepository topicRepository = getTopicRepository(timeline);
-            topicRepository.deleteTopic(timeline.getTopic());
+            if (topicRepository.topicExists(timeline.getTopic())) {
+                topicRepository.deleteTopic(timeline.getTopic());
```
v-stepanov: What if between the calls `topicRepository.topicExists` and `topicRepository.deleteTopic` topic will be deleted by the job that removes expired topics? Then you deletion process will fail with exception.
That's why I think it's better to call `deleteTopic` without checking that the topic exists and just catch the exception in case topic deletion failed.

lmontrieux: Good point. I will change this.

lmontrieux: de92095

v-stepanov:
👍

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -120,44 +127,49 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
...
                     ImmutableSet.of(eventTypeName), Optional.empty(), 0, 1);
             if (!subscriptions.isEmpty()) {
-                return Result.conflict("Not possible to remove event-type as it has subscriptions");
+                throw new UnableProcessException("Can't remove event type " + eventTypeName
+                        + ", as it has subscriptions");
             }
-            final TopicRepository topicRepository = timelineService.getTopicRepository(eventType);
-            // TODO: Cascading delete. Event Type must be deleted with all it's timelines.
+            timelineService.deleteAllTimelinesForEventType(eventTypeName);
             eventTypeRepository.removeEventType(eventTypeName);
```
antban: I don't like this place. One should use transactional approach to this, and then delete topics from underlying storages. Right now we can get situation with event type that was not deleted, but all the timelines were deleted.

antban: I don't even know how to validate this place, if it is always possible to hack around: 
1. Create 10 timelines
2. Simultaneously delete event type and create subscription for it.
Timeline deletion takes time and probably after 10 attempts it will be possible to get event type that wasn't deleted because of constraints, but with all the topics deleted.

lmontrieux: Good points. I will fix this.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -120,44 +132,55 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
...
+            final Multimap<TopicRepository, String> topicsToDelete = transactionTemplate.execute(action -> {
+                return deleteEventType(eventTypeName);
+            });
+            for (final TopicRepository topicRepository : topicsToDelete.keySet()) {
+                for (final String topic : topicsToDelete.get(topicRepository)) {
+                    try {
+                        topicRepository.deleteTopic(topic);
+                    } catch (TopicDeletionException e) {
+                        // If a timeline was marked as deleted, then the topic does not exist, and we should proceed.
+                        LOG.info("Could not delete topic " + topic);
```
antban: Maybe it makes sense to log `e`

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -227,6 +232,21 @@ public void delete(final String eventTypeName, final String timelineId, final Cl
...
+            throws TopicDeletionException, TimelineException, NotFoundException {
+        LOG.info("Deleting all timelines for event type {}", eventTypeName);
+        final Multimap<TopicRepository, String> topicsToDelete = ArrayListMultimap.create();
+        for (final Timeline timeline : getTimelines(eventTypeName)) {
+            topicsToDelete.put(getTopicRepository(timeline), timeline.getTopic());
+            timelineDbRepository.deleteTimeline(timeline.getId());
+        }
+        final List<Timeline> timelines = getTimelines(eventTypeName);
+        for (final Timeline timeline : timelines) {
+            timelineDbRepository.deleteTimeline(timeline.getId());
```
antban: Ha:) Is it control shot to timeline head?

antban:
In general I'm fine with it. We can terminate application exceptions further, there is code duplication

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -120,44 +132,55 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
...
             }
-            final TopicRepository topicRepository = timelineService.getTopicRepository(eventType);
-            // TODO: Cascading delete. Event Type must be deleted with all it's timelines.
-            eventTypeRepository.removeEventType(eventTypeName);
-            topicRepository.deleteTopic(eventType.getTopic());
-            return Result.ok();
+            final Multimap<TopicRepository, String> topicsToDelete = transactionTemplate.execute(action -> {
+                return deleteEventType(eventTypeName);
+            });
+            for (final TopicRepository topicRepository : topicsToDelete.keySet()) {
```
antban: And forgot one thing - it's better to perform termination after calling deleteCloser

antban:
:+1:

lmontrieux:
👍

lmontrieux:
👍

antban:
:+1:

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -227,6 +232,17 @@ public void delete(final String eventTypeName, final String timelineId, final Cl
         }
     }
 
+    public Multimap<TopicRepository, String> deleteAllTimelinesForEventType(final String eventTypeName)
+            throws TopicDeletionException, TimelineException, NotFoundException {
+        LOG.info("Deleting all timelines for event type {}", eventTypeName);
+        final Multimap<TopicRepository, String> topicsToDelete = ArrayListMultimap.create();
+        for (final Timeline timeline : getTimelines(eventTypeName)) {
```
adyach: @lmontrieux I think you should use getActiveTimelinesOrdered instead of getTimelines

lmontrieux: 613359c

adyach:
👍

lmontrieux:
👍

v-stepanov:
👍

lmontrieux:
👍

lmontrieux:
👍

v-stepanov:
👍

PR #617, opened 2017-03-29
ARUHA-724 Allow to consume from several topics using NakadiKafkaConsumer

antban:


antban:
the bug is not seen yet, will be fixed in different pr

PR #620, opened 2017-04-03
ARUHA-723 Update README since we support multiple ETs in subscriptions

lmontrieux:
https://techjira.zalando.net/browse/ARUHA-723

The README (and likely the manual) state that only 1 event type can be used per subscription.
This is no longer the case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -465,7 +465,7 @@ The typical workflow when using subscriptions is:
 
 1. Commit the cursors found in the event batches back to Nakadi, which will store the offsets. 
 
-_\* Note: the API signature can support subscribing to multiple event types with a single subscription, but this is not implemented yet; it's planned to be enabled soon._
+_\* Note: the API signature supports subscribing to multiple event types with a single subscription._
 
 
 If the connection is closed, and later restarted, clients will get events from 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -465,7 +465,7 @@ The typical workflow when using subscriptions is:
 
 1. Commit the cursors found in the event batches back to Nakadi, which will store the offsets. 
 
-_\* Note: the API signature can support subscribing to multiple event types with a single subscription, but this is not implemented yet; it's planned to be enabled soon._
+_\* Note: the API signature supports subscribing to multiple event types with a single subscription._
```
adyach: @lmontrieux do you think we really need this note? In my opinion it is clear in the description, that there is a several even type can be used.
> Create a Subscription specifying the event-types you want to read

lmontrieux: @adyach That's true, I guess we can get rid of the note entirely.

lmontrieux: 5013e25

adyach: and then you do not need a star(*) here
>  1. Create a Subscription specifying the event-types\* you want to read.

lmontrieux: 82f9bd1

adyach:
👍

lmontrieux:
👍

PR #621, opened 2017-04-04
aruha-386: Event type retention time not reflected in Kafka

adyach:
ARUHA-386

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -24,6 +24,7 @@ import org.zalando.nakadi.exceptions.NakadiException;
 import org.zalando.nakadi.exceptions.NoSuchEventTypeException;
 import org.zalando.nakadi.exceptions.NoSuchPartitionStrategyException;
 import org.zalando.nakadi.exceptions.TopicDeletionException;
+import org.zalando.nakadi.exceptions.runtime.TopicConfigException;
 import org.zalando.nakadi.partitioning.PartitionResolver;
 import org.zalando.nakadi.repository.EventTypeRepository;
 import org.zalando.nakadi.repository.TopicRepository;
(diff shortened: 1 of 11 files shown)


lmontrieux:
If you merge master, you'll fix travis :)

Also, you need the ticket number in the PR description to make zappr happy

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -184,6 +184,13 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
             final EventType eventType = schemaEvolutionService.evolve(original, eventTypeBase);
             eventType.setDefaultStatistic(
                     validateStatisticsUpdate(original.getDefaultStatistic(), eventType.getDefaultStatistic()));
+            final Long newRetentionTime = eventTypeBase.getOptions().getRetentionTime();
+            final Long oldRetentionTime = original.getOptions().getRetentionTime();
+            if (newRetentionTime != null && !newRetentionTime.equals(oldRetentionTime)) {
```
lmontrieux: Here, it is possible that the change in retention time will succeed, but the event type will fail to update later (line 194). In this case the client will think everything has failed, even though the retention time was updated.

I think it might be better to change the retention time after updating the event type.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -184,6 +184,13 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
...
+                timelineService.getActiveTimelinesOrdered(eventTypeName)
+                        .forEach(timeline -> timelineService.getTopicRepository(timeline)
+                                .setRetentionTime(timeline.getTopic(), newRetentionTime));
```
lmontrieux: Here you don't catch `TopicConfigException`, and it's not in the method signature either (if I remember correctly, we had agreed to include unchecked exceptions in method signatures)

src/main/java/org/zalando/nakadi/service/EventTypeService.java
lmontrieux: Actually this would cause the same problem, but the other way around. It should be possible to rollback the first operation (update topics' retention time, or update ET) if the second fails.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
adyach: cf3b1c8

src/main/java/org/zalando/nakadi/service/EventTypeService.java
adyach: cf3b1c8

lmontrieux:
👍

lmontrieux:
👍

adyach:
👍

adyach:
👍

antban:
:+1:

PR #624, opened 2017-04-11
ARUHA-715 Recreate version node if is was deleted because of found.

antban:
ARUHA-715 In case when nakadi looses connection to zk, ephemeral nodes are deleted. In this case there will be notification in timeline api `versionChanged`, but all ephemeral nodes could be deleted at that moment. In order to fix it, TimelineSyncImpl will try to recreate node while writing new information in zk.


rcillo:
@antban do you want to add a test that removes the node to check that it's recreated?

rcillo:
👍

antban:
@rcillo I believe that it is not needed

antban:
:+1:

PR #626, opened 2017-04-18
Aruha 732 dont delete event types

lmontrieux:
[ARUHA-732
](https://techjira.zalando.net/browse/ARUHA-732)
* everyone should be able to delete events on staging
* only Aruha team should be able to delete events on live
* the API definition should reflect this
* this should be under a feature flag

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -278,8 +278,8 @@ paths:
 
          * Events in the stream may be visible for a short period of time before being removed.
 
-         On nakadi-live, the oauth resource owner username has to be equal to 'nakadi.oauth2.adminClientId' property
-         to be able to access this endpoint.
+         Depending on the Nakadi configuration, the oauth resource owner username may have to be equal to
+         'nakadi.oauth2.adminClientId' property to be able to access this endpoint.
 
       parameters:
         - name: name
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -278,6 +278,9 @@ paths:
 
          * Events in the stream may be visible for a short period of time before being removed.
 
+         On nakadi-live, the oauth resource owner username has to be equal to 'nakadi.oauth2.adminClientId' property
+         to be able to access this endpoint.
+
       parameters:
         - name: name
           in: path
@@ -295,6 +298,10 @@ paths:
           description: EventType is successfuly removed
         '401':
           description: Client is not authenticated
+          schema:
+            $ref: '#/definitions/Problem'
+        '403':
+          description: Access forbidden
           schema:
             $ref: '#/definitions/Problem'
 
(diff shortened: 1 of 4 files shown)


api/nakadi-event-bus-api.yaml
```diff
@@ -278,6 +278,9 @@ paths:
 
          * Events in the stream may be visible for a short period of time before being removed.
 
+         On nakadi-live, the oauth resource owner username has to be equal to 'nakadi.oauth2.adminClientId' property
```
v-stepanov: I don't think we should mix open-source stuff with our private operational information. "nakadi-live" doesn't say anything for people outside Zalando.

src/main/java/org/zalando/nakadi/controller/EventTypeController.java
```diff
@@ -186,4 +190,8 @@ private void setDefaultEventTypeOptions(final EventTypeBase eventType) {
         return Responses.create(Response.Status.SERVICE_UNAVAILABLE, exception.getMessage(), request);
     }
 
+    private boolean isNotAdmin(final Client client) {
```
v-stepanov: I'm a bit skeptical about naming boolean methods `isNotSomething`. Because when you will need to check for `something` you will have 'if (!isNotSomething()) {...}'. I think the more clear approach is to have `isSomething` method, even if you use it in only place with inverse check.

api/nakadi-event-bus-api.yaml
lmontrieux: Good point. Would it be ok to replace this with `Depending on the nakadi configuration, the oauth resource owner username may have to be equal [...]`?

src/main/java/org/zalando/nakadi/controller/EventTypeController.java
lmontrieux: a694241

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -472,6 +473,43 @@ public void whenDeleteEventTypeThen403() throws Exception {
     }
 
     @Test
+    public void whenDeleteEventTypeNotAdminAndDeletionDeactivatedThenForbidden() throws Exception {
```
v-stepanov: These two methods are almost the same except the very last line. 
Maybe you can extract a helper method here.

api/nakadi-event-bus-api.yaml
v-stepanov: Yes, this is good.

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
lmontrieux: 1dbdd49

api/nakadi-event-bus-api.yaml
lmontrieux: 895938d

v-stepanov:
👍

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
lmontrieux: @v-stepanov there was a bug in my test 😮  e2bb753

v-stepanov:
👍

lmontrieux:
👍

PR #628, opened 2017-04-20
ARUHA-746: Added update of cleanup-time when changing retention time …

v-stepanov:
…for EventType

https://techjira.zalando.net/browse/ARUHA-746

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -275,8 +275,8 @@ public class EventTypeService {
                                      final Long oldRetentionTime) throws NakadiException {
         final NakadiException exception = transactionTemplate.execute(action -> {
             try {
-                eventTypeRepository.update(eventType);
                 updateTimelinesCleanup(eventType.getName(), newRetentionTime, oldRetentionTime);
+                eventTypeRepository.update(eventType);
                 return null;
             } catch (final NakadiException e) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -269,7 +271,43 @@ public void delete(final String eventTypeName, final Client client)
         }
     }
 
-    private void updateRetentionTime(final String eventTypeName, final Long retentionTime)
+    private void updateEventTypeInDB(final EventType eventType, final Long newRetentionTime,
+                                     final Long oldRetentionTime) throws NakadiException {
+        final NakadiException exception = transactionTemplate.execute(action -> {
+            try {
+                eventTypeRepository.update(eventType);
+                updateTimelinesCleanup(eventType.getName(), newRetentionTime, oldRetentionTime);
```
adyach: @v-stepanov it is not clear how the event type cache will be updated? Maybe it make sense just to switch it with the previous line.

v-stepanov: Very good point 👍  I switched the order https://github.com/zalando/nakadi/pull/628/commits/7c4fd7a566ac2dc2a3698de9c096ed06fc239bee
Now the case will be updated with latest state of timelines.

adyach:
👍

v-stepanov:
👍

PR #629, opened 2017-04-20
aruha-667 fix schema versioning of composed types

rcillo:
Previous to this PR, changing the number of itens in a composed
schema (anyOf, oneOf, allOf) was possible due to an error in how
comparisons are made.

I replacing containsAll by a simple size comparison, so to be sure that
the number of items remain the same. Checks for changes in the items
themselves is done correctly, as demoed in the tests.


rcillo:
Wrong ticket code.

PR #630, opened 2017-04-20
ARUHA-677 Fix schema comparison

rcillo:
1. Previous to this PR, changing the number of itens in a composed
schema (anyOf, oneOf, allOf) was possible due to an error in how
comparisons are made. I replaced containsAll by a simple size comparison, so to be sure that
the number of items remain the same. Checks for changes in the items
themselves is done correctly, as demoed in the tests.

2. undetected changes causing db inconsistence

3. when creating or updating a schema, it will no longer allow for non-specified-json-schema attributes. So, "example" attributes will be rejected, for example.

It will now generate 500 when schema changes goes unnoticed, forcing us to fix them.


src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
```diff
@@ -97,6 +102,12 @@ public EventType evolve(final EventType original, final EventTypeBase eventType)
 
         final Version.Level changeLevel = semanticOfChange(changes, original.getCompatibilityMode());
 
+        if (changeLevel == NO_CHANGES && !original.getSchema().getSchema().equals(eventType.getSchema().getSchema())) {
+            LOG.error("undetected schema changes from {} to {}", original.getSchema().getSchema(),
+                    eventType.getSchema().getSchema());
+            throw new MyNakadiRuntimeException1("undetected schema changes");
```
lmontrieux: I don't really like it when we throw `MyNakadiRuntimeException1`, because the name doesn't say anything about what problem there was. Also, if we do it a lot, it might be difficult for the caller to find out where it comes from. I prefer to subclass the exception with something that has a clear name

v-stepanov: This exception is not supposed to be thrown as it doesn't mean anything. It's just a common type for our exceptions.

src/test/java/org/zalando/nakadi/validation/SchemaEvolutionServiceTest.java
```diff
@@ -124,6 +126,20 @@ public void whenPatchChangesBumpVersion() throws Exception {
...
+        final EventTypeTestBuilder builder = EventTypeTestBuilder.builder();
+        final EventType oldEventType = builder.schema("{}").build();
+        final EventType newEventType = builder.schema("{\"example\":\"something\"}").build();
+
+        Mockito.doReturn(Optional.empty()).when(evolutionConstraint).validate(oldEventType, newEventType);
+        Mockito.doReturn(NO_CHANGES).when(compatibleChanges).get(any());
+        Mockito.doReturn(Lists.newArrayList(new SchemaChange(TITLE_CHANGED, "#/"))).when(schemaDiff)
+                .collectChanges(any(), any());
+
+        final EventType eventType = service.evolve(oldEventType, newEventType);
```
lmontrieux: I think you don't need to declare eventType, as it is never used

src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
rcillo: I agree with you in case we needed to handle this exception. But in this situation, this is really like a NullPointer. Something unexpected happened: something changed in the schema but our algorithm for comparison was unable to precisely tell why. So we just 💣 forcefully to halt all further operations.

Later, we check the logs, see what the input looked like, fix the bug and tell the user to try again.

rcillo: What about this comment here https://github.com/zalando/nakadi/pull/630#discussion_r112670086 does it make sense?

src/test/java/org/zalando/nakadi/validation/SchemaEvolutionServiceTest.java
rcillo: 👍 911b813

src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
v-stepanov: @rcillo I don't really see how it is connected. When we throw an exception we should try it to be as informative as possible so that the one who catches it will make the most use of it. Throwing default MyNakadiRuntimeException1 just makes it very uninformative like just throwing a plain RuntimeException.

Also, I'm not sure that it is completely unexpected. At least in the ticket I see a description of the case when this can happen. https://techjira.zalando.net/browse/ARUHA-677

rcillo: @v-stepanov do you think that if I created a specific exception type for this it would be better? Like `UnexpectedSchemaChangeException`?

About the ticket, the way to solve this is not by improving the `diff` operation, but to avoid users from defining weird properties like `example`. Since the ticket tile was about consistency between the information in tables event_type and event_type_schema, I just added enough code to avoid this inconsistency from happening.

But I'll keep coding and implement additional validations that will keep users from defining such weird schemas with invalid properties.

v-stepanov: @rcillo yes, for me that would be much better to have `UnexpectedSchemaChangeException` as it describes the problem by its name 👍

lmontrieux: I'd be happy with that too 👍

rcillo: 👍 4b48b93

rcillo:
@lmontrieux @v-stepanov I implemented some additional checks when creating/updating schema 61f804c.

This commit checks that the schema provided by the user doesn't contain any weird properies, in case of `forward` mode.

It already did this for `compatible` mode, with the additional that it considered `not`, `additionalItems`, `additionalProperties` and `patterProperties` to be invalid, as well. I removed these constraints from the `forward` validation, not to over work this change. I think that ideally the same set of attributes should be allowed, and Nakadi should support a "nakadi-json-schema" spec. But this is outside the scope of this ticket.

rcillo:
One more commit e52f75b just to make sure we apply schema validation for all schemas, regardless of compatibility mode.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -344,7 +344,7 @@ private void validateSchema(final EventTypeBase eventType) throws InvalidEventTy
             }
 
             if (eventType.getCompatibilityMode() != CompatibilityMode.NONE) {
-                validateJsonSchemaConstraints(schemaAsJson);
+                validateJsonSchemaConstraints(eventType, schemaAsJson);
```
lmontrieux: Here it seems that only the compatibility mode is used to validate the constraints, rather than the whole event type. Would it make sense to pass the compatibility mode only?

rcillo: @lmontrieux done 8e195b4

lmontrieux:
@rcillo there's a checkstyle violation, but otherwise this is good

rcillo:
@lmontrieux 😪 1638b18

lmontrieux:
👍

rcillo:
👍

rcillo:
@v-stepanov @lmontrieux closed this in favor of https://github.com/zalando/nakadi/pull/641

PR #634, opened 2017-04-25
Aruha 769 subscription stats performance

v-stepanov:
https://techjira.zalando.net/browse/ARUHA-769

Basically I improved 2 things here:
1) Now we load topic stats for all topics at once (per repository) not separately for each topic like before
2) For stats endpoint only end offsets are needed, so now we don't seek for beginning offsets any more


src/main/java/org/zalando/nakadi/exceptions/runtime/MyNakadiRuntimeException1.java
```diff
@@ -4,7 +4,7 @@
  * Parent class for Nakadi runtime exceptions
  * Name NakadiRuntimeException was already taken for some kind of wrapper. This name is a nice alternative ;)
  */
-public class MyNakadiRuntimeException1 extends RuntimeException {
+public abstract class MyNakadiRuntimeException1 extends RuntimeException {
```
lmontrieux: I've been meaning to do that for a while, great! 🎉  🍾

src/main/java/org/zalando/nakadi/repository/kafka/KafkaPartitionStartStatistics.java
```diff
@@ -0,0 +1,49 @@
...
+    @Override
+    public NakadiCursor getFirst() {
+        // TODO: Support several timelines
+        return first.toNakadiCursor(getTimeline());
+    }
+
+    @Override
+    public NakadiCursor getBeforeFirst() {
+        // TODO: Support several timelines
+        return first.addOffset(-1).toNakadiCursor(getTimeline());
```
lmontrieux: What happens if `first` is `0`? I think that, in that case, we should get the last offset of the previous non-empty timeline, correct?

v-stepanov: I just copy/pasted it from here https://github.com/zalando/nakadi/blob/3d551da20b2fa4b5ed99a77a8aad0e120c38e3e6/src/main/java/org/zalando/nakadi/repository/kafka/KafkaPartitionStatistics.java#L34
together with TODO. I believe this problem will be fixed when implementing these TODOs.
But I don't think that this PR is a good place for that.

src/main/java/org/zalando/nakadi/domain/PartitionBaseStatisticsImpl.java
```diff
@@ -0,0 +1,28 @@
+package org.zalando.nakadi.domain;
+
+public class PartitionBaseStatisticsImpl implements PartitionBaseStatistics {
+
+    private Timeline timeline;
```
v-stepanov: fixed https://github.com/zalando/nakadi/pull/634/commits/0b3e09407f65ae52da1312dea03dc2cfad274d1c

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -339,6 +342,32 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
     }
 
     @Override
+    public List<PartitionEndStatistics> loadTopicEndStatistics(final Collection<Timeline> timelines)
```
rcillo: If I understood correctly, all these timelines must relate to the same repository. So it means that the caller needs to be aware of it.

Thinking about future users of this method, it could be useful to write a comment, warning them of this fact. Even though it should be expected, since this method is in the repository layer, it's bound to a single repo.

rcillo:
👍

v-stepanov:
👍

antban:
:+1:

PR #635, opened 2017-04-26
ARUHA-771: Fixed the problem that forbids create subscription with to…

v-stepanov:
…o long name event-types
https://techjira.zalando.net/browse/ARUHA-771


database/nakadi/10_data/04_tables/02_subscription.sql
```diff
@@ -5,18 +5,15 @@ SET ROLE zalando_nakadi_data_owner;
...
 );
 
-CREATE INDEX ON zn_data.subscription USING gin (s_subscription_object);
-
 CREATE INDEX ON zn_data.subscription ((s_subscription_object->>'created_at'));
 
-CREATE UNIQUE INDEX ON zn_data.subscription ((s_subscription_object->>'owning_application'),
-                                             (s_subscription_object->>'event_types'),
-                                             (s_subscription_object->>'consumer_group'));
+ALTER TABLE zn_data.subscription ADD CONSTRAINT subscription_key_fields_hash_is_unique UNIQUE (s_key_fields_hash);
```
adyach: What do you think about adding it in create table statement?

src/main/java/org/zalando/nakadi/util/HashGenerator.java
```diff
@@ -0,0 +1,25 @@
...
+public class HashGenerator {
+
+    public String generateSubscriptionKeyFieldsHash(final SubscriptionBase subscription) {
+        final Hasher hasher = Hashing.murmur3_128()
+                .newHasher()
+                .putString(subscription.getOwningApplication(), UTF_8)
+                .putString(subscription.getConsumerGroup(), UTF_8);
+        subscription.getEventTypes()
+                .stream()
+                .sorted()
```
adyach: Why sorted ? :)

database/nakadi/10_data/04_tables/02_subscription.sql
v-stepanov: good idea 👍 
fixed here https://github.com/zalando/nakadi/pull/635/commits/8a65d8d7f0469fbfae906d7cd6d3e735e99a176f

v-stepanov:
@adyach I now also added file that will help with migration https://github.com/zalando/nakadi/pull/635/files#diff-d65bd7e7a089f8056a8674103039b168R15

src/main/java/org/zalando/nakadi/util/Aruha771MigrationHelper.java
```diff
@@ -0,0 +1,63 @@
...
+            while (rs.next()) {
+                final String jsonSubscription = rs.getString("s_subscription_object");
+                final Subscription subscription = jsonMapper.readValue(jsonSubscription, Subscription.class);
+                final String hash = hashGenerator.generateSubscriptionKeyFieldsHash(subscription);
+
+                final Statement statement = con.createStatement();
+                statement.execute("UPDATE zn_data.subscription SET s_key_fields_hash='" + hash + "' " +
+                        "WHERE s_id='" + subscription.getId() + "'");
+
+                System.out.println(++i + " | " + subscription.getId() + " | " + hash);
```
adyach: 😮

src/main/java/org/zalando/nakadi/util/Aruha771MigrationHelper.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.Statement;
+
+/**
+ * Temporary class to be used for DB migration of ARUHA-771
+ */
+public class Aruha771MigrationHelper {
```
adyach: I do not really like this class, because It is only "one time run" class, but it will be in our opened repository. Probably, it is better to move it closed repository.

src/main/java/org/zalando/nakadi/util/Aruha771MigrationHelper.java
```diff
@@ -0,0 +1,63 @@
...
+                final String hash = hashGenerator.generateSubscriptionKeyFieldsHash(subscription);
+
+                final Statement statement = con.createStatement();
+                statement.execute("UPDATE zn_data.subscription SET s_key_fields_hash='" + hash + "' " +
+                        "WHERE s_id='" + subscription.getId() + "'");
+
+                System.out.println(++i + " | " + subscription.getId() + " | " + hash);
+            }
+        } catch (final Exception e) {
+            e.printStackTrace();
```
adyach: 😮

src/main/java/org/zalando/nakadi/util/Aruha771MigrationHelper.java
rcillo: I'm not sure. Let's suppose someone uses Nakadi outside of our stack (I know no one does). This user would be interested in this. Also, it would be useful to have some upgrade instructions for this user to perform while upgrading.

But, yeah, given no one else uses it yet, it would be better to have it outside.

src/test/java/org/zalando/nakadi/util/HashGeneratorTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.junit.Test;
+import org.zalando.nakadi.domain.SubscriptionBase;
+import org.zalando.nakadi.view.SubscriptionCursorWithoutToken;
+
+import static com.google.common.collect.Sets.newHashSet;
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.CoreMatchers.not;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class HashGeneratorTest {
```
rcillo: I missed the output of the method `generateSubscriptionKeyFieldsHash` being written at least once in tests.

v-stepanov: What do you mean by `being written`?

src/main/java/org/zalando/nakadi/util/Aruha771MigrationHelper.java
v-stepanov: @adyach @rcillo 
For me the purpose of this class is the same as the purpose of scripts which we put to https://github.com/zalando/nakadi/tree/master/database/migration
It helps to migrate database from one state to another. It is not an SQL-script just because it needs some more complex logic.
Maybe it will make sense to put it not inside nakadi java code but to make a separate source-root for it in this repo. What do you think?

src/main/java/org/zalando/nakadi/util/Aruha771MigrationHelper.java
v-stepanov: what? :)

adyach: 21st century,  we use loggers for that :)

v-stepanov: it's a script, it's not a software :)
It has DB settings written right inside the code, it has no dependency injection and no tests and it is run by main() method :)

adyach: Yes, exactly because of that reason I am not good having it in opensource repo in the main tree :)

src/test/java/org/zalando/nakadi/util/HashGeneratorTest.java
rcillo: Something like `equalsTo("event-type-1|event-type-2|event-type-3")`.

src/main/java/org/zalando/nakadi/util/Aruha771MigrationHelper.java
v-stepanov: 1) It's difficult to do as it requires:
a) Nakadi JsonConfig
b) HashGenerator
c) Subscription domain class
2) For me this class is just a script like sql scripts in https://github.com/zalando/nakadi/tree/master/database/migration - but it is written in different language than SQL.

I agree that I maybe put it to wrong place in our repo so that people can think that it is part of Nakadi source code. I will put it to another source set.

v-stepanov:
@rcillo @adyach I created a separate source-set for migration scripts written in java https://github.com/zalando/nakadi/pull/635/commits/07cc848831358198899b63b1aa54e3e5213532b5
I think it is better this way not to trash actual Nakadi sources.

src/test/java/org/zalando/nakadi/util/HashGeneratorTest.java
v-stepanov: ahh, you mean to test the actual output, so that if the implementation changes - the test will fail, right?

adyach:
:shipit:

src/test/java/org/zalando/nakadi/util/HashGeneratorTest.java
v-stepanov: ok, I added that now https://github.com/zalando/nakadi/pull/635/commits/7a935eda2f2b8a5e9648ab3c8cf3f70ef74f548f
(if this is what you meant :) )

rcillo: 👍

v-stepanov:
👍

adyach:
:shipit:

rcillo:
👍

v-stepanov:
:+1:

lmontrieux:
👍

PR #636, opened 2017-04-26
aruha-747 fix zappr violations

rcillo:
This project is currently non compliant, even though we followed all the
instructions on how to properly setup zappr.

I have been instructed by the team responsible that I should remove
everything else but the 2 necessary lines with X-Zalando-Team and
X-Zalando-Type.


lmontrieux:
👍

rcillo:
👍

PR #641, opened 2017-05-03
Aruha 677 versioning dummy changes

rcillo:
This PR is a different approach to https://github.com/zalando/nakadi/pull/630

Instead of blocking all not semantical changes (changes to schema that have no impact to validation), we are going to accept them and bump patch version.

### DIFF
--- a/src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
+++ b/src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
@@ -97,8 +97,8 @@ public class SchemaEvolutionService {
 
         final List<SchemaChange> changes = schemaDiff.collectChanges(schema(original), schema(eventType));
 
-        final Version.Level changeLevel = semanticOfChange(original, eventType, changes,
-                original.getCompatibilityMode());
+        final Version.Level changeLevel = semanticOfChange(original.getSchema().getSchema(),
+                eventType.getSchema().getSchema(), changes, original.getCompatibilityMode());
 
         if (isForwardToCompatibleUpgrade(original, eventType)) {
             validateCompatibilityModeMigration(original, eventType, changes);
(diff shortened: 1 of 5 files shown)


src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
```diff
@@ -140,18 +143,27 @@ private String formatErrorMessage(final SchemaChange change) {
         return change.getJsonPath() + ": " + errorMessages.get(change.getType());
     }
 
-    private Version.Level semanticOfChange(final List<SchemaChange> changes,
+    private Version.Level semanticOfChange(final EventTypeBase original, final EventTypeBase eventType,
```
lmontrieux: Also I wonder if it would be better to change the name of the second parameter. The first two parameters are event types, so calling the second one `eventType` is a bit confusing.

src/main/java/org/zalando/nakadi/validation/schema/diff/CombinedSchemaDiff.java
```diff
@@ -17,8 +17,7 @@
     static void recursiveCheck(final CombinedSchema combinedSchemaOriginal, final CombinedSchema combinedSchemaUpdate,
                                        final Stack<String> jsonPath,
                                        final List<SchemaChange> changes) {
-        if(!(combinedSchemaOriginal.getSubschemas().containsAll(combinedSchemaUpdate.getSubschemas())
-                && combinedSchemaUpdate.getSubschemas().containsAll(combinedSchemaOriginal.getSubschemas()))) {
+        if(combinedSchemaOriginal.getSubschemas().size() != combinedSchemaUpdate.getSubschemas().size()) {
```
lmontrieux: Is there any chance that both list contain the same number of elements, but that they are different?

rcillo: Yes. They could be reordered. I considered performing some kind of "json sorting" before comparing. But for the sake of simplicity it doesn't support reordering of elements in arrays.

src/main/java/org/zalando/nakadi/validation/SchemaEvolutionService.java
rcillo: 👍  changes made in 71cdb0d

lmontrieux:
👍

lmontrieux:
👍

SergKam:
:+1:

lmontrieux:
👍

v-stepanov:
:+1:

SergKam:
:+1:

PR #653, opened 2017-05-09
Aruha 778 db timeout

lmontrieux:
ARUHA-778 add timeout to database connections

I'm opening this PR now, so we can test and argue over (and eventually agree on) the best value for each timeout.


src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
```diff
@@ -252,7 +252,7 @@ public void whenGetEventsWithUknownTopicThenTopicNotFound() {
                 .then()
                 .statusCode(HttpStatus.NOT_FOUND.value())
                 .and()
-                .contentType(equalTo("application/problem+json;charset=UTF-8"))
```
adyach: Could you elaborate on that ?

src/main/java/org/zalando/nakadi/Application.java
```diff
@@ -2,8 +2,10 @@
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.context.annotation.ComponentScan;
 
 @SpringBootApplication
+@ComponentScan(basePackages = {"org.zalando.nakadi", "org.zalando.nakadi.config"})
```
adyach: Could explain why it requires component scan with the new version of spring boot ?

adyach:
looks cool 👍 
also I think we should update gradle plugin version:
https://github.com/zalando/nakadi/pull/653/files#diff-c197962302397baf3a4cc36463dce5eaL73

src/main/java/org/zalando/nakadi/Application.java
lmontrieux: @adyach After upgrading spring boot, it would not find the org.zalando.nakadi.config.SecuritySettings component to create a bean, during tests. Authorization tests failed as a result.

build.gradle
```diff
@@ -98,7 +98,7 @@ dependencies {
         dropwizardVersion = '3.1.2'
     }
 
-    pgsql 'org.postgresql:postgresql:9.4.1207'
+    pgsql 'org.postgresql:postgresql:42.1.1'
```
adyach: the same entry down there

adyach: and quite strange versioning :)

lmontrieux: @adyach I'm not sure I understand your first question. For the versioning, yes it is strange, but here is the documentation that explains it (I also included it in the commit message): https://jdbc.postgresql.org/documentation/faq.html#versioning

src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
lmontrieux: For some reason (I need to investigate), upgrading spring boot meant that `charset=UTF-8` is not returned anymore. I don't think the change makes a difference for clients.

lmontrieux:
@adyach the only thing that I didn't update was the Problem library - libraries, in fact. Upgrading to the latest version breaks the code in many places, which I think is really too much for a PR that was originally supposed to be about a few JDBC parameters.

build.gradle
adyach: @lmontrieux on the line 124 we have entry:
+    compile 'org.postgresql:postgresql:42.1.1'
looks the same as on the line 101

lmontrieux: @adyach It was there already, I'm looking into it now

lmontrieux: @adyach turns out `pgsql ...` was unnecessary: 3e98e76

lmontrieux:
@adyach I've also extracted the spring framework version number to a variable, for good measure

adyach:
very clean, thank you

adyach:
👍

lmontrieux:
👍

adyach:
👍

lmontrieux:
👍

build.gradle
```diff
@@ -76,7 +76,7 @@ buildscript {
     }
 
     dependencies {
-        classpath "org.springframework.boot:spring-boot-gradle-plugin:${springBootVersion}"
+        classpath "org.springframework.boot:spring-boot-gradle-plugin:1.2.5.RELEASE"
```
adyach: what do you think about filing a ticket against spring boot plugin ?

lmontrieux: @adyach it's possible that we're doing something wrong, or that some change or another is documented in any version of spring-boot-gradle-plugin between 1.2.5 and 1.5.3. So, I first want to look at that. But I think this should be in another ticket.

adyach:
👍

src/main/resources/application.yml
```diff
@@ -18,15 +18,17 @@ spring:
...
+      initial-size: 2
+      max-active: 2
+      min-idle: 1
+      max-idle: 1
+      test-on-borrow: true
+      validation-query: SELECT 1
+      test-while-idle: true
+      time-between-eviction-runs-millis: 5000
+      min-evictable-idle-time-millis: 60000
+      connection-properties: socketTimeout=2;connectTimeout=2;loginTimeout=2
```
adyach: in my opinion it should be lesser than that,  < 1s

lmontrieux: @adyach unfortunately these timeouts are in seconds, and the type is `int`, so the minimum would be 1 second.

lmontrieux:
👍

adyach:
👍

antban:
:+1:

adyach:
👍

lmontrieux:
👍

zrobgar:
👍

PR #656, opened 2017-05-11
Code style diff file

adyach:
Looking at many PRs in our repo, there is a lot of unnecessary movements of imports. IDEA allows to export code style difference between your settings and IDEA. I decided to finish it exporting my idea code style diff.
So lets try to adopt that format to avoid huge import movements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,8 +21,11 @@ hs_err_pid*
 *.iml
 
 ## Directory-based project format:
-.idea/
+.idea/*
 # if you remove the above rule, at least ignore the following:
+
+# exclude code style diff
+!.idea/codeStyleSettings.xml
 
 # User-specific stuff:
 # .idea/workspace.xml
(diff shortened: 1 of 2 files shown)


lmontrieux:
:+1:

lmontrieux:
@adyach are we going to merge this? Does everyone agree? @antban @v-stepanov @rcillo @zrobgar

adyach:
@lmontrieux you know, I would love to try it out

antban:
:+1:

.gitignore
```diff
@@ -21,9 +21,12 @@ hs_err_pid*
 *.iml
 
 ## Directory-based project format:
-.idea/
+.idea/*
```
v-stepanov: For me Idea doesn't really handle that correctly. Idea considers the files that are located in this DIR as git-files. So when I change something in one of these files - the file then is put to the list of files to commit ("Local changes"). Sometimes after some time Idea marks modified file as none-repo file but then again detects the changes. :(
@adyach

adyach: hm, honestly, I do not commit from IDEA, because of that I have not seen a such issue :) also, I thought the IDEA does not really change these files very often. Maybe it is a bug in IDEA?
Also my setup is:
git version 2.10.0
IDEA 2017.1.4

PR #658, opened 2017-05-12
Aruha 792 log slos

v-stepanov:
https://techjira.zalando.net/browse/ARUHA-792

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
@@ -1,5 +1,6 @@
 package org.zalando.nakadi.controller;
 
+import com.google.common.base.Charsets;
 import org.json.JSONException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -25,7 +26,7 @@ import org.zalando.problem.ThrowableProblem;
 import org.zalando.problem.spring.web.advice.Responses;
 
 import javax.ws.rs.core.Response;
-import java.nio.charset.StandardCharsets;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import static org.springframework.http.ResponseEntity.status;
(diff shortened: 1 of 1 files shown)


src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -100,19 +106,23 @@ private ResponseEntity postEventInternal(final String eventTypeName,
         }
     }
 
+    private void reportSLOs(final long startingNanos, final int totalSizeBytes, final int eventCount) {
+        final long msSpent = TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startingNanos);
```
adyach: Probably, you transform from nanos to millis you lose precision. Maybe just take System.currentTimeMillis() ?

v-stepanov: hmm.. what's the difference?

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -100,19 +106,23 @@ private ResponseEntity postEventInternal(final String eventTypeName,
...
+        LOG.info("[SLO] [publishing-latency] time={} size={} count={}", msSpent, totalSizeBytes, eventCount);
```
adyach: Shouldn't we report only successfully published events? or somehow indicate that not all of them were published in slo log entry ..

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
adyach: I will rephrase the question: "why do you convert from ns to ms, just use ms right away" :)

v-stepanov: Because the startTime I already have in nanos (this existed before my changes)
So to get everything in ms I will have to do an extra call.

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
v-stepanov: Yes, you are right. I think we agreed to measure this only for success requests. I will add changes here.

v-stepanov: Fixed https://github.com/zalando/nakadi/pull/658/commits/c827015adb2f8b7971c31b93fa7487d18cdce7a9

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
adyach: It is not clear why we need ns in one other method, maybe we can refactor it a bit. What do you think?

v-stepanov: I actually like that it is nanoseconds - that gives you better precision. 
`toMS(nano - nano)` gives better precision than `ms - ms`

adyach: So, we agreed to leave it as it is, since nanos is used in another place

adyach:
👍

v-stepanov:
👍

v-stepanov:
👍

adyach:
👍

PR #659, opened 2017-05-15
Reduce allocations writing batch events for event consumers 

antban:
If SEND_BATCH_VIA_OUTPUT_STREAM is enabled, send the batch and its events directly to the consumer's OutputStream as bytes instead of building up an intermediate batch String via createStreamEvent. The standard parts of the batch are allocated statically as bytes instead of per call.

This is done to reduce the number of allocations generated for a batch and is based on some performance tuning investigations.

EventStreamTest is cloned as is to EventStreamTestSendBatchViaOutputStream, to exercise the current tests without changing them (to avoid adding test bugs), with some extra tests add to EventStreamTest that check the new writeStreamEvent method directly. The clone can be removed once the flagged path becomes the default and the flag is removed.
This is copy of PR #589 , to make it simpler to test and build
Related to ARUHA-641 and ARUHA-793


src/main/java/org/zalando/nakadi/util/FeatureToggleService.java
```diff
@@ -29,7 +29,9 @@
         CHECK_APPLICATION_LEVEL_PERMISSIONS("check_application_level_permissions"),
         CHECK_PARTITIONS_KEYS("check_partitions_keys"),
         CHECK_OWNING_APPLICATION("check_owning_application"),
-        LIMIT_CONSUMERS_NUMBER("limit_consumers_number");
+        LIMIT_CONSUMERS_NUMBER("limit_consumers_number"),
+        ZERO_PADDED_OFFSETS("zero_padded_offsets"),
```
v-stepanov: ghost from the past

v-stepanov:
👍

antban:
:+1:

PR #661, opened 2017-05-16
feature: get event type by topic

adyach:
Admin api to get ET by topic + fixed errors swagger file

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -1634,7 +1634,7 @@ paths:
           required: true
       responses:
         '200':
-          description: Event type wihch is represented by provided topic
+          description: Event type which is represented by provided topic
           schema:
             $ref: '#/definitions/EventType'
         '404':
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -583,7 +583,7 @@ paths:
           schema:
             type: array
             items:
-              $ref: "#/definitions/CursorDistanceQuery"
+              $ref: '#/definitions/CursorDistanceQuery'
           required: true
       responses:
         '200':
@@ -745,6 +745,7 @@ paths:
           description: |
             Offset to query for unconsumed events. Depends on `partition` parameter. When present adds the property
             `unconsumed_events` to the response partition object.
+          type: string
       responses:
         '200':
           description: OK
@@ -817,7 +818,7 @@ paths:
               - if the `shift` provided leads to a cursor that is not yet existent, i.e. it's pointing to
               some cursor yet to be generated in the future.
             schema:
-              - $ref: '#/definitions/Problem'
+              $ref: '#/definitions/Problem'
 
   /subscriptions:
     post:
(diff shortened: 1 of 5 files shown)


api/nakadi-event-bus-api.yaml
```diff
@@ -1617,6 +1618,34 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+  /topics/{name}:
+    get:
+      tags:
+        - topics-api
+      description: |
+        Retrieves the event type by provided topic.
+        It searches for event type using the latest active timeline.
```
antban: Why only latest

adyach: It was my requirement

api/nakadi-event-bus-api.yaml
```diff
@@ -1617,6 +1618,34 @@ paths:
...
+        - name: name
+          in: path
+          description: Name of the topic
+          type: string
+          required: true
+      responses:
+        '200':
+          description: Event type wihch is represented by provided topic
+          schema:
+            $ref: '#/definitions/EventType'
```
antban: Theoretically it should be an array

src/main/java/org/zalando/nakadi/service/TopicsService.java
```diff
@@ -0,0 +1,61 @@
...
+        this.timelineDbRepository = timelineDbRepository;
+        this.eventTypeRepository = eventTypeRepository;
+    }
+
+    public EventType getEventTypeByTopic(final String topic, final Client client) {
+        if (!client.getClientId().equals(securitySettings.getAdminClientId())) {
+            throw new ForbiddenAccessException("Request is forbidden for user " + client.getClientId());
+        }
+
+        final Optional<String> etName = timelineDbRepository.listTimelinesOrdered().stream()
```
antban: I understand that this operation is really rare, but you will fetch all Timelines in db. At some point nakadi will just die.

adyach: My point was that there are no timelines in database

api/nakadi-event-bus-api.yaml
```diff
@@ -1617,6 +1618,34 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+  /topics/{name}:
+    get:
```
v-stepanov: This is not very RESTful. When I do GET /topics/myTopic I expect to get some topic object but instead I get EventType (for which there is a separate resource)
So the endpoint should be more like "get event-type by topic"

I think the endpoint
`GET /event-types?topic=9bc91968-101b-435b-b697-33d47f8c69c4` will fit more

src/main/java/org/zalando/nakadi/controller/TopicsController.java
```diff
@@ -0,0 +1,54 @@
...
+import org.zalando.nakadi.exceptions.runtime.NoTopicFoundException;
+import org.zalando.nakadi.security.Client;
+import org.zalando.nakadi.service.TopicsService;
+import org.zalando.problem.Problem;
+import org.zalando.problem.spring.web.advice.Responses;
+
+import javax.ws.rs.core.Response;
+
+@RestController
+@RequestMapping(value = "/topics/")
```
v-stepanov: I am missing security configuration for the new endpoint. (And the test for that)

lmontrieux:
@adyach are you still working on this?

adyach:
@lmontrieux yep, give me sometime for that.

PR #664, opened 2017-05-18
ARUHA-802 Update lastCommitTime for subscription

antban:
when stream switches from state where everything was committed to state with something is uncommitted.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
@@ -48,7 +48,6 @@ class StreamingState extends State {
     private ZKSubscription topologyChangeSubscription;
     private EventConsumer.ReassignableEventConsumer eventConsumer;
     private boolean pollPaused;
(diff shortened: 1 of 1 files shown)


antban:
:+1:

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -192,12 +203,15 @@ private void checkBatchTimeouts() {
     private void streamToOutput() {
         final long currentTimeMillis = System.currentTimeMillis();
         int messagesAllowedToSend = (int) getMessagesAllowedToSend();
+        final boolean wasCommitted = isEverythingCommitted();
```
adyach: is it the same as `messagesAllowedToSend == 0` ?

antban: no, it's not messagesAllowedToSend is > 0 even if something was uncommitted

adyach:
👍

PR #666, opened 2017-05-23
aruha-815: changed description of the stats object

adyach:
This commit clarifies on subscription stats object about
unconsumed_events property absence

### DIFF
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -2354,9 +2354,9 @@ definitions:
             unconsumed_events:
               type: number
               description: |
-                The amount of events in this partition that are not yet consumed within this subscription. May be not
-                determined at the moment when no events were yet consumed from the partition in this subscription (in
-                that case the property will be absent).
+                The amount of events in this partition that are not yet consumed within this subscription.
+                The property may be absent at the moment when no events were yet consumed from the partition in this
+                subscription (In case of `read_from` is `BEGIN` the `unconsumed events` is not shown)
             stream_id:
               type: string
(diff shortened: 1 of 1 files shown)


api/nakadi-event-bus-api.yaml
```diff
@@ -2354,9 +2354,9 @@ definitions:
             unconsumed_events:
               type: number
               description: |
-                The amount of events in this partition that are not yet consumed within this subscription. May be not
-                determined at the moment when no events were yet consumed from the partition in this subscription (in
-                that case the property will be absent).
+                The amount of events in this partition that are not yet consumed within this subscription.
+                The property may be absent at the moment when no events were yet consumed from the partition in this
+                subscription (In case of `read_from` is `BEGIN` the `unconsumed events` is not shown)
```
v-stepanov: What about read_from=END?

v-stepanov: Also for the the phrase "In case of `read_from` is `BEGIN` the `unconsumed events` is not shown" sounds like it is never shown for such a subscription.

adyach: 1729395

adyach: That's why it is in curly brackets :)

v-stepanov: Then I think it's better to remove `"the `unconsumed events` is not shown"` at the end

adyach: 3733f71

v-stepanov:
👍

adyach:
👍

PR #667, opened 2017-05-23
Aruha 756 authorization api

lmontrieux:
[ARUHA-756](https://techjira.zalando.net/browse/ARUHA-756)

NOTE: This PR *does not* include a description of the types of subject attributes that will be accepted by Nakadi. This is on purpose, as these will be implemented separately, and will depend on a deployment configuration. Examples of types of attributes are usernames, team names, application names, scopes, roles, or permissions.


api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,61 @@ definitions:
...
+      - read
+      - write
+
+  AccessControlListItem:
+    type: object
+    description: |
+      An attribute for authorization. This object includes the type of attribute (team, permission, etc.),
+      and the attribute value. A catch-all can be represented with type 'ALL', and value 'ALL'.
+    properties:
+      attribute_type:
```
v-stepanov: I think we should define what are the possible values for `attribute_type`
Because currently it's not exactly clear what I can specify here.

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,61 @@ definitions:
...
+    description: |
+      Authorization section for an event type. This section defines three access control lists: one for producing
+      events ('write'), one for consuming events ('read'), and one for administering an event type ('admin').
+      Regardless of the values of the authorization properties, administrator accounts will always be authorized.
+    properties:
+      admin:
+        type: array
+        description: |
+          An array of subject attributes that are required for updating the event type. Any one of the attributes
+          defined in this array is sufficient to be authorized. The catch-all item takes precedence over all others,
```
v-stepanov: I didn't quite get it about catch-all item.

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,61 @@ definitions:
...
+          all others, i.e. if it is present, all users are authorized.
+        items:
+          $ref: '#/definitions/AccessControlListItem'
+        minItems: 1
+    required:
+      - admin
+      - read
+      - write
+
+  AccessControlListItem:
```
v-stepanov: I think on alignment item we had some nicer name for this entity.
I think to call it ListItem is strange because everything can be a list item :)

api/nakadi-event-bus-api.yaml
lmontrieux: fb2a3c7

api/nakadi-event-bus-api.yaml
lmontrieux: d63bbb2

api/nakadi-event-bus-api.yaml
lmontrieux: @v-stepanov I'm not sure about this. Sure, we should have it somewhere, but the list of supported attribute types will vary, and come from the authorization plugin - so it is deployment-specific. What do you think?

api/nakadi-event-bus-api.yaml
lmontrieux: We did call them roles, but... they're not roles :)
So I want to use another name. I think `AuthorizationAttribute` is better?

api/nakadi-event-bus-api.yaml
v-stepanov: Ah, ok. I didn't know that it is deployment-specific. 
I just think that API guild will have the same question.

api/nakadi-event-bus-api.yaml
v-stepanov: Ok, but I still don't get it. Should one just use "*" as an element of this array?

api/nakadi-event-bus-api.yaml
```diff
@@ -263,6 +271,11 @@ paths:
           description: Unprocessable Entity
           schema:
             $ref: '#/definitions/Problem'
+        '403':
+          description: Access forbidden
```
adyach: Do you think it makes sense to add lines 243-244 to description of the error?

lmontrieux: It won't hurt, and it will be clearer to users. I will add it to all 403s that can be caused by event type authorization failures.

api/nakadi-event-bus-api.yaml
lmontrieux: @v-stepanov I've updated the PR description to address this question.

api/nakadi-event-bus-api.yaml
lmontrieux: @v-stepanov swagger doesn't support `oneOf` yet, so I'm not sure how to achieve this other than using a discriminator, which I think is even more confusing. Is it ok if I leave it as it is?

api/nakadi-event-bus-api.yaml
lmontrieux: @adyach 714aaa8

api/nakadi-event-bus-api.yaml
v-stepanov: Maybe we should then change it? Because that will confuse users a lot as it now confuses me :)
Maybe "*" can be a possible value for attribute_value ?

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,62 @@ definitions:
...
+        minItems: 1
+      read:
+        type: array
+        description: |
+          An array of subject attributes that are required for reading events from the event type. Any one of the
+          attributes defined in this array is sufficient to be authorized. The wildcard item takes precedence over
+          all others, i.e., if it is present, all users are authorized.
+        items:
+          $ref: '#/definitions/AuthorizationAttribute'
+        minItems: 1
```
v-stepanov: one more question. How to make it forbidden-for-all?

lmontrieux: I don't think it should be. What would be the point of an event type that nobody can administer, read from, or write to? Plus, admins are always authorized anyway, so it would not be a real forbiddent-for-all.

api/nakadi-event-bus-api.yaml
lmontrieux: @v-stepanov fixed in 5c99708

v-stepanov: I mean if we try to do something that is not swagger-compatible then we maybe shouldn't do it and implemented it in some other way.

api/nakadi-event-bus-api.yaml
v-stepanov: The only use-case I see is that team created ET and pushing events but they are not read by anyone yet. But maybe it's not that often use-case.

lmontrieux: In this case, I don't think it is a problem that the users who can write to the ET can also read from it. They already know what's in there

v-stepanov:
👍

api/nakadi-event-bus-api.yaml
lmontrieux: I see 2 other options:
- an object whose fields are all optional, and each field represents a type of attribute. So you'd have for example:
```
{ "read":
  { "users": ["slava", "lionel", "*"],
    "teams": ["aruha", "flux"]
  }
}
```
I don't like it because it's hard to exhaustively define what the fields are, since types are implemented in the authorization plugin. But the wildcard could be anywhere.

The other option I see is to use a single array of strings, formatted as `<type>::<value>`:
```
{
"read": ["user::slava", "user::lionel", "*", "team::aruha", "team::flux"]
}
```
We'll have to split the strings, but otherwise it works well with wildcards. I would prefer this option over the one above.

lmontrieux: I discussed this offline with @v-stepanov and we agreed that it is fine as it is.

lmontrieux:
As far as I can tell, Zally only reports one new issue caused by the introduction of authorization:
SHOULD – Array names should be pluralized
#/definitions/EventTypeAuthorization: 'admin','read','write'

I believe that the names 'admin', 'read', and 'write' should not be changed in this case, as they are widely used in authorization.

api/nakadi-event-bus-api.yaml
```diff
@@ -754,6 +790,10 @@ paths:
           description: Client is not authenticated
           schema:
             $ref: '#/definitions/Problem'
+        '403':
+          description: Access forbidden because of EventType authorization failure.
```
conorclifford: is this the only reason for this response code? Or can this code also be returned for callers without the required oauth scope? if the latter, it might be worth including that possibility in the description.

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,62 @@ definitions:
...
+    description: |
+      Authorization section for an event type. This section defines three access control lists: one for producing
+      events ('write'), one for consuming events ('read'), and one for administering an event type ('admin').
+      Regardless of the values of the authorization properties, administrator accounts will always be authorized.
+    properties:
+      admin:
+        type: array
+        description: |
+          An array of subject attributes that are required for updating the event type. Any one of the attributes
+          defined in this array is sufficient to be authorized. The wildcard item takes precedence over all others,
```
rcillo: Have you considered using an empty list instead of having a wildcard? This way, all existent event types would be open by default.

api/nakadi-event-bus-api.yaml
```diff
@@ -2215,6 +2269,8 @@ definitions:
           contains at least one of the scopes in this list.
           If no scopes provided then anyone can publish to this event type.
 
+          Usage of write_scopes is deprecated.
```
ePaul: Does this mean it is ignored, or it is mapped to the EventTypeAuthorization.read?

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,62 @@ definitions:
...
+  AuthorizationAttribute:
+    type: object
+    description: |
+      An attribute for authorization. This object includes the type of attribute (team, permission, etc.),
+      and the attribute value. A wildcard can be represented with type '*', and value '*'. It means that all
+      authenticated users are allowed to perform an operation.
+    properties:
+      attribute_type:
+        type: string
+        description: the type of attribute (e.g., 'team', or 'permission')
```
conorclifford: could this be an `extensible-enum` listing the current set of possible values?

api/nakadi-event-bus-api.yaml
lmontrieux: It will be ignored, and hopefully eventually removed. In Zalando deployments, the feature has been disabled months ago, so they are already ignored.

api/nakadi-event-bus-api.yaml
ePaul: Or you could use an attribute name-value pair which is not given to anyone.

api/nakadi-event-bus-api.yaml
lmontrieux: I have, but I think we discussed this in alignment, and decided that it would be too easy for a user to accidentally open their event type to all. Having to explicitly include a wildcard makes such errors less likely.

api/nakadi-event-bus-api.yaml
lmontrieux: Good point, I'll add that

api/nakadi-event-bus-api.yaml
```diff
@@ -568,6 +584,8 @@ paths:
 
         Calculate the distance between two offsets. This is useful for performing checks for data completeness, when
         a client has read some batches and wants to be sure that all delivered events have been correctly processed.
+
+        If per-EventType authorization is enabled, the called must be authorized to read from the EventType.
```
musiKk: This should be "caller" instead of "called", no? There are many occurrences of this in this document.

api/nakadi-event-bus-api.yaml
lmontrieux: I'm a little worried about that, because possible values will depend on the particular deployment. The authorization module is actually in a separate repository, as it is Zalando-specific, so it is service-dependent, in my opinion. We are however looking into exposing various configuration values, including a list of accepted attribute types, either as a new endpoint, or in our GHE wiki.

api/nakadi-event-bus-api.yaml
lmontrieux: Correct, that's a typo.

lmontrieux: @musiKk fixed in 360ad57

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,62 @@ definitions:
...
 
+  EventTypeAuthorization:
+    type: object
+    description: |
+      Authorization section for an event type. This section defines three access control lists: one for producing
+      events ('write'), one for consuming events ('read'), and one for administering an event type ('admin').
+      Regardless of the values of the authorization properties, administrator accounts will always be authorized.
+    properties:
+      admin:
+        type: array
```
ePaul: @lmontrieux wrote (in a top-level comment):

> As far as I can tell, Zally only reports one new issue caused by the introduction of authorization:
SHOULD – Array names should be pluralized
#/definitions/EventTypeAuthorization: 'admin','read','write'

> I believe that the names 'admin', 'read', and 'write' should not be changed in this case, as they are widely used in authorization.

I guess the point is that the value of those properties are not an "admin", a "read" or a "write", but lists of authorizations. So maybe better names would be `admin_authorizations`, `read_authorizations` and `write_authorizations`?

api/nakadi-event-bus-api.yaml
lmontrieux: @conorclifford fixed in aebee3e

api/nakadi-event-bus-api.yaml
lmontrieux: We could have attribute_type: "none", attribute_value: "none", but then we need someone will eventually mix a 'none' with a '*', and it will be confusing. Unless we only allow 'none' to be the only element in the array.

api/nakadi-event-bus-api.yaml
```diff
@@ -722,7 +755,9 @@ paths:
       security:
         - oauth2: ['nakadi.event_stream.read']
 
-      description: Returns the given `Partition` of this EventType
+      description: |
+        Returns the given `Partition` of this EventType. If per-EventType authorization is enabled, the called must
```
musiKk: One "called" -> "caller" still left. ;)

api/nakadi-event-bus-api.yaml
```diff
@@ -597,6 +615,10 @@ paths:
           description: Unprocessable Entity
           schema:
             $ref: '#/definitions/Problem'
+        '403':
+          description: Access forbidden because of missing scope of EventType authorization failure.
```
musiKk: Typo? "of" should be "or", right? 6 occurrences total.

api/nakadi-event-bus-api.yaml
conorclifford: ah ok, understandable.

api/nakadi-event-bus-api.yaml
lmontrieux: Doh! 86399dd

api/nakadi-event-bus-api.yaml
lmontrieux: Sorry :( fixed in 86399dd

api/nakadi-event-bus-api.yaml
musiKk: I think the approach is sensible. We also had to make a similar decision in the past and also went with the explicit wildcard. However, the `authorization` property in the event type is optional and defaults to "open for all" if absent so I'm not sure if much is won w.r.t. avoiding accidental event type opening.

lmontrieux: @musiKk Eventually we want to make the `authorization` property mandatory, but that is something that may break some clients who create event types automatically. Starting with an optional property allows us to deploy the feature quickly, and progressively enforce its usage.

musiKk: > that may break some clients who create event types automatically

As someone who has been bitten by exactly that in the past I appreciate that a lot.

api/nakadi-event-bus-api.yaml
lmontrieux: If we go down that route, I would prefer something like `admin_attributes`, `read_attributes`, and `write_attributes`, which is shorter and more precise (these are lists of authorization attributes). But I still think, imho, that `admin`, `read`, and `write` are sufficient, and clearer.

api/nakadi-event-bus-api.yaml
tkrop: Yes, agreed! But from this point of view you should throw out the **e.g. ...** and **wildcard sentence above** and state explicitly that supported types and values are depending on the deployed/configured *attribute-based access control* module.

Anyhow, from a Nakadi user perspective I still like to have more guidance on how to work with the authorization concept. Can we link to an additional internal manual/api?

api/nakadi-event-bus-api.yaml
ePaul: Hmm, `reads` and `writes` sounds wrong, even though it complies to the letter of the guideline. The entries there are not reads/writes.

Maybe `readers` and `writers` (together with `admins`), with the meaning that each item in that list indicates who is allowed to read/write/administrate?

tkrop: @ePaul Yes, `readers` and `writers` is even better.

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,62 @@ definitions:
...
+    type: object
+    description: |
+      Authorization section for an event type. This section defines three access control lists: one for producing
+      events ('write'), one for consuming events ('read'), and one for administering an event type ('admin').
+      Regardless of the values of the authorization properties, administrator accounts will always be authorized.
+    properties:
+      admin:
+        type: array
+        description: |
+          An array of subject attributes that are required for updating the event type. Any one of the attributes
```
lmontrieux: In the literature on access control, attributes are categorised in `subject attributes`, `action attributes`, `resource attributes`, and `environment attributes` (see e.g. the XACML standard: http://docs.oasis-open.org/xacml/3.0/xacml-3.0-core-spec-os-en.html#_Toc325047262). In this case we try to keep things as simple as possible, so we do not want to write complex rules. In particular, attributes will always be subject attributes.

But, if it is less confusing to simply call them `attributes` or `authorization attributes`, then I think we should go with the least confusing option.

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,62 @@ definitions:
...
+      - write
+
+  AuthorizationAttribute:
+    type: object
+    description: |
+      An attribute for authorization. This object includes the type of attribute (team, permission, etc.),
+      and the attribute value. A wildcard can be represented with type '*', and value '*'. It means that all
+      authenticated users are allowed to perform an operation.
+    properties:
+      attribute_type:
```
lmontrieux: @tkrop I don't really like the `attribute_*` prefix either, but found that only using `type` is very confusing when reading the specification (it's used to denote the type of a field).
The attribute type here, however, is not `read`, `write`, or `admin`. It is rather something like `username`, `team`, `scope`, or `permission`.

lmontrieux:
@tkrop We have a concept document, a link to which you can find in the API review spreadsheet (it's not public, which is why I don't add the link here. Maybe it should be public, I'll see if we can make that happen). I agree with your point about explicitly stating that supported types and values depend on the authorization module.

api/nakadi-event-bus-api.yaml
lmontrieux: @ePaul @tkrop I can live with `readers`, `writers`, and `admins` :)

api/nakadi-event-bus-api.yaml
tlossen: could also be called `key` and `value`, as the `AuthorizationAttribute` is basically a key-value-pair?

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,62 @@ definitions:
...
+          all others, i.e., if it is present, all users are authorized.
+        items:
+          $ref: '#/definitions/AuthorizationAttribute'
+        minItems: 1
+      write:
+        type: array
+        description: |
+          An array of subject attributes that are required for writing events to the event type. Any one of the
+          attributes defined in this array is sufficient to be authorized. The wildcard item takes precedence over
+          all others, i.e. if it is present, all users are authorized.
```
tlossen: in other words, the attributes are implicitly joined by OR. i don't think that the wildcard needs any kind of precedence, then.

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,62 @@ definitions:
...
+        items:
+          $ref: '#/definitions/AuthorizationAttribute'
+        minItems: 1
+    required:
+      - admin
+      - read
+      - write
```
tlossen: what happened to the `list` permission described in the concept?

api/nakadi-event-bus-api.yaml
lmontrieux: That's correct

api/nakadi-event-bus-api.yaml
lmontrieux: There was some debate over whether or not `list` should be included in the initial version. It was eventually decided that it would be added later, if users ask for it.

api/nakadi-event-bus-api.yaml
tlossen: i would maybe avoid this term then. the wildcard behaviour is already explained by the previous sentence:
> Any one of the attributes defined in this array is sufficient to be authorized.

api/nakadi-event-bus-api.yaml
tlossen: ah, ok.

api/nakadi-event-bus-api.yaml
lmontrieux: @tkrop 95a93f5

api/nakadi-event-bus-api.yaml
lmontrieux: @tlossen 95a93f5

api/nakadi-event-bus-api.yaml
lmontrieux: @tlossen b14f463

api/nakadi-event-bus-api.yaml
lmontrieux: @ePaul @tkrop 1ae025b

lmontrieux:
@tlossen Thank you for your review. I believe that we have addressed all your comments?

lmontrieux:
👍

api/nakadi-event-bus-api.yaml
```diff
@@ -2328,6 +2386,62 @@ definitions:
...
+      readers:
+        type: array
+        description: |
+          An array of subject attributes that are required for reading events from the event type. Any one of the
+          attributes defined in this array is sufficient to be authorized. The wildcard item takes precedence over
+          all others, i.e., if it is present, all users are authorized.
+        items:
+          $ref: '#/definitions/AuthorizationAttribute'
+        minItems: 1
+      writeers:
```
tkrop: `writeers` -> `writers`

lmontrieux: @tkrop thank you! d407db6

api/nakadi-event-bus-api.yaml
ePaul: Could you add a "and is ignored" to this sentence, to make this clear? (Similarly for the read_scopes 12 lines below).

lmontrieux:
@tlossen I've made a small name change in `AuthorizationAttribute`, updated field `key` to `data_type`. This is to be consistent with the vocabulary in the XACML standard.

lmontrieux:
👍

v-stepanov:
👍

PR #672, opened 2017-05-31
ARUHA-817 Validate occured_at in a more accurate way

antban:
ARUHA-817

### DIFF
--- a/src/main/java/org/zalando/nakadi/validation/RFC3339DateTimeValidator.java
+++ b/src/main/java/org/zalando/nakadi/validation/RFC3339DateTimeValidator.java
@@ -13,7 +13,7 @@ public class RFC3339DateTimeValidator implements FormatValidator {
     private final String errorMessage = "must be a valid date-time";
 
     // Valid offsets are either Z or hh:mm. The format hh:mm:ss is not valid
-    private final String dateTimeOffsetPattern = "^.*T\\d{2}:\\d{2}:\\d{2}.*(Z|((\\+|-)\\d{2}:\\d{2}))$";
+    private final String dateTimeOffsetPattern = "^.*[Tt]\\d{2}:\\d{2}:\\d{2}.*([zZ]|([+-]\\d{2}:\\d{2}))$";
     private final Pattern pattern = Pattern.compile(dateTimeOffsetPattern);
     private final Optional<String> error = Optional.of(errorMessage);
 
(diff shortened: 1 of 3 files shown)


rcillo:
👍

antban:
:+1:

zrobgar:
👍

antban:
:+1:

src/main/java/org/zalando/nakadi/validation/RFC3339DateTimeValidator.java
```diff
@@ -1,21 +1,19 @@
...
-import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
+import org.everit.json.schema.FormatValidator;
 
 public class RFC3339DateTimeValidator implements FormatValidator {
 
     private final String errorMessage = "must be a valid date-time";
 
     // Valid offsets are either Z or hh:mm. The format hh:mm:ss is not valid
-    private final String dateTimeOffsetPattern = "^.*(Z|((\\+|-)\\d\\d:\\d\\d))$";
+    private final String dateTimeOffsetPattern = "^.*[Tt]\\d{2}:\\d{2}:\\d{2}.*([zZ]|([+-]\\d{2}:\\d{2}))$";
```
adyach: it matches to 1996-10-15T16:39:57.123+**99:99**

antban: @adyach It won't be accepted because of `OffsetDateTime.parse`

adyach:
👍

antban:
:+1:

src/main/java/org/zalando/nakadi/validation/RFC3339DateTimeValidator.java
ePaul: Do you really want a `.*` here for the fractional seconds?

rcillo: Even if it's improved to be just numbers, this is also covered by the Java `OffsetDateTime.parse`.

Unfortunately, this parser is not as strict as we would like, so we need to have this additional regex to cover corner cases.

ePaul: Hmm, maybe indicate this purpose in the comment above, like `// regular expression to catch validation cases not caught by OffsetDateTime.parse`.

antban:
:+1:

adyach:
👍

antban:
:+1:

antban:
:+1:

lmontrieux:
👍

antban:
:+1:

PR #673, opened 2017-06-06
Aruha 840: Optimize subscription streaming

rcillo:
This PR is a "rebase" of #590.

Since a lot has changed in the code, it's not a true rebase but an
adaptation. The new subscription streaming code no longer uses `List<String>`
events but instead a `List<ConsumedEvent>`.

In order to avoid transforming the later into the former and allocating
a new list on every call to write to output stream, I created a new
method writeSubscriptionBatch that accepts lists of `ConsumedEvent`s.

This code is not the most straightforward, but the intention is to avoid
allocating extra lists and objects. If you have a better idea on how to
implement it, please, share.


src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -110,10 +110,8 @@ public void onException(final Exception ex) {
         }
 
         @Override
-        public void streamData(final byte[] data) throws IOException {
-            headersSent = true;
```
rcillo: @antban it wasn't clear why `headersSent` should be set in this method. It's already set in line 96 in the method above `onException` and also in line 84 in method `onInitialized`. It was not clear which scenario required it to be set in the `streamData`.

Also, this new streaming code requires access to the output stream directly, since it writes arrays of bytes to it directly. This code is shared between low level and high level, and only high level encapsulates the output stream.

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
```diff
@@ -48,7 +68,7 @@ public int writeBatch(final OutputStream os, final Cursor cursor, final List<Str
         if (!events.isEmpty()) {
             os.write(B_EVENTS_ARRAY_BEGIN);
             for (int i = 0; i < events.size(); i++) {
-                final byte[] event = events.get(i).getBytes(StandardCharsets.UTF_8);
+                final byte[] event = extractor.apply(events.get(i)).getBytes(StandardCharsets.UTF_8);
```
adyach: @rcillo what do you think about having slightly more code duplication in favour avoiding type casting ?

rcillo: That's a possibility I didn't consider. I could duplicate this entire method in order to avoid this "extrator" thing. The advantage would be to have less type casting. On the other hand we would duplicate some code, in both binary and string classes.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -239,41 +235,26 @@ private void flushData(final EventTypePartition pk, final List<ConsumedEvent> da
...
+                    sentOffset,
+                    getContext().getCursorTokenService().generateToken());
+
+            final int batchSize = getContext().getWriter().writeSubscriptionBatch(
+                    getOut().getOutputStream(),
+                    cursor,
+                    data,
+                    metadata);
+
+            getOut().getOutputStream().flush();
```
adyach: Could we do it in the EventStreamWriter?

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
adyach: @rcillo also I expected that we will avoid doing getBytes, since Kafka has ByteArraySerializer/Deserializer

rcillo: @adyach I removed casting by duplicating the code. It was not a "minimal" duplication but a full duplication.

As to Kafka byte array, this is not part of this PR. This PR is just about adapting another PR mentioned in the description. Further optimisations should be done in another ticket.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
rcillo: I think this is the right place to be, since the writer is just a writer and should have no opinion on when something should be flushed or not.

adyach: In my opinion it is the work of the writer, because it writes data to the output stream, and we never know when the system will flush its buffers. From other hand work with output stream in different parts of the code dilutes the usage of the output stream which is not good for the future work with the code.

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
antban: @rcillo Actually I do not remember why it is set here ...

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
```diff
@@ -62,4 +82,59 @@ public int writeBatch(final OutputStream os, final Cursor cursor, final List<Str
...
+                }
+            }
+        }
+        if (metadata.isPresent()) {
+            os.write(B_DEBUG_BEGIN);
+            byteCount += B_DEBUG_BEGIN.length;
+
+            final byte[] debug = metadata.get().getBytes(StandardCharsets.UTF_8);
+            os.write(debug);
+            byteCount += debug.length;
```
antban: Can metadata or event type name possibly contain something that should be escaped? (probably not, just to ensure that you checked debug calls)

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
rcillo: 👍  moved call of flush to the writer 029c39d

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
rcillo: @antban I thought about it. Event type name is safe and I hope that debug will remain safe since it's generated by the Nakadi and is not result of user input.

adyach:
👍

adyach:
👍

rcillo:
👍

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
```diff
@@ -44,11 +63,59 @@ public int writeBatch(final OutputStream os, final Cursor cursor, final List<Str
...
+    }
+
+    @Override
+    public int writeSubscriptionBatch(final OutputStream os, final SubscriptionCursor cursor,
+                                      final List<ConsumedEvent> events,
+                                      final Optional<String> metadata) throws IOException {
+        int byteCount = B_FIXED_BYTE_COUNT_SUBSCRIPTION;
+
+        os.write(B_CURSOR_PARTITION_BEGIN);
+        final byte[] partition = cursor.getPartition().getBytes(StandardCharsets.UTF_8);
```
antban: @rcillo There is `import static com.google.common.base.Charsets.UTF_8` in this file

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
```diff
@@ -44,11 +63,59 @@ public int writeBatch(final OutputStream os, final Cursor cursor, final List<Str
...
+                final byte[] event = events.get(i);
+                os.write(event);
+                byteCount += event.length;
+                if (i < (events.size() - 1)) {
+                    os.write(B_COMMA_DELIM);
+                } else {
+                    os.write(B_CLOSE_BRACKET);
+                }
+            }
+        }
```
antban: Is it possible not to duplicate this part of events write process?

rcillo: In my first proposal I avoided duplication but had to do some type casting. After some feedback, I decided that duplicating would be better than casting. @antban you can take a look at the `extractor` thing https://github.com/zalando/nakadi/pull/673/commits/687dc11690fdc5cb4eddf6d7b0193be6d1d93741

Another approach was proposed by @adyach but it would require creating a lambda with a close on every call to this method. Since it's a hotspot and this change is about optimization, I thought it would be acceptable to have some duplication.

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
rcillo: @antban fixed 7dd2dcf

rcillo:
@adyach I changed an import in this commit. I had accidentally used the wrong UTF_8 and I just replaced it by the original import.

rcillo:
👍

adyach:
👍

PR #675, opened 2017-06-06
aruha-840 use Kafka byte array deserializer

rcillo:
*This is a complimentary PR. It was not implemented in the original one and is out of the scope of the task. But I felt like it would be a valuable contribution.*

In order to reduce the unecessary allocation of String objects, we are
going to adopt Kafka byte array deserializer on the consumption path.

We expect to alleviate GC with this change.


adyach:
👍

adyach:
👍

rcillo:
👍

PR #677, opened 2017-06-07
Do not override flow_id when present

### LINKED ISSUE #391: Nakadi silently overrides metadata.flow_id
When publishing an event, if the header `X-Flow-Id` is not provided, then Nakadi will silently override this field with a randomly generated value.

I would expect Nakadi not to override it, if it was provided, since it's mentioned in the API that "No preexisting value might be changed".

> 1. Once the validation succeeded, the content of the Event is updated according to the
>        enrichment rules in the order the rules are defined in the `EventType`.  No preexisting
>        value might be changed (even if added by an enrichment rule). Violations on this will force
>        the immediate **rejection** of the Event. The invalid overwrite attempt will be included in
>        the item's `BatchItemResponse` object.

It should either reject such event (which doesn't make much sense) or keep it, if present.

rcillo:
Nakadi used to expect the flow_id to be taken from a header but not even
a single user was happy that their flow_id has been
overriten. E.g. users do not expect their event to be changed if they
are sending something already.

It's also a general guideline from Nakadi, not to "enrich" events,
always favoring user defined data. We do not change users data and this
bug broke this commitment.

Fixes #391

### DIFF
--- a/src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
+++ b/src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
@@ -30,7 +30,7 @@ public class MetadataEnrichmentStrategy implements EnrichmentStrategy {
     }
 
     private void setFlowId(final JSONObject metadata) {
-        if (metadata.optString("flow_id") == null || "".equals(metadata.optString("flow_id"))) {
+        if ("".equals(metadata.optString("flow_id"))) {
             metadata.put("flow_id", FlowIdUtils.peek());
         }
     }
--- a/src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
+++ b/src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
@@ -30,7 +30,9 @@ public class MetadataEnrichmentStrategy implements EnrichmentStrategy {
     }
 
     private void setFlowId(final JSONObject metadata) {
-        metadata.put("flow_id", FlowIdUtils.peek());
+        if (metadata.optString("flow_id") == null || "".equals(metadata.optString("flow_id"))) {
+            metadata.put("flow_id", FlowIdUtils.peek());
+        }
     }
 
     private void setEventTypeName(final JSONObject metadata, final EventType eventType) {
--- a/src/test/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategyTest.java
+++ b/src/test/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategyTest.java
@@ -91,6 +91,34 @@ public class MetadataEnrichmentStrategyTest {
     }
 
     @Test
+    public void whenFlowIdIsPresentDoNotOverride() throws Exception {
+        final EventType eventType = buildDefaultEventType();
+        final JSONObject event = buildBusinessEvent();
+        event.getJSONObject("metadata").put("flow_id", "something");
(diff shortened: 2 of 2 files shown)


src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
```diff
@@ -30,7 +30,9 @@ private void setVersion(final JSONObject metadata, final EventType eventType) {
     }
 
     private void setFlowId(final JSONObject metadata) {
-        metadata.put("flow_id", FlowIdUtils.peek());
+        if (metadata.optString("flow_id") == null || "".equals(metadata.optString("flow_id"))) {
```
adyach: optString does not return null, it can return "null", "" or "value"
https://github.com/stleary/JSON-java/blob/ef7a5e40be2e6109b31957bc56b5b80943cc1f92/JSONObject.java#L1345

and I think it makes sense to move `metadata.optString("flow_id")` to var since you use it twice.

src/test/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategyTest.java
```diff
@@ -91,6 +91,34 @@ public void setFlowId() throws Exception {
...
+        event.getJSONObject("metadata").put("flow_id", "");
+        final BatchItem batch = createBatchItem(event);
+
+        final String flowId = randomString();
+        FlowIdUtils.push(flowId);
+        strategy.enrich(batch, eventType);
+
+        assertThat(batch.getEvent().getJSONObject("metadata").getString("flow_id"), equalTo(flowId));
+    }
+
```
adyach: No test case for absent flowId key

src/test/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategyTest.java
```diff
@@ -91,6 +91,34 @@ public void setFlowId() throws Exception {
     }
 
     @Test
+    public void whenFlowIdIsPresentDoNotOverride() throws Exception {
+        final EventType eventType = buildDefaultEventType();
+        final JSONObject event = buildBusinessEvent();
+        event.getJSONObject("metadata").put("flow_id", "something");
+        final BatchItem batch = createBatchItem(event);
+
+        final String flowId = randomString();
```
adyach: it is not cleat why you generate id here

src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
rcillo: You're right. I removed the unnecessary check for null 24154c7

src/test/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategyTest.java
rcillo: It's already tested by existent test, that's why I didn't add a test for that.

src/test/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategyTest.java
rcillo: I generated an id so to prove that it's not used. Makes sense? The idea is to show that even though there is a flow id for this thread, it's no longer used in case the value is already specified in the event.

adyach:
👍

lmontrieux:
@rcillo are you ready to merge this?

rcillo:
👍

PR #679, opened 2017-06-14
ARUHA-671 Remove part of timelines migration code 

antban:
ARUHA-671 Use timelines everywhere, remove migration code


src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
```diff
@@ -122,48 +115,28 @@ public final void deleteSubscription() {
...
-        }
     }
 
     @Override
-    public final void fillEmptySubscription(final Collection<SubscriptionCursorWithoutToken> cursorsIn) {
+    public final void fillEmptySubscription(final Collection<SubscriptionCursorWithoutToken> cursors) {
         try {
-            final Collection<SubscriptionCursorWithoutToken> cursors;
-            // Delete root subscription node
+            // Delete root subscription node, if it was erroneously created
```
rcillo: When should this happen?

rcillo:
👍

src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
antban: When initialization didn't finish (for example initialization started, but then someone killed nakadi with -9)

antban:
:+1:

PR #680, opened 2017-06-14
ARUHA-838: Added validation of empty consumer group

v-stepanov:
https://techjira.zalando.net/browse/ARUHA-838

### DIFF
--- a/src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
+++ b/src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
@@ -22,6 +22,7 @@ public class SubscriptionBase {
     }
 
     @NotNull
+    @Size(min = 1, message = "must contain at least one character")
     private String owningApplication;
 
     @NotNull
--- a/src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
+++ b/src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
@@ -29,6 +29,7 @@ public class SubscriptionBase {
     private Set<String> eventTypes;
 
     @NotNull
+    @Size(min = 1, message = "must contain at least one character")
     private String consumerGroup = "default";
 
     @NotNull
--- a/api/nakadi-event-bus-api.yaml
+++ b/api/nakadi-event-bus-api.yaml
@@ -745,6 +745,7 @@ paths:
           description: |
             Offset to query for unconsumed events. Depends on `partition` parameter. When present adds the property
             `unconsumed_events` to the response partition object.
+          type: string
       responses:
         '200':
           description: OK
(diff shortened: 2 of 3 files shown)


src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
```diff
@@ -29,6 +29,7 @@
     private Set<String> eventTypes;
 
     @NotNull
+    @Size(min = 1, message = "must contain at least one character")
```
adyach: I thought it should be something that make sense, like 3 chars, because you can form the work from them, like `cat` :)

rcillo: What if users want to have consumer groups with a single char, like "1", "2", etc. They would be forced to do something like "001" which could be frustranting. I'm fine with single char.

rcillo:
@v-stepanov do you think it's worth it to update the API by adding something like `minLength` to the definition of `consumer_group` string?

src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
adyach: Such consumer group is meaningless, that's why I want to forbid creating them.

v-stepanov: For me 1 and 2 chars can also make sense. And 3 chars is not a guarantee that it will make sense :)
So I would prefer to keep it as it is.

adyach: Maybe then adjusting max value? http://docs.oracle.com/javaee/7/api/javax/validation/constraints/Size.html#max--

v-stepanov:
@rcillo I added minLength to API now https://github.com/zalando/nakadi/pull/680/commits/6941085f9bbc24d5261a6ee55e52b095a3eed2b5
also fixed some errors in swagger file.

rcillo:
👍

src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
v-stepanov: @adyach why do you want to adjust max value?

adyach: @v-stepanov because the default max value is 2147483647, in my opinion it is too much for consumer group. And it is good practice to validate borders of the input.

v-stepanov:
👍

rcillo:
👍

src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
v-stepanov: Well, in general yes, but we don't really validate the max size of properties anywhere except the event-type name. So if we decide to do that we should add that for all fields and as a separate task. Doing it right now for this only field will not bring much value as for me.

PR #681, opened 2017-06-14
aruha-858 allow calculation of negative distances

rcillo:
We got some early feedback that it would be more useful for users to get
negative distance as a result instead of just an error. This way they
don't need to order cursors on their side before querying Nakadi.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
+++ b/src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
@@ -24,8 +24,6 @@ import java.util.Map;
 import java.util.stream.Collectors;
 
 import static org.zalando.nakadi.exceptions.runtime.InvalidCursorOperation.Reason.CURSORS_WITH_DIFFERENT_PARTITION;
-import static org.zalando.nakadi.exceptions.runtime.InvalidCursorOperation.Reason.INVERTED_OFFSET_ORDER;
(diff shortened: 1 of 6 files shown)


src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
```diff
@@ -44,7 +44,7 @@ public Long calculateDistance(final NakadiCursor initialCursor, final NakadiCurs
         if (!initialCursor.getPartition().equals(finalCursor.getPartition())) {
             throw new InvalidCursorOperation(CURSORS_WITH_DIFFERENT_PARTITION);
         } else if (initialCursor.getTimeline().getOrder() > finalCursor.getTimeline().getOrder()) {
-            throw new InvalidCursorOperation(INVERTED_TIMELINE_ORDER);
+            return - getDistanceDifferentTimelines(finalCursor, initialCursor);
```
adyach: Why is it `-` ?

rcillo: The code I had implemented expects that parameters are correctly sorted. So I just inverted the parameters order and marked the result as negative.

adyach:
👍

rcillo:
@adyach I forgot to clean up these constants c0d5ed1

adyach:
@rcillo do you plan to update api doc?

rcillo:
@adyach you're right. I had scanned it once but I missed this 91995ab

adyach:
👍

rcillo:
👍

PR #683, opened 2017-06-19
Aruha 759 per event type authz

lmontrieux:
ARUHA 759

This PR implements a dummy authorization "plugin", that always approves all authorization requests. It is equivalent to no authorization at all. More useful authorization can be implemented as a separate plugin.


adyach:
It is not clear how your plugin is going to be called when Nakadi runs

lmontrieux:
@adyach fixed that in ac69aa0

src/main/java/org/zalando/nakadi/config/PluginsConfig.java
```diff
@@ -0,0 +1,51 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.io.DefaultResourceLoader;
+import org.zalando.nakadi.plugin.api.ApplicationService;
+import org.zalando.nakadi.plugin.api.ApplicationServiceFactory;
+import org.zalando.nakadi.plugin.api.SystemProperties;
+import org.zalando.nakadi.plugin.api.authz.AuthorizationService;
+import org.zalando.nakadi.plugin.api.authz.AuthorizationServiceFactory;
+
+@Configuration
+public class PluginsConfig {
```
adyach: Probably, makes sense to move system props here as well, because they are exactly for these two beans

lmontrieux: @adyach 9d5c6f8

adyach:
It looks good, however I would choose more generic approach in favour that one, because it requires a lot of places to modify to introduce a new plugin in the system. From other side who knows when we are gonna need it.

adyach:
👍

adyach:
👍

lmontrieux:
👍

PR #684, opened 2017-06-21
aruha-876: kafka client 10.1

adyach:
aruha-876


src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -126,7 +186,8 @@ private void createTopic(final String topic, final int partitionsNum, final int
                 final Properties topicConfig = new Properties();
                 topicConfig.setProperty("retention.ms", Long.toString(retentionMs));
                 topicConfig.setProperty("segment.ms", Long.toString(rotationMs));
-                AdminUtils.createTopic(zkUtils, topic, partitionsNum, replicaFactor, topicConfig);
+                AdminUtils.createTopic(zkUtils, topic, partitionsNum, replicaFactor, topicConfig,
+                        RackAwareMode.Enforced$.MODULE$);
```
rcillo: I'm reading this part of the client code https://github.com/apache/kafka/blob/cae5977ed0d6b63f992973800273769c970b0a0a/core/src/main/scala/kafka/admin/RackAwareMode.scala#L22-L39 and trying to understand the change in behaviour with `RackAwareMode.Enforced`.

Do you understand the difference between `Safe`, `Disabled` and `Enforced`? It's not clear for me.

adyach: There is about how you configure racks in kafka config, more here 
https://cwiki.apache.org/confluence/display/KAFKA/KIP-36+Rack+aware+replica+assignment

adyach: In context of aws it will be az

rcillo: Yes, I know the feature, but do we use rack awareness? Why choosing `Enforced` here?

rcillo:
👍

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: Ok, so what I got is that once we add rack awareness, by specifying the `broker.rack` property in `server.properties` this is going to be immediately used by Nakadi when creating new topics. So sounds like a good thing. It's just not clear if it has any difference right now, since we do not have `broker.rack` configured.

rcillo:
👍

v-stepanov:
👎

v-stepanov:
I had to put a veto, because we need first to check how it influences https://techjira.zalando.net/browse/ARUHA-854 (hash partitioning problem)

v-stepanov:
Ricardo reported that ETs created with a new client version have a different order of partitions that the ones created with old client version. I will check this more deeply but for now I think that the fix for https://techjira.zalando.net/browse/ARUHA-854 should be released before this PR is merged.

adyach:
@rcillo @v-stepanov since the aruha-854 is merged can I get a green light ?

adyach:
👍

adyach:
@rcillo @v-stepanov @lmontrieux @antban could you approve it? just merged from upstream

v-stepanov:
👍

adyach:
👍

PR #685, opened 2017-06-22
Aruha 854 hash partitioning problem

v-stepanov:
https://techjira.zalando.net/browse/ARUHA-854

This is not yet a fix but just a PR with extended logging.
I plan to deploy this version to live and switch traffic to it for a couple of hours.

### DIFF
--- a/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
+++ b/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
@@ -53,9 +53,9 @@ public class HashPartitionStrategy implements PartitionStrategy {
             final int partitionIndex = abs(hashValue) % partitions.size();
             final String partition = partitions.get(partitionIndex);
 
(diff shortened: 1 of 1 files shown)


rcillo:
What are you going to look for in these logs? There will be a huge amount of then, since they are called for each event partitioned. If we are out of luck these are 10k lines of logs per second. That's why I ask, how are you going to make sense out of so much data?

v-stepanov:
@rcillo 
The first search I wanted to do is:
`"[PNUM:8]:" !"2,5,4,7,1,3,6,0"`
To check if there is any other distribution of partitions for 8-partition event-types.

Do you think that there can be some performance problems in nakadi because of such a big amount of logs? I wanted to switch traffic to this stack for a small amount of time actually. 
The other option will be to collect and aggregate that data into some internal structure and log it for example each 5 min.

rcillo:
@v-stepanov ok, got it. I think it's safe to deploy and switch only a part of the traffic for some time.

rcillo:
👍

valgog:
You can also log only the difference. So keeping track of known
distributions, and logging out only unknowns.

PR #687, opened 2017-06-27
Aruha 854 hash partitioning problem

v-stepanov:
https://techjira.zalando.net/browse/ARUHA-854

### DIFF
--- a/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
+++ b/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
@@ -53,9 +53,9 @@ public class HashPartitionStrategy implements PartitionStrategy {
             final int partitionIndex = abs(hashValue) % partitions.size();
             final String partition = partitions.get(partitionIndex);
 
-            LOG.info("[HASH_BUG] hash - values:{} -> hash:{} -> index:{} -> partition:{}; partitions of '{}' [PNUM:{}]: {}",
-                    fieldValues, hashValue, partitionIndex, partition, eventType.getName(), partitions.size(),
-                    partitions.stream().collect(Collectors.joining(",")));
+            LOG.info("[HASH_BUG] hash - values:{} -> hash:{} -> index:{} -> partition:{}; " +
+                    "partitions of '{}' [PNUM:{}]: {}", fieldValues, hashValue, partitionIndex, partition,
+                    eventType.getName(), partitions.size(), partitions.stream().collect(Collectors.joining(",")));
 
             return partition;
 
--- a/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
+++ b/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
@@ -53,8 +53,8 @@ public class HashPartitionStrategy implements PartitionStrategy {
             final int partitionIndex = abs(hashValue) % partitions.size();
             final String partition = partitions.get(partitionIndex);
 
-            LOG.info("[HASH_BUG] hash - values:{} -> hash:{} -> index:{} -> partition:{}; partitions of '{}': {}",
-                    fieldValues, hashValue, partitionIndex, partition, eventType.getName(),
+            LOG.info("[HASH_BUG] hash - values:{} -> hash:{} -> index:{} -> partition:{}; partitions of '{}' [PNUM:{}]: {}",
+                    fieldValues, hashValue, partitionIndex, partition, eventType.getName(), partitions.size(),
                     partitions.stream().collect(Collectors.joining(",")));
 
             return partition;
@@ -68,6 +68,5 @@ public class HashPartitionStrategy implements PartitionStrategy {
             }
         }
     }
-
 
 }
--- a/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
+++ b/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
@@ -49,14 +49,12 @@ public class HashPartitionStrategy implements PartitionStrategy {
 
             final JsonPathAccess traversableJsonEvent = new JsonPathAccess(event);
 
-            final ArrayList<String> fieldValues = new ArrayList<>();
             final int hashValue = partitionKeyFields.stream()
                     // The problem is that JSONObject doesn't override hashCode(). Therefore convert it to
                     // a string first and then use hashCode()
                     .map(pkf -> EventCategory.DATA.equals(eventType.getCategory()) ? DATA_PATH_PREFIX + pkf : pkf)
                     .map(Try.wrap(okf -> {
                         final String fieldValue = traversableJsonEvent.get(okf).toString();
-                        fieldValues.add(fieldValue);
                         return stringHash.hashCode(fieldValue);
                     }))
                     .map(Try::getOrThrow)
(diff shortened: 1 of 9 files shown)


src/main/java/org/zalando/nakadi/partitioning/StringHash.java
```diff
@@ -0,0 +1,18 @@
+package org.zalando.nakadi.partitioning;
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class StringHash {
+
+    public int hashCode(final String string) {
```
adyach: Just to mention here, that this function will return the same hash codes for "Aa" and "BB".
Probably, we are not allowed to change it.

src/main/resources/application.yml
```diff
@@ -89,6 +89,16 @@ nakadi:
     checkRunMs: 600000 # 10 min
     timelineCleanup.runPeriodMs: 3600000 # 1 hour
     consumerNodesCleanup.runPeriodMs: 21600000 # 6 hours
+  hashPartitioning.overrideOrder:
```
adyach: @v-stepanov is it possible to move it to the private repo?

src/main/java/org/zalando/nakadi/partitioning/StringHash.java
rcillo: @adyach you're right, I think we cannot change it since that would change how partitioning is working right now. This might not be the most sophisticated hashing out there, but there are some probability checks that show this hashing spreads quite well https://github.com/zalando/nakadi/blob/4b771e748af3cce5b8ff7519d28e1461d51d64a1/src/test/java/org/zalando/nakadi/partitioning/HashPartitionStrategyTest.java#L63-L76

Another thing I realised is that it overflows for strings bigger than 32 chars in length. But it seems that losing the bytes that don't fit is not an issue.

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -30,17 +49,33 @@ public String calculatePartition(final EventType eventType, final JSONObject eve
...
                     .mapToInt(hc -> hc)
                     .sum();
 
-            final int partitionIndex = abs(hashValue) % partitions.size();
-            return partitions.get(partitionIndex);
+
+            int partitionIndex = abs(hashValue) % partitions.size();
+            partitionIndex = hashPartitioningCrutch.adjustPartitionIndex(partitionIndex, partitions.size());
+
+            final List<String> sortedPartitions = partitions.stream().sorted().collect(Collectors.toList());
```
rcillo: Do you think it would be better to find a way of doing this without sorting a collection on every call?

src/main/java/org/zalando/nakadi/partitioning/StringHash.java
v-stepanov: @adyach @rcillo I just extracted String.hashCode to this function, I believe we can't change it without breaking partitioning.

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategyCrutch.java
```diff
@@ -0,0 +1,60 @@
...
+    @Autowired
+    @SuppressWarnings("unchecked")
+    public HashPartitionStrategyCrutch(final Environment environment,
+                                       @Value("${" + PROPERTY_PREFIX + ".max}") final int maxPartitionNum) {
+
+        final ImmutableMap.Builder<Integer, List<Integer>> mapBuilder = ImmutableMap.builder();
+        for (int pCount = 1; pCount <= maxPartitionNum; pCount++) {
+
+            final String propertyName = PROPERTY_PREFIX + ".p" + pCount;
+            final List<String> predefinedOrder = (List<String>) environment.getProperty(propertyName, List.class);
```
adyach: I see that you do map to integer later. why don't you want to cast it to (List<Integer>) right here?

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
v-stepanov: I think there is no a simple solution for that. I don't really want to make one more creepy cache here. 
I think sorting a collection of <100 string elements will not take any significant amount of time.
Do you have any ideas regarding that?

adyach: One of the improvements could be to sort in place using Collections.sort

rcillo: I don't know. Why not just return "partitionIndex" as a string directly? Kafka partitions are always "0, 1, 2, 3...". I think it's a reasonable compromise.

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategyCrutch.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private static final String PROPERTY_PREFIX = "nakadi.hashPartitioning.overrideOrder";
+
+    // due to initial problem with order of partitions we need to keep the original order now
+    private final Map<Integer, List<Integer>> partitionsOrder;
+
+    @Autowired
+    @SuppressWarnings("unchecked")
+    public HashPartitionStrategyCrutch(final Environment environment,
+                                       @Value("${" + PROPERTY_PREFIX + ".max}") final int maxPartitionNum) {
```
rcillo: If this property is not defined, what happens here?

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategyCrutch.java
v-stepanov: I don't exactly get what you mean. I don't cast it to list of integers here because it's not a list of integers :)

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategyCrutch.java
v-stepanov: I think it will be 0, but I need to check

src/main/resources/application.yml
v-stepanov: Yes, makes sense. I will move it to our private configs.

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategyCrutch.java
adyach: According to the application.yml it is the list of integers:
https://github.com/zalando/nakadi/pull/687/files#diff-256c57f5b357be82840a9b2207f0dcd5R94

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
v-stepanov: @adyach `Collections.sort` sorts the initial collection, it is unmodifiable in our case so it will fail.

@rcillo I decided to use indexes instead of actual partitions not to be bind to Kafka specific stuff.

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategyCrutch.java
v-stepanov: @adyach yes, but unfortunately it returns list of strings.

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategyCrutch.java
v-stepanov: @rcillo good catch, it was actually failing. I added default value now https://github.com/zalando/nakadi/pull/687/commits/e9193762008a31488fac413d651580ea7ebd6a3a

src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
rcillo: @v-stepanov ok. Let's go with this approach.

src/main/resources/application.yml
v-stepanov: I removed it from here, will place it to private configuration
https://github.com/zalando/nakadi/pull/687/commits/e9193762008a31488fac413d651580ea7ebd6a3a

rcillo:
👍

v-stepanov:
👍

rcillo:
👍

zrobgar:
👍

v-stepanov:
👍

PR #691, opened 2017-07-05
aruha-767: enforce admin authorization for event types

adyach:
aruha-767: Only allow users in the admin list to edit event types.


src/main/java/org/zalando/nakadi/exceptions/runtime/ServiceTemporaryUnavailableException.java
```diff
@@ -4,4 +4,8 @@
     public ServiceTemporaryUnavailableException(final Exception cause) {
```
lmontrieux: This should be called `ServiceTemporarilyUnavailableException` (-ily)

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -211,65 +211,41 @@ public void delete(final String eventTypeName, final Client client)
...
-            return Result.problem(e.asProblem());
-        } catch (final NoSuchPartitionStrategyException e) {
-            LOG.debug("Partition strategy does not exist", e);
-            return Result.problem(e.asProblem());
+            throw new ServiceTemporaryUnavailableException(
+                    "Event type is currently in maintenance, please repeat request", e);
         } catch (final NakadiException e) {
             LOG.error("Unable to update event type", e);
-            return Result.problem(e.asProblem());
+            throw new NakadiRuntimeException(e);
```
lmontrieux: Aren't we losing some details about the exact issue here, compared to before?

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -94,4 +100,23 @@ private void checkAuthAttributesAreValid(final Map<String, List<AuthorizationAtt
...
         }
     }
+
+    public void authorizeEventTypeUpdate(final EventType eventType)
+            throws ForbiddenAccessException, ServiceTemporaryUnavailableException {
+        if (eventType.getAuthorization() == null) {
+            return;
+        }
+
+        final Resource resource = new EventTypeResource("/event-type", "event-type",
```
lmontrieux: I think it should be the event type name rather than `/event-type` here

src/test/java/org/zalando/nakadi/controller/EventTypeAuthorizationTest.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.content;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+public class EventTypeAuthorizationTest extends EventTypeControllerTestCase {
+
+    public EventTypeAuthorizationTest() throws IOException {
+    }
+
+    @Test
+    public void whenPUTAuthorized200() throws Exception {
```
lmontrieux: It seems that this test is dependent on the implementation of the default authorization plugin (which always returns true). Should we perhaps mock the call to `isAuthorized`? Or am I over-complicating this?

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -92,95 +62,18 @@
...
-        doReturn(SecuritySettings.AuthMode.OFF).when(settings).getAuthMode();
-        doReturn("nakadi").when(settings).getAdminClientId();
-        doReturn(true).when(featureToggleService).isFeatureEnabled(CHECK_PARTITIONS_KEYS);
-
-        mockMvc = standaloneSetup(controller)
-                .setMessageConverters(new StringHttpMessageConverter(), jackson2HttpMessageConverter)
-                .setCustomArgumentResolvers(new ClientResolver(settings, featureToggleService))
-                .setControllerAdvice(new ExceptionHandling())
-                .build();
+public class EventTypeControllerTest extends EventTypeControllerTestCase {
```
lmontrieux: Why did you decide to separate EventTypeControllerTest and EventTypeControllerTestCase?

src/main/java/org/zalando/nakadi/service/EventTypeService.java
adyach: Which details? it will be unpacked in ExceptionHandling creating the same result

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
adyach: Because EventTypeControllerTest is huge and I want to have another test which uses the same fields and methods, EventTypeAuthorizationTest

lmontrieux: I agree that it is huge, but how should we decide, now and later, where a test for the event type controller should go? For me it's not quite clear how to decide, based on the class names.

src/test/java/org/zalando/nakadi/controller/EventTypeAuthorizationTest.java
adyach: the whole AuthorizationValidator is mocked here :)

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
adyach: I just moved out authz, so, tests are for authz go to the EventTypeAuthorizationTest, we can move out other parts later, if needed.

src/test/java/org/zalando/nakadi/controller/EventTypeAuthorizationTest.java
lmontrieux: 👍

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
lmontrieux: 👍

src/main/java/org/zalando/nakadi/exceptions/runtime/ServiceTemporaryUnavailableException.java
adyach: 0ef2685

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
adyach: 0ef2685

src/main/java/org/zalando/nakadi/service/EventTypeService.java
lmontrieux: Nevermind, I looked more carefully and this is fine.

lmontrieux:
👍

adyach:
👍

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -281,6 +257,29 @@ public void delete(final String eventTypeName, final Client client)
...
+            throw new InconsistentStateException("Empty value for retention time in existing EventType");
+        }
+        boolean retentionTimeUpdated = false;
+        try {
+            if (newRetentionTime != null && !newRetentionTime.equals(oldRetentionTime)) {
+                updateTopicRetentionTime(original.getName(), newRetentionTime);
+            } else {
+                eventType.setOptions(original.getOptions());
+            }
+            updateEventTypeInDB(eventType, newRetentionTime, oldRetentionTime);
```
v-stepanov: now `updateEventTypeInDB` is inside `updateRetentionTime` which is not cool. 
I would expect that `updateRetentionTime` does only the update of retention time, but it actually does the full update of event-type in DB.

adyach: The point is to segregate this logic from the update method, because it is quite big and has a lot of different functionality. I would just rename it as well as updateEventTypeInDB, because it does update of timeline in addition to update of et.

PR #694, opened 2017-07-05
aruha-856 authorize stream read

rcillo:
Every endpoint related to reading from a partition must be
authorized. Not only getting events, but getting information about how
many events there are or any sort of glimpse over a partition.


src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
```diff
@@ -0,0 +1,79 @@
...
+import org.zalando.nakadi.plugin.api.authz.AuthorizationService;
+import org.zalando.nakadi.plugin.api.authz.Resource;
+import org.zalando.nakadi.plugin.api.authz.Subject;
+import org.zalando.nakadi.repository.EventTypeRepository;
+import org.zalando.nakadi.security.Client;
+
+import java.util.List;
+import java.util.Optional;
+
+public class AuthorizationUtils {
```
adyach: We have AuthorizationValidator in master. It is implemented as bean, I think it is better to use it.

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -154,11 +156,15 @@ public StreamingResponseBody streamEvents(
...
                         nakadiSettings.getDefaultCommitTimeoutSeconds(), client.getClientId());
                 streamer = subscriptionStreamerFactory.build(subscriptionId, streamParameters, output,
-                        connectionReady, blacklistService);
+                        connectionReady, blacklistService, client);
                 streamer.stream();
             } catch (final InterruptedException ex) {
                 LOG.warn("Interrupted while streaming with " + streamer, ex);
                 Thread.currentThread().interrupt();
+            } catch (final AccessDeniedException e) {
+                writeProblemResponse(response, output.getOutputStream(), Problem.valueOf(Response.Status.FORBIDDEN,
```
adyach: maybe it is better to put this logic in SubscriptionOutputImpl.onException it will be simpler to refactor exceptions later

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
```diff
@@ -0,0 +1,79 @@
...
+                                                 final AuthorizationService authorizationService,
+                                                 final Client client, final SubscriptionBase subscriptionBase) {
+        subscriptionBase.getEventTypes().stream().forEach(
+                (eventTypeName) -> {
+                    try {
+                        eventTypeRepository.findByNameO(eventTypeName).ifPresent(eventType -> {
+                            authorizeStreamRead(authorizationService, client, eventType);
+                        });
+                    } catch (final InternalNakadiException e) {
+                        e.printStackTrace();
```
adyach: If we failed to check permissions, we should fail streaming, otherwise it is a security threat.

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
```diff
@@ -0,0 +1,79 @@
...
+            }
+
+            @Override
+            public String getType() {
+                return "event-type";
+            }
+
+            @Override
+            public Optional<List<AuthorizationAttribute>> getAttributesForOperation(
+                    final AuthorizationService.Operation operation) {
```
adyach: Interface is supposed to work with operation, but you do not use it :)

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionStreamerFactory.java
```diff
@@ -56,17 +65,23 @@ public SubscriptionStreamerFactory(
...
             final SubscriptionOutput output,
             final AtomicBoolean connectionReady,
-            final BlacklistService blacklistService) throws NoSuchSubscriptionException, ServiceUnavailableException,
+            final BlacklistService blacklistService,
+            final Client client) throws NoSuchSubscriptionException, ServiceUnavailableException,
             InternalNakadiException, NoSuchEventTypeException {
 
         final Subscription subscription = subscriptionDbRepository.getSubscription(subscriptionId);
+
+        authorizeSubscriptionRead(eventTypeRepository, authorizationService, client, subscription);
```
adyach: The permission check is inside build method, which is to build the object, not to check the permission. It would be nice to have it in service layer, but we do not have it exactly here. We can move it to controller level, but there we need subscription. Maybe we can move retrieval of subscription to controller level or create service level for that?

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
```diff
@@ -54,21 +56,29 @@
...
     @RequestMapping(path = "/event-types/{eventTypeName}/cursor-distances", method = RequestMethod.POST)
     public ResponseEntity<?> getDistance(@PathVariable("eventTypeName") final String eventTypeName,
                                          @Valid @RequestBody final ValidListWrapper<CursorDistance> queries,
-                                         final Client client) {
+                                         final Client client) throws InternalNakadiException, NoSuchEventTypeException {
+        // TODO: remove once new authorization is in place
         checkReadScopes(eventTypeName, client);
 
+        final EventType eventType = eventTypeRepository.findByName(eventTypeName);
+        authorizeStreamRead(authorizationService, client, eventType);
```
adyach: shouldn't be these calls on the service layer?

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
rcillo: god damn code generation. Forgot to rethrow c58a792 thanks for the catch

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionStreamerFactory.java
rcillo: I'll extract subscription retrieval to the controller.

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
```diff
@@ -55,7 +56,7 @@ public static void authorizeSubscriptionRead(final EventTypeRepository eventType
                             authorizeStreamRead(authorizationService, client, eventType);
                         });
                     } catch (final InternalNakadiException e) {
-                        e.printStackTrace();
+                        throw new MyNakadiRuntimeException1("Failed authorizing");
```
adyach: As we agreed not to use MyNakadiRuntimeException1 for throwing, I suggest ServiceTemporarilyUnavailableException

src/main/java/org/zalando/nakadi/controller/CursorOperationsController.java
rcillo: I don't know. I gave it some thought and it seems wrong to have to edit every single method signature to include a client. It feels like coupling the execution of an operation with checking if that operation is authorized. It might be that I already have a fixed idea in mind because when I implemented this kind of feature before, it would usually be implemented using some aspects or something outside of the actual operation.

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
rcillo: I considered doing this but I would have to check the type of exception there. Do you think it would be fine?

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionStreamerFactory.java
rcillo: Done a82a804

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
rcillo: yeah. I'll maybe just use the resource you created in your PR. I'll wait for it to be merged and then fix it here, what do you think?

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
rcillo: 👍  d0c2979

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
rcillo: I thought this class is responsible for validating some parameters and not actually implementing code that performs authorization.

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
adyach: Adding TODO about it would be good

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
adyach: it is already in master, but I propose to use resource which was created by @antban, it is `almost` in master :)

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
adyach: One could think of it as of permissions validation :) anyway, it is better to inject object rather than using helpers, in the end we use spring

src/main/java/org/zalando/nakadi/controller/PartitionsController.java
```diff
@@ -102,9 +117,13 @@ public PartitionsController(final TimelineService timelineService,
     public ResponseEntity<?> getPartition(@PathVariable("name") final String eventTypeName,
                                           @PathVariable("partition") final String partition,
                                           @Nullable @RequestParam(value = "consumed_offset", required = false)
-                                              final String consumedOffset, final NativeWebRequest request) {
+                                              final String consumedOffset, final NativeWebRequest request,
```
v-stepanov: broken formatting

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionStreamerFactory.java
```diff
@@ -59,16 +56,15 @@ public SubscriptionStreamerFactory(
...
             final AtomicBoolean connectionReady,
-            final BlacklistService blacklistService) throws NoSuchSubscriptionException, ServiceUnavailableException,
+            final BlacklistService blacklistService,
+            final Client client) throws NoSuchSubscriptionException, ServiceUnavailableException,
             InternalNakadiException, NoSuchEventTypeException {
-
-        final Subscription subscription = subscriptionDbRepository.getSubscription(subscriptionId);
         final Session session = Session.generate(1);
-        final String loggingPath = "subscription." + subscriptionId + "." + session.getId();
+        final String loggingPath = "subscription." + subscription.getId() + "." + session.getId();
```
v-stepanov: what is the purpose of this change?

rcillo: @v-stepanov to move the authorization outside of the builder, which is a very weird layer to make authorization.

src/main/java/org/zalando/nakadi/controller/PartitionsController.java
rcillo: Not actually, this is a continuation of the line above, right?

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
rcillo: Done 5c947b1

src/main/java/org/zalando/nakadi/util/AuthorizationUtils.java
rcillo: Done

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
rcillo: Done dfa5fe5

rcillo:
@adyach I have addressed the comments

adyach:
👍

rcillo:
@adyach I accidentally committed a file which I just removed 95b1491

adyach:
👍

rcillo:
👍

PR #697, opened 2017-07-06
ARUHA-923 Stop streaming when event type is changed

antban:
ARUHA-923

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/EventStreamController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventStreamController.java
@@ -7,6 +7,24 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.OutputStream;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.core.Response;
+import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
+import static javax.ws.rs.core.Response.Status.FORBIDDEN;
+import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
+import static javax.ws.rs.core.Response.Status.PRECONDITION_FAILED;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -34,10 +52,12 @@ import org.zalando.nakadi.exceptions.ServiceUnavailableException;
 import org.zalando.nakadi.exceptions.UnparseableCursorException;
 import org.zalando.nakadi.exceptions.runtime.AccessDeniedException;
 import org.zalando.nakadi.metrics.MetricUtils;
+import static org.zalando.nakadi.metrics.MetricUtils.metricNameFor;
 import org.zalando.nakadi.repository.EventConsumer;
 import org.zalando.nakadi.repository.EventTypeRepository;
 import org.zalando.nakadi.repository.TopicRepository;
 import org.zalando.nakadi.security.Client;
+import org.zalando.nakadi.service.AuthorizationChangeListener;
 import org.zalando.nakadi.service.AuthorizationValidator;
 import org.zalando.nakadi.service.BlacklistService;
 import org.zalando.nakadi.service.ClosedConnectionsCrutch;
@@ -50,29 +70,9 @@ import org.zalando.nakadi.service.EventStreamFactory;
 import org.zalando.nakadi.service.timeline.TimelineService;
 import org.zalando.nakadi.util.AuthorizationUtils;
 import org.zalando.nakadi.util.FeatureToggleService;
+import static org.zalando.nakadi.util.FeatureToggleService.Feature.LIMIT_CONSUMERS_NUMBER;
 import org.zalando.nakadi.view.Cursor;
 import org.zalando.problem.Problem;
-
-import javax.annotation.Nullable;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-import javax.ws.rs.core.Response;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.stream.Collectors;
-
-import static javax.ws.rs.core.Response.Status.BAD_REQUEST;
-import static javax.ws.rs.core.Response.Status.FORBIDDEN;
-import static javax.ws.rs.core.Response.Status.INTERNAL_SERVER_ERROR;
-import static javax.ws.rs.core.Response.Status.NOT_FOUND;
-import static javax.ws.rs.core.Response.Status.PRECONDITION_FAILED;
-import static org.zalando.nakadi.metrics.MetricUtils.metricNameFor;
-import static org.zalando.nakadi.util.FeatureToggleService.Feature.LIMIT_CONSUMERS_NUMBER;
 
 @RestController
 public class EventStreamController {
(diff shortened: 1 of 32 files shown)


rcillo:
@antban when a stream terminates because of an internal error, how does the listener gets unregistered?

antban:
@rcillo There is try with resource everywhere, so it will be called automagicaly

v-stepanov:
@antban I don't see any new tests for new functionality, only fixes of old tests.

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -237,7 +242,17 @@ public StreamingResponseBody streamEvents(
...
 
-                eventStream.streamEvents(connectionReady);
+                eventStream.streamEvents(connectionReady, () -> {
+                    if (needCheckAuthorization.getAndSet(false)) {
+                        // in a case of exception all exceptions are treated in the same manner - all are internal
+                        // errors
+                        try {
+                            authorizationValidator.authorizeStreamRead(eventTypeRepository.findByName(eventTypeName));
+                        } catch (final InternalNakadiException | NoSuchEventTypeException ex) {
+                            throw new RuntimeException(ex);
```
adyach: ServiceTemporarilyUnavailableException

src/main/java/org/zalando/nakadi/service/AuthorizationChangeListener.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Consumer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.zalando.nakadi.repository.db.EventTypeCache;
+
+@Component
+public class AuthorizationChangeListener {
```
adyach: please, rename to EventTypeChangeListener

antban:
@v-stepanov Regarding tests - I didn't want to create tests on functionality that is not approved or may be changed a lot.
I will add tests for AuthorizationChangeListener

v-stepanov:
@antban I think it's the other way round: if you deliver something without tests - it has not so good chances to be approved :)

antban:
@v-stepanov Added tests  https://github.com/zalando/nakadi/pull/697/commits/8ffbf249e663a849e2d2a066e2006b84a9a74ef3

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -237,7 +243,17 @@ public StreamingResponseBody streamEvents(
...
+                    if (needCheckAuthorization.getAndSet(false)) {
+                        // in a case of exception all exceptions are treated in the same manner - all are internal
+                        // errors
+                        try {
+                            authorizationValidator.authorizeStreamRead(eventTypeRepository.findByName(eventTypeName));
+                        } catch (final InternalNakadiException | NoSuchEventTypeException ex) {
+                            throw new ServiceTemporarilyUnavailableException(ex);
+                        }
+                    }
+                });
```
v-stepanov: I don't really like that authorization now is spread through the controller. Is it somehow possible to compact it and keep in one place?

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
```diff
@@ -235,6 +245,10 @@ private String getZNodePath(final String eventTypeName) {
         return ZKPaths.makePath(ZKNODE_PATH, eventTypeName);
     }
 
+    public void addInvalidationListener(final Consumer<String> onEventTypeInvalidated) {
```
adyach: If you call method, while iterating over it in line 223, you got exception. Also you access arraylist from different threads ...

src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -175,11 +180,26 @@ public StreamingResponseBody streamEvents(
...
+
+                try (Closeable ignore = eventTypeChangeListener.registerListener(
+                        eventTypeModificationListener, subscription.getEventTypes())) {
+                    authorizationValidator.authorizeSubscriptionRead(eventTypeRepository, subscription);
 
-                streamer = subscriptionStreamerFactory.build(subscription, streamParameters, output,
-                        connectionReady, blacklistService, client);
-                streamer.stream();
+                    streamer.get().stream();
+                }
```
v-stepanov: the same comment as for Low Level controller. Is it somehow possible to separate authorization, because now method `streamEvents` really exploded and it's difficult to read its code.

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -47,7 +47,7 @@ public EventStream(final EventConsumer eventConsumer,
         this.writer = writer;
     }
 
-    public void streamEvents(final AtomicBoolean connectionReady) {
+    public void streamEvents(final AtomicBoolean connectionReady, final Runnable checkAuthorization) {
```
v-stepanov: I think to make it more consistent it would make sense to pass here a second atomic boolean instead of runnable. And you eve have that atomic boolean already in controller code.

antban: @v-stepanov Initially I did it like that. but because of the difference of how it calculated (and called on zk thread, actually), I have to perform authorization check here. It will require adding several additional parameters here (eventTypeCache + authorizationProvider), which, in turn will require modification of tons of unit tests, that are all passing all the parameters to construct object. It is too much, that is why I decided to go with runnable (I do not like it as well)

src/main/java/org/zalando/nakadi/repository/db/EventTypeCache.java
antban: True

src/test/java/org/zalando/nakadi/service/EventTypeChangeListenerTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+public class EventTypeChangeListenerTest {
+
+    private EventTypeChangeListener listener;
+    private Consumer<String> notificationTrigger;
+
+    @Before
+    public void setUp() {
+        final EventTypeCache cache = mock(EventTypeCache.class);
+        final ArgumentCaptor<Consumer<String>> changeCall = new ArgumentCaptor<>();
```
v-stepanov: this constructor is deprecated

antban: @v-stepanov I know, and have read reasoning for that. But failed to find correct way of creating correct captor (templated class for lambda function)

src/test/java/org/zalando/nakadi/service/EventTypeChangeListenerTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+        listener = new EventTypeChangeListener(cache);
+        verify(cache, times(1)).addInvalidationListener(changeCall.capture());
+
+        notificationTrigger = changeCall.getValue();
+    }
+
+    @Test
+    public void testPossibleToCallUnknownET() {
+        notificationTrigger.accept("any");
```
v-stepanov: what exactly do you test here?

antban: authorizeStreamRead - I test that it won't explode

v-stepanov: what? :)

src/test/java/org/zalando/nakadi/service/EventTypeChangeListenerTest.java
v-stepanov: I think it should be like this
```java
final Class<Consumer<String>> stringConsumerClass = (Class) Consumer.class;
final ArgumentCaptor<Consumer<String>> changeCall = ArgumentCaptor.forClass(stringConsumerClass);
```

src/test/java/org/zalando/nakadi/service/EventTypeChangeListenerTest.java
antban: In other places I test what will happen if I have listener.
In this place I test what will happen if there is no listener

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -199,7 +208,7 @@ public StreamingResponseBody streamEvents(
                 // TODO: deprecate and remove previous authorization strategy
                 client.checkScopes(eventType.getReadScopes());
 
-                authorizationValidator.authorizeStreamRead(eventType);
+                authorizeStreamRead(eventTypeName);
```
v-stepanov: Will the correct exception be thrown here? So that user receives 403.

v-stepanov: Ok, now I see that it will. Everything is good.

v-stepanov:
👍

antban:
:+1:

src/main/java/org/zalando/nakadi/service/subscription/state/StartingState.java
```diff
@@ -35,6 +35,7 @@ public void onEnter() {
                             return Response.Status.FORBIDDEN;
                         }
                     }));
+            return;
```
adyach: 👍

adyach:
👍

lmontrieux:
👍

antban:
:+1:

lmontrieux:
👍

PR #699, opened 2017-07-07
aruha-932: check admin permission when delete et

adyach:
aruha-932


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -168,6 +169,9 @@ public void delete(final String eventTypeName, final Client client)
             if (!client.idMatches(eventType.getOwningApplication())) {
```
antban: Why you left this thing here?

antban: Oh, yeah, this check is constantly turned off, and we will remove it later.

antban:
:+1:

adyach:
👍

PR #700, opened 2017-07-10
ARUHA-933 Document per-resource authorization feature

lmontrieux:
ARUHA-933

This PR adds documentation about the new per-resource authorization feature.

### DIFF
--- a/docs/ResourceAuthorization.md
+++ b/docs/ResourceAuthorization.md
@@ -115,7 +115,7 @@ implemented in a separate plugin); and a Policy Enforcement Point (PEP), which e
 
 ### Default plugin
 
-The default plugin is part of the Nakadi codebase. It authorized all requests, and validates all attributes. If you 
+The default plugin is part of the Nakadi codebase. It authorizes all requests, and validates all attributes. If you 
 want to deploy Nakadi and use the resource authorization feature, you should provide your own plugin.
 
 ### Building your own
--- a/docs/ResourceAuthorization.md
+++ b/docs/ResourceAuthorization.md
@@ -75,7 +75,7 @@ Updating an event type is similar to creating one. Here is a sample request, tha
 to the same application:
 
 ```bash
-curl -v -XPOST -H "Content-Type: application/json" http://localhost:8080/event-types -d '{
+curl -v -XPUT -H "Content-Type: application/json" http://localhost:8080/event-types/order_received -d '{
   "name": "order_received",
   "owning_application": "acme-order-service",
   "category": "business",
--- a/docs/ResourceAuthorization.md
+++ b/docs/ResourceAuthorization.md
@@ -1,0 +1,172 @@
+# Resource Authorization
+
+Nakadi allows users to restrict access to resources they own - currently, event types are the only resources supported, 
+but we plan to extend this feature to subscriptions in the near future.
+
+The authorization model is simple: policies can be attached to resources. A policy `P` defines which `subjects` can 
+perform which `operations` on a resource `R`. To do so, the policy contains, for each operation, a list of attributes that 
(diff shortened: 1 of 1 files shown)


docs/ResourceAuthorization.md
```diff
@@ -0,0 +1,172 @@
...
+to match at least one of these attributes (not necessarily all of them).
+
+There are three kinds of operation: `admin`, to update the resource and delete it; `read`, to read events from the 
+resource; and `write`, to write events to the resource.
+
+An authorization request is represented by the tuple
+
+`R(subject, operation, resource)`
+
+The request will be approved iff the `resource` policy has at least one attribute for `operation` that matches the
```
phibrex: change "iff" to "if"

lmontrieux: No :)

`iff` means `if and only if`: https://en.wikipedia.org/wiki/If_and_only_if

docs/ResourceAuthorization.md
```diff
@@ -0,0 +1,172 @@
...
+## Implementation
+
+The implementation of per-resource authorization comprises of three components: a Policy Administration Point (PAP), 
+where users can define and update the authorization policy for an event type; a Policy Decision Point (PDP), where 
+a decision is made whether or not to authorize a subject to perform an operation in a resource (this component is 
+implemented in a separate plugin); and a Policy Enforcement Point (PEP), which enforces the PDP's decisions.
+
+### Default plugin
+
+The default plugin is part of the Nakadi codebase. It authorized all requests, and validates all attributes. If you
```
rcillo: s/authorized/authorizes/

rcillo:
@lmontrieux in order to minimize the amount of duplicated documentation and make maintenance easier, I think we should clarify what kind of documentation should go into "/docs", what goes in the README, what goes in the manual and what goes in the API. What do you think of:

- docs: only implementation details for complex features. It's mostly for developers and covers design decisions and implementation.
- README: something for users in a rush. Maybe change the examples there to include authorization, so that new users would be aware that this feature exists.
- API: well, API. Should not be too length, so in case a field requires a lot of explanation, like "compatibility_mode" one should write about it the manual.
- manual: more in detail user documentation.

I feel like this doc mixes a little bit of some of these.

Maybe just leave the plugin details here, since it might be useful for future Nakadi developers, explaining why it's done as a plugin.

Consider updating README with curl examples. Consider moving the rest to the manual.

What do you think?

docs/ResourceAuthorization.md
lmontrieux: @rcillo  8c1a43f

lmontrieux:
@rcillo very good point.

I did put this in here, at @zrobgar 's request. As he is currently working on re-organising our docs, I understand that he plans on retiring the nakadi manual, and moving it between docs/ and the readme ( @zrobgar , correct me if I'm wrong). What I'm less clear about, however, is what the purpose of each 'thing' will be after that reorganisation.

rcillo:
@zrobgar @lmontrieux sorry, but this should not be a single person decision and should not be done like this. I'll reject this PR by now.

rcillo:
👎

rcillo:
@lmontrieux @zrobgar as we discussed offline, ideally this PR would be merged after the migration of nakadi manual. So that it would be placed in its final destination, fitting a bigger structure of user facing documentation.

Right now, it does not fit the "/docs" folder since it's used for internal documentation of features.

So I see the following alternatives:

- open a PR in nakadi manual, and once its migration is done, it will be moved naturally to the right place.
- first migrate nakadi manual, then modify this PR to fit this new structure.

lmontrieux:
Closing because this is addressed is #714

PR #702, opened 2017-07-11
Aruha 929 authorization refactoring

rcillo:
When first implementing event type authorization feature, we parallelised the work to an extent that caused duplicated code to be written.

This PR unifies the code for resource creation and exception handling.

- MyEventTypeResource2 has been replaced by EventTypeResource
- ForbiddenAccessException and ResourceAccessNotAuthorizedException have been replaced by AccessDeniedException
- all error messages regarding authorization are now done in a single place

If you think there is something else that should be unified, let me know.


src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -154,9 +155,6 @@ public Result listSubscriptions(@Nullable final String owningApplication, @Nulla
     public Result<Void> deleteSubscription(final String subscriptionId, final Client client) {
```
antban: Why have you left client here?

v-stepanov:
@rcillo I think that also the following thing should be done:
I think `AuthorizationUtils.errorMessage` should be removed as it duplicates `AccessDeniedException.explain`. As it is the only method in `AuthorizationUtils` I think the class itself should also be removed.

src/main/java/org/zalando/nakadi/service/MyEventTypeResource2.java
```diff
@@ -1,48 +0,0 @@
-package org.zalando.nakadi.service;
-
-import java.util.List;
-import java.util.Optional;
-import org.zalando.nakadi.domain.EventTypeAuthorization;
-import org.zalando.nakadi.plugin.api.authz.AuthorizationAttribute;
-import org.zalando.nakadi.plugin.api.authz.AuthorizationService;
-import org.zalando.nakadi.plugin.api.authz.Resource;
-
-class MyEventTypeResource2 implements Resource {
```
v-stepanov: I liked this name 😢

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -152,7 +152,7 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
 
     public void delete(final String eventTypeName, final Client client)
```
v-stepanov: `client` is not needed here anymore.

v-stepanov:
@rcillo one more thing. `AccessDeniedException` has a parameter `subject` in constructor but in never used inside the class and it's always passed as `null`. Do we actually need that?

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -110,51 +109,53 @@ private void checkAuthAttributesAreValid(final Map<String, List<AuthorizationAtt
...
-        final Subject subject = null;
-        if (!authorizationService.isAuthorized(subject, AuthorizationService.Operation.READ, resource)) {
-            throw new AccessDeniedException(subject, AuthorizationService.Operation.READ, resource);
+        if (eventType.getAuthorization() == null) {
+            return;
+        }
+
+        final Resource resource = new EventTypeResource(eventType.getName(), eventType.getAuthorization());
+        try {
+            if (!authorizationService.isAuthorized(null, AuthorizationService.Operation.READ, resource)) {
```
v-stepanov: What is this `subject`? Why it's always null?

rcillo: @v-stepanov this is defined in the API. The fact that we use application context information for performing authorization is an implementation detail. I'm not sure that changing this API is desirable. @lmontrieux what do you think?

rcillo: Maybe, we should even extract the context here and pass it as the subject and then in the plugin implementation use the parameter instead of reading from a singleton.

lmontrieux: @rcillo @v-stepanov the subject is definitely redundant with getting the principal from the security context, as we do in the plugin. However, of the 2 solutions, I think the security context may be the most generic one, as it may contain more data then what we have in the Subject interface (name and token, iirc). I can see, for example, delegation scenarios where several prinicpals are used for a single subject, which we can't represent with the Subject interface as it is.

Changing the API is not that bad, actually. It's only a plugin API, so we don't have to go through the API guild for approval.

lmontrieux: However, it means that we need to change code in nakadi, nakadi-plugin-api, and our authz plugin, simultaneously.

rcillo: Well, that's not an issue (changing it in several places). But in the documentation we say that authorization is composed of a tuple with 3 items. That will no longer be reflected in the API, since we are retrieving data from a singleton in the context. But ok. I understand that to be pragmatic let's just remove it.

v-stepanov: Well, If we see that we really will not use it, I would prefer to remove it.

rcillo:
I addressed the comments, with exception for changing the API for the authorization plugin. This comes next.

v-stepanov:
@rcillo by "next" you mean next PR or next commit in this PR ?

lmontrieux:
@v-stepanov that would be in another repository :)

v-stepanov:
@lmontrieux But in this repo changes are not done yet. `AuthorizationService.isAuthorized` still expects subject as a first param.

lmontrieux:
@v-stepanov ah, then it is both this repo and the other one :).

rcillo:
@v-stepanov I removed subject from the method signature.

v-stepanov:
👍

rcillo:
@v-stepanov @lmontrieux I'll wait for nakadi plugin implementation in private repo to make a new staging release.

lmontrieux:
👍

antban:
:+1:

PR #710, opened 2017-07-17
ARUHA-654: Fix. OpenAPI creates broken link in the Nakadi manual.

SergKam:
ARUHA-654:Update API documentation in nakadi-manual

## Description
If summary present in the OpenAPI file the swagger2markdown uses the summary instead of endpoint name as an anchor.
The script that generates Index files uses only the endpoint name.
It is better to unify the API and use "description" everywhere.

## Review
- [x] Tests
- [x] Documentation
- [x] CHANGELOG


lmontrieux:
@SergKam can you please add these changes to the changelog?

lmontrieux:
👍

SergKam:
:+1:

PR #712, opened 2017-07-20
Aruha 954 log realm

v-stepanov:
[ARUHA-954: Log realm for each request](https://techjira.zalando.net/browse/ARUHA-954)

## Description
We need this to know if there's someone who uses Nakadi with the realm other than /services

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG - I will not add changelog as these changes are temporary

## Deployment Notes
We will switch the traffic to this stack only for some small period of time


lmontrieux:
👍

PR #715, opened 2017-07-24
ARUHA-906: Fix documentations build

SergKam:
ARUHA-906: Merge Documentation
https://techjira.zalando.net/browse/ARUHA-906

## Description
To build API documentation the swagger file should be inside /docs folder.
So I moved it there but leave the symlink on the old place


antban:
:+1:

SergKam:
:+1:

PR #729, opened 2017-08-15
aruha-1015 identifying source of error

rcillo:
## Description

This PR is not meant to be merged to master, nor released on live. It's
going to be deployed only on staging, for testing purposes.

With this data we should be able to prove if there is a problem with ELB
buffering data.

## Deployment Notes
This branch is meant to be deployed only for testing purposes.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
+++ b/src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
@@ -1,13 +1,19 @@
 package org.zalando.nakadi.service;
 
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.stereotype.Component;
 import org.zalando.nakadi.domain.ConsumedEvent;
+import org.zalando.nakadi.exceptions.runtime.MyNakadiRuntimeException1;
 import org.zalando.nakadi.view.Cursor;
 import org.zalando.nakadi.view.SubscriptionCursor;
 
 import java.io.IOException;
 import java.io.OutputStream;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.util.Date;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
```diff
@@ -85,6 +94,31 @@ public int writeBatch(final OutputStream os, final Cursor cursor, final List<byt
...
+    private int writeDebugInfo(final OutputStream os) throws IOException {
+        int byteCount = 0;
+
+        os.write(B_DEBUG_BEGIN);
+        byteCount += B_DEBUG_BEGIN.length;
+
+        final DateTime now = new DateTime(DateTimeZone.UTC);
+        final InetAddress ip;
+        try {
+            ip = InetAddress.getLocalHost();
```
v-stepanov: will it show the real IP, or some docker-local IP?

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
```diff
@@ -85,6 +94,31 @@ public int writeBatch(final OutputStream os, final Cursor cursor, final List<byt
...
+        } catch (final UnknownHostException e) {
+            throw new MyNakadiRuntimeException1("problem getting servers ip", e);
+        }
+        final String debugString = "Batch written at " + now.toString() + " from server " + ip.getHostAddress();
```
v-stepanov: Maybe it will make sense to write it as a structured json, like
```
{ 
  "sent_at": "2017-08-15T13:57:07.634Z",
  "nakadi_ip": "145.34.12.89"
}
```
That will allow to parse it easier in consumer (but I actually I don't know if we will want to parse it)

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
rcillo: Good point. I'll pass an env variable that identifies the stack.

src/main/java/org/zalando/nakadi/service/EventStreamWriterBinary.java
v-stepanov: ok, then it's fine as it is.

lmontrieux:
@rcillo since this PR was for testing only, do you think it should be closed now?

lmontrieux:
Closing since it wasn't meant to be merged. @rcillo , feel free to re-open if you think I should not have closed this.

PR #730, opened 2017-08-16
aruha-1032: enabled lz4 for producer

adyach:
[ARUHA-1032: Enable lz4 compression type for Kafka producer](https://techjira.zalando.net/browse/ARUHA-1032)

## Description
We want kafka compression feature to be tested and released, so we enable lz4 compression.

## Review
- [x] CHANGELOG


src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -108,6 +109,7 @@ public Properties getKafkaProducerProperties() {
         producerProps.put("request.timeout.ms", kafkaSettings.getRequestTimeoutMs());
         producerProps.put("batch.size", kafkaSettings.getBatchSize());
         producerProps.put("linger.ms", kafkaSettings.getLingerMs());
+        producerProps.put(ProducerConfig.COMPRESSION_TYPE_CONFIG, "lz4");
```
lmontrieux: It's great that you use this constant instead of a String like the few lines above. Do you think you could do the same for the other properties?

lmontrieux: And, would it make sense to make `lz4` part of kafkaSettings?

adyach: @lmontrieux cda0b7ab
About moving it to KafkaSettings: I think it is not going to change from env to env(at least I do not see it that way). There are also other properties, which are not in that config.

lmontrieux:
👍

adyach:
👍

src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -77,8 +80,10 @@ private static String buildBootstrapServers(final List<Broker> brokers) {
     private Properties buildKafkaProperties(final List<Broker> brokers) {
         final Properties props = new Properties();
         props.put("bootstrap.servers", buildBootstrapServers(brokers));
-        props.put("value.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
-        props.put("key.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
+        props.put(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG,
+                "org.apache.kafka.common.serialization.ByteArrayDeserializer");
+        props.put(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG,
+                "org.apache.kafka.common.serialization.ByteArrayDeserializer");
```
v-stepanov: @adyach are these two properties used at all? As I understand they are overwritten in getKafkaConsumerProperties.
I ask this because I see that you changed it from StringDeserializer to ByteArrayDeserializer.

adyach: you are right, they are overridden, I did not want to refactor that class at all, but @lmontrieux asked for changing properties to use kafka clients constants. If you insist I can factor this out.

v-stepanov: Yes, I think we should
- remove that if we are sure that these are not used 
or
- switch it back to StringDeserializer if we are not sure :)

adyach: @v-stepanov c2e6a9af

adyach:
👍

v-stepanov:
👍

PR #731, opened 2017-08-18
Aruha 1015 jetty 500 threads

rcillo:
This PR is not meant to be merged.

I just would like to share the code from the jetty experiment. I'm forcefully setting min and max threads to 500 so that we achieve the same level of parallelism from tomcat. This could trigger the bug from aruha-1015.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,6 +87,7 @@ jar {
 
 configurations {
     all*.exclude module : 'spring-boot-starter-logging'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -87,6 +87,7 @@ jar {
 
 configurations {
     all*.exclude module : 'spring-boot-starter-logging'
+    compile.exclude module: "spring-boot-starter-tomcat"
```
adyach: but it is as well LGTM

PR #732, opened 2017-08-21
aruha-1083 change Tomcat by Jetty

rcillo:
[ARUHA-1083: Deploy Nakadi with Jetty](https://techjira.zalando.net/browse/ARUHA-1083)

## Description
This PR replaces Tomcat by Jetty as the default servlet container. The main motivation is to solve a rare but critical concurrency bug that could lead to wrong event delivery. The investigation is detailed in this [postmortem](https://docs.google.com/document/d/1d1ryipAuUGUqOYAfZBhQZRiQTT5BnVce4BOiHJU61vM/edit).

As a positive side effect, we also observed much lower CPU usage with Jetty.

## Review
- [ ] Tests
- [ ] Documentation
- [x] CHANGELOG

## Deployment Notes
No special deployment instruction. But since it changes a significant component, we should first try it with a smaller part of traffic before shipping to all users.

### DIFF
--- a/src/main/java/org/zalando/nakadi/config/JettyConfig.java
+++ b/src/main/java/org/zalando/nakadi/config/JettyConfig.java
@@ -15,7 +15,8 @@ public class JettyConfig {
             @Value("${jetty.threadPool.maxThreads:200}") final String maxThreads,
             @Value("${jetty.threadPool.minThreads:8}") final String minThreads,
             @Value("${jetty.threadPool.idleTimeout:60000}") final String idleTimeout) {
-        final JettyEmbeddedServletContainerFactory factory =  new JettyEmbeddedServletContainerFactory(Integer.valueOf(port));
+        final JettyEmbeddedServletContainerFactory factory =
+                new JettyEmbeddedServletContainerFactory(Integer.valueOf(port));
         factory.addServerCustomizers((JettyServerCustomizer) server -> {
             final QueuedThreadPool threadPool = server.getBean(QueuedThreadPool.class);
             threadPool.setMaxThreads(Integer.valueOf(maxThreads));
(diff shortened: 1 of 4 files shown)


lmontrieux:
👍

v-stepanov:
@rcillo we have a section with tomcat properties https://github.com/zalando/nakadi/blob/master/src/main/resources/application.yml#L21
Should it be now removed?

rcillo:
@v-stepanov these configs are related to the tomcat.jdbc.pool packaged, which is not related to the embedded tomcat servlet container. As far as I can tell, they are not dependent on each other and it still makes sense to keep these configs here.

v-stepanov:
@rcillo ok, good.

v-stepanov:
👍

rcillo:
@v-stepanov @lmontrieux I added one more commit to make it possible to configure jetty, even though we are using the defaults. I did it because the number of threads is important for Nakadi, since it's used to configure Hystrix. So we need to have this property available, even though Jetty still uses the default values.

lmontrieux:
👍

src/main/java/org/zalando/nakadi/config/JettyConfig.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class JettyConfig {
+    @Bean
+    public JettyEmbeddedServletContainerFactory jettyEmbeddedServletContainerFactory(
+            @Value("${server.port:8080}") final String port,
+            @Value("${jetty.threadPool.maxThreads:200}") final String maxThreads,
+            @Value("${jetty.threadPool.minThreads:8}") final String minThreads,
+            @Value("${jetty.threadPool.idleTimeout:60000}") final String idleTimeout) {
```
v-stepanov: In general I'm fine with that but as you already specified the default properties in application.yml - I think you should not specify the defaults here.

v-stepanov:
👍

rcillo:
👍

v-stepanov:
👍

lmontrieux:
👍

PR #736, opened 2017-08-22
aruha-1083 fix gzip flush

rcillo:
## Description
This is a continuation from https://github.com/zalando/nakadi/pull/732

By default, Jetty output stream flushing behaves differently from Tomcat
when using compression. Calling flush would not cause the actual bytes
to be immediately flushed. This change in behaviour introduced a
problem, where subscription consumers would not get batches on time in
order to commit them. This way, their connections would be constantly
closed and they were not able to make progress.

With this change, all batches are flushed to the wire at the exact same
time as they used to be before switching to Jetty.

## Review
- [ ] Tests
- [ ] Documentation
- [x] CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- Sync flush batches when using gzip streams.
+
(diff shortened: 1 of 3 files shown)


adyach:
👍

zrobgar:
👍

PR #739, opened 2017-08-23
Aruha 998 admin endpoint

lmontrieux:
ARUHA-998: Admin endpoints

## Description
2 New endpoints for getting and posting lists of administrators

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG


src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -185,4 +199,43 @@ public void validateAuthorization(final EventType original, final EventTypeBase
         validateAuthorization(newAuth);
     }
 
+    public AdminAuthorization getAdmins() {
```
adyach: Why do you think this methods belong to validator? It looks more like service level logic of settigns controller

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -185,4 +199,43 @@ public void validateAuthorization(final EventType original, final EventTypeBase
...
+                .collect(Collectors.toList());
+        final List<AuthorizationAttribute> writePermissions = allPermissions.stream()
+                .filter(p -> p.getOperation() == AuthorizationService.Operation.WRITE)
+                .map(p -> p.getAuthorizationAttribute())
+                .collect(Collectors.toList());
+        return new AdminAuthorization(adminPermissions, readPermissions, writePermissions);
+    }
+
+    public void updateAdmins(final AdminAuthorization newAdmins) throws InsufficientAuthorizationException {
+        if (newAdmins.getAdmins().isEmpty() || newAdmins.getWriters().isEmpty() || newAdmins.getReaders().isEmpty()) {
```
adyach: Isn't it checked on controller level by spring? Also you pass view objects to service level.

src/main/java/org/zalando/nakadi/domain/AdminAuthorization.java
```diff
@@ -0,0 +1,89 @@
...
+    public List<AuthorizationAttribute> getList(final AuthorizationService.Operation operation) {
+        switch (operation) {
+            case ADMIN:
+                return admins;
+            case READ:
+                return readers;
+            case WRITE:
+                return writers;
+            default:
+                return null;
```
adyach: shouldn't be that exception that someone trying to pass unprocessable operation?

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -185,4 +199,43 @@ public void validateAuthorization(final EventType original, final EventTypeBase
...
+            throw new InsufficientAuthorizationException("There must be at least one administrator in each list");
+        }
+        final AdminAuthorization currentAdmins = getAdmins();
+        transactionTemplate.execute(action -> {
```
adyach: Do you think you could move trasaction in the place where you requests to db?

src/main/java/org/zalando/nakadi/domain/AdminAuthorization.java
lmontrieux: Actually this would never happen: Operation is an enum and 'ADMIN', 'READ' and 'WRITE' are the only possible values

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
lmontrieux: Good point. I will move it.

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
lmontrieux: @adyach I could, but it seems to me that in other places, we run transactions in the service. For example, there are transactions in EventTypeService, and in TimelineService.
Moving the transaction to the repository will make the repository code more complex, but it can be done. What do you think?

src/main/java/org/zalando/nakadi/domain/AdminAuthorization.java
adyach: and that's why it usually trhows an exception ) that later you don't dig into the stack trace understanding where this npe came from.

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
adyach: In TimelineService it is done that way, because the value has to be updated once it is created in DB, but ussual way is to have it in repository. As a bonus you will not need to use TransactionTemplate replacing it with @Transactional on the method. I do not see it that comlex, you will basically merge this two methods in one.

lmontrieux: 425df1d

src/main/java/org/zalando/nakadi/domain/AdminAuthorization.java
lmontrieux: 68129ae

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
lmontrieux: 4e4d08a

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
lmontrieux: 7933617
Still an issue with spring not checking the minimum size as I expected, I'm on it

lmontrieux: 62eb15b

src/main/java/org/zalando/nakadi/domain/AdminAuthorization.java
```diff
@@ -0,0 +1,124 @@
...
+        final List<AuthorizationAttribute> writers = permissions.stream()
+                .filter(p -> p.getOperation().equals(AuthorizationService.Operation.WRITE))
+                .map(p -> p.getAuthorizationAttribute())
+                .collect(Collectors.toList());
+
+        return new AdminAuthorization(admins, readers, writers);
+    }
+
+    @Override
+    public boolean equals(final Object o) {
```
adyach: Why do you define it? I do not see any usage of it, but I see that Permission class is used by Predicate.isEquals, which uses equals method, which is not defined for Permission.

lmontrieux: 79d774e and fb90914

adyach:
👍

adyach:
👍

antban:
:+1:

PR #743, opened 2017-08-29
Aruha 1079 admin override ET authorization

lmontrieux:
ARUHA-1079: Admin override of Event Type authorization

## Description
This PR allows administrators to bypass event type authorization. Therefore, admins are always allowed to perform all actions on all event types.

## Review
- [ ] Tests
- [ ] Documentation
- [x] CHANGELOG


src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -119,7 +122,9 @@ public void authorizeEventTypeWrite(final EventType eventType)
                     AuthorizationService.Operation.WRITE,
                     resource);
             if (!authorized) {
-                throw new AccessDeniedException(AuthorizationService.Operation.WRITE, resource);
+                if (!adminService.isAdmin(AuthorizationService.Operation.WRITE)) {
+                    throw new AccessDeniedException(AuthorizationService.Operation.WRITE, resource);
+                }
```
antban: @lmontrieux 
Despite the fact that I think that reading everything with administrator is a wrong, but valuable thing, I strongly believe that WRITE operation should be forbidden for administrators.

antban: I mean for all event types. It will lead to huge unrecoverable errors.

lmontrieux: @antban I have removed the ability to read and write everything. It is still possible for admins to add themselves to readers and/or writers, which will (eventually) be logged.

57a4328

antban:
:+1:

adyach:
deploy validation please

SergKam:
:+1:

PR #765, opened 2017-10-24
aruha-1242 add event type and application to SLO log

rcillo:
This is part of https://techjira.zalando.net/browse/ARUHA-1242

## Description
By adding event type and the name of the "user" publishing, we are able
to easily identify top publishing applications using scalyr.

We already expose this data in our /metrics endpoint, but as we all know
very well, it has more than 2MB and zmon is just not able to handle it,
specially when running 400 instances for Black Friday.

On the other hand, scalyr, when properly used, is reliable and
performant.

I'm gonna use this new information to identify event types that would
require repartitioning for Black Friday.

## Deployment Notes
It will require a modification in scalyr parser in order to properly extract the new fields and make queries faster.

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
@@ -71,7 +71,7 @@ public class EventPublishingController {
                     request, eventTypeMetrics, client);
             eventTypeMetrics.incrementResponseCount(response.getStatusCode().value());
             return response;
-        } catch (RuntimeException ex) {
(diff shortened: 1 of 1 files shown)


rcillo:
One could argue that the event type and application name are already present in the access log line but that would not allow us to easily compute numeric queries based on it since it would require correlation of flow id between different log lines. I needed all this information in a single log line.

adyach:
👍

rcillo:
👍

rcillo:
deploy validation please

PR #766, opened 2017-10-24
ARUHA-1255 Optimize subscription reaction on changed offsets

antban:
[ARUHA-1255: Do not call zk twice while receiving new offset of commit while streaming]

## Description
While creating subscription for nodes modification through curator library 2 actions accomplished in 1 request - creating subscription and getting latest data. Therefore, there is no need to additionally query data.
## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes


src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionImpl.java
```diff
@@ -0,0 +1,137 @@
...
+
+import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+
+public abstract class ZkSubscriptionImpl<ReturnType, ZkType> implements ZkSubscription<ReturnType>, Watcher {
+    protected final CuratorFramework curatorFramework;
+    protected final String key;
+    private final AtomicReference<Runnable> listener;
+    private final AtomicReference<ExceptionOrData<ReturnType>> data = new AtomicReference<>();
```
v-stepanov: Why do you use AtomicReferences here and not just volatile properties?

src/main/java/org/zalando/nakadi/service/subscription/state/ClosingState.java
```diff
@@ -68,19 +75,19 @@ private void onTopologyChanged() {
...
         reactOnTopologyChange();
     }
 
-    private void reactOnTopologyChange() {
+    private void reactOnTopologyChange() throws NakadiRuntimeException {
+        final ZkSubscriptionClient.Topology topology = topologyListener.getData();
 
         // Collect current partitions state from Zk
         final Map<EventTypePartition, Partition> partitions = new HashMap<>();
-        getZk().runLocked(() -> Stream.of(getZk().listPartitions())
```
v-stepanov: Is it the call to ZK which was removed?

v-stepanov:
Everything else looks good for me (it's strange, I expected to write at least 10 comments)

src/main/java/org/zalando/nakadi/service/subscription/state/ClosingState.java
antban: no, i will add comment where the actual stuff happend

src/main/java/org/zalando/nakadi/service/subscription/state/ClosingState.java
```diff
@@ -108,21 +115,17 @@ private void registerListener(final EventTypePartition key) {
         listeners.put(
                 key,
                 getZk().subscribeForOffsetChanges(
-                        key, () -> addTask(() -> this.offsetChanged(key))));
-        reactOnOffset(key);
-    }
-
-    private void offsetChanged(final EventTypePartition key) {
-        if (listeners.containsKey(key)) {
-            listeners.get(key).refresh();
```
antban: @v-stepanov this call was removed

src/main/java/org/zalando/nakadi/service/subscription/state/ClosingState.java
```diff
@@ -108,21 +115,17 @@ private void registerListener(final EventTypePartition key) {
...
     }
 
     private void reactOnOffset(final EventTypePartition key) {
+        if (!listeners.containsKey(key)) {
+            return;
+        }
         final NakadiCursor newCursor;
         try {
-            newCursor = getContext().getCursorConverter().convert(key.getEventType(), getZk().getOffset(key));
+            newCursor = getContext().getCursorConverter().convert(key.getEventType(), listeners.get(key).getData());
```
antban: @v-stepanov And that is where the second call is

v-stepanov:
👍

antban:
:+1:

antban:
deploy validation please

PR #770, opened 2017-10-27
aruha-1247: increased minimum fetch bytes

adyach:
[ARUHA-1247: Adjust params Kafka params](https://techjira.zalando.net/browse/ARUHA-1247)

## Description
Kafka broker sends data to consumer once it received 1 byte of data or 500ms passed by default. This PR changes the default behaviour of that.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG


rcillo:
@adyach I agree with this change, since going through the loop of fetch/flush in Nakadi for a single byte costs too much. But have you noticed any difference in CPU or heap usage in Nakadi with this change in place?

rcillo:
@adyach if you observed a lower CPU usage, I would say that this parameter should be derived from batch_timeout, since it might be that we would be adding latency of 500ms messages that used to be delivered as soon as they were available. I'm not aware of any realtime use case with Nakadi, but I cannot state this for sure.

adyach:
@rcillo I have not noticed any CPU usage changes, I thought It would be nice to verify it in production, because we have much more use cases there. I am also not concerned about 500ms for low latency applications.

PR #778, opened 2017-11-09
Aruha 1313

v-stepanov:
[ARUHA-1313](https://techjira.zalando.net/browse/ARUHA-1313)

## Description
Fixed subscription /stats endpoint

## Review
- [x] Tests
- [ ] Documentation
- [x] CHANGELOG

## Deployment Notes
Deploy and enjoy

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionNode.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionNode.java
@@ -1,6 +1,5 @@
 package org.zalando.nakadi.service.subscription.zk;
 
-import org.zalando.nakadi.domain.EventTypePartition;
 import org.zalando.nakadi.service.subscription.model.Partition;
 import org.zalando.nakadi.service.subscription.model.Session;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionNode.java
```diff
@@ -39,23 +38,24 @@ public void setSessions(final Session[] sessions) {
...
                 .findAny();
     }
 
     @Nullable
-    public String guessStream(final String partition) {
-        return getPartitionWithActiveSession(partition).map(Partition::getSession).orElse(null);
+    public String guessStream(final String eventType, final String partition) {
+        return getPartitionWithActiveSession(eventType, partition)
+                .map(Partition::getSession)
+                .orElse(null);
```
adyach: Could you return empty string here? In that case you won't need Optional at SubscriptionService.java:270

src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionNode.java
```diff
@@ -39,23 +38,24 @@ public void setSessions(final Session[] sessions) {
...
+        return getPartitionWithActiveSession(eventType, partition)
+                .map(Partition::getState)
+                .orElse(Partition.State.UNASSIGNED);
     }
 
-    private Optional<Partition> getPartitionWithActiveSession(final String partition) {
+    private Optional<Partition> getPartitionWithActiveSession(final String eventType, final String partition) {
         return Stream.of(partitions)
-                .filter(p -> p.getPartition().equals(partition))
+                .filter(p -> p.getPartition().equals(partition) && p.getEventType().equals(eventType))
```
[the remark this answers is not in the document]
v-stepanov: yes

src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionNode.java
v-stepanov: For me this is actually an anti-pattern to return empty String when there is no value. That's why I would not like to change it. We replace `null` with empty String before showing that to user - which is more or less fine. But I think this method should not return empty String to indicate no value.

adyach: Returning `null` is even worse! Return optional then :)

v-stepanov: No, I think returning `null` is much better than returning empty String to indicate that there is no value.
Also, the purpose of my PR was a bugfix, not a refactoring. I don't really understand why I should refactor here :)

adyach:
👍

v-stepanov:
👍

v-stepanov:
deploy validation please

lmontrieux:
👍

v-stepanov:
👍

PR #779, opened 2017-11-10
aruha-1211: change default storage

adyach:
[aruha-1211: Prescale Kafka](https://techjira.zalando.net/browse/ARUHA-1211)

## Description
The ability to change default timeline storage at runtime and by providing env var.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

### DIFF
--- a/src/main/java/org/zalando/nakadi/config/NakadiConfig.java
+++ b/src/main/java/org/zalando/nakadi/config/NakadiConfig.java
@@ -1,5 +1,6 @@
 package org.zalando.nakadi.config;
 
+import org.apache.curator.framework.CuratorFramework;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Qualifier;
@@ -67,15 +68,16 @@ public class NakadiConfig {
 
     private String getStorageId(final ZooKeeperHolder zooKeeperHolder,
                                 final Environment environment) {
+        final CuratorFramework curator = zooKeeperHolder.get();
         try {
-            zooKeeperHolder.get().create().creatingParentsIfNeeded()
+            curator.create().creatingParentsIfNeeded()
                     .forPath(StorageService.ZK_TIMELINES_DEFAULT_STORAGE, null);
         } catch (final Exception e) {
             LOGGER.debug("Node {} is already there", StorageService.ZK_TIMELINES_DEFAULT_STORAGE);
         }
 
         try {
-            final byte[] storageIdBytes = zooKeeperHolder.get().getData()
+            final byte[] storageIdBytes = curator.getData()
                     .forPath(StorageService.ZK_TIMELINES_DEFAULT_STORAGE);
             if (storageIdBytes != null) {
                 return new String(storageIdBytes);
--- a/src/main/java/org/zalando/nakadi/config/NakadiConfig.java
+++ b/src/main/java/org/zalando/nakadi/config/NakadiConfig.java
@@ -16,6 +16,9 @@ import org.zalando.nakadi.exceptions.runtime.DuplicatedStorageException;
 import org.zalando.nakadi.repository.db.StorageDbRepository;
 import org.zalando.nakadi.repository.zookeeper.ZooKeeperHolder;
 import org.zalando.nakadi.repository.zookeeper.ZooKeeperLockFactory;
+import org.zalando.nakadi.service.StorageService;
+
+import java.util.Optional;
 
 @Configuration
 @EnableScheduling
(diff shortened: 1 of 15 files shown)


src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -28,16 +35,40 @@
...
+        try {
+            curator.getData().usingWatcher((CuratorWatcher) event -> {
+                final byte[] defaultStorageId = curator.getData().forPath(ZK_TIMELINES_DEFAULT_STORAGE);
+                if (defaultStorageId != null) {
+                    final Result<Storage> storageResult = getStorage(new String(defaultStorageId));
+                    if (storageResult.isSuccessful()) {
+                        defaultStorage.setStorage(storageResult.getValue());
+                    }
+                }
+                watchDefaultStorage();
```
v-stepanov: why it calls itself?

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -28,16 +35,40 @@
...
+            }).forPath(ZK_TIMELINES_DEFAULT_STORAGE);
+        } catch (final Exception e) {
+            LOG.warn(e.getMessage(), e);
```
v-stepanov: maybe you can add some context here in the log. Like "Error occurred when updating default storage info" or something like that.

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -125,4 +155,16 @@ public StorageService(final ObjectMapper objectMapper,
         return Result.ok();
     }
 
+    public Result<Storage> setDefaultStorage(final String defaultStorageId) {
+        final Result<Storage> storageResult = getStorage(defaultStorageId);
+        if (storageResult.isSuccessful()) {
+            try {
+                curator.setData().forPath(ZK_TIMELINES_DEFAULT_STORAGE, defaultStorageId.getBytes(Charsets.UTF_8));
+            } catch (Exception e) {
```
v-stepanov: classic

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -125,4 +155,16 @@ public StorageService(final ObjectMapper objectMapper,
...
+                LOG.error(e.getMessage(), e);
```
v-stepanov: can you please put some context in this log as well

src/main/java/org/zalando/nakadi/config/NakadiConfig.java
```diff
@@ -34,22 +39,54 @@ public ZooKeeperLockFactory zooKeeperLockFactory(final ZooKeeperHolder zooKeeper
...
+    private String getStorageId(final ZooKeeperHolder zooKeeperHolder,
+                                final Environment environment) {
+        final CuratorFramework curator = zooKeeperHolder.get();
         try {
-            storageDbRepository.createStorage(storage);
-        } catch (final DuplicatedStorageException e) {
-            LOGGER.info("Creation of default storage failed: {}", e.getMessage());
+            curator.create().creatingParentsIfNeeded()
+                    .forPath(StorageService.ZK_TIMELINES_DEFAULT_STORAGE, null);
+        } catch (final Exception e) {
```
v-stepanov: I think you should catch more specific exception here (I think its NodeExistsException or something like that).

src/main/java/org/zalando/nakadi/config/NakadiConfig.java
```diff
@@ -34,22 +39,54 @@ public ZooKeeperLockFactory zooKeeperLockFactory(final ZooKeeperHolder zooKeeper
...
+            LOGGER.debug("Node {} is already there", StorageService.ZK_TIMELINES_DEFAULT_STORAGE);
         }
-        return storage;
+
+        try {
+            final byte[] storageIdBytes = curator.getData()
+                    .forPath(StorageService.ZK_TIMELINES_DEFAULT_STORAGE);
+            if (storageIdBytes != null) {
+                return new String(storageIdBytes);
```
v-stepanov: it's better specify encoding here.

src/main/resources/application.yml
```diff
@@ -92,6 +92,7 @@ nakadi:
       runPeriodMs: 3600000 # 1 hour
       deletionDelayMs: 2000 # 2 seconds, to be on the safe side
     consumerNodesCleanup.runPeriodMs: 21600000 # 6 hours
+  timelines.storage.default: "default"
```
v-stepanov: why is it in quotes?

src/main/java/org/zalando/nakadi/service/StorageService.java
adyach: the watcher is one time thing, you have to create watch once it was triggered

src/main/java/org/zalando/nakadi/controller/StoragesController.java
```diff
@@ -88,4 +88,16 @@ public StoragesController(final SecuritySettings securitySettings, final Storage
         }
         return Responses.create(result.getProblem(), request);
     }
+
+    @RequestMapping(value = "/storages/default/{id}", method = RequestMethod.PUT)
```
v-stepanov: I think you can add some tests to `StoragesControllerTest`

src/main/java/org/zalando/nakadi/service/StorageService.java
adyach: Actually, you can see that by just checking the stacktrace in the logs

v-stepanov:
what about swagger?

src/main/java/org/zalando/nakadi/service/StorageService.java
adyach: checkstyle says nothing about it :)

src/main/resources/application.yml
adyach: because it is a string

src/main/java/org/zalando/nakadi/config/NakadiConfig.java
adyach: 👍

src/main/java/org/zalando/nakadi/controller/StoragesController.java
adyach: This is covered by StorageControllerAT.java

src/main/java/org/zalando/nakadi/config/NakadiConfig.java
adyach: Basically, I do not care what happens on that step, if it fails I will take from env var. I can fix the message here

v-stepanov: With current implementation each time Nakadi instance starts it will log ERROR in the logs. I think it's better to make difference between NodeExists exception and other exceptions.

src/main/java/org/zalando/nakadi/controller/StoragesController.java
v-stepanov: There you cover the basic use case (which is ok for acceptance test)
But you don't cover the error cases (not authenticated; error from storageService), I think unit test for controller is a perfect place for that.

src/main/java/org/zalando/nakadi/service/StorageService.java
v-stepanov: Yes, you are right, it's a one-time trigger https://zookeeper.apache.org/doc/r3.4.5/zookeeperProgrammers.html#ch_zkWatches

src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -28,16 +35,40 @@
...
+                          @Qualifier("default_storage") final DefaultStorage defaultStorage,
+                          final ZooKeeperHolder zooKeeperHolder) {
         this.objectMapper = objectMapper;
         this.storageDbRepository = storageDbRepository;
+        this.defaultStorage = defaultStorage;
+        this.curator = zooKeeperHolder.get();
+    }
+
+    @PostConstruct
+    private void watchDefaultStorage() {
```
v-stepanov: It may happen that functionality that creates `/nakadi/timelines/default_storage` node will be called later than this `@PostConstruct` and in this case the watcher will fail to create so all updates of this node will be ignored.

v-stepanov:
@adyach 
For me everything looks good. I have only one open question https://github.com/zalando/nakadi/pull/779/files#r150561572

v-stepanov:
👍

adyach:
👍

adyach:
deploy validation please

adyach:
👍

rcillo:
👍

PR #780, opened 2017-11-11
ARUHA-1316: added default stream_timeout;

v-stepanov:
[ARUHA-1316: Limit stream_timeout for LoLA and HiLA](https://techjira.zalando.net/browse/ARUHA-1316)

## Description
Limited stream timeout to 1h ±10min.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
We should watch carefully in a case it brings problems to our users.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventStreamConfig.java
+++ b/src/main/java/org/zalando/nakadi/service/EventStreamConfig.java
@@ -6,15 +6,18 @@ import org.zalando.nakadi.exceptions.UnprocessableEntityException;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.Immutable;
 import java.util.List;
+import java.util.Random;
 
 @Immutable
 public class EventStreamConfig {
 
+    public static final int MAX_STREAM_TIMEOUT = 3600 + 600; // 1h 10m
+
     private static final int BATCH_LIMIT_DEFAULT = 1;
     private static final int STREAM_LIMIT_DEFAULT = 0;
     private static final int BATCH_FLUSH_TIMEOUT_DEFAULT = 30;
-    private static final int STREAM_TIMEOUT_DEFAULT = 0;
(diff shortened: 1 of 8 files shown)


src/main/java/org/zalando/nakadi/service/EventStreamConfig.java
```diff
@@ -150,7 +157,7 @@ public Builder withBatchTimeout(@Nullable final Integer batchTimeout) {
         }
 
         public Builder withStreamTimeout(@Nullable final Integer streamTimeout) {
-            if (streamTimeout != null) {
+            if (streamTimeout != null && streamTimeout <= MAX_STREAM_TIMEOUT && streamTimeout > 0) {
```
adyach: Do you think the users use case will be satisfied with that constraint? E.g. they require 1.5 hours to process some queries. I mean leave it for users, if they provide it they need it

v-stepanov: Well, it will be strange if we define the max. limit but allow users to use value greater than this limit. Also if we allow that - that can kill the whole idea of connections shuffling (e.g. if couple of heavy subscriptions will set limit to 1 week)

adyach: Then do not define max value, define only default one. I do not think that there are such limits, mostly people use default, which is infinity, which causing the issues.

v-stepanov: Ok, I see your point. But I actually wanted to limit the max value not only introduce the default one. If we leave the possibility to create super long living connections then we might still have the problem we try to solve. But let's discuss this point tomorrow after standup to get opinion of other team members.

adyach:
👍

rcillo:
👍

rcillo:
deploy validation please

PR #783, opened 2017-11-16
aruha-1322: changelog.md

adyach:
aruha-1322: changelog.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [2.3.0] - 2017-11-14
+
 ### Added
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+## [2.3.0] - 2017-11-14
+
 ### Added
+- Switch from local token info service to remote one
```
rcillo: @adyach I think it's the other way around "switch from remote token info to local token info service"

adyach: but the default one is local, so you can switch to remote by turning the feature on

rcillo: ok, got it

adyach:
👍

PR #785, opened 2017-11-22
ARUHA-1304: Nakadi collects the metric

adyach:
[ARUHA-1304: Nakadi publishes access log to itself](https://techjira.zalando.net/browse/ARUHA-1304)

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
The Event Type has to be created and the feature toggle has to be turned on


src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -50,6 +59,8 @@ protected void doFilterInternal(final HttpServletRequest request,
                     timing,
                     contentEncoding,
                     acceptEncoding);
+
+            publisher.publish(method,path, query,user, response.getStatus(), timing);
```
v-stepanov: `method,path` no space

src/main/java/org/zalando/nakadi/service/kpi/NakadiAccessLogPublisher.java
```diff
@@ -0,0 +1,38 @@
...
+            final FeatureToggleService featureToggleService,
+            final NakadiPublisher publisher,
+            @Value("${nakadi.kpi.event-types.nakadi.access.log:nakadi.access.log}") final String etName) {
+        super(featureToggleService, publisher);
+        this.nakadiAccessLogEventType = etName;
+    }
+
+    public void publish(final String method, final String path, final String query,
+                        final String user, final int status, final Long timing) {
+        publish(nakadiAccessLogEventType, method,path, query,user, status, timing);
```
v-stepanov: somebody has stolen your spaces

src/main/java/org/zalando/nakadi/service/kpi/AbstractNakadiPublisher.java
```diff
@@ -0,0 +1,28 @@
...
+    protected void publish(final String etName, final Object... data) {
+        if (!featureToggleService.isFeatureEnabled(FeatureToggleService.Feature.KPI_COLLECTION)) {
+            return;
+        }
+
+        final JSONObject event = prepareData(data);
+        publisher.enrichAndSubmit(event, etName);
+    }
+
+    protected abstract JSONObject prepareData(final Object[] data);
```
v-stepanov: I don't really get why you decided to extract this to a separate method. Because of this you have to operate with this array of objects. I think the architecture will be more clear if the publish method would accept already JSONObject
`protected void publish(final String etName, final JSONObject data)`
And the child classes would have a separate publish method (as it is now) in which they will convert parameters to JSONObject and pass to this `publish`

src/main/java/org/zalando/nakadi/service/kpi/NakadiPublisher.java
```diff
@@ -0,0 +1,127 @@
...
+                    return;
+                }
+                eventPublisher.publish(jsonArray.toString(), etName, null);
+                LOG.trace("Published batch of {} to {}", eventsCount, etName);
+            } catch (final Exception e) {
+                LOG.error("Error occurred while publishing events to {}, {}", etName, e.getMessage(), e);
+            }
+        } finally {
+            if (featureToggleService.isFeatureEnabled(FeatureToggleService.Feature.KPI_COLLECTION)) {
+                executorService.submit(() -> sendEventBatch(etName));
```
v-stepanov: I don't exactly get how it works.
It submits itself for execution but at the same time it is also submitted for execution from `scheduleEventsBatchesSend` method.

src/main/java/org/zalando/nakadi/service/kpi/NakadiPublisher.java
```diff
@@ -0,0 +1,127 @@
...
+
+    private void scheduleEventsBatchesSend() {
+        LOG.trace("Schedule events collection on {}", Thread.currentThread().getName());
+        if (featureToggleService.isFeatureEnabled(FeatureToggleService.Feature.KPI_COLLECTION)) {
+            for (final String etName : eventTypeEvents.keySet()) {
+                executorService.submit(() -> sendEventBatch(etName));
+            }
+        }
+        try {
+            Thread.sleep(scheduleTimeout);
```
v-stepanov: You can use `ScheduledExecutorService` instead and schedule `scheduleEventsBatchesSend` to be run every X seconds. And you will avoid writing your own scheduler.

src/main/java/org/zalando/nakadi/service/kpi/publisher/EventsProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+        final JSONObject metadata = new JSONObject()
+                .put("occurred_at", Instant.now())
+                .put("eid", uuidGenerator.randomUUID());
+        event.put("metadata", metadata);
+
+        final BlockingQueue<JSONObject> events =
+                eventTypeEvents.computeIfAbsent(etName, etn ->
+                {
+                    LOG.trace("Schedule events collection");
+                    executorService.submit(() -> sendEventBatch(etName));
```
v-stepanov: I think there will be a problem if `sendEventBatch(etName)` will be called before the value is put in the map

src/main/java/org/zalando/nakadi/service/kpi/publisher/NakadiKpiPublisher.java
```diff
@@ -0,0 +1,32 @@
...
+    private final EventsProcessor eventsProcessor;
+
+    @Autowired
+    protected NakadiKpiPublisher(final FeatureToggleService featureToggleService,
+                                 final EventsProcessor eventsProcessor) {
+        this.featureToggleService = featureToggleService;
+        this.eventsProcessor = eventsProcessor;
+    }
+
+    public void publish(final String etName, final Supplier<JSONObject> data) {
```
v-stepanov: can you please rename `data` parameter

src/main/java/org/zalando/nakadi/service/kpi/NakadiAccessLogPublisher.java
```diff
@@ -0,0 +1,29 @@
...
+
+    @Autowired
+    public NakadiAccessLogPublisher(
+            final NakadiKpiPublisher nakadiKpiPublisher,
+            @Value("${nakadi.kpi.event-types.nakadi.access.log:nakadi.access.log}") final String etName) {
+        this.nakadiKpiPublisher = nakadiKpiPublisher;
+        this.nakadiAccessLogEventType = etName;
+    }
+
+    public void publishAccessLog(final Supplier<JSONObject> data) {
```
v-stepanov: can you please rename `data` parameter

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -50,6 +60,14 @@ protected void doFilterInternal(final HttpServletRequest request,
...
                     contentEncoding,
                     acceptEncoding);
+
+            publisher.publishAccessLog(() -> new JSONObject()
+                    .put("method", method)
+                    .put("path", path)
+                    .put("query", query)
+                    .put("app", user)
+                    .put("status_code", response.getStatus())
+                    .put("response_time_ms", timing));
```
v-stepanov: I like it 👍

src/main/java/org/zalando/nakadi/service/kpi/NakadiAccessLogPublisher.java
```diff
@@ -0,0 +1,29 @@
...
+@Service
+public class NakadiAccessLogPublisher {
+
+    private final NakadiKpiPublisher nakadiKpiPublisher;
+    private final String nakadiAccessLogEventType;
+
+    @Autowired
+    public NakadiAccessLogPublisher(
+            final NakadiKpiPublisher nakadiKpiPublisher,
+            @Value("${nakadi.kpi.event-types.nakadi.access.log:nakadi.access.log}") final String etName) {
```
rcillo: Do you plan to provide sensible default values for these properties in the `application.yml`? Are they documented somewhere? It's not evident for me what this property means by just reading its name.

src/main/java/org/zalando/nakadi/service/kpi/publisher/NakadiKpiPublisher.java
rcillo: One suggestion would be `kpiEntry`. What do you think?

rcillo: Or `kpiEvent`

src/main/java/org/zalando/nakadi/service/kpi/publisher/EventsProcessor.java
adyach: what kind of problem ?

src/main/java/org/zalando/nakadi/service/kpi/NakadiAccessLogPublisher.java
adyach: what is wrong with provided default value here?

adyach: we do not document any other values, why this one should be exceptional?

rcillo: Forget it. I just noticed that the default is right there. It's that we sometimes provide defaults in the code and sometimes in the application yml.

src/main/java/org/zalando/nakadi/service/kpi/publisher/EventsProcessor.java
```diff
@@ -0,0 +1,119 @@
...
+        try {
+            final BlockingQueue<JSONObject> events = eventTypeEvents.get(etName);
+            if (events == null) {
+                return;
+            }
+
+            final long finishAt = System.currentTimeMillis() + batchCollectionTimeout;
+            int eventsCount = 0;
+            JSONArray jsonArray = null;
+            while (eventsCount != batchSize && System.currentTimeMillis() < finishAt) {
```
rcillo: I'm not sure if we need batching here since batching is already done by the KafkaProducer.

I would even say, let's create a different Kafka producer, not to use the same producer as user events, and just fire and forget events on it. I mean, we don't need to use the "sync" publisher for this feature.

We could remove all the async code from here and just queue the events in the the KafkaProducer directly since it's already async.

rcillo: Or if we are using the same producer, I would consider using it on a lower level, not to go through `EventPublisher`. Or is there a particular reason why you chose to use the `EventPublisher` with all of its features: validation, enrichment, partitioning and so on?

rcillo: Ah ok, timelines... hunnn I would consider adding a new method to the existing publisher to allow for async post in order to reuse all the queue and async done by the KafkaProducer not to have another buffer zone here.

src/main/java/org/zalando/nakadi/service/kpi/NakadiAccessLogPublisher.java
rcillo: It's fine 👍 I just wonder how someone figures out that it's possible to change some property. Do people have to read the code?

src/main/java/org/zalando/nakadi/service/kpi/publisher/EventsProcessor.java
rcillo: I think @adyach covered this case in line 62, or not? I mean, if `eventTypeEvents.get(etName)` returns `null` because the executor was faster than this code, than it returns and runs again on line 95 `executorService.submit(() -> sendEventBatch(etName))`.

But I would avoid writing more concurrent code if we already have a buffer and queue feature in KafkaProducer. I would try to leverage it the most because then we don't need to fine tune configuration for 2 buffers. All optimizations we do to `KafkaProducer` in term of buffer size and other params would also benefit this feature.

rcillo:
@adyach to summarise my comments, I would consider adding a new call to `EventPublisher`. Something like `postAsync`. That would push the complexity and optmization to the layer that interacts with the storage and could possibly just use kafka producer buffering.

But I understand that there is timelines and short-circuiting and that it might not be trivial. I would like to know what do you think, but maybe it's better to talk tomorrow about it.

src/main/java/org/zalando/nakadi/service/kpi/publisher/NakadiKpiPublisher.java
v-stepanov: Well, I actually suggested to rename it in order to reflect that it is a supplier, not the actual data

src/main/java/org/zalando/nakadi/service/kpi/NakadiAccessLogPublisher.java
adyach: Even moving this property to application.yml won't add more understanding about the property 'value'. In my opinion it should be documented, like e.g. Kafka describes their producer, consumer and broker config params.

src/main/java/org/zalando/nakadi/service/kpi/publisher/EventsProcessor.java
adyach: I understand your point, in the draft version I even tried to implement a separate method, which helps to avoid validation step, but there are still other steps, which taking most of the time.
The purpose of the buffers here is to avoid spending time on the client thread. The time spent is not only about publishing to Kafka, but also about timelines, validation(?), enrichment, authz, short circuiting.

rcillo: Ok, understood. If we did not go though unnecessary steps like validation, enrichment and so on, we would be able to reduce the time spent on user thread considerably, since it would be just a matter of queueing the event in the kafka producer.

But I think your solution is good and I'm not argue any further.

src/main/java/org/zalando/nakadi/service/kpi/publisher/EventsProcessor.java
v-stepanov: ah ok, as @rcillo mentioned - there is a check for null. So everything will work fine.

rcillo:
@adyach are you going to add tests?

src/main/java/org/zalando/nakadi/service/kpi/NakadiAccessLogPublisher.java
v-stepanov: I also like the idea to move this value to application.yml, it's strange to have the value right here in the code.

adyach: moved 3686ca2

adyach:
@rcillo @v-stepanov tests 3686ca2

src/test/java/org/zalando/nakadi/service/kpi/publisher/EventsProcessorTest.java
```diff
@@ -0,0 +1,35 @@
...
+        final JSONObject event = new JSONObject().put("path", "/path/to/event").put("user", "adyachkov");
+        Mockito.when(featureToggleService.isFeatureEnabled(FeatureToggleService.Feature.KPI_COLLECTION))
+                .thenReturn(true);
+
+        eventsProcessor.enrichAndSubmit("test_et_name", event);
+        TestUtils.waitFor(() -> {
+            try {
+                Mockito.verify(eventPublisher).publish(Mockito.any(), Mockito.any(), Mockito.any());
+            } catch (final Exception e) {
+                e.printStackTrace();
```
v-stepanov: I think that you should also throw AssertionError here. If you just print the stacktrace - the test will be green.

src/main/java/org/zalando/nakadi/service/kpi/NakadiPublishingController.java
```diff
@@ -0,0 +1,29 @@
...
+import org.json.JSONObject;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Service;
+import org.zalando.nakadi.service.kpi.publisher.NakadiKpiPublisher;
+
+import java.util.function.Supplier;
+
+@Service
+public class NakadiPublishingController {
```
v-stepanov: Can we have a different name than a "controller"? In a context of Spring the "controller" usually means the handler of some URL requests.

adyach: @v-stepanov name it :)

src/test/java/org/zalando/nakadi/service/kpi/publisher/EventsProcessorTest.java
adyach: This is verification step, the method won't be called, afaiu

src/main/java/org/zalando/nakadi/service/kpi/NakadiPublishingController.java
v-stepanov: To say the truth, I don't understand the purpose of this class. For me it seems that NakadiKpiPublisher could be directly called from LoggingFilter. 
That's why I can't find a name for it :)

adyach: I thought about it as well, but decided to keep in order to collect related calls to publish in one place and do not spread them all over the code.

v-stepanov:
👍

adyach:
👍

adyach:
deploy validation please

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -50,6 +63,14 @@ protected void doFilterInternal(final HttpServletRequest request,
                     timing,
                     contentEncoding,
                     acceptEncoding);
+
+            nakadiKpiPublisher.publish(accessLogEventType, () -> new JSONObject()
+                    .put("method", method)
+                    .put("path", path)
+                    .put("query", query)
+                    .put("app", user)
```
SergKam: We cannot use it for statistics in datalake.  
We need to add some hashing. As here https://techjira.zalando.net/browse/ARUHA-1321

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -50,6 +69,14 @@ protected void doFilterInternal(final HttpServletRequest request,
                     timing,
                     contentEncoding,
                     acceptEncoding);
+
+            nakadiKpiPublisher.publish(accessLogEventType, () -> new JSONObject()
+                    .put("method", method)
+                    .put("path", path)
+                    .put("query", query)
+                    .put("app", Hex.encodeHexString(sha256MessageDigest.digest(user.getBytes(Charsets.UTF_8))))
```
SergKam: I would add both original and hashed app name. One we need for audit and some specific cases of KPI later,
and hash we need for the count of unique users KPI right now.

lmontrieux: I would also add a salt

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -69,12 +69,12 @@ protected void doFilterInternal(final HttpServletRequest request,
                     timing,
                     contentEncoding,
                     acceptEncoding);
-
             nakadiKpiPublisher.publish(accessLogEventType, () -> new JSONObject()
                     .put("method", method)
                     .put("path", path)
                     .put("query", query)
-                    .put("app", Hex.encodeHexString(sha256MessageDigest.digest(user.getBytes(Charsets.UTF_8))))
+                    .put("app", user)
```
v-stepanov: I think that un-encrypted user data should have some prefix so that Data Lake automatically screens it.
@SergKam, should it be started with "user..." ?

SergKam: I don't know the rules of datalake detector, but user_id was automatically hidden as sensitive data.
Better ask guys from datalake, but they can mark it later manually anyway.

src/main/java/org/zalando/nakadi/service/EventPublisher.java
```diff
@@ -118,6 +119,33 @@ public EventPublishResult publish(final String events, final String eventTypeNam
...
     }
 
+    EventPublishResult publishInternal(final String events, final String eventTypeName) {
+        Closeable publishingCloser = null;
+        final List<BatchItem> batch = BatchFactory.from(events);
+        try {
+            publishingCloser = timelineSync.workWithEventType(eventTypeName, nakadiSettings.getTimelineWaitTimeoutMs());
+            final EventType eventType = eventTypeCache.getEventType(eventTypeName);
+            partition(batch, eventType);
+            submit(batch, eventType);
```
v-stepanov: why have you also skipped validation and enrichment? I thought you will just add one param to `publish` method which will indicate if you need authorization or not.

v-stepanov:
👍

adyach:
deploy validation please

adyach:
deploy validation please

v-stepanov:
👍

adyach:
👍

adyach:
@v-stepanov thanks for merging upstream changes, could you approve one more time? I am gonna release it

v-stepanov:
👍

adyach:
👍

adyach:
deploy validation please

adyach:
👍

v-stepanov:
👍

adyach:
👍

rcillo:
👍

v-stepanov:
👍

PR #791, opened 2017-11-29
Aruha 1347 - do not fail when getting subscriptions stats with deleted storage

lmontrieux:
ARUHA-1347: /stats returns 500s, after Kafka cluster is removed

## Description
In this PR, we do not try to create a TopicRepository when calling /stats, which fails when the cluster has been deleted.

This PR is a quick fix. A more thorough refactoring of Cursors should be planned

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

### DIFF
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/SubscriptionAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/SubscriptionAT.java
@@ -388,12 +388,6 @@ public class SubscriptionAT extends BaseAT {
         }
     }
 
-    @Test
-    public void whenStatsWithOffsetInDeletedStorageThenDoNotFail() throws IOException {
-        final String et = createEventType().getName();
-        final Subscription s = createSubscriptionForEventType(et);
-    }
-
     private Response commitCursors(final Subscription subscription, final String cursor, final String streamId) {
         return given()
                 .body(cursor)
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/SubscriptionAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/SubscriptionAT.java
@@ -388,6 +388,12 @@ public class SubscriptionAT extends BaseAT {
         }
     }
 
+    @Test
+    public void whenStatsWithOffsetInDeletedStorageThenDoNotFail() throws IOException {
+        final String et = createEventType().getName();
+        final Subscription s = createSubscriptionForEventType(et);
+    }
+
     private Response commitCursors(final Subscription subscription, final String cursor, final String streamId) {
         return given()
                 .body(cursor)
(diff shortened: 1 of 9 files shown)


src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
```diff
@@ -195,8 +196,13 @@ private NakadiCursor moveBack(final ShiftedNakadiCursor cursor) throws InvalidCu
     }
 
     private long numberOfEventsBeforeCursor(final NakadiCursor initialCursor) {
-        final TopicRepository topicRepository = getTopicRepository(initialCursor.getTimeline());
-        return topicRepository.numberOfEventsBeforeCursor(initialCursor);
+        final Storage.Type storageType = initialCursor.getTimeline().getStorage().getType();
+        switch (storageType) {
+            case KAFKA:
+                return KafkaCursor.toKafkaOffset(initialCursor.getOffset());
```
v-stepanov: I understand the fix you did but I don't exactly understand how the code works at a bigger scale. 
E.g. why we need to know number of events before cursor that is located in timeline which is already deleted? Or do we need to be able to calculate the distance even between expired offsets/cursors?

src/main/java/org/zalando/nakadi/repository/TopicRepository.java
```diff
@@ -77,8 +77,6 @@ void validateReadCursors(List<NakadiCursor> cursors) throws InvalidCursorExcepti
 
     long totalEventsInPartition(Timeline timeline, String partition);
 
-    long numberOfEventsBeforeCursor(NakadiCursor cursor);
```
adyach: The proposed solution will work as a quick fix. Do you think we should file a ticket to refactor it later?

adyach:
@lmontrieux do you think it is possible to write a test? It should be failing later in the code on CursorOperationsService.java:61.

src/main/java/org/zalando/nakadi/repository/TopicRepository.java
lmontrieux: @adyach absolutely, I even mention it in the PR description. I will create a Jira ticket for it.

src/main/java/org/zalando/nakadi/service/CursorOperationsService.java
lmontrieux: Well, we could of course improve this, but in my opinion this falls into the scope of the to-be-created ticket about refactoring this portion of the code. Yesterday, some of us discussed that the cursors code should probably be refactored entirely, so this PR is just a quick fix.

Since it is a quick fix, I prefer to keep it as small as possible.

lmontrieux:
@adyach test in 8018280

src/acceptance-test/java/org/zalando/nakadi/webservice/hila/SubscriptionAT.java
```diff
@@ -388,6 +388,12 @@ public void whenStatsOnNotInitializedSubscriptionThanCorrectResponse() throws IO
         }
     }
 
+    @Test
+    public void whenStatsWithOffsetInDeletedStorageThenDoNotFail() throws IOException {
```
adyach: it is not clear what it checks

adyach:
👍

lmontrieux:
👍

lmontrieux:
deploy validation please

PR #795, opened 2017-12-06
Aruha-1373: allow to narrow additional properties schema

rcillo:
## Description
When changing additional properties from "true" (implicit or explicit)
to some other schema, which further reduces the set of valid events,
Nakadi used to throw a NullPointer.

This change fixes it and allows the redefinition of additional
properties under "forward" mode. This is interpreted as an allowed
change because the set of accepted events is compatible for consumers.

It might introduce incompatible changes for producers since the
validation might become more strict. But since producers are the ones
changing this schema, it's expected that they are doing it on purpose to
provide more guarantees for consumers.

## Review
- [x] Tests
- [x] Documentation
- [x] CHANGELOG


antban:
:+1:

rcillo:
👍

rcillo:
deploy validation please

PR #796, opened 2017-12-07
aruha-1375 fix concurrency issue on timelines switch

rcillo:
## Description

When switching timelines, under high load with multiple concurrent
producers, events would be published to the old timeline, making them
unaccessible to consumers.

The problem was caused by a concurrency issue, since cache expiration
and event publishing would be accidentally executed in parallel.

In this PR we first expire the cache to only then allow producers to
proceed with publishing.
## Review
- [x] CHANGELOG


src/main/java/org/zalando/nakadi/service/timeline/LocalLocking.java
```diff
@@ -70,4 +45,34 @@ public Closeable workWithEventType(final String eventType, final long timeoutMs)
...
+        final Set<String> unlockedEventTypes = new HashSet<>();
+        synchronized (lock) {
+            for (final String item : this.lockedEventTypes) {
+                if (!lockedEventTypes.contains(item)) {
+                    unlockedEventTypes.add(item);
+                }
+            }
+        }
+        return unlockedEventTypes;
+    }
```
antban: Isn't it simplier 
```
synchronized (lock) {
    return this.lockedEventTypes.stream()
            .filter(v -> !lockedEventTypes.contains(v))
            .collect(Collectors.toSet));
}
```

antban:
:+1:

src/main/java/org/zalando/nakadi/service/timeline/LocalLocking.java
rcillo: 👍 done here https://github.com/zalando/nakadi/pull/796/commits/0c2f80fa08a4b8087474b97673f34fa13ffef7b3

antban:
:+1:

rcillo:
👍

rcillo:
deploy validation please

rcillo:
👍

antban:
:+1:

PR #799, opened 2017-12-11
Aruha 1311 collect et metrics

v-stepanov:
[ARUHA-1311: Nakadi collects the metric](https://techjira.zalando.net/browse/ARUHA-1311)

## Description
This functionality sends KPI metrics for event-types modifications

## Review
- [x] Tests
- [ ] Documentation
- [x] CHANGELOG

## Deployment Notes
Feature toggle should be ON.
Should not influence performance as the ET modifications happen not so often.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -10,6 +10,7 @@ import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Component;
 import org.springframework.transaction.support.TransactionTemplate;
 import org.zalando.nakadi.config.NakadiSettings;
@@ -82,6 +83,8 @@ public class EventTypeService {
     private final TimelineSync timelineSync;
     private final NakadiSettings nakadiSettings;
     private final TransactionTemplate transactionTemplate;
+    private final NakadiKpiPublisher nakadiKpiPublisher;
+    private final String etLogEventType;
 
     @Autowired
     public EventTypeService(final EventTypeRepository eventTypeRepository,
@@ -95,7 +98,9 @@ public class EventTypeService {
                             final AuthorizationValidator authorizationValidator,
                             final TimelineSync timelineSync,
                             final TransactionTemplate transactionTemplate,
-                            final NakadiSettings nakadiSettings) {
+                            final NakadiSettings nakadiSettings,
+                            final NakadiKpiPublisher nakadiKpiPublisher,
+                            @Value("${nakadi.kpi.event-types.nakadiEventTypeLog}") final String etLogEventType) {
         this.eventTypeRepository = eventTypeRepository;
         this.timelineService = timelineService;
         this.partitionResolver = partitionResolver;
@@ -108,6 +113,8 @@ public class EventTypeService {
         this.timelineSync = timelineSync;
         this.transactionTemplate = transactionTemplate;
         this.nakadiSettings = nakadiSettings;
+        this.nakadiKpiPublisher = nakadiKpiPublisher;
+        this.etLogEventType = etLogEventType;
     }
 
     public List<EventType> list() {
@@ -135,6 +142,13 @@ public class EventTypeService {
                 LOG.error("Error creating event type {}", eventType, e1);
             }
             throw e;
+        }
+        try {
+            nakadiKpiPublisher.publish(etLogEventType, () -> new JSONObject()
+                    .put("event_type", eventType)
+                    .put("status", "created"));
+        } catch (final Exception e) {
+            LOG.error("Exception occurred when submitting KPI metrics event", e);
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -136,6 +143,14 @@ public void create(final EventTypeBase eventType) throws TopicCreationException,
             }
             throw e;
         }
+        try {
+            nakadiKpiPublisher.publish(etLogEventType, () -> new JSONObject()
+                    .put("event_type", eventType.getName())
+                    .put("status", "created")
+                    .put("category", eventType.getCategory().toString()));
+        } catch (final Exception e) {
```
adyach: It throws no exceptions, why did you wrap it in catch ?

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -170,9 +187,7 @@ public void delete(final String eventTypeName) throws EventTypeDeletionException
                 throw new ConflictException("Can't remove event type " + eventTypeName
                         + ", as it has subscriptions");
             }
-            topicsToDelete = transactionTemplate.execute(action -> {
-                return deleteEventType(eventTypeName);
-            });
+            topicsToDelete = transactionTemplate.execute(action -> deleteEventType(eventTypeName));
```
adyach: 👍

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -206,6 +221,16 @@ public void delete(final String eventTypeName) throws EventTypeDeletionException
                 }
             }
         }
+        try {
+            if (category != null) {
```
adyach: is there any specific reason, that you did not put the code inside this `if` within `try/catch` where event type object is available ?

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -136,6 +143,14 @@ public void create(final EventTypeBase eventType) throws TopicCreationException,
             }
             throw e;
         }
+        try {
+            nakadiKpiPublisher.publish(etLogEventType, () -> new JSONObject()
+                    .put("event_type", eventType.getName())
+                    .put("status", "created")
+                    .put("category", eventType.getCategory().toString()));
```
adyach: There is no need to call toString(), it will be the same as it is defined in the enum: `UNDEFINED` `DATA` `BUSINESS`

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTestCase.java
```diff
@@ -103,7 +105,8 @@ public void init() throws Exception {
 
         final EventTypeService eventTypeService = new EventTypeService(eventTypeRepository, timelineService,
                 partitionResolver, enrichment, subscriptionRepository, schemaEvolutionService, partitionsCalculator,
-                featureToggleService, authorizationValidator, timelineSync, transactionTemplate, nakadiSettings);
+                featureToggleService, authorizationValidator, timelineSync, transactionTemplate, nakadiSettings,
+                nakadiKpiPublisher, "");
```
adyach: maybe name it, because it is not clear what it is

src/main/java/org/zalando/nakadi/service/EventTypeService.java
v-stepanov: Just in case some unchecked exception is thrown. My point here was that if `nakadiKpiPublisher.publish` fails for whatever reason - that should not fail the actual request to Nakadi (creating/updating/deleting ET)

adyach: it was designed not to fail during putting event into the queue, but if you are not certain, than I suggest to catch exceptions inside publish method, it will be cleaner.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
v-stepanov: ok, thanks. Makes sense. Changed here https://github.com/zalando/nakadi/pull/799/commits/70d110aa249bc846a043d55816abf6451ba6deda

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTestCase.java
v-stepanov: done https://github.com/zalando/nakadi/pull/799/commits/79da36f9ee405ae77b7e075811aeec25bcbab9d2

src/main/java/org/zalando/nakadi/service/EventTypeService.java
v-stepanov: Yes, I wanted to publish KPI event only after the ET deletion is finished (including topic deletion). That's why I put it in a place after all topics are deleted.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
v-stepanov: Ok, I moved the catch inside publish method https://github.com/zalando/nakadi/pull/799/commits/b7ebac82d12a99880f07c753b412fdfa6e42edb3

adyach:
👍

v-stepanov:
👍

v-stepanov:
deploy validation please

adyach:
👍

v-stepanov:
👍

PR #803, opened 2017-12-13
Add timeout for subscription locking

antban:
[ARUHA-1388: Subscription is locked in ZK]

## Description
There is an unreproducable bug in curator library with obtaining a lock with this actions: 
1. Create lease
2. Loose zk connection
3. Try to delete lease and fail

It is impossible to reproduce, and it is really really really rare. 
In order to reduce the impact on system because of this bug and provide log entries to filter on, we will just log it for further investigation.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
@@ -34,6 +34,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


antban:
:+1:

adyach:
👍

antban:
deploy validation please

src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
```diff
@@ -85,7 +87,11 @@ public final void runLocked(final Runnable function) {
                 lock = new InterProcessSemaphoreMutex(curatorFramework, "/nakadi/locks/subscription_" + subscriptionId);
             }
 
-            lock.acquire();
+            final boolean acquired = lock.acquire(SECONDS_TO_WAIT_FOR_LOCK, TimeUnit.SECONDS);
+            if (!acquired) {
+                throw new ServiceUnavailableException("Failed to acquire subscription lock within " +
```
v-stepanov: maybe also log some specific log message here?

antban: I use 2 techniques, that help me a lot: 
1. throw Exception
2. log message
I try to never mix it. And here is the same. Log message will be logged in place, where exceptions are logged.

v-stepanov: Oh yes, sorry. I forgot that it will be caught and logged anyway.

antban:
:+1:

lmontrieux:
👍

PR #806, opened 2017-12-14
Correct FAQ section title

PetrGlad:
Also it is not clear what this phrase means "Because Nakadi has exclusive offset handling,".
I would probably remove this altogether and only said that newest_available_offset always holds offset of the latest received event even if all events are expired.

### DIFF
--- a/docs/_documentation/faq.md
+++ b/docs/_documentation/faq.md
@@ -106,7 +106,7 @@ If you get the message "Is the docker daemon running on this host?" first check 
 eval "$(docker-machine env default)"
 ```
 
-#### What's the reason for newest available offset being bigger than oldest offset?
+#### What's the reason for newest available offset being smaller than oldest offset?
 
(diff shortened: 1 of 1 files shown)


lmontrieux:
To be honest it doesn't make much sense to me either. Generally, I prefer to think of offsets as pointing between events, rather than at events. Makes it easier.

Thanks for the PR!

lmontrieux:
👍

adyach:
@PetrGlad thank you for contribution!

adyach:
👍

PetrGlad:
@lmontrieux when one needs to split something that is linearly ordered into non intersecting ranges it is always a good idea to use semi-open ranges. That is ranges that include exactly one of ends: `[a,b)` or `(a,b]`. Conversely `[a,b]` and `(a,b)`complicate all operations significantly as one have to use +/-1s (or limits, if items are rational numbers) for splits, joins and comparisons.

In fact Nakadi allows handling event offset ranges as `(a, b]` (although see #816). This sort of goes away from usual conventions of including starting position instead, but I can live with this :)

PR #807, opened 2017-12-15
ARUHA-1398 reduce logging when user committed with unknown stream ID

lmontrieux:
ARUHA-1398: Reduce logging when user commits with unknown stream ID

## Description
HOTFIX - when user commits with unknown stream ID, only log exception message

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/CursorsController.java
+++ b/src/main/java/org/zalando/nakadi/controller/CursorsController.java
@@ -162,7 +162,7 @@ public class CursorsController {
     @ExceptionHandler(InvalidStreamIdException.class)
     public ResponseEntity<Problem> handleInvalidStreamId(final InvalidStreamIdException ex,
                                                          final NativeWebRequest request) {
-        LOG.warn("Stream id {} is not found", ex.getStreamId(), ex.getMessage());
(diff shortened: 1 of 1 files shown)


src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -162,7 +162,7 @@ public CursorsController(final CursorsService cursorsService,
     @ExceptionHandler(InvalidStreamIdException.class)
     public ResponseEntity<Problem> handleInvalidStreamId(final InvalidStreamIdException ex,
                                                          final NativeWebRequest request) {
-        LOG.warn("Stream id {} is not found", ex.getStreamId(), ex);
+        LOG.warn("Stream id {} is not found", ex.getStreamId(), ex.getMessage());
```
antban: @lmontrieux 
Wow... Strange. I think in this manner you need to add addional `{}` to log message.

lmontrieux: 088ddfd

antban:
плюс один

antban:
:+1:

lmontrieux:
plus un

lmontrieux:
👍

PR #809, opened 2017-12-18
Refactor logging messages

lmontrieux:
- do not log traces
- unify style
- fix a few typos

## Description
This PR refactors logging in Nakadi.
- do not log traces, which pollutes logs and may cause issues if a stack trace is printed a lot
- unify style, for better consistency
- fix a few typos

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

### DIFF
--- a/src/main/java/org/zalando/nakadi/ShutdownHooks.java
+++ b/src/main/java/org/zalando/nakadi/ShutdownHooks.java
@@ -29,7 +29,7 @@ public class ShutdownHooks {
                 try {
                     hook.run();
                 } catch (final RuntimeException ex) {
-                    LOG.warn("Failed to call on shutdown hook for {}", hook, ex);
+                    LOG.warn("Failed to call on shutdown hook for {}: {}", hook, ex.getMessage());
                 }
             }
         }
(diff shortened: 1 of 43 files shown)


src/main/java/org/zalando/nakadi/ShutdownHooks.java
```diff
@@ -29,7 +29,7 @@ private static void onNodeShutdown() {
                 try {
                     hook.run();
                 } catch (final RuntimeException ex) {
-                    LOG.warn("Failed to call on shutdown hook for {}", hook, ex);
+                    LOG.warn("Failed to call on shutdown hook for {}: {}", hook, ex.getMessage());
```
adyach: It wont't be possible to debug the error if `Runnable` failed and there is only message: `java.lang.NullPointerException`.

adyach:
Thank you for contribution, Lionel :)

The change itself is a good way to clean error paths and make our logs not that messy, I understand that. We may skip stacktrace for client error codes, but the errors, which does not relate to the them should be logged to improve investigation process of the failure.

Although, I agree with the change, it is questionable whenever it is a problem of the logging, which turned out to be our problem or the client who generates that errors, because in some cases it is possible, that even client errors were produced, because of the internal errors (like npe etc.) and without stacktraces it is harder to figure out.

v-stepanov:
I also think that with this change it will be very difficult to track the source of exceptions as we will loose the stacktrace.

lmontrieux:
@adyach @v-stepanov this is a good point. Yet some traces really shouldn't be there. We should agree on when to log traces, and when to log messages only.
I like @adyach 's proposal to skip traces for client error codes, but keeping them for "unexpected" things like NPEs.

lmontrieux:
I am closing this PR until I have time to take a less brutal approach to fixing some of our logs.

PR #811, opened 2017-12-19
Do not create TopicRepository when it is not needed.

antban:
[ARUHA-1397: 503 when patching subscriptions with cursor from deleted timeline]

## Description
With timelines deletion nakadi introduced many places that are breaking while relaying on a fact that kafka is always available.
This PR tries to fix this places.


src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -133,12 +133,13 @@ public EventStreamController(final EventTypeRepository eventTypeRepository,
         List<Cursor> cursors = null;
         if (cursorsStr != null) {
             try {
-                cursors = jsonMapper.readValue(cursorsStr, new TypeReference<ArrayList<Cursor>>() {});
```
v-stepanov: Was there anything changed in behavior of `getStreamingStart`? Or are there only improvements and refactoring? Because I could not identify any fix/change in this method.

antban: @v-stepanov There was small refactoring, that didn't change anything. It was made because it was hard to analyse what is going on, therefore I had to separate cursor conversion and topic repository requests.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionValidationService.java
```diff
@@ -101,6 +103,9 @@ private void validateInitialCursors(final SubscriptionBase subscription,
         try {
             for (final SubscriptionCursorWithoutToken cursor : subscription.getInitialCursors()) {
                 final NakadiCursor nakadiCursor = cursorConverter.convert(cursor);
+                if (nakadiCursor.getTimeline().isDeleted()) {
+                    throw new InvalidCursorException(UNAVAILABLE, nakadiCursor);
```
v-stepanov: Will not it result in 422 when providing expired cursor? I understand that now it's even worse (probably the returned status code is 500) but I think it's an important use-case when creating a subscription based on cursors of another subscription. It can be that one of event-types of old subscription has latest cursors committed in some old timeline.

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -147,15 +141,17 @@ private void validateStreamId(final List<NakadiCursor> cursors, final String str
...
+            InternalNakadiException, NoSuchEventTypeException, InvalidCursorException {
         if (cursors.isEmpty()) {
             throw new UnableProcessException("Cursors are absent");
         }
         final Subscription subscription = subscriptionRepository.getSubscription(subscriptionId);
-        validateSubscriptionResetCursors(subscription, cursors);
-
+        validateCursorsBelongToSubscription(subscription, cursors);
+        for (final NakadiCursor cursor : cursors) {
+            cursor.validate();
```
v-stepanov: this method validates much less than `validateSubscriptionResetCursors` used to validate.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionValidationService.java
antban: @v-stepanov If I understood correctly - we decided that it is a way to go now - not allowed to put cursors that are not available.

src/main/java/org/zalando/nakadi/service/NakadiCursorComparator.java
```diff
@@ -50,8 +50,7 @@ private int compareOrdered(final NakadiCursor c1, final NakadiCursor c2) {
         }
 
         while (first.getTimeline().getOrder() != c2.getTimeline().getOrder()) {
-            final boolean isFirstAtEndOfTimeline = StaticStorageWorkerFactory.get(first.getTimeline())
-                    .isLastOffsetForPartition(first.getTimeline(), first.getPartition(), first.getOffset());
+            final boolean isFirstAtEndOfTimeline = first.isLast();
```
v-stepanov: I think this variable can be inlined now

src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -150,8 +155,32 @@ public void resetCursors(final String subscriptionId, final List<NakadiCursor> c
...
+                            v -> v.getFirst().getEventTypePartition(),
+                            v -> v
+                    ));
+            for (final NakadiCursor toCheck : entry.getValue()) {
+                final PartitionStatistics etpStats = statsMapped.get(toCheck.getEventTypePartition());
+                if (!comparator.isOrdered(etpStats.getBeforeFirst(), toCheck, etpStats.getLast())) {
+                    throw new InvalidCursorException(CursorError.UNAVAILABLE, toCheck);
+                }
+            }
+        }
```
v-stepanov: I'm not really sure why you created this additional code to check cursors availability. I think we already have code in `TopicRepository.validateReadCursors` that does it (it is currently used to check initial_cursors of newly created subscription)

src/main/java/org/zalando/nakadi/service/NakadiCursorComparator.java
```diff
@@ -32,6 +32,21 @@ public int compare(final NakadiCursor c1, final NakadiCursor c2) {
         }
     }
 
+    public boolean isOrdered(final NakadiCursor... cursors) {
+        NakadiCursor old = null;
+        for (int i = 0; i < cursors.length; ++i) {
+            final NakadiCursor current = cursors[i];
+            if (null == current) {
```
v-stepanov: I thought we have none-null-by-default policy

src/main/java/org/zalando/nakadi/repository/TopicRepositoryHolder.java
```diff
@@ -27,40 +30,57 @@
     private static final Logger LOG = LoggerFactory.getLogger(TopicRepositoryHolder.class);
 
     private final Map<Storage.Type, TopicRepositoryCreator> repositoryCreators;
-    private final ReadWriteLock readWriteLock;
     private final Map<Storage, TopicRepository> storageTopicRepository;
 
+    private final Lock lock = new ReentrantLock();
```
v-stepanov: why have you refactored locking?

antban: To allow getting topic repositories in parallel while it is being created

antban:
@v-stepanov I addressed all the comments

v-stepanov:
I like everything now

v-stepanov:
👍

antban:
deploy validation please

antban:
deploy validation please

antban:
deploy validation please

src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -201,4 +205,10 @@ public CursorsController(final CursorsService cursorsService,
         return Responses.create(Problem.valueOf(Response.Status.NOT_IMPLEMENTED, "Feature is disabled"), request);
     }
 
+    public static class CursorsAreEmptyException extends RuntimeException {
```
lmontrieux: Why did you decide to define this exception here?

antban: @lmontrieux 
I reviewed the way how underlying services treat this situation and found out, that services are not relying by any means on the amount of cursors (they may work with empty collections).
So the only reasoning for making this situation exceptional is making better user experience (which I believe should be done on a view level).
Therefore I decided to check for it there.

antban: @lm just updated: https://github.com/zalando/nakadi/pull/811/commits/5a556e048feb272721d0b61434e2be7f3ad98369

antban:
:+1:

lmontrieux:
👍

PR #812, opened 2017-12-19
aruha-1399 collect event publishing KPI

rcillo:
https://techjira.zalando.net/browse/ARUHA-1401

## Description

Collecting data necessary to calculate event publishing KPIs. Further details can be found in the data modeling document https://docs.google.com/document/d/1d4Z4FzSSrVCl2oy2G-bAPEcFbyAed_ZOxQqzr2rBGpw/edit#heading=h.c9gt0cqkozqy

It was necessary to refactor out the `hash` computation so that it could
be reused.

## Deployment Notes
The corresponding event type should be created in the targeted deployment environments.

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
+++ b/src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
@@ -133,6 +133,7 @@ public class EventPublishingController {
                     .put("app", applicationName)
                     .put("app_hashed", nakadiKpiPublisher.hash(applicationName))
                     .put("number_of_events", eventCount)
+                    .put("ms_spent", msSpent)
                     .put("batch_size", totalSizeBytes));
         }
     }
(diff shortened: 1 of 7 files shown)


src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -117,6 +127,13 @@ private void reportSLOs(final long startingNanos, final int totalSizeBytes, fina
 
             LOG.info("[SLO] [publishing-latency] time={} size={} count={} eventTypeName={} app={}", msSpent,
                     totalSizeBytes, eventCount, eventTypeName, applicationName);
+
+            nakadiKpiPublisher.publish(kpiBatchPublishedEventType, () -> new JSONObject()
```
v-stepanov: I understand that "msSpent" is not needed for your task, but do you think that we may need it in the future? 
Or maybe we can keep the event-type forward compatible and add it later if needed?

rcillo: I'll add right now, just because we can =)

rcillo: @v-stepanov done https://github.com/zalando/nakadi/pull/812/commits/a62648ee65310078e2ef236c4396288358d9d0cc

v-stepanov:
👍

PR #813, opened 2017-12-20
Aruha 1403 nakadi consumption kpi

rcillo:
https://techjira.zalando.net/browse/ARUHA-1403

## Description
Collect streaming data every X milliseconds. By default 30s.

I based this PR on top of the "publisher kpi" because I wanted to use some changes made there.


src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -237,6 +262,20 @@ private void streamToOutput() {
         }
     }
 
+    private void publishKpi(final String appName, final NakadiKpiPublisher kpiPublisher, final String et,
+                            final long bytes, final long count) {
+        kpiPublisher.publish(
+                getContext().getKpiDataStreamedEventType(),
+                () -> new JSONObject()
+                        .put("api", "hila")
+                        .put("subscription", getContext().getSubscription())
```
v-stepanov: `.getId()`

rcillo: https://github.com/zalando/nakadi/pull/813/commits/5c679907ff705be3103cda29d0b57a35dbaef1fc

src/main/java/org/zalando/nakadi/metrics/StreamKpiData.java
```diff
@@ -0,0 +1,24 @@
+package org.zalando.nakadi.metrics;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+public class StreamKpiData {
+        private AtomicLong bytesSent = new AtomicLong(0);
+        private AtomicLong numberOfEventsSent = new AtomicLong(0);
```
v-stepanov: Why do you use atomics here? It seems that these values are accessed/set within one thread only.

rcillo: Yes, I went home yesterday and realised it's completely useless. I'll change it to usual long.

rcillo: https://github.com/zalando/nakadi/pull/813/commits/c946a4403f66393729112d2d7f0edcca8a423224

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -109,6 +123,18 @@ public void streamEvents(final AtomicBoolean connectionReady, final Runnable che
                     }
                 }
 
+                if (lastKpiEventSent + kpiFrequencyMs < System.currentTimeMillis()) {
+                    final long count = kpiData.getAndResetNumberOfEventsSent();
+                    final long bytes = kpiData.getAndResetBytesSent();
+
+                    if (count > 0) {
```
v-stepanov: as I understand you skip the bytes which were sent by keep-alives (within empty batches), right?

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -237,6 +262,20 @@ private void streamToOutput() {
         }
     }
 
+    private void publishKpi(final String appName, final NakadiKpiPublisher kpiPublisher, final String et,
+                            final long bytes, final long count) {
+        kpiPublisher.publish(
```
rcillo: @v-stepanov would you be fine if I added an "[SLO] [data-streamed]" log here? It would not increase our log amount that much and we could use scalyr to visualize stream data per event type, per subscription, per application and so on.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -265,6 +309,16 @@ public void onExit() {
...
 
+        final String appName = getContext().getParameters().getConsumingAppId();
+        final NakadiKpiPublisher kpiPublisher = getContext().getKpiPublisher();
+        final Subscription subscription = getContext().getSubscription();
+        subscription.getEventTypes().stream()
+                .forEach(et -> {
+                    final long bytes = kpiDataPerEventType.get(et).getAndResetBytesSent();
+                    final long count = kpiDataPerEventType.get(et).getAndResetNumberOfEventsSent();
+                    publishKpi(appName, kpiPublisher, et, bytes, count);
+                });
```
v-stepanov: This code is very similar to the code you have some lines above. Maybe it is possible to make a function from it and use it in 2 places? Also, here you don't check if count is zero.

src/main/java/org/zalando/nakadi/service/EventStream.java
rcillo: Yes, exactly. I thought that this is not actual event data and that it's mostly and operational overhead that the client has not much control over. What do you think? It wasn't specified but I think that this is what is expected.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
rcillo: I wanted to count at least once per stream. Even if no events were transferred. This way we would not miss zombie consumers from the count, since they are actually using the system.

Ok, I'll try to improve the reutilisation.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
v-stepanov: Yes, sure. I was also thinking about something like that. If we calculate this data anyway it makes sense to have it in the logs.

src/main/resources/application.yml
```diff
@@ -125,6 +128,9 @@ nakadi:
       commitTimeout: 5 # seconds
   subscription:
     maxPartitions: 30
+kpi:
+  config:
+    stream-data-collection-frequency-ms: 100
```
v-stepanov: Why do you specify a different value for the test environment? As I see you don't have a test for consumption part. BTW, do you want to have a test for that? :)

rcillo: I tried to write the test. I stopped after I spent 2 hours. It's impossible to unit test the subscription API. I thought about writing an acceptance. I think I could do that =)

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
v-stepanov: Ok I see now. For me it was just strange that you count empty batches data in one case and don't count in another.

rcillo: Improved a little https://github.com/zalando/nakadi/pull/813/commits/3ceea68d643556a8572e43c1c7ee1018d36e960c

rcillo:
deploy validation please

v-stepanov:
👍

rcillo:
👍

rcillo:
deploy validation please

v-stepanov:
👍

antban:
:+1:

PR #819, opened 2018-01-09
ARUHA-1442 Limit memory consumption of subscription streamer to 50 Mb

antban:
[ARUHA-1442: Limit memory consumption of subscription streamer to 50 Mb](link to ticket)

## Description
Now single subscription stream can kill nakadi with not wisely set streaming configuration parameters. 
This PR allows to collect batches only in case when for single stream it takes less then 50Mb, otherwise heaviest partitions will be dumped.

### DIFF
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
@@ -39,7 +39,6 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
@@ -462,7 +461,7 @@ public class EventStreamReadingAT extends BaseAT {
     }
 
     @Test(timeout = 10000)
-    public void whenMemoryOverflowItIsUsed() throws IOException {
+    public void whenMemoryOverflowEventsDumped() throws IOException {
         // Create event type
         final EventType loadEt = EventTypeTestBuilder.builder()
                 .defaultStatistic(new EventTypeStatistics(PARTITIONS_NUM, PARTITIONS_NUM))
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
@@ -4,9 +4,11 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.Sets;
+import com.jayway.restassured.RestAssured;
 import com.jayway.restassured.response.Header;
 import com.jayway.restassured.response.Response;
 import org.hamcrest.Matchers;
+import org.junit.Assert;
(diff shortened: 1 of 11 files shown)


src/main/java/org/zalando/nakadi/controller/EventStreamController.java
```diff
@@ -111,7 +113,9 @@ public EventStreamController(final EventTypeRepository eventTypeRepository,
                                  final FeatureToggleService featureToggleService,
                                  final CursorConverter cursorConverter,
                                  final AuthorizationValidator authorizationValidator,
-                                 final EventTypeChangeListener eventTypeChangeListener) {
+                                 final EventTypeChangeListener eventTypeChangeListener,
+                                 @Value("${nakadi.stream.maxStreamMemoryBytes}")
+                                     @Nullable final Long maxMemoryUsageBytes) {
```
v-stepanov: Why is it nullable?

src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -122,6 +123,23 @@ public void streamEvents(final AtomicBoolean connectionReady, final Runnable che
                         batchStartTimes.put(partition, currentTimeMillis());
                     }
                 }
+                // Dump some data that is exceeding memory limits
+                long memoryUsed = currentBatches.values().stream()
+                        .mapToLong(partition -> partition.stream().mapToLong(event -> event.length).sum()).sum();
```
v-stepanov: This will actually be run at least 10 times a second. Do you think it will make sense to somehow limit the period of running this calculation?

antban: Optimized

src/main/java/org/zalando/nakadi/controller/EventStreamController.java
antban: Removed `@Nullable`

antban:
@v-stepanov Provided acceptance test, that shows that limiting is working. Also, reviewed the logs, and it is always triggered when memory limit is reached https://github.com/zalando/nakadi/pull/819/commits/881a3cf41369279d48aa3aaaf55ab734d4116f79

src/acceptance-test/java/org/zalando/nakadi/webservice/EventStreamReadingAT.java
```diff
@@ -458,6 +461,43 @@ public void whenReadEventsConsumerIsBlocked() throws Exception {
         }
     }
 
+    @Test(timeout = 10000)
+    public void whenMemoryOverflowItIsUsed() throws IOException {
```
v-stepanov: What is used?

v-stepanov:
@antban In general I like everything but the checkstyle failed.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -235,6 +236,34 @@ private void streamToOutput() {
                 messagesAllowedToSend -= toSend.size();
             }
         }
+
+        long memoryConsumed = offsets.values().stream().mapToLong(PartitionData::getBytesInMemory).sum();
+        while (memoryConsumed > getContext().getStreamMemoryLimitBytes()) {
```
adyach: Please, log warning, as it is done for lola

antban:
:+1:

v-stepanov:
👍

antban:
deploy validation please

v-stepanov:
@antban I forgot one thing - we should update API documentation to reflect that batch_flush_timeout and batch_limit are only desired values now.

v-stepanov:
👍

antban:
:+1:

PR #825, opened 2018-01-16
KPI token realm

rcillo:
The realm is needed to distinguish an employee from an application.

https://techjira.zalando.net/browse/ARUHA-1406

### DIFF
--- a/src/main/java/org/zalando/nakadi/security/ClientResolver.java
+++ b/src/main/java/org/zalando/nakadi/security/ClientResolver.java
@@ -15,9 +15,8 @@ import org.zalando.nakadi.config.SecuritySettings;
 import org.zalando.nakadi.util.FeatureToggleService;
 
 import java.security.Principal;
-import java.util.Collections;
+import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 
 import static org.zalando.nakadi.config.SecuritySettings.AuthMode.OFF;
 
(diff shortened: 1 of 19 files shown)


src/main/java/org/zalando/nakadi/security/ClientResolver.java
```diff
@@ -50,15 +49,25 @@ public Client resolveArgument(final MethodParameter parameter,
...
-    private Set<String> getScopes() {
+    public String getRealm() {
+        String realmString = "";
+
         final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
-        if (authentication instanceof OAuth2Authentication) {
-            return ((OAuth2Authentication) authentication).getOAuth2Request().getScope();
+        if (OAuth2Authentication.class.equals(authentication.getClass())) {
+            final Object details = ((OAuth2Authentication) authentication).getUserAuthentication().getDetails();
+            if (details instanceof Map) {
```
adyach: do you think we could avoid 2 additional inference of type here? it is gonna be done on a lot requests, but does it really necessary?

rcillo: That sounds like a good idea. I'll give it a try.

src/main/java/org/zalando/nakadi/security/ClientResolver.java
```diff
@@ -15,9 +15,8 @@
 import org.zalando.nakadi.util.FeatureToggleService;
 
 import java.security.Principal;
-import java.util.Collections;
+import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 
 import static org.zalando.nakadi.config.SecuritySettings.AuthMode.OFF;
```
adyach: could you please remove the injection of unused FeatureToggleService? thank you

src/main/java/org/zalando/nakadi/security/ClientResolver.java
rcillo: What do you think about a5c3ded

rcillo: What do you think about a5c3ded

adyach:
👍

rcillo:
deploy validation please

adyach:
👍

rcillo:
👍

rcillo:
deploy validation please

src/main/java/org/zalando/nakadi/security/ClientResolver.java
rcillo: Just for the record, it was removed 👍

rcillo:
👍

adyach:
👍

rcillo:
👍

rcillo:
👍

adyach:
👍

PR #829, opened 2018-01-19
ARUHA-1452 Allow migration from EmptySchema to ObjectSchema and back.

antban:
[ARUHA-1452: Not possible to add "properties" object to schema when updating ET]

The problem with the schema library is that EmptySchema is not empty ObjectSchema (as it may seem).
In order to be able to treat it in the same way, emptySchema is casted to the actual type that is being checked.

### DIFF
--- a/src/main/java/org/zalando/nakadi/validation/schema/diff/SchemaDiff.java
+++ b/src/main/java/org/zalando/nakadi/validation/schema/diff/SchemaDiff.java
@@ -50,9 +50,6 @@ public class SchemaDiff {
             if (originalIn instanceof EmptySchema && updateIn instanceof ObjectSchema) {
                 original = replaceWithEmptyObjectSchema(originalIn);
                 update = updateIn;
-            } else if (originalIn instanceof ObjectSchema && updateIn instanceof EmptySchema) {
-                original = originalIn;
-                update = replaceWithEmptyObjectSchema(updateIn);
             } else {
                 state.addChange(TYPE_CHANGED);
                 return;
--- a/src/main/java/org/zalando/nakadi/validation/schema/diff/SchemaDiff.java
+++ b/src/main/java/org/zalando/nakadi/validation/schema/diff/SchemaDiff.java
@@ -2,6 +2,7 @@ package org.zalando.nakadi.validation.schema.diff;
 
 import org.everit.json.schema.ArraySchema;
 import org.everit.json.schema.CombinedSchema;
+import org.everit.json.schema.EmptySchema;
 import org.everit.json.schema.EnumSchema;
 import org.everit.json.schema.NumberSchema;
 import org.everit.json.schema.ObjectSchema;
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/nakadi/validation/schema/diff/SchemaDiff.java
```diff
@@ -29,25 +30,39 @@
...
-            state.addChange(TYPE_CHANGED);
-            return;
+        final Schema original;
+        final Schema update;
+        if (!originalIn.getClass().equals(updateIn.getClass())) {
+            // Tricky part. EmptySchema is the same as an empty ObjectSchema.
+            if (originalIn instanceof EmptySchema && updateIn instanceof ObjectSchema) {
+                original = replaceWithEmptyObjectSchema(originalIn);
+                update = updateIn;
+            } else if (originalIn instanceof ObjectSchema && updateIn instanceof EmptySchema) {
```
rcillo: Is this correct? If I understood it correctly, this would allow changing:

```javascript
{
  "properties": {
    "foo": {
      "type":"string"
    }
  }
}
```

to

```javascript
{
}
```

The first schema does not match the event `[]` but the later does.

I think that lines 48 to 52 are correct, but I'm not sure about 53 to 55.

antban: @rcillo Yes, this is correct. The purpose of this 2 steps are to replace EmptySchema with empty ObjectSchema, later it will call usual process of validation. 
In the case you pointed it will actually compare 2 schemas: 
```
{
  "properties": {
    "foo": {
      "type":"string"
    }
  }
}
```
and 
```
{
  "properties": {}
}
```
And the possibility of this change is regulated by ObjectSchema checks.

rcillo: @antban ok, now I got it. So in this case it would reject the change because the user would have removed one key.

But then I think there is another problem. It would accept the change:

```javascript
{
  "properties": {}
}
```

To

```javascript
{}
```

This change should not be allowed because the second schema accepts and array event `[]` as valid but the former doesn't. 

Makes sense?

antban: @rcillo Yep, makes sense. Changed this behavior: https://github.com/zalando/nakadi/pull/829/commits/91431cbe835f3bb433b05ff21ea4043745784f7f

rcillo:
👍

antban:
:+1:

antban:
deploy validation please

src/test/java/org/zalando/nakadi/validation/schema/SchemaDiffTest.java
```diff
@@ -57,4 +58,13 @@ public void testRecursiveCheck() throws IOException {
...
+
+    @Test
+    public void testSchemaAddsProperties() {
+        final Schema first = SchemaLoader.load(new JSONObject("{}"));
+
+        final Schema second = SchemaLoader.load(new JSONObject("{\"properties\": {}}"));
+        final List<SchemaChange> changes = service.collectChanges(first, second);
+        Assert.assertTrue(changes.isEmpty());
+    }
 }
```
adyach: No new line

antban: yes

antban:
:+1:

v-stepanov:
👍

PR #830, opened 2018-01-23
Aruha 1462: Feature toggle for database modifying operations

lmontrieux:
# ARUHA-1462: Feature toggle for database modifying operations

## Description
This PR adds a new feature flag: DISABLE_DB_WRITE_OPERATIONS
This flag disables all UPDATE, CREATE, and DELETE queries to the database, making it read-only. This is useful when doing database migrations, where we need Nakadi to connect to a read-only slave.

This PR also includes two refactorings:
- instances now react immediately to feature being toggled, using curator's `PathChildrenCache`
- Nakadi now sets features to default values at startup if instructed to do so. This allows to switch features on and off, both in local dev environment, and in acceptance tests.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
If option `nakadi.featureToggle.default` is true (which is the new default), Nakadi will set the features to their default value. Be careful when deploying!


src/main/java/org/zalando/nakadi/controller/ExceptionHandling.java
```diff
@@ -156,4 +157,12 @@ private String generateErrorTraceId() {
         return Responses.create(Response.Status.SERVICE_UNAVAILABLE, exception.getMessage(), request);
     }
 
+    @ExceptionHandler(DbWriteOperationsBlockedException.class)
+    public ResponseEntity<Problem> handleDbWriteOperationsBlockedException(
+            final DbWriteOperationsBlockedException exception, final NativeWebRequest request) {
+        LOG.error(exception.getMessage());
```
v-stepanov: I think it can be logged with WARN as this is situation that we create ourselves and we are aware of it, so it's not really an error.

v-stepanov:
@lmontrieux But in fact this is 2 in 1: feature-toggle refactoring + db-write feature toggle, right?

v-stepanov:
@lmontrieux 
I think at least two places that modify DB are missing in this PR:
1) Creating a timeline
2) Timeline cleanup (TimelineCleanupJob)

src/main/java/org/zalando/nakadi/config/RepositoriesConfig.java
```diff
@@ -13,28 +15,37 @@
...
+            final ZooKeeperHolder zooKeeperHolder, final FeaturesConfig featuresConfig) {
+        final FeatureToggleService featureToggleService = new FeatureToggleServiceZk(zooKeeperHolder);
         if (forceDefault) {
-            return new FeatureToggleServiceDefault();
-        } else {
-            return new FeatureToggleServiceZk(zooKeeperHolder);
+            final Set<String> features = featuresConfig.getDefaultFeatures().keySet();
+            for (final String feature: features) {
+                LOG.info("Setting feature {} to {}", feature, featuresConfig.getDefault(feature));
+                featureToggleService.setFeature(
```
v-stepanov: But that will actually mean that it will populate zookeeper with default values. Why have you decided to change this behavior?

src/main/java/org/zalando/nakadi/Application.java
```diff
@@ -2,10 +2,13 @@
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.ComponentScan;
+import org.zalando.nakadi.config.FeaturesConfig;
 
 @SpringBootApplication
 @ComponentScan(basePackages = {"org.zalando.nakadi", "org.zalando.nakadi.config"})
+@EnableConfigurationProperties(FeaturesConfig.class)
```
adyach: Usually, it is in configuration, where you use it, in this case I would put on RepositoriesConfig and moved FeaturesConfig into RepositoriesConfig, it will be more localised to the class where it is used. What do you think?

src/main/java/org/zalando/nakadi/config/RepositoriesConfig.java
lmontrieux: There are 2 reasons:
- when running Nakadi locally, I want to be able to toggle features on and off, to test stuff
- when running acceptance tests, I want to test the effect of a toggle on an endpoint

The previous behaviour had the toggles set at startup, and you had to restart Nakadi to change the configuration, which was in 2 different classes (all the features in one class are enabled, except those that are specifically disabled in another class).

src/main/java/org/zalando/nakadi/service/AdminService.java
```diff
@@ -38,6 +42,10 @@ public AdminService(final AuthorizationDbRepository authorizationDbRepository,
     }
 
     public void updateAdmins(final List<Permission> newAdmins) {
```
adyach: throws `DbWriteOperationsBlockedException`

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -123,6 +123,10 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
 
     public void create(final EventTypeBase eventType) throws TopicCreationException, InternalNakadiException,
```
adyach: throws `DbWriteOperationsBlockedException`
Same for other methods in the class

src/main/resources/application.yml
```diff
@@ -132,6 +145,19 @@ nakadi:
       commitTimeout: 5 # seconds
   subscription:
     maxPartitions: 30
+  features.defaultFeatures:
```
adyach: do we really need this duplication? I understand the purpose, make it visible for other.

src/main/java/org/zalando/nakadi/service/FeatureToggleServiceZk.java
```diff
@@ -1,57 +1,55 @@
...
+        } catch (IOException e) {
+            LOG.error("Could not close features cache", e);
+        }
     }
 
     public boolean isFeatureEnabled(final Feature feature) {
         try {
-            return cache.get(feature);
-        } catch (final ExecutionException e) {
+            return featuresCache.getCurrentData(PREFIX + "/" + feature.getId()) != null;
```
adyach: I thought about create the path beforehand, since there is limit amount of paths, but from other side it could be optimised by jvm

src/main/java/org/zalando/nakadi/config/RepositoriesConfig.java
v-stepanov: ok, makes sense

adyach:
@lmontrieux there are also `TimelineService` and `SubscriptionService`, which access database on write, do you think we should protect them as well?

src/main/java/org/zalando/nakadi/config/RepositoriesConfig.java
```diff
@@ -13,28 +15,37 @@
...
-            final ZooKeeperHolder zooKeeperHolder) {
+            final ZooKeeperHolder zooKeeperHolder, final FeaturesConfig featuresConfig) {
+        final FeatureToggleService featureToggleService = new FeatureToggleServiceZk(zooKeeperHolder);
         if (forceDefault) {
-            return new FeatureToggleServiceDefault();
-        } else {
-            return new FeatureToggleServiceZk(zooKeeperHolder);
+            final Set<String> features = featuresConfig.getDefaultFeatures().keySet();
+            for (final String feature: features) {
+                LOG.info("Setting feature {} to {}", feature, featuresConfig.getDefault(feature));
```
adyach: hard to get the naming `getDefault`, what do you think to rename it to `isEnabled`, `getDefaultState` etc..

src/main/java/org/zalando/nakadi/config/RepositoriesConfig.java
```diff
@@ -13,28 +15,37 @@
...
     public FeatureToggleService featureToggleService(
             @Value("${nakadi.featureToggle.default}") final boolean forceDefault,
-            final ZooKeeperHolder zooKeeperHolder) {
+            final ZooKeeperHolder zooKeeperHolder, final FeaturesConfig featuresConfig) {
+        final FeatureToggleService featureToggleService = new FeatureToggleServiceZk(zooKeeperHolder);
         if (forceDefault) {
-            return new FeatureToggleServiceDefault();
-        } else {
-            return new FeatureToggleServiceZk(zooKeeperHolder);
+            final Set<String> features = featuresConfig.getDefaultFeatures().keySet();
```
adyach: you could incorporate it inside FeaturesConfig, and give access to data it requires from outside

src/main/java/org/zalando/nakadi/service/FeatureToggleServiceZk.java
```diff
@@ -1,57 +1,55 @@
...
-                    public Boolean load(final Feature key) throws Exception {
-                        return isFeatureEnabledInZk(key);
-                    }
-                });
-
+    @PreDestroy
+    public void cleanUp() {
+        try {
+            this.featuresCache.close();
+        } catch (IOException e) {
```
v-stepanov: final
200x200

src/main/resources/application.yml
adyach: what is about this one ?

adyach:
👍

adyach:
LGTM, except there is conflict and travis build is not passing

src/main/java/org/zalando/nakadi/service/FeatureToggleServiceZk.java
lmontrieux: Fixed

src/main/java/org/zalando/nakadi/config/RepositoriesConfig.java
lmontrieux: Good point, exposing internals is ugly

src/main/java/org/zalando/nakadi/config/RepositoriesConfig.java
lmontrieux: I like `getDefaultState`

src/main/java/org/zalando/nakadi/service/FeatureToggleServiceZk.java
lmontrieux: Doesn't take much to create the paths beforehand, so I did that

src/main/java/org/zalando/nakadi/service/EventTypeService.java
lmontrieux: Fixed

src/main/java/org/zalando/nakadi/service/AdminService.java
lmontrieux: Fixed

src/main/java/org/zalando/nakadi/Application.java
lmontrieux: You are right, I moved it to RepositoriesConfig

src/main/java/org/zalando/nakadi/service/FeatureToggleServiceZk.java
```diff
@@ -1,57 +1,62 @@
...
         this.zkHolder = zkHolder;
+        try {
+            this.featuresCache = new PathChildrenCache(zkHolder.get(), PREFIX, false);
+            this.featuresCache.start();
+        } catch (final Exception e) {
+            LOG.error(e.getMessage(), e);
+        }
+        featurePaths = new HashMap<>(Feature.values().length);
+        for (final Feature feature: Feature.values()) {
+            featurePaths.put(feature, PREFIX + "/" + feature.getId());
```
v-stepanov: What is the idea behind this map? To say the truth I don't understand why we need it.

lmontrieux: @adyach suggested that, since there are a small number of features, we build a map of paths at startup, in order to speed up lookups at runtime.

v-stepanov: @lmontrieux @adyach 
To say the truth I don't know what is faster to concat 3 strings or to make a hash and lookup element in a hashmap; for me both of these operations sound like nothing. But I know which code is simpler and doesn't confuse :) 
(BTW there is a helper method from Zookeeper ZKPaths.makePath(s1, s2) which uses StringBuilder inside)

adyach: @v-stepanov suggestion was to not concatenate strings "here and there". Since this code path is heavily used, pre creating them before the usage should be beneficial. And I have no idea if JVM will optimise it for us. I do not see any complexity in Lionel's approach.

v-stepanov: @adyach Well, it will be beneficial only if it is faster. If it is slower - it will not :)
I don't say that current code is complex I say that it's more complicated than the one without this pre-created map.
I think it's a good example of premature optimization.

v-stepanov:
👍

src/main/java/org/zalando/nakadi/service/FeatureToggleServiceZk.java
v-stepanov: It's good for me.

lmontrieux:
👍

lmontrieux:
Deploy validation please

adyach:
👍

adyach:
👍

lmontrieux:
👍

lmontrieux:
Deploy validation please

PR #834, opened 2018-01-29
ARUHA-1500: API design for choosing partition to read from subscription

v-stepanov:
# API design for specifying partitions when reading from a subscription

> Zalando ticket : ARUHA-1500 (only if appropriate)

## Description
This is an API design for specifying the partitions to read from a subscription. I checked different possibilities but according to our API guidelines the best way to do that is [GET with body](http://zalando.github.io/restful-api-guidelines/#get-with-body). If there are some alternative suggestions - feel free to suggest.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
After the team approves - this should be reviewed by the API guild.


docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1178,15 +1178,83 @@ paths:
...
+    post:
+      tags:
+        - subscription-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        GET with body.
+        Starts a new stream for reading events from this subscription. Using this endpoint it is possible to specify
+        which partitions of subscription to read in this stream. If at least one of these partitions is already
+        assigned to another stream - the request will fail. It is possible to specify only the partitions that belong
```
lmontrieux: Do you think it would be possible that, for the same subscription, some consumers use this endpoint, and others use the rebalanced one? If that is the case, should it be that auto-rebalance consumers will be rebalanced if this calls arrives?
If both endpoints cannot be used simultaneously in the same subscription, then I think it should be stated here.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1178,15 +1178,83 @@ paths:
...
+        - name: partitions
+          in: body
+          schema:
+            type: object
+            properties:
+              items:
+                description: |
+                  List of partitions to read from in this stream.
+                  If not specified or if `items` array is empty - Nakadi will automatically choose partitions to assign
+                  to this stream.
```
lmontrieux: How? And how many?

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2718,6 +2807,17 @@ parameters:
     required: false
     default: 0
 
+  MaxUncommittedEvents:
+    name: max_uncommitted_events
+    in: query
+    description: |
+      The amount of uncommitted events Nakadi will stream before pausing the stream. When in paused
```
lmontrieux: the maximum amount

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: Yes, I already thought about that problem. There are different ways to solve it. The easiest way is to forbid using both endpoints for the same subscription. 

The other option is create a new notion like assignment-type: hard-assign and soft-assign. When you specify which partitions to stream - this is hard-assign, if not - this is soft assign. The hard-assigned partitions can't be stolen, the soft-assigned partitions can be stolen. So when you connect without specifying which partitions to read - Nakadi does a rebalance of soft-assigned partitions.

v-stepanov: Maybe we can discuss it on a small meeting with a team.

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: I thought that this is the actual number that will be sent before pause. Is there the usecase when Nakadi will pause before `max_uncommitted_events` is reached?

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: I provided some ideas in the comment above. Probably we need to discuss it with team to come to an agreement regarding this point.

docs/_data/nakadi-event-bus-api.yaml
lmontrieux: What if there are no more events to send?

v-stepanov: Ok, you win :)
https://github.com/zalando/nakadi/pull/834/commits/ea4706bee6c92e524ea4f7b3cd71df03ceea9975

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: As we have decided we will go with the solution to forbid using hybrid mode. Also, when creating subscription the user should specify which balance mode he wants to use. I applied appropriate changes to API already.

SergKam: We know that moving from one implementation to another (blue/green) deployment is hard already. We making even harder because we are forcing consumers to create a new subscription which will have their own state(offsets).
I this setup i will be hard for the teams to switch to this new model without significant downtime and possible errors with the offset calculation leading to lost events.
For me for example, it is easy to understand how nakadi rebalance the partitions that were not requested directly.
At least, this value should be editable without deleting subscription. So users can stop their old consumers and start new from the same offsets.

rcillo: @SergKam this feature is meant a complement for the migration from low-level to high-level API. So clients using this new extension are not expected to migrate any subscriptions, since they are not using them at the moment (they are low-level users).

For the only user that is going to "migrate" from "auto-assigned" to "self-assigned" (Diga) they already confirmed that it will not be a "blue/green switch" but they are building a completely new application that will use new subscriptions.

That said, the "blue/green" deployment is a hypothetical requirement and that's why we are using this more straightforward approach. But it doesn't mean that in the future we might allow such migration, if needed.

v-stepanov: @SergKam I think we don't expect people to migrate from auto-balancing to custom-balancing. Auto-balancing is the way to go in 99% of cases.
This custom-balancing feature is more for those who still use Low Level API and is missing this functionality in Subscription API.

SergKam: @rcillo If we designing our API only for Diga and Low-level API Users than yes. I agree and commit.
But for me, it looks very short-sighted. Why do you think Diga needs it? This feature significantly simplifies client architecture. They can aggregate and compact data inside consumers without complex orchestration. I think many many users will use this feature as soon as it will be available and for them migration is important.

v-stepanov: @SergKam I think using custom balancing is actually anti-pattern of using Nakadi Subscription API.
Data Lake needs it because they want to store data by partition - the same way as it was stored in Nakadi, that's why it's very important for them to keep specific partitions streams on specific instances.

And this feature doesn't simplify client architecture because you will have to write additional code that balances partitions among your clients/instances and makes sure that all partitions are consumed etc. So it actually makes client architecture more complicated. 
So I don't think that we will promote this feature a lot. It's a specific feature for those who really need it.

lmontrieux: @v-stepanov one of the issues that are bound to happen is that people will use this feature and not notice when repartitioning happens, and then they will lose data. Do you think it makes sense to add a warning about that in the documentation?

v-stepanov: @lmontrieux I added some info regarding that:
https://github.com/zalando/nakadi/pull/834/commits/6f8d27dbbb4cbb2d8b8b438088a7c01863a37934

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,158 @@ paths:
             $ref: '#/definitions/Problem'
         '409':
           description: |
-            Conflict. There are no empty slots for this subscriptions. The amount of consumers for this subscription
+            Conflict. There are several possible reasond for receiving this status code:
```
rcillo: resons

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,158 @@ paths:
...
+
+            - There are no empty slots for this subscriptions. The amount of consumers for this subscription
+            already equals the maximal value - the total amount of this subscription partitions.
```
rcillo: maximum

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,158 @@ paths:
...
+
+            - Resetting of subscription cursors request is still in the progress.
+
+            - This endpoint was called for a subscription with balance_mode set to 'custom'.
```
rcillo: Are you going to keep `custom` mode?

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,158 @@ paths:
...
+      tags:
+        - subscription-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        GET with body.
+        Starts a new stream for reading events from this subscription. Using this endpoint it is possible to specify
+        which partitions of subscription to read in this stream.
+
+        It is possible to specify partition that is already consumed by another stream, in this case there are two
```
rcillo: This is what I was afraid of. It's very hard to understand.

docs/_data/nakadi-event-bus-api.yaml
lmontrieux: total number of partitions in this subscription

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,158 @@ paths:
             $ref: '#/definitions/Problem'
         '409':
           description: |
-            Conflict. There are no empty slots for this subscriptions. The amount of consumers for this subscription
+            Conflict. There are several possible reasond for receiving this status code:
+
+            - There are no empty slots for this subscriptions. The amount of consumers for this subscription
+            already equals the maximal value - the total amount of this subscription partitions.
+
+            - Resetting of subscription cursors request is still in the progress.
```
lmontrieux: - Request to reset subscription cursors is still in progress.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,158 @@ paths:
...
+
+    post:
+      tags:
+        - subscription-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        GET with body.
+        Starts a new stream for reading events from this subscription. Using this endpoint it is possible to specify
+        which partitions of subscription to read in this stream.
```
lmontrieux: which partitions to read in this stream.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,158 @@ paths:
...
+
+        It is possible to specify partition that is already consumed by another stream, in this case there are two
+        possible outcomes. If the partition was assigned to another stream as a result of automatic assignment - it will
+        be possible to "steal" it. But if the partition is assigned to another stream because it was directly requested
+        by it - then it will not be possible to transfer this partition to current stream - such request will fail.
```
lmontrieux: It is possible to specify a partition that is already consumed by another stream. In this case, there can be two outcomes, depending on how the current assignment was obtained. (1) If the partition is already assigned to another stream through automatic assignment, it will be possible to "steal" it, and the request will succeed. (2) If the partition is already assigned to another stream because of a request to stream from this partition specifically, then it will not be possible to "steal" the partition, and the request will fail.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,156 @@ paths:
...
+
+        It is possible to specify only the partitions that belong to this subscription.
```
lmontrieux: It is only possible to specify partitions that belong to this subscription.

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: fixed

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: fixed

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: no sorry, just forgot to remove it. I fixed it now

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,156 @@ paths:
...
+
+        When using this endpoint the balancing of partitions becomes user responsibility. User should track that all
+        partitions are consumed (if this is the case to consume all partitions) and also react appropriately when the
+        number of partitions of subscription's event-type changes due to repartitioning.
+
+        The rest of behavior is identical to endpoint GET /subscriptions/{subscription_id}/events.
```
lmontrieux: The rest of the behaviour

lmontrieux: Sorry, behavior. 'behaviour' is British English

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,156 @@ paths:
...
+                description: |
+                  List of partitions to read from in this stream. If absent or empty - then the partitions will be
+                  automatically assigned by Nakadi in the same way as when using endpoint GET
+                  /subscriptions/{subscription_id}/events
+                type: array
+                items:
+                  $ref: '#/definitions/EventTypePartition'
+              max_uncommitted_events:
+                description: |
+                  The maximum amount of uncommitted events Nakadi will stream before pausing the stream. When in paused
```
lmontrieux: The maximum number

lmontrieux: events that Nakadi will stream

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: you also haven't defined "steal" :-P

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,156 @@ paths:
...
+                  state and commit comes - the stream will resume. Minimal value is 1.
```
lmontrieux: you should use the minimum: 1 property

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,156 @@ paths:
...
+        '409':
+          description: |
+            Conflict. There are several possible reasons for receiving this status code:
+            
+            - There are no empty slots for this subscriptions. The amount of consumers for this subscription
+            already equals the maximum value - the total amount of this subscription partitions.
 
-            This status code is also returned in the case of resetting subscription cursors request still in the
-            progress.
+            - Resetting of subscription cursors request is still in the progress.
```
lmontrieux: Same as above

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2718,6 +2868,17 @@ parameters:
     required: false
     default: 0
 
+  MaxUncommittedEvents:
+    name: max_uncommitted_events
+    in: query
+    description: |
+      The maximum amount of uncommitted events Nakadi will stream before pausing the stream. When in paused
```
lmontrieux: maximum number

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2538,6 +2682,12 @@ definitions:
             stream_id:
               type: string
               description: the id of the stream that consumes data from this partition
+            assign_type:
+              type: string
+              description: |
+                - `hard`: partition can't be transferred to another stream until the stream is closed;
+                - `soft`: partition can be transferred to another stream in a case of rabalance or if another stream
```
lmontrieux: in case of rebalance, or if (3 changes)

docs/_data/nakadi-event-bus-api.yaml
lmontrieux: Yes, I saw @SergKam 's comment after I wrote this :)

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: sorry, didn't get it

lmontrieux: in `a` case of r`a`balance or if-> in case of r`e`balance`,` or if

v-stepanov:
@lmontrieux @rcillo @SergKam I think I addressed all your comments

SergKam:
:+1:

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,159 @@ paths:
...
+
+        It is only possible to specify partitions that belong to this subscription.
+
+        When using this endpoint, the balancing of partitions is the user's responsibility. It is also the user's
+        responsibility to detect, and react to, changes in the number of partitions in the subscription (following the
+        re-partitioning of an event type). Using the GET /subscriptions/{subscription_id}/stats endpoint can be helpful.
+
+        The rest of the behavior is identical to endpoint GET /subscriptions/{subscription_id}/events.
+      parameters:
+        - name: streamParameters
```
tkrop: @v-stepanov Hi, @tfrauenstein and I have talked about this API change. We would suggest to keep the `get-with-body` as symmetric to the `get` as possible. Therefore, we would suggest to reuse the query parameter of the `get` instead of appending every parameter to the body. We think that this would be more to the expectations of the consumers and help them easier to adapt to the new endpoint.

antban: @tkrop 
In terms of adoption - I can not fully agree with you. One will have to change only:
 - method to call
 - append non-empty body with specific parameter
 - check that library that is being used for http interaction will actually put parameters to a query string, instead of post body (or will not miss it, because post body is already there).

Probably in most cases it will work without a lot of modifications (but because of uncertainty will not simplify onboarding). 

From the other hand it will produce api, that is not well-looking, and we will not be able to get rid of it. Not well looking is because all the parameters that are passed to streaming endpoint are on the same level of hierarchy - it is all streaming parameters. The reason of putting them to different places is unclear.

Also, in case if we are talking about simplification, we should (of course) think about unification - only one of this methods will be used by particular client (I hope), and for this particular client using unified way of specifying parameters is actually the simplest approach, meaning simplification of onboarding.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,167 @@ paths:
...
+      description: |
+        GET with body.
+
+        Starts a new stream for reading events from this subscription. The minimal consumption unit is a partition, so
+        it is possible to start as many streams as the total number of partitions in event-types of this subscription.
+        The position of the consumption is managed by Nakadi. The client is required to commit the cursors he gets in
+        a stream.
+
+        If you create a stream without specifying the partitions to read from - Nakadi will automatically assign
+        partitions to this new stream. By default Nakadi distributes partitions among clients trying to give the equal
```
rcillo: `s/the equal amount/an equal number/`

rcillo:
👍 for me it's clear how it behaves. I also reconsidered our discussion about marking `GET` as deprecated. I know that with the addition of `POST` we have some redundancy, but deprecating `GET` would require the entire company to redeploy their consumers. Even if we are still support `GET` till end-of-life, there would riots. So, I think it's fine to maintain both.

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: fixed https://github.com/zalando/nakadi/pull/834/commits/db3635b998900d5554a1ba7c5f8e704fb3dff20d

SergKam:
@rcillo I think it's fine to deprecate GET. We are not "retiring" it anytime soon. "Deprecate" only means do not use it for any new project. http://zalando.github.io/restful-api-guidelines/#191 we do not force anyone (yet :) ) to redeploy. And another important point is that we don't need to change all applications, almost all are using some limited number of clients.

antban:
:+1:

v-stepanov:
I will send a call for API review by API guild.

docs/_data/nakadi-event-bus-api.yaml
etorreborre: > Data Lake needs it because they want to store data by partition - the same way as it was stored in Nakadi, that's why it's very important for them to keep specific partitions streams on specific instances.

I was actually looking a place where the rationale for this feature is explained. Is it anywhere aside from this comment?

v-stepanov: @etorreborre Good point. I added motivation to description of this PR
> Motivation for this feature is that we are deprecating Low Level API and the only feature that exists in Low Level API and is missing in Subscription API is specifying from which partition(s) to read.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1221,17 +1213,167 @@ paths:
...
+            type: object
+            properties:
+              partitions:
+                description: |
+                  List of partitions to read from in this stream. If absent or empty - then the partitions will be
+                  automatically assigned by Nakadi.
+                type: array
+                items:
+                  $ref: '#/definitions/EventTypePartition'
+              max_uncommitted_events:
```
a1exsh: Why not refer to `MaxUncommittedEvents`?

v-stepanov: @a1exsh Is it possible to refer to a parameter from object property? I could not find any example of that in swagger spec.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2538,6 +2693,12 @@ definitions:
             stream_id:
               type: string
               description: the id of the stream that consumes data from this partition
+            assign_type:
```
a1exsh: Did you mean `assignment_type`?

v-stepanov: Good point. Fixed https://github.com/zalando/nakadi/pull/834/commits/c02b2d1d92264a318695c6bcb2aa90c610d98f43

v-stepanov:
👍

lmontrieux:
👍

PR #839, opened 2018-02-14
ARUHA-1085 Concept for archiving all event types

lmontrieux:
# Concept for archiving all event types

> Zalando ticket : ARUHA-1085

## Description
This is a concept to allow some applications to archive events from all
event types, regardless of the event types' authorization policy.

### DIFF
--- a/docs/_concepts/archive-authz.md
+++ b/docs/_concepts/archive-authz.md
@@ -30,4 +30,4 @@ To make it clear that archiving applications are able to read from all event typ
 ### List of tickets (to be created)
 
 - Add check for archiving applications when consumers connect (HiLA only as LoLA is being deprecated);
-- Add warning to response body when new event types are created and existing event types are updated.
+- Add warning to response headers when new event types are created and existing event types are updated.
--- a/docs/_concepts/archive-authz.md
+++ b/docs/_concepts/archive-authz.md
@@ -4,7 +4,7 @@
 
 Organisations may wish to archive all the data that passes through Nakadi, for example in a data lake, using one or several applications that consume all events from all event types. Our experience has been that it is cumbersome, confusing, and error-prone, to ensure that event types that restrict READ allow the archiving application(s) to read their events. Changes in the archiving application also require modification of the authorization sections of all protected event types, which takes time and is confusing for event type owners.
 
-This concept includes 2 possible solutions. Once a solution is chosen, the concept will be updated accordingly.
+This concept outlines a temporary solution to allow a set of applications to read from all event types. The solution described here represents technical debt that will have to be repaid later, which will involve a re-design of authorization in Nakadi.
 
 ## Requirements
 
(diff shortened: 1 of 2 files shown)


docs/_concepts/archive-authz.md
```diff
@@ -0,0 +1,96 @@
...
+
+- Relatively easy to implement: no changes to the authorization logic;
+- Makes it obvious which applications can read from the event type.
+
+### Cons
+
+- Confusing for users: they are not able to remove some applications from the readers, and there is nothing that marks them as archiving applications;
+- Updating the list of archiving applications could be cumbersome, as it may require updating a lot of event types.
+
+## Solution 2: a new resource for archiving applications
```
v-stepanov: I'm strongly for this approach 👍

docs/_concepts/archive-authz.md
```diff
@@ -0,0 +1,96 @@
...
+
+In this solution, we create a new resource, `archive`, for which applications can only get READ permissions (write and admin must be empty).
+
+When a client tries to read from an event type with an authorization section, if the client is not in the list of clients authorized to read from the event type, Nakadi will check if the client is in the list of archiving applications.
+
+Nakadi will have to know the list of archiving applications; to do this, a new endpoint will be created:
+```
+POST /settings/resources/archivers
```
v-stepanov: What will be posted here? How the structure will look like?

docs/_concepts/archive-authz.md
```diff
@@ -0,0 +1,96 @@
...
+```
+and its complement:
+```
+GET /settings/resources/archivers
+```
+
+These endpoints will be accessible to administrators with ADMIN rights only.
```
v-stepanov: ADMIN-read, ADMIN-write or ADMIN-admin ?

lmontrieux: ADMIN-ADMIN

docs/_concepts/archive-authz.md
lmontrieux: same as /settings/admins, but only for read (I'll update the concept)

docs/_concepts/archive-authz.md
```diff
@@ -0,0 +1,96 @@
...
+
+When a client tries to read from an event type with an authorization section, if the client is not in the list of clients authorized to read from the event type, Nakadi will check if the client is in the list of archiving applications.
+
+Nakadi will have to know the list of archiving applications; to do this, a new endpoint will be created:
+```
+POST /settings/resources/archivers
+```
+and its complement:
+```
+GET /settings/resources/archivers
```
v-stepanov: for admins endpoint we don't have this "resources" path segment, it is just `/settings/admins` Do you think that will make sense to move admin endpoint also to resources?

lmontrieux: Yes, I think it makes sense to have all the resources in the same place

docs/_concepts/archive-authz.md
```diff
@@ -0,0 +1,96 @@
...
+- Makes it obvious which applications can read from the event type.
+
+### Cons
+
+- Confusing for users: they are not able to remove some applications from the readers, and there is nothing that marks them as archiving applications;
+- Updating the list of archiving applications could be cumbersome, as it may require updating a lot of event types.
+
+## Solution 2: a new resource for archiving applications
+
+In this solution, we create a new resource, `archive`, for which applications can only get READ permissions (write and admin must be empty).
```
SergKam: I don't like the 'archive' word as it will be then 'audit', or 'monitor' etc. I see the need for 'roles' here. We already have regular users, admins, power readers. Maybe we can unify them somehow. Like extending /admins endpoint to accept readers, writers, admins, and data_readers lists.
At least 'archive' should be like 'data_access' or 'global_readers' to not point to the function("why") but to the access pattern ("what").

lmontrieux:
Please add a thumbs up reaction to this message to vote for **solution 1**

lmontrieux:
Please add a thumbs up reaction to this message to vote for **solution 2**

docs/_concepts/archive-authz.md
v-stepanov: I don't know, for me the current Lionel's suggestion is pretty clear and it doesn't break the general idea that there are resources and for each resource there are three possible types of permission: read/write/admin. Introducing of new permission type "data_readers" will look weird.

SergKam: @v-stepanov yes, my idea was to introduce some kind of auth role-based approach. It can be done extending access levels or extending access to particular resources. I like Lionel's idea to put all roles/resources in one "folder" so it can be extended later if needed. The only problem is that "resources" name does not suggest anything about access control.

docs/_concepts/archive-authz.md
SergKam: I like Lionel's idea to put all roles/resources in one "folder" so it can be extended later if needed. The only problem is that "resources" name does not suggest anything about access control.
Another concern, I was writing clients for auth and I learn how difficult is to work with deeply nested structures. I would prefer to have a simple list of objects with fields like:
"""
{
data_type: "service",
data_value: "stups_nakadi"
resource: "admin",
access: "read"
}
"""
I am doing it in the client anyway. also it would simplify all CRUD operations.

v-stepanov: @SergKam I think the idea was to have a uniform json-structure for everything that involves authorization. We have this for event-types authorization and admins definition. I think it would be better to follow the common approach.

docs/_concepts/archive-authz.md
lmontrieux: @SergKam @v-stepanov I'm ok with renaming ('what' is indeed better than 'why' in our current authorization model), but I'm not sure about extending admins. Admins has 3 lists of permissions (read, write, admin), and something like 'global_readers' isn't really a permission.
What could work would be `all_data_access`, for example.

Another option, as @SergKam suggests, is to go for a RBAC model, but that would require more changes. We would have to define roles, and assign permissions to roles (a permission being the right to perform an operation on an object), and assign users (both employees and applications) to roles. In this case, we'd have something like:
`/settings/authz/roles/admin_read`
`/settings/authz/roles/admin_write`
`/settings/authz/roles/admin_all_event_types`
`/settings/authz/roles/all_data_access`
etc.
Current, per-event type authorization could be considered as individual roles, so for each protected event type (eg `etXXX`) there would be 3 roles: `etXXX_read`, `etXXX_write`, `etXXX_admin`. This is where the RBAC model breaks down, as there will be a lot of roles.

I think it is quite a bit more work, and I'm not convinced that we need that yet.

SergKam: But we will have a broken RBAC anyway but like:
`/settings/authz/admins/(read, write, admin)`
`/settings/authz/all_data_access/(read, write, admin)`
it doesn't look intuitive for me as it is our own reinvented wheel.

lmontrieux:
**Decision**: we will implement solution 2, but without an additional endpoint. Changes will have to be done directly to the database. This is because we want to deliver the ability for archiving applications to read all events from all event types very quickly. We accept the additional technical debt and will come back with another concept to pay it down.

Now I will update this PR to reflect our decision. Once approved I will get started with the implementation.

v-stepanov:
👍

docs/_concepts/archive-authz.md
```diff
@@ -0,0 +1,33 @@
...
+When a client tries to read from an event type with an authorization section, if the client is not in the list of clients authorized to read from the event type, Nakadi will check if the client is in the list of archiving applications.
+
+Nakadi will have to know the list of archiving applications; they will be added to the database, with the resource name 'all_data_access'
+
+To make it clear that archiving applications are able to read from all event types, a warning will be included to the response for each successful creation of a new event type, or update of an existing event type.
+
+### List of tickets (to be created)
+
+- Add check for archiving applications when consumers connect (HiLA only as LoLA is being deprecated);
+- Add warning to response body when new event types are created and existing event types are updated.
```
SergKam: It should go not in the body but in headers https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Warning

lmontrieux: d29fdc3

lmontrieux:
👍

SergKam:
:+1:

PR #842, opened 2018-02-15
ARUHA-1504 Global read access for all_data_access

lmontrieux:
# Allow READ access to applications with READ permission for all_data_access resource

> Zalando ticket : ARUHA-1504

## Description
If an application holds a READ permission for resource all_data_access, it will be able to read from all event types, regardless of their authorization policy.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
The list of subjects with permissions for all_data_access must be inserted directly into the DB.

### DIFF
--- a/src/main/java/org/zalando/nakadi/repository/db/AuthorizationDbRepository.java
+++ b/src/main/java/org/zalando/nakadi/repository/db/AuthorizationDbRepository.java
@@ -39,13 +39,14 @@ public class AuthorizationDbRepository extends AbstractDbRepository {
         return listAllFromResource(ALL_DATA_ACCESS_RESOURCE);
     }
 
-    private List<Permission> listAllFromResource(final String resource) {
+    private List<Permission> listAllFromResource(final String resource) throws RepositoryProblemException {
         final List<Permission> permissions;
         try {
             permissions = jdbcTemplate.query("SELECT * FROM zn_data.authorization WHERE az_resource=?",
                     new Object[] { resource }, permissionRowMapper);
         } catch (final DataAccessException e) {
-            throw new RepositoryProblemException("Error occurred when fetching permissions for ", e);
+            throw new RepositoryProblemException(
+                    String.format("Error occurred when fetching permissions for resource %s", resource), e);
         }
 
         return permissions;
--- a/src/main/java/org/zalando/nakadi/repository/db/AuthorizationDbRepository.java
+++ b/src/main/java/org/zalando/nakadi/repository/db/AuthorizationDbRepository.java
@@ -15,6 +15,9 @@ import org.zalando.nakadi.plugin.api.authz.AuthorizationService;
 
 import java.util.List;
 
+import static org.zalando.nakadi.domain.AdminResource.ADMIN_RESOURCE;
+import static org.zalando.nakadi.domain.AllDataAccessResource.ALL_DATA_ACCESS_RESOURCE;
+
 @DB
 @Repository
 public class AuthorizationDbRepository extends AbstractDbRepository {
@@ -29,15 +32,23 @@ public class AuthorizationDbRepository extends AbstractDbRepository {
     }
 
     public List<Permission> listAdmins() throws RepositoryProblemException {
-        final List<Permission> admins;
+        return listAllFromResource(ADMIN_RESOURCE);
+    }
+
+    public List<Permission> listAllDataAccess() throws RepositoryProblemException {
+        return listAllFromResource(ALL_DATA_ACCESS_RESOURCE);
+    }
+
+    private List<Permission> listAllFromResource(final String resource) {
+        final List<Permission> permissions;
         try {
-            admins = jdbcTemplate.query("SELECT * FROM zn_data.authorization WHERE az_resource='nakadi'",
-                    permissionRowMapper);
+            permissions = jdbcTemplate.query("SELECT * FROM zn_data.authorization WHERE az_resource=?",
+                    new Object[] { resource }, permissionRowMapper);
         } catch (final DataAccessException e) {
-            throw new RepositoryProblemException("Error occurred when fetching administrators", e);
+            throw new RepositoryProblemException("Error occurred when fetching permissions for ", e);
         }
 
-        return admins;
+        return permissions;
     }
 
     public void deletePermission(final Permission permission) {
--- a/src/main/java/org/zalando/nakadi/service/AdminService.java
+++ b/src/main/java/org/zalando/nakadi/service/AdminService.java
@@ -4,6 +4,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.zalando.nakadi.config.NakadiSettings;
 import org.zalando.nakadi.domain.AdminResource;
+import org.zalando.nakadi.domain.AllDataAccessResource;
 import org.zalando.nakadi.domain.Permission;
 import org.zalando.nakadi.domain.ResourceAuthorization;
 import org.zalando.nakadi.exceptions.UnableProcessException;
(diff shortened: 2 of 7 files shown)


src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -155,7 +155,9 @@ public void authorizeStreamRead(final EventType eventType) throws AccessDeniedEx
         final Resource resource = new EventTypeResource(eventType.getName(), eventType.getAuthorization());
         try {
             if (!authorizationService.isAuthorized(AuthorizationService.Operation.READ, resource)) {
-                throw new AccessDeniedException(AuthorizationService.Operation.READ, resource);
+                if (!adminService.hasAllDataAccess(AuthorizationService.Operation.READ)) {
```
SergKam: Maybe it is normal in Java, just curious, Why two IF instead of AND?

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -155,7 +155,9 @@ public void authorizeStreamRead(final EventType eventType) throws AccessDeniedEx
...
+                    throw new AccessDeniedException(AuthorizationService.Operation.READ, resource);
+                }
             }
         } catch (final PluginException e) {
             throw new ServiceTemporarilyUnavailableException("Error calling authorization plugin", e);
```
SergKam: Maybe it is normal in Java, just learning, again,
Why you Throw and Catch and Throw again. Instead one Throw, maybe complex like:
```
... e =new AccessDeniedException(AuthorizationService.Operation.READ, resource);
throw new ServiceTemporarilyUnavailableException("Error calling authorization plugin", e);
```

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
lmontrieux: The `&&` does short-circuit evaluation so actually it would be equivalent

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
lmontrieux: Here we catch PluginException, which is thrown from inside a plugin (so we may not necessarily know how it behaves). We catch it, then throw ServiceTemporarilyUnavailableException, so we don't expose implementation details to the caller. Or at least that's how I see it

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
lmontrieux: It would be nicer with a `&&`, you are right

lmontrieux: 7927b54

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
SergKam: Ah, I see, you are catching possible errors from authorizationService and adminService.

src/main/java/org/zalando/nakadi/repository/db/AuthorizationDbRepository.java
```diff
@@ -29,15 +32,23 @@ public AuthorizationDbRepository(final JdbcTemplate jdbcTemplate, final ObjectMa
...
     public List<Permission> listAdmins() throws RepositoryProblemException {
-        final List<Permission> admins;
+        return listAllFromResource(ADMIN_RESOURCE);
+    }
+
+    public List<Permission> listAllDataAccess() throws RepositoryProblemException {
+        return listAllFromResource(ALL_DATA_ACCESS_RESOURCE);
+    }
+
+    private List<Permission> listAllFromResource(final String resource) {
```
adyach: throws RepositoryProblemException

lmontrieux: 943f590

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -154,7 +154,8 @@ public void authorizeStreamRead(final EventType eventType) throws AccessDeniedEx
 
         final Resource resource = new EventTypeResource(eventType.getName(), eventType.getAuthorization());
         try {
-            if (!authorizationService.isAuthorized(AuthorizationService.Operation.READ, resource)) {
+            if (!authorizationService.isAuthorized(AuthorizationService.Operation.READ, resource)
```
v-stepanov: @lmontrieux But you have implemented it only for Low Level API and not for Subscription API. I thought you wanted to do vise versa.

v-stepanov: Ah, ok. You implemented it for both.

src/test/java/org/zalando/nakadi/controller/EventStreamControllerTest.java
```diff
@@ -466,6 +472,24 @@ public void testAccessDenied() throws Exception {
         assertThat(responseToString(responseBody), TestUtils.JSON_TEST_HELPER.matchesObject(expectedProblem));
     }
 
+    @Test
+    public void testAccessAllowedForAllDataAccess() throws Exception {
```
v-stepanov: This class test Low Level API stream. What about a test for Subscription API?

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
lmontrieux: I realised that implementing it here means that it will work for all endpoints that are protected with read permissions (for example, stats). It's already a hack solution and I'd rather not make it worse, I think it is fine to have it for LoLA as well, even though it is deprecated.

v-stepanov: Yes definitely. I think it doesn't make sense to spend additional effort to disable that for LoLA.

src/test/java/org/zalando/nakadi/controller/EventStreamControllerTest.java
lmontrieux: @v-stepanov Sorry, I'm not going to write a test for Subscription API. The reason is:
- to write a unit test means a nightmare of mocks (I spent too much time trying, it's still not working and there are a lot of mocks already)
- ideally we should have an acceptance test. I think this should be written elegantly when we have an endpoint to set applications that can read from all. I could populate the database with some values on startup... but that's also ugly.

What do you think?

v-stepanov: @lmontrieux 
Yes, probably it will be difficult to write AT checking authorization because we run ATs without any tokens. And unit test will have too much to mock. Ok, let's do it without a test.

src/main/java/org/zalando/nakadi/service/AdminService.java
```diff
@@ -62,6 +64,13 @@ public boolean isAdmin(final AuthorizationService.Operation operation) {
         return authorizationService.isAuthorized(operation, resource);
     }
 
+    public boolean hasAllDataAccess(final AuthorizationService.Operation operation) {
+        final List<Permission> permissions = authorizationDbRepository.listAllDataAccess();
```
v-stepanov: I don't remember: have we planned to have simple guava cache in this implementation or not?
I checked - currently reading events endpoint (both hila and lola) is called ~25 times/sec.

lmontrieux: Good point. Here is a very simple cache, where data expires after 1 minute, as we discussed: 625c2e2

v-stepanov:
👍

lmontrieux:
👍

lmontrieux:
Deploy validation please

lmontrieux:
👍

v-stepanov:
👍

PR #843, opened 2018-02-15
ARUHA-1136 Warning header for event type creation and update

lmontrieux:
# Warning header for event type creation and update

> Zalando ticket : ARUHA-1136

## Description
When creating or updating event types, it is possible to add a warning
header to the response. The purpose of this header is to warn users
that events in their event type may be read by archiving applications,
regardless of the authorization policy for the event type.

The warning header can be disabled by setting property
nakadi.authz.warnAllDataAccessMessage to an empty string.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG


src/main/java/org/zalando/nakadi/controller/EventTypeController.java
```diff
@@ -145,7 +154,11 @@ public EventTypeController(final EventTypeService eventTypeService,
         }
 
         eventTypeService.update(name, eventType);
-        return status(HttpStatus.OK).build();
+        final HttpHeaders headers = new HttpHeaders();
+        if (!nakadiSettings.getWarnAllDataAccessMessage().isEmpty()) {
+            headers.add(HttpHeaders.WARNING, nakadiSettings.getWarnAllDataAccessMessage());
```
SergKam: According to the standard, the warning header format should contain code and app name.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Warning
It can be added here or in senza files but I don't see it there either.

lmontrieux: I saw that page, but it seems that this format is specifically for cache-related warnings, isn't it? Do you think it applies here as well?

SergKam: No, we should use 299 code "Miscellaneous Warning | Same as 199(Arbitrary, non-specific warning), but indicating a persistent warning"
 standard very specific about it https://tools.ietf.org/html/rfc7234#section-5.5
```
Warnings can be used for other purposes, both cache-related and
   otherwise.  The use of a warning, rather than an error status code,
   distinguishes these responses from true failures.
```

lmontrieux: @SergKam dc56340

src/main/java/org/zalando/nakadi/controller/EventTypeController.java
```diff
@@ -157,6 +163,15 @@ public EventTypeController(final EventTypeService eventTypeService,
         return status(HttpStatus.OK).body(result.getValue());
     }
 
+    private HttpHeaders generateWarningHeaders() {
+        final HttpHeaders headers = new HttpHeaders();
+        if (!nakadiSettings.getWarnAllDataAccessMessage().isEmpty()) {
+            headers.add(HttpHeaders.WARNING,
+                    String.format("299 nakadi \"%s\"", nakadiSettings.getWarnAllDataAccessMessage()));
```
v-stepanov: what does it mean?

src/main/java/org/zalando/nakadi/config/NakadiSettings.java
```diff
@@ -21,6 +21,7 @@
     private final long eventMaxBytes;
     private final int maxSubscriptionPartitions;
     private final AuthorizationAttribute defaultAdmin;
+    private final String warnAllDataAccessMessage;
 
     @Autowired
     public NakadiSettings(@Value("${nakadi.topic.max.partitionNum}") final int maxTopicPartitionCount,
```
v-stepanov: Now I think this monster class was a big mistake.

src/main/java/org/zalando/nakadi/controller/EventTypeController.java
v-stepanov: Ok, I know now. Already read Sergey's comment.

src/test/java/org/zalando/nakadi/controller/EventTypeControllerTestCase.java
```diff
@@ -93,7 +93,7 @@ public void init() throws Exception {
 
         final NakadiSettings nakadiSettings = new NakadiSettings(0, 0, 0, TOPIC_RETENTION_TIME_MS, 0, 60,
                 NAKADI_POLL_TIMEOUT, NAKADI_SEND_TIMEOUT, 0, NAKADI_EVENT_MAX_BYTES,
-                NAKADI_SUBSCRIPTION_MAX_PARTITIONS, "service", "nakadi");
+                NAKADI_SUBSCRIPTION_MAX_PARTITIONS, "service", "nakadi", "");
```
v-stepanov: maybe you can set it here to some value and write a test in EventTypeControllerTest to check that header is actually set?

lmontrieux: @v-stepanov b09c4f3

v-stepanov:
👍

lmontrieux:
👍

lmontrieux:
👍

lmontrieux:
👍

adyach:
👍

PR #851, opened 2018-02-22
Results of profiling

adyach:
Results of profiling with https://github.com/jvm-profiling-tools/async-profiler

## Description
It is possible that this PR request will be never merged, once we reviewed it has to be tested.

The profiling was done in two different ways: local and on production environment. You can find flamegraphs attached to analyse it and conclude the decisions I made or maybe find more things, which I could misinterpret.
https://drive.google.com/file/d/1DvWqOFQr_7QC1aHYkP_PFdmicSBC_FvD/view?usp=sharing

It was observed that polling to Kafka takes a lot of CPU time `StreamingState.pollDataFromKafka`, which means it just using CPU without the need, because usually there is data for streaming from previous call, so we could tolerate it and wait a bit before polling the next time, so the scheduling of `StreamingState.pollDataFromKafka` was changed by adding very small timeout.

`KafkaCursor.toNakadiOffset` is in very hot path of the code and it uses a lot `String.format`, which could be replaced with the more efficient implementation of Apache Commons.

Nakadi consumes a lot of CPU decompressing and compressing request first time during HTTP request/response and the second time during requests to/from Kafka. It is not possible to replace or remove this functionality, but it is possible to improve the request path to Nakadi, because we use custom implementation on the level of filters, replacing it with Jetty `GzipHandler` level decompression it should improve the performance since it reuses some parts of the code on the container level.

The `RFC3339DateTimeValidator.validate` spends a lot of time on cpu and it is obvious, because it is a part of validation. We use regex to validate the time + `OffsetDateTime.parse(dateTime, ISO_OFFSET_DATE_TIME)` to make sure it is really correct time, I replaced it with only one regex it is definitely not completely final check, but it covers the existing tests, so I would like to test it in production as well.

It is noticed that we use logging request twice to track access and slos. It showed up in different parts of profiling and in general it could around 1% of total cpu time used. I decided to improve it anyway extending our logger with async functionality.

`BlacklistService` was noticed in the profiling, because it is called on every call to publish or consume. Actually, the reason for that is curator `TreeCache`, which is used for caching requests to Zookeeper, surprisingly it generates new array on every call to `TreeCache.getChildrenData`. `TreeCache` should be replaced with Zookeeper watchers to avoid this calls, which are around 0.2-0.4% of total time on the processor.

String.getBytes showed in very hot path during publishing, which could be replaced with String.length with the same result, but it is not possible I have tested uit manually, for some reason unicode symbols are not transformed to unicode code, otherwise we could reduce around 1% of time spent on CPU for that.

`PartitionData.getUnconfirmed` was noticed in the profiling, which uses `TreeMap` underneath to calculate cursors distance and it could be replaced with `CursorOperationsService.calculateDistance`, but this quite questionable, because `calculateDistance` usees timelenes and event type cache, so it could be tested a bi later.

Another good candidate is metrics endpoint, which showed up in profiling and consumed around 1.2% of CPU time, I think it is due to the nature of the huge amount of data we have to deserialize before sending requests.

There are things, which we are not able to fix since they are not in our code:
Validation of the events using third party library 
Compression of responses 
Compression/decompression of Kafka messages
Local tokeninfo service (~7-10% at CPU 70-90%)

Overall usage patterns I have since in the profiling looks not bad from my point of view, they are expected behaviour, with little things which I mentioned before. Another big part is usage of Spring could be additional cost we pay, it should be considered migrating from it to more low level implementation to have more control over the stream of data we send to clients if we really consider improving performance of Nakadi.


src/main/java/org/zalando/nakadi/validation/RFC3339DateTimeValidator.java
```diff
@@ -2,40 +2,30 @@
...
 
-    private final String errorMessage = "must be a valid date-time";
-
-    // Valid offsets are either Z or hh:mm. The format hh:mm:ss is not valid
-    // Pattern is used to catch situations that are not covered by {@link OffsetDateTime#parse(CharSequence)}
-    private final String dateTimeOffsetPattern = "^.*[Tt]\\d{2}:\\d{2}:\\d{2}.*([zZ]|([+-]\\d{2}:\\d{2}))$";
-    private final Pattern pattern = Pattern.compile(dateTimeOffsetPattern);
-    private final Optional<String> error = Optional.of(errorMessage);
+    private static final String DATE_TIME_OFFSET_PATTERN =
+            "^\\d{4}-([0][1-9]|[1][0-2])-[0-3]\\d{1}[Tt][0-2]\\d{1}:[0-5]\\d{1}:[0-5]\\d{1}.?\\d{0,9}([zZ]|([+-]\\d{2}:\\d{2}))$";
```
SergKam: How costly a regexp is? Maybe we could write our own custom validation function that will be even more efficient.

antban: If I'm not mistaken - there is a datetime validator that is used in library as well. here we are just checking format. As @SergKam maybe we should test exactly for differences?

adyach: on local machine it looks like that:
Regex:
1: 849ms
2: 886 ms
3: 2728ms

OffsetDateTime:
1: 6539ms
2: 11627ms
3: 6674ms

adyach: About more efficient, we could start testing regex approach and approve it later on if required

antban:
:+1:

adyach:
deploy validation please

PR #868, opened 2018-05-03
Removing deprecated/unmaintained Nakadi clients

mdedetrich:
# Summary

This PR removes both the Nakadi Klients and Reactive Nakadi clients as suggested clients for Nakadi, reasons being

1. Both clients are unmaintained. The last commit for Reactive Nakadi is over a year ago and Nakadi Klients is deprecated
2. Nakadi Klients is already marked as deprecated
3. Both clients implement the low level Nakadi API (which is also deprecated) and there is no hint that either of the clients have plans to upgrade to the high level API


ePaul:
:+1:

mdedetrich:
👍

lmontrieux:
👍

PR #873, opened 2018-05-08
ARUHA-1664: Subscription time lag stats

v-stepanov:
Extended subscription statistics endpoint with optional time-lag information.

Zalando ticket: ARUHA-1664

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
@@ -1,6 +1,7 @@
 package org.zalando.nakadi.service.subscription;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import org.echocat.jomon.runtime.concurrent.RetryForSpecifiedTimeStrategy;
 import org.json.JSONObject;
@@ -24,6 +25,7 @@ import org.zalando.nakadi.domain.Subscription;
 import org.zalando.nakadi.domain.SubscriptionBase;
 import org.zalando.nakadi.domain.SubscriptionEventTypeStats;
 import org.zalando.nakadi.domain.Timeline;
+import org.zalando.nakadi.exceptions.ErrorGettingCursorTimeLagException;
 import org.zalando.nakadi.exceptions.InternalNakadiException;
 import org.zalando.nakadi.exceptions.InvalidCursorException;
 import org.zalando.nakadi.exceptions.NakadiException;
@@ -176,7 +178,7 @@ public class SubscriptionService {
                     new PaginationWrapper<>(subscriptions, paginationLinks);
             if (showStatus) {
                 final List<Subscription> items = paginationWrapper.getItems();
-                items.forEach(s -> s.setStatus(createSubscriptionStat(s, false)));
+                items.forEach(s -> s.setStatus(createSubscriptionStat(s, false, false)));
             }
             return Result.ok(paginationWrapper);
         } catch (final ServiceTemporarilyUnavailableException e) {
@@ -229,7 +231,8 @@ public class SubscriptionService {
     }
 
     public ItemsWrapper<SubscriptionEventTypeStats> getSubscriptionStat(final String subscriptionId,
-                                                                        final boolean includeDistance)
+                                                                        final boolean includeDistance,
+                                                                        final boolean showTimeLag)
             throws InconsistentStateException, NoSuchSubscriptionException, ServiceTemporarilyUnavailableException {
         final Subscription subscription;
         try {
@@ -237,19 +240,21 @@ public class SubscriptionService {
         } catch (final ServiceTemporarilyUnavailableException ex) {
             throw new InconsistentStateException(ex.getMessage());
         }
-        final List<SubscriptionEventTypeStats> subscriptionStat = createSubscriptionStat(subscription, includeDistance);
+        final List<SubscriptionEventTypeStats> subscriptionStat = createSubscriptionStat(subscription, includeDistance,
+                showTimeLag);
         return new ItemsWrapper<>(subscriptionStat);
     }
 
     private List<SubscriptionEventTypeStats> createSubscriptionStat(final Subscription subscription,
-                                                                    final boolean includeDistance)
+                                                                    final boolean includeDistance,
+                                                                    final boolean showTimeLag)
             throws InconsistentStateException, ServiceTemporarilyUnavailableException {
         final List<EventType> eventTypes = getEventTypesForSubscription(subscription);
         final ZkSubscriptionClient subscriptionClient = createZkSubscriptionClient(subscription);
         final Optional<ZkSubscriptionNode> zkSubscriptionNode = subscriptionClient.getZkSubscriptionNode();
 
         if (includeDistance) {
-            return loadStats(eventTypes, zkSubscriptionNode, subscriptionClient);
+            return loadStats(eventTypes, zkSubscriptionNode, subscriptionClient, showTimeLag);
         } else {
             return loadLightStats(eventTypes, zkSubscriptionNode);
         }
@@ -287,8 +292,8 @@ public class SubscriptionService {
         return topicPartitions;
     }
 
-    private Map<EventTypePartition, Optional<Duration>> getTimeLags(final Collection<NakadiCursor> committedPositions,
-                                                                    final List<PartitionEndStatistics> endPositions) {
+    private Map<EventTypePartition, Duration> getTimeLags(final Collection<NakadiCursor> committedPositions,
+                                                          final List<PartitionEndStatistics> endPositions) {
 
         final NakadiCursorComparator cursorComparator = new NakadiCursorComparator(eventTypeCache);
         final MultiTimelineEventConsumer consumer = timelineService.createMultiTimelineEventConsumer(
@@ -306,7 +311,7 @@ public class SubscriptionService {
                                     .orElse(false);
 
                             if (isAtTail) {
-                                return Optional.of(Duration.ZERO);
+                                return Duration.ZERO;
                             } else {
                                 return getNextEventTimeLag(c, consumer);
                             }
@@ -314,16 +319,11 @@ public class SubscriptionService {
                 ));
     }
 
-    private Optional<Duration> getNextEventTimeLag(final NakadiCursor cursor,
-                                                   final MultiTimelineEventConsumer consumer) {
+    private Duration getNextEventTimeLag(final NakadiCursor cursor, final MultiTimelineEventConsumer consumer) {
         try {
             consumer.reassign(ImmutableList.of(cursor));
-        } catch (NakadiException e) {
-            e.printStackTrace();
-            return Optional.empty();
-        } catch (InvalidCursorException e) {
-            e.printStackTrace();
-            return Optional.empty();
+        } catch (final NakadiException | InvalidCursorException e) {
+            throw new ErrorGettingCursorTimeLagException(cursor, e);
         }
 
         final ConsumedEvent nextEvent = executeWithRetry(
@@ -335,10 +335,10 @@ public class SubscriptionService {
                         .withResultsThatForceRetry((ConsumedEvent) null));
 
         if (nextEvent == null) {
-            return Optional.of(Duration.ZERO);
+            return Duration.ZERO;
+        } else {
+            return Duration.ofMillis(new Date().getTime() - nextEvent.getTimestamp());
         }
-        final Duration duration = Duration.ofMillis(new Date().getTime() - nextEvent.getTimestamp());
-        return Optional.of(duration);
     }
 
     private Map<TopicRepository, List<Timeline>> getTimelinesByRepository(final Collection<EventType> eventTypes) {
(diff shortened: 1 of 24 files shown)


v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
```diff
@@ -0,0 +1,129 @@
...
+        } catch (final Throwable e) {
+            throw new InconsistentStateException("Unexpected error occurred when getting subscription time lag", e);
+        }
+    }
+
+    private Duration getNextEventTimeLag(final NakadiCursor cursor) throws ErrorGettingCursorTimeLagException,
+            InconsistentStateException {
+        try {
+            final EventConsumer consumer = timelineService.createEventConsumer(
+                    "time-lag-checker-" + UUID.randomUUID().toString(), ImmutableList.of(cursor));
```
rcillo: I did not even remember this "identification" feature still existed. I thought it was "commented out" as an emergency reaction to an incident which we never followed up.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
```diff
@@ -0,0 +1,129 @@
...
+
+            final ConsumedEvent nextEvent = executeWithRetry(
```
rcillo: I thought this execute with retries was exclusively used by tests. Would you consider adding a parameter to `readEvents` to pass the amount of time you are willing to wait? Then you give control to that method and let it use Kafka to manage it. I think that KafkaConsumer accepts some parameters that define precisely "fetch_wait_time_ms". You could use it, when building a new KafkaConsumer. Then you just need to pool and read whatever comes with it.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
```diff
@@ -0,0 +1,129 @@
...
+            throw e;
+        } catch (final Throwable e) {
+            throw new InconsistentStateException("Unexpected error occurred when getting subscription time lag", e);
+        }
+    }
+
+    private Duration getNextEventTimeLag(final NakadiCursor cursor) throws ErrorGettingCursorTimeLagException,
+            InconsistentStateException {
+        try {
+            final EventConsumer consumer = timelineService.createEventConsumer(
```
rcillo: Creating a consumer is an expensive thing. Would you consider to try to "pack" many partitions in a single consumer? I'm not sure we have control over independent partitions reading, but I would like to ask you anyway, why did you decide not to do it.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
```diff
@@ -0,0 +1,129 @@
...
+
+            final ConsumedEvent nextEvent = executeWithRetry(
+                    () -> {
+                        final List<ConsumedEvent> events = consumer.readEvents();
+                        return events.isEmpty() ? null : events.iterator().next();
+                    },
+                    new RetryForSpecifiedTimeStrategy<ConsumedEvent>(EVENT_FETCH_WAIT_TIME_MS)
+                            .withResultsThatForceRetry((ConsumedEvent) null));
+
+            if (nextEvent == null) {
```
rcillo: Shouldn't this be an error condition? If I understood correctly, before invoking this method you check if the cursor is not the END. So there should always be a "next". Then this should throw an error because we were not able to fetch the event on time.

src/test/java/org/zalando/nakadi/service/SubscriptionTimLagServiceTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.greaterThan;
+import static org.hamcrest.Matchers.hasSize;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class SubscriptionTimLagServiceTest {
```
rcillo: There's a time in `TimLag`.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
v-stepanov: The problem here is exactly no control over reading from specific partitions. So if I create a consumer for many partitions I don't know when I will get data for some specific partition. It may happen that I will have to read a lot of data from partition A before I get at least something for partition B. 
That's why I decided to make a consumer per partition. And as that happens in multi-threaded way - I can't share the same consumer of course.

rcillo: Ok, got it.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
v-stepanov: I think I was just lazy to do all that refactoring to be able to pass that parameter :) 
I will do it

rcillo: @v-stepanov I think it might be there already as the last parameter in the constructor. It's called "pollTimeout"

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
rcillo: @v-stepanov I was checking Kafka Consumer parameters and maybe we can create a single consumer that will read a single event from each partition by using `max.partition.fetch.bytes = 1` http://kafka.apache.org/documentation.html#consumerconfigs

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
v-stepanov: It is commented out in a place where this parameter is passed to kafka consumer.

v-stepanov:
deploy validation please

src/test/java/org/zalando/nakadi/service/SubscriptionTimLagServiceTest.java
v-stepanov: Fixed https://github.com/zalando/nakadi/pull/873/commits/36ff06ad6adb6c505d566d59419174244969d7e7

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
```diff
@@ -0,0 +1,129 @@
...
+
+    private Duration getNextEventTimeLag(final NakadiCursor cursor) throws ErrorGettingCursorTimeLagException,
+            InconsistentStateException {
+        try {
+            final EventConsumer consumer = timelineService.createEventConsumer(
+                    "time-lag-checker-" + UUID.randomUUID().toString(), ImmutableList.of(cursor));
+
+            final ConsumedEvent nextEvent = executeWithRetry(
+                    () -> {
+                        final List<ConsumedEvent> events = consumer.readEvents();
```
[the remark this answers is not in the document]
v-stepanov: Oh yes! Thanks for noticing. Fixed https://github.com/zalando/nakadi/pull/873/commits/36ff06ad6adb6c505d566d59419174244969d7e7

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
v-stepanov: I agree. Fixed https://github.com/zalando/nakadi/pull/873/commits/36ff06ad6adb6c505d566d59419174244969d7e7

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
v-stepanov: Another thing is that it's not so nice to pass kafka-specific parameter from a place which is abstracted from kafka.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
```diff
@@ -0,0 +1,130 @@
...
+                                      final NakadiCursorComparator cursorComparator) {
+        this.timelineService = timelineService;
+        this.cursorComparator = cursorComparator;
+    }
+
+    public Map<EventTypePartition, Duration> getTimeLags(final Collection<NakadiCursor> committedPositions,
+                                                         final List<PartitionEndStatistics> endPositions)
+            throws ErrorGettingCursorTimeLagException, InconsistentStateException {
+
+        final ExecutorService executor = Executors.newFixedThreadPool(LAG_CALCULATION_PARALLELISM);
```
antban: Does it make sense to constantly create threads in packs of 10? 
Why not to use single CachedThreadPool for this service?

v-stepanov: Fixed thread pool will create 10 Threads only if they are really needed (subscription has >= 10 partitions). It will anyway add threads 1 by 1 until reaches 10. The only problem it has is that it will never decrease thread num after that, but for us it's not a problem because the lifetime of this threadpool is very short.
 
And I don't see how to limit CachedThreadPool size.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -214,27 +219,30 @@ public Result listSubscriptions(@Nullable final String owningApplication, @Nulla
...
+                                                                    final boolean includeDistance,
+                                                                    final boolean showTimeLag)
             throws InconsistentStateException, ServiceTemporarilyUnavailableException {
         final List<EventType> eventTypes = getEventTypesForSubscription(subscription);
         final ZkSubscriptionClient subscriptionClient = createZkSubscriptionClient(subscription);
         final Optional<ZkSubscriptionNode> zkSubscriptionNode = subscriptionClient.getZkSubscriptionNode();
 
         if (includeDistance) {
-            return loadStats(eventTypes, zkSubscriptionNode, subscriptionClient);
+            return loadStats(eventTypes, zkSubscriptionNode, subscriptionClient, showTimeLag);
```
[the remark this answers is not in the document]
v-stepanov: Yes, I agree. I will try to replace these two booleans with enum.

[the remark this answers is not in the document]
v-stepanov: fixed https://github.com/zalando/nakadi/pull/873/commits/8a2eb564af944f6c7640b1e388114ec94b155c15

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
antban: @v-stepanov 
Sometimes I feel myself as alien. I mean why not to use single cached thread pool for instance, not for request 
```
new ThreadPoolExecutor(
    0, // Minimum number of threads
    1000, // Maximum number of threads
    60L, // Idle timeout to kill threads after
    TimeUnit.SECONDS, // Idle timeout unit
    new LinkedBlockingQueue<>()) // Queue, to which tasks are scheduled
```
Let's suppose that I am doing 2000 requests targeting 8000 partitions. You will create definitely more than 2000 threads, at most - 8000 (will be pretty close to max), and after starting - will destroy it. With single cached thread pool you will create up to max parallel requests* avg partition count = 20 *4 = 80 (it is really expected value, in total). 
We can calculate difference and think about time spent for threads creation/termination. 

I don't really see the reason of single request limit to 10 threads. You are just saying that overall I have possibility of infinite thread creation (across different requests), but I will not try to accomplish fast this huge request, because I do not want to. 
This 10 limit will not save instance from 1000 concurrent requests, but will definitely make some requests slower. 
Also, having limit of threads dedicated for this operation will guarantee that nakadi will not be overloaded with this amount of requests.

v-stepanov:
deploy validation please

v-stepanov:
deploy validation please

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionTimeLagService.java
v-stepanov: @antban I reworked it.
Now there is a common thread pool. But I anyway implemented the throttling per request (one request will not occupy more than 20 threads).
I hope this makes you satisfied.

v-stepanov:
@antban I updated PR according to your offline comments.

v-stepanov:
deploy validation please

antban:
:+1:

v-stepanov:
👍

PR #874, opened 2018-05-09
ARUHA-1641 Refresh zk subscriptions on timeout

antban:
ARUHA-1641
Every now and then while streaming nakadi looses information about latest committed cursors. In order to avoid this situation nakadi will try to refresh information about currently committed offsets.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
@@ -214,11 +214,29 @@ public class StreamingContext implements SubscriptionStreamer {
         });
     }
 
+    public void refreshZkListeners() {
+        // Already stopped, nothing to do.
+        if (null == sessionListSubscription) {
+            return;
+        }
+        getZkClient().refreshListeners(this.getParameters().commitTimeoutMillis / 4);
+
+        rescheduleZkListenersRefresh();
+    }
+
+    private void rescheduleZkListenersRefresh() {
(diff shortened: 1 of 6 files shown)


src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
```diff
@@ -214,11 +214,29 @@ public void switchState(final State newState) {
...
+                this.getParameters().commitTimeoutMillis / 8,
+                TimeUnit.MILLISECONDS);
+    }
+
     public void registerSession() throws NakadiRuntimeException {
         log.info("Registering session {}", session);
         // Install rebalance hook on client list change.
         sessionListSubscription = zkClient.subscribeForSessionListChanges(() -> addTask(this::rebalance));
         zkClient.registerSession(session);
+        rescheduleZkListenersRefresh();
```
adyach: maybe move it to `StartingState.onEnter()`? It seems to be related more to start of subscription in general rather than registration of session.

src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
```diff
@@ -214,11 +214,29 @@ public void switchState(final State newState) {
         });
     }
 
+    public void refreshZkListeners() {
+        // Already stopped, nothing to do.
+        if (null == sessionListSubscription) {
```
adyach: it is not clear how it is related to refreshing zk listeners

PR #882, opened 2018-05-30
ARUHA-1680 Use strict json parsing for event type publishing

antban:
# One-line summary

ARUHA-1680 Use strict json parsing for event type publishing


src/test/resources/very_complex.json
```diff
@@ -0,0 +1,3316 @@
...
+                                                                                                                                                             "▟GL K2i뛱iＱ\"̠.옛1X$}涺]靎懠ڦ늷?tf灟ݞゟ{": 1.227740268699265E-19,
+                                                                                                                                                             "꒶]퓚%ฬK❅": [{
+                                                                                                                                                              "(ෛ@Ǯっ䧼䵤[aﾃൖvEnAdU렖뗈@볓yꈪ,mԴ|꟢캁(而첸죕CX4Y믅": "2⯩㳿ꢚ훀~迯?᪑\\啚;4X\u20c2襏B箹)俣eỻw䇄",
+                                                                                                                                                              "75༂f詳䅫ꐧ鏿 }3\u20b5'∓䝱虀f菼Iq鈆﨤g퍩)BFa왢d0뮪痮M鋡nw∵謊;ꝧf美箈ḋ*\u001c`퇚퐋䳫$!V#N㹲抗ⱉ珎(V嵟鬒_b㳅\u0019": null,
+                                                                                                                                                              "e_m@(i㜀3ꦗ䕯䭰Oc+-련0뭦⢹苿蟰ꂏSV䰭勢덥.ྈ爑Vd,ᕥ=퀍)vz뱊ꈊB_6듯\"?{㒲&㵞뵫疝돡믈%Qw限,?\r枮\"? N~癃ruࡗdn&": null,
+
```
v-stepanov: no newline 
:)

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
```diff
@@ -0,0 +1,287 @@
+package org.zalando.nakadi.domain;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+public class StrictJsonParser {
```
v-stepanov: Can you please describe which part of this file was written by you and which was borrowed? So that I know where to focus the review.

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
```diff
@@ -0,0 +1,63 @@
...
+        StrictJsonParser.parse("{name\"}");
+    }
+
+    @Test(expected = JSONException.class)
+    public void testBadJson4() {
+        StrictJsonParser.parse("{\"name\"NaN}");
+    }
+
+    @Test(expected = JSONException.class)
+    public void testBadJson5() {
```
v-stepanov: I think it would be better if the test method names describe the actual test case they test.

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
```diff
@@ -0,0 +1,287 @@
...
+        void back() {
+            this.currentPosition--;
+        }
+
+        static boolean isEmptyCharacter(final char c) {
+            return (c == ' ' || c == '\t' || c == '\n' || c == '\r');
+        }
+    }
+
+    public static Object parse(final String value) throws JSONException {
```
v-stepanov: why does it return Object, not the JSONObject?

v-stepanov: Ok, I see now that the next method returns JSONObject. Do we need both of them?

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.Charset;
+
+public class StrictJsonParserTest {
+
+    private void testSingleString(final String value) {
+
+        final JSONObject orthodoxJson = new JSONObject(value);
+        final JSONObject anarchyJson = (JSONObject) StrictJsonParser.parse(value);
```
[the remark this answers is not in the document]
v-stepanov: StrictJsonParser.parseObject returns JSONObject. No casting will be needed.

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
```diff
@@ -0,0 +1,63 @@
...
+    @Test(expected = JSONException.class)
+    public void testBadJson4() {
+        StrictJsonParser.parse("{\"name\"NaN}");
+    }
+
+    @Test(expected = JSONException.class)
+    public void testBadJson5() {
+        StrictJsonParser.parse("{\"name\":NaN}");
+    }
+
```
v-stepanov: One more case I remember is extra "," like
```
{
  "blah": "foo",
}
```
Can we add a test for that?

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
antban: I took readStringTillTheEnd method completely. Everything else is more or less rewritten

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
antban: @v-stepanov The problem with this approach is that I can not describe what I am testing against. But if it really matters, I will rewrite tests for failing scenarios completely into 1 and check exception by hands.

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
```diff
@@ -0,0 +1,287 @@
...
+        } else {
+            throw syntaxError("Expected false value", tokenizer);
+        }
+    }
+
+    private static Object readTrueTillTheEnd(final StringTokenizer tokenizer) {
+        if (tokenizer.next(3).equals("rue")) {
+            return Boolean.TRUE;
+        } else {
+            throw syntaxError("Expected false value", tokenizer);
```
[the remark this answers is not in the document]
antban: yep, mistake

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
```diff
@@ -0,0 +1,63 @@
...
+}
```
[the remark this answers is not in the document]
antban: ok

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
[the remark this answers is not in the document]
antban: ok

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
antban: yep

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
```diff
@@ -35,29 +35,24 @@ public void testVeryComplexJson() throws IOException {
...
-    public void testBadJson4() {
-        StrictJsonParser.parse("{\"name\"NaN}");
-    }
-
-    @Test(expected = JSONException.class)
-    public void testBadJson5() {
-        StrictJsonParser.parse("{\"name\":NaN}");
+    @Test
+    public void testInvalidJsonFormats() {
+        final String[] testsData = new String[]{
```
v-stepanov: this is already invented https://github.com/junit-team/junit4/wiki/parameterized-tests

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
v-stepanov: @antban what about this point?

v-stepanov:
@antban the parser itself looks good for me. Of course I'm not a compiler so I can't say if it will work or not :)
What is your plan regarding releasing of this?

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
antban: @v-stepanov Do you think that it pays to readability or simplifies development?

v-stepanov: The problem with your approach is that if you test A, B, C, D, E, F in your test and the test fails for B, it will not even run for C, D, E, F. So you will not have the full picture of which cases fail/successful.

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
antban: @v-stepanov Fixed https://github.com/zalando/nakadi/pull/882/commits/4e9aa2a8d524d83b342d30b524e91cfd5028d9d2

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
antban: I know about this, but honestly, I think that it is not the case here. It either fails or succeeds. But as you insist, I will change it.

v-stepanov: No, for me it's fine like this, I think it's not a big issue.

v-stepanov:
👍

src/test/java/org/zalando/nakadi/domain/StrictJsonParserTest.java
antban: @v-stepanov too late https://github.com/zalando/nakadi/pull/882/commits/b3058d7174bc3dacafafe29805eeaf5b32ab2ae5

v-stepanov:
👍

antban:
:+1:

antban:
deploy validation please

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
```diff
@@ -0,0 +1,283 @@
...
+                if (separator != ':') {
+                    throw syntaxError("Waiting for name-value separator : while parsing object", tokenizer);
+                }
+                final Object value = parse(tokenizer);
+                result.putOnce(name, value);
+                final char nextToken = tokenizer.nextUnskippable();
+                if (nextToken == '}') {
+                    finished = true;
+                } else if (nextToken != ',') {
+                    throw syntaxError("Unexpected symbol " + nextToken + " while parsing object", tokenizer);
```
v-stepanov: can you please put that symbol in single quotes?

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+        tokenizer.back();
+        boolean finished = false;
+        while (!finished) {
+            result.put(parse(tokenizer));
+            final char separator = tokenizer.nextUnskippable();
+            if (separator == ']') {
+                finished = true;
+            } else if (separator != ',') {
+                throw syntaxError("Unexpected separator " + separator, tokenizer);
```
v-stepanov: can you please put the separator in single quotes?

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
```diff
@@ -0,0 +1,283 @@
...
+        final StringTokenizer stringTokenizer = new StringTokenizer(value, startIdx, endIdx);
+        final Object result = parse(stringTokenizer);
+        if (!allowMore) {
+            final char unexpectedValue;
+            try {
+                unexpectedValue = stringTokenizer.nextUnskippable();
+            } catch (JSONException ignore) {
+                return result;
+            }
+            throw syntaxError("Unexpected symbol " + unexpectedValue, stringTokenizer);
```
v-stepanov: can you please put that symbol in single quotes?

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
```diff
@@ -0,0 +1,283 @@
...
+        static boolean isEmptyCharacter(final char c) {
+            return (c == ' ' || c == '\t' || c == '\n' || c == '\r');
+        }
+    }
+
+    public static JSONObject parseObject(final String value) throws JSONException {
+        return (JSONObject) parse(value, true);
+    }
+
+    public static Object parse(final String value, final boolean allowMore) throws JSONException {
```
v-stepanov: There is still the public parse method which returns plain Object. I believe we don't need it as public method here, right?

src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
```diff
@@ -0,0 +1,283 @@
...
+        void back() {
+            this.currentPosition--;
+        }
+
+        static boolean isEmptyCharacter(final char c) {
+            return (c == ' ' || c == '\t' || c == '\n' || c == '\r');
+        }
+    }
+
+    public static JSONObject parseObject(final String value) throws JSONException {
```
v-stepanov: Is it possible to provide more context when showing error to user. Currently it looks like:
```
{
  "type": "http://httpstatus.es/400",
  "title": "Bad Request",
  "status": 400,
  "detail": "Expected false value at pos 164"
}
```
I think it would be better if we show something like "Error occurred when parsing event(s). Expected false value at pos 164"

antban:
@v-stepanov Fixed all the comments https://github.com/zalando/nakadi/pull/882/commits/3e03aa5224daebadeb172a3b506ec1e04d981310

src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -64,15 +68,29 @@ public static EmptyInjectionConfiguration build(final int position, final boolea
...
-        this.event = strictParsing ? StrictJsonParser.parseObject(event) : new JSONObject(event);
-        this.eventSize = event.getBytes(StandardCharsets.UTF_8).length;
+        this.event = new JSONObject(rawEvent);
+
+        // We will first release in a shadow mode when the strict parser does second parsing
+        // and we compare if the result is the same as with the old parser; we also log invalid events;
+        if (strictParsing) {
+            try {
+                final JSONObject shadowEvent = StrictJsonParser.parseObject(rawEvent);
+                if (!shadowEvent.toString().equals(this.event.toString())) {
```
antban: If the order of the fields in object will be the same here?

v-stepanov: I used the same approach as you used in unit test. I thought that it is safe :)

src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -64,15 +68,29 @@ public static EmptyInjectionConfiguration build(final int position, final boolea
...
+                    LOG.debug("[STRICT_JSON_DIFF] Strict parser produced different output for event: {}", rawEvent);
```
adyach: could you please log both parsed events to compare them later?

src/main/java/org/zalando/nakadi/domain/BatchItem.java
rcillo: @v-stepanov This `toString` operation is very expensive and is the core of the optimization that @antban did that modified the behaviour couple of months ago. I think we cannot go this far. The best we can do is to verify if it throws exception or not.

v-stepanov: Then we will not be able to validate the correctness. We can check the performance on staging and cancel if it consumes too much resources.

v-stepanov: I mean JSONObject doesn't really have equals method. So I can't compare them in a simple way.

antban: Then I agree with Ricardo that we don't need shadow parsing

antban: Then I agree with Ricardo that we don't need shadow parsing

rcillo: ok, I get it. Then we need to be mindful and run it for a limited amount of time. 1 day I would say.

rcillo:
👍

src/main/java/org/zalando/nakadi/domain/BatchItem.java
v-stepanov: why?

src/main/java/org/zalando/nakadi/domain/BatchItem.java
v-stepanov: ok, will do

v-stepanov: https://github.com/zalando/nakadi/pull/882/commits/dc5fe41e80cbfc7ba40e8129b6b407e7ee565973

v-stepanov:
👍

v-stepanov:
👍

antban:
:+1:

antban:
:+1:

antban:
:+1:

lmontrieux:
👍

rcillo:
👍

PR #884, opened 2018-06-05
Subscription authorization

rcillo:
https://techjira.zalando.net/browse/ARUHA-1681

Since the introduction of Event Type Authorization, it was planned that
Subscriptions would also be authorized. Even though data is already
protected through Event Type Authorization, users can still accidentally
commit offsets to subscriptions owned by other users, eventually causing
applications to not process all the data. Subscriptions could also be
accidentally deleted or have its slots assigned to some other
application impeding applications from consuming data.

With this feature we address all of the problems above. This is just the
API definition.

### DIFF
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -1019,6 +1019,29 @@ paths:
       responses:
         '204':
           description: Subscription was deleted
+        '403':
+          description: Access forbidden
+          schema:
+            $ref: '#/definitions/Problem'
+        '404':
+          description: Subscription not found
+          schema:
+            $ref: '#/definitions/Problem'
+    put:
+      tags:
+        - subscription-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: Updates a subscription.
+      parameters:
+        - $ref: '#/parameters/SubscriptionId'
+      responses:
+        '204':
+          description: Subscription was updated
+        '403':
+          description: Access forbidden
+          schema:
+            $ref: '#/definitions/Problem'
         '404':
           description: Subscription not found
           schema:
@@ -1158,6 +1181,10 @@ paths:
       responses:
         '204':
           description: Offsets were reset
+        '403':
+          description: Access forbidden
+          schema:
+            $ref: '#/definitions/Problem'
         '404':
           description: Subscription not found
           schema:
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -1032,9 +1032,21 @@ paths:
         - subscription-api
       security:
         - oauth2: ['nakadi.event_stream.read']
-      description: Updates a subscription.
+      description: |
+        This endpoint only allows to update the authorization section of a subscription. All other properties are
+        immutable. This operation is restricted to subjects with administrative role.
       parameters:
-        - $ref: '#/parameters/SubscriptionId'
+        - name: subscription
+          in: body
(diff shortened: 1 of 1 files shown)


docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1109,6 +1109,10 @@ paths:
                   $ref: '#/definitions/CursorCommitResult'
             required:
               - items
+        '403':
```
v-stepanov: can you please also add it to reset cursors endpoin

v-stepanov: and also to delete endpoint

v-stepanov:
I'm missing update-subscription endpoint :)

rcillo:
@v-stepanov addressed the comments.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1019,6 +1019,29 @@ paths:
...
+        '404':
+          description: Subscription not found
+          schema:
+            $ref: '#/definitions/Problem'
+    put:
+      tags:
+        - subscription-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: Updates a subscription.
```
v-stepanov: Do you think it makes sense to describe here which parameters of subscription it is possible to update? I'm afraid users will try to update owning app and list of event-types.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1019,6 +1019,29 @@ paths:
...
+      parameters:
```
v-stepanov: I think you also need to describe here the subscription itself that comes in body

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1019,6 +1019,38 @@ paths:
...
+        This endpoint only allows to update the authorization section of a subscription. All other properties are
+        immutable. This operation is restricted to subjects with administrative role.
+      parameters:
+        - name: subscription
+          in: body
+          description: Subscription to create
+          schema:
+            $ref: '#/definitions/Subscription'
+          required: true
+      parameters:
```
v-stepanov: you have `parameters` properties doubled

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1019,6 +1019,38 @@ paths:
...
+        - subscription-api
+      security:
+        - oauth2: ['nakadi.event_stream.read']
+      description: |
+        This endpoint only allows to update the authorization section of a subscription. All other properties are
+        immutable. This operation is restricted to subjects with administrative role.
+      parameters:
+        - name: subscription
+          in: body
+          description: Subscription to create
```
v-stepanov: not to create

v-stepanov:
@rcillo Now everything looks good for me. The only thing that can be improved - is to specify which role is needed to reset cursors. I think now it will be not clear for the user reading the API description which role he needs to reset cursors.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1019,6 +1019,41 @@ paths:
...
+          type: string
+          description: Id of subscription
+          required: true
+      responses:
+        '204':
+          description: Subscription was updated
+        '403':
+          description: Access forbidden
+          schema:
+            $ref: '#/definitions/Problem'
```
antban: But where is 
```
         '400':
           description: Bad Request
           schema:
             $ref: '#/definitions/Problem'
```

PR #886, opened 2018-06-06
Upgrade zalando problem

lmontrieux:
# Upgrade zalando problem library

> Zalando ticket : ARUHA-1695

## Description
This PR upgrades the 3 zalando problem dependencies to their latest version. This has required a lot of refactoring, because the way problems and responses are created has changed. Hence, this upgrade was done in multiple steps:

- move all problem creation to controllers, where it should be
- as a consequence, the "old" style of exception, which included the `.asProblem()` method, are gone and replaced by runtime exceptions, as decided quite some time ago
- upgrade zalando problem dependencies
- fix problem and response creation

And as bonus:
- refactor some of the unit and acceptance tests. Now, when a controller endpoint is very simple, we have to following:
  - tests for expected behaviour in acceptance tests only
  - what is difficult to test in acceptance tests (eg, failure of underlying storage) is tested in the service
- `MyNakadiRuntimeException1` is dead. Long live `NakadiRuntimeBaseException`!

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG


lmontrieux:
This has been separated in multiple PRs for readabilty (and also because it works)

PR #890, opened 2018-06-14
Set `network_mode` to bridge in docker-compose to support OSX.

bitnot:
Changing network mode in `docker-compose.yml` from `host` to `bridge`
allows running it on Mac. This also requires services to refer to
dependencies by their names (e.g. `zookeeper` instead of 'localhost').
Documentation has been updated to reflect the fact, that it is possible
to run Nakadi in Docker on OSX.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
Updated `docker-compose.yml` for local development and documentation. No code changes.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,11 +20,10 @@ services:
       - NAKADI_FEATURES_DEFAULT_FEATURES_KPI_COLLECTION
       - NAKADI_FEATURES_DEFAULT_FEATURES_DISABLE_DB_WRITE_OPERATIONS
       - NAKADI_FEATURES_DEFAULT_FEATURES_DISABLE_LOG_COMPACTION
-      - NAKADI_ZOOKEEPER_BROKERS=localhost:2181
-      - SPRING_DATASOURCE_URL=jdbc:postgresql://localhost:5432/local_nakadi_db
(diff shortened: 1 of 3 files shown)


lmontrieux:
Thanks for your contribution, @bitnot . This has been a pain point for quite some time. I'll check that everything works asap. Basically I want to make sure that this works:
- on linux
- on mac os x, with docker for mac
- on mac os x, with the 'old' docker machine

bitnot:
Tested on old "Docker Toolbox". Works, except for API being available on VirtualBox interface (`docker-machine ip`, "192.168.99.100" in my case) and not on localhost.
So `curl localhost:8080/health` does not work, but `curl $(docker-machine ip):8080/health` does.

lmontrieux:
@bitnot Had a quick look at our travis build, I can see that the acceptance tests are failing. Unsurprisingly it's because some tests try to find Nakadi on localhost. Could you please have a look?

bitnot:
Bridge mode seems to affect integration tests (performance?). 
Although `gradlew fullAcceptanceTest` passes locally, Travis CI build fails. I give up. 😭

PR #892, opened 2018-06-18
Implementation of ordering key fields

rcillo:
https://techjira.zalando.net/browse/ARUHA-1723

Implementation of ordering key fields. The ticket refers to a more detailed discussion on why this field was requested.

I extended the original description with information I considered relevant to reduce the support burden in case users misunderstand the use case for this field.


v-stepanov:
Looks good!

v-stepanov:
👍

rcillo:
deploy validation please

rcillo:
👍

v-stepanov:
👍

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2468,6 +2468,36 @@ definitions:
...
+          This is only an informational field. The events are delivered to consumers in the order they were published.
+          No reordering is done by Nakadi.
+
+          This field is useful in case the producer wants to communicate the complete order accross all the events
+          published to all partitions. This is the case when there is an incremental generator on the producer side,
+          for example.
+
+          It differs from `partition_key_fields` in the sense that it's not used for partitioning (known as sharding in
+          some systems). Also, the order of the events published in the same partition is not guaranteed to be
+          monotonically increased as with `partition_key_fields`. For `ordering_key_fields`, the consumer should handle
```
ePaul: > Also, the order of the events published in the same partition is not guaranteed to be monotonically increased as with `partition_key_fields`.

This sentence sounds as if the `partition_key_fields` indicates ordering (which it doesn't, as far as I know).
Maybe instead add this to the previous paragraph:

> The order indicated by the fields can differ from the order the events are in each partition, in case of out-of-order submission.

Another difference would be that this property applies to the whole event, not just the part inside of `data`, so it is possible to refer to metadata fields.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2468,6 +2468,36 @@ definitions:
...
+
+          In most cases, this would have just a single item (the path of the field
+          by which this is to be ordered), but can have multiple items, in which case
+          those are considered as a compound key, with lexicographic ordering (first
+          item is most significant).
+        items:
+          type: string
+          description: |
+            Indicates a single ordering field. This is a dot separated string, which is applied
+            onto the whole event object, including the contained metadata and data (in
```
ePaul: In the API guideline proposal I wrote "JSON pointer", because that is a thing which is defined and where it is clear how it would be applied to a JSON object – for "dot separated string" this is less the case.

As I understand (from looking at the code), for `partition_key_fields` you are replacing the `.` by `/` and then using it as a JSON pointer – is this meant to be an implementation detail? (A JSON pointer allows some more things, like accessing an array, too.)

Differences might happen if there are field names containing a `/` or `.`.

rcillo: @ePaul the decision to use dot notation is more than 2 years and a half old. It would be hard to tell who was involved and the reason for doing so. What I can tell that in all this time, it has served users well. I don't remember users having any hard time to use it, since it's simple. You just need to describe the path to the value using dots. Based on this experience and having in mind that consistency is important, I made the choice to change it.

docs/_data/nakadi-event-bus-api.yaml
rcillo: @ePaul thanks, changed here 1804d4d

Regarding the possibility to access metadata fields, it's better described in the array item description.

antban:
:+1:

rcillo:
👍

antban:
:+1:

PR #894, opened 2018-06-20
Increase max request size

rcillo:
https://techjira.zalando.net/browse/ARUHA-1739

In order to send larger messages we need to also allow for bigger
requests. This is not only the case for bigger messages but also if we
want to send larger batches to kafka, which in some cases can increase
throughput.


v-stepanov:
👍

rcillo:
👍

PR #896, opened 2018-06-22
ARUHA-1693: Added functionality to delete subscriptions together with event-type (if feature is toggled);

v-stepanov:
# Added functionality to delete subscriptions together with event-type (if feature is toggled)

> Zalando ticket : ARUHA-1693

## Description
On test environments we may allow users to delete event-types together with subscriptions

## Review
- [ ] Tests
- [ ] Documentation
- [x] CHANGELOG

## Deployment Notes
We should toggle this feature to make it work

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -1,7 +1,9 @@
 package org.zalando.nakadi.service;
 
+import com.google.common.collect.HashMultimap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
+import org.echocat.jomon.runtime.concurrent.RetryForSpecifiedTimeStrategy;
 import org.everit.json.schema.Schema;
 import org.everit.json.schema.SchemaException;
 import org.everit.json.schema.loader.SchemaLoader;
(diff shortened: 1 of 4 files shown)


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -258,6 +261,51 @@ public void delete(final String eventTypeName) throws EventTypeDeletionException
...
+    private Multimap<TopicRepository, String> deleteEventTypeIfNoSubscriptions(final String eventType) {
+        if (hasSubscriptions(eventType)) {
+            throw new ConflictException("Can't remove event type " + eventType
+                    + ", as it has subscriptions");
+        }
+        return transactionTemplate.execute(action -> deleteEventType(eventType));
+    }
+
+    @SuppressWarnings("unchecked")
+    private Multimap<TopicRepository, String> deleteEventTypeWithSubscriptions(final String eventType) {
```
rcillo: @v-stepanov why not delete the event type inside the transaction? Since there is a database consistency check, it would not be possible to commit the transaction if a subscription was created concurrently, right? Then it would fail the request and the caller could try to delete again manually. Attempting to perform the retry by yourself doesn't seem to cover all the possible race conditions, since it theoretically could retry for 1000 miliseconds and still fail.

But, theory apart, I think the way you did works. So I'm just commenting to know if you considered this other option and why you decided to go this way.

v-stepanov: I already run everything from within a transaction. (inside `transactionTemplate.execute(...)`)
But I think I handle errors in a wrong way. You are right that if the new subscription is created - the transaction will fail: not in the place where ET is removed but in the place of transaction execution. Ok, I will fix it.

v-stepanov: @rcillo I changed it. I remove the retry as it anyway doesn't give 100% guarantee of deletion. So I think the user should retry.
https://github.com/zalando/nakadi/pull/896/commits/07427f751624bad5cc4ab8a57bc4827f356a8fa5

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -258,6 +259,40 @@ public void delete(final String eventTypeName) throws EventTypeDeletionException
                 .put("compatibility_mode", eventType.getCompatibilityMode()));
     }
 
+    private Multimap<TopicRepository, String> deleteEventTypeIfNoSubscriptions(final String eventType) {
+        if (hasSubscriptions(eventType)) {
+            throw new ConflictException("Can't remove event type " + eventType
+                    + ", as it has subscriptions");
+        }
+        return transactionTemplate.execute(action -> deleteEventType(eventType));
```
rcillo: Do you need to do it in a transaction?

v-stepanov: It used to run inside a transaction, so I didn't change anything here. 
I think it is needed because delete event-type also handles the deletion of timelines.

v-stepanov:
deploy validation please

v-stepanov:
👍

PR #898, opened 2018-06-26
Change audience from underscore to dash

rcillo:
https://techjira.zalando.net/browse/ARUHA-1528


v-stepanov:
@rcillo Hm... Have you tried something like here: https://stackoverflow.com/questions/12468764/jackson-enum-serializing-and-deserializer
I think it will look nicer like that

rcillo:
@v-stepanov I can try. I had problem because there is a modifier registered for all enums and it somehow conflicted with other approaches. But I did not try the one you pointed out yet.

rcillo:
@v-stepanov I tried the proposed alternative and it does not work because in our project we have a global override on how enums are handled. This overrides everything else. In order to use this cleaner approach, I would have to explicitly state for every enum in the project how it should be serialized and deserialized. I think in the end it would not pay off.

src/main/java/org/zalando/nakadi/config/JsonConfig.java
```diff
@@ -35,7 +36,7 @@
     @Primary
     public ObjectMapper jacksonObjectMapper() {
         final ObjectMapper objectMapper = new ObjectMapper()
-                .setPropertyNamingStrategy(CAMEL_CASE_TO_LOWER_CASE_WITH_UNDERSCORES);
+                .setPropertyNamingStrategy(SNAKE_CASE);
```
v-stepanov: will not it break anything from existing behavior?

rcillo: I modified it because the old constant has been deprecated and replaced by this new one. They mean the same thing.

v-stepanov:
@rcillo ok, thanks for checking. Then let's keep it as it is currently implemented.

v-stepanov:
👍

rcillo:
👍

rcillo:
deploy validation please

v-stepanov:
👍

rcillo:
👍

rcillo:
deploy validation please

rcillo:
deploy validation please

PR #903, opened 2018-06-27
Docs. Better Event model description.

SergKam:
The event categories don't have any common base object/model instead, an Event schema can be one of the three categories
In my humble opinion, this way of describing Event is much cleaner and easier to read and better looks in the manual.


SergKam:
:+1:

PR #907, opened 2018-07-09
Aruha 1757 log compaction

v-stepanov:
# Log Compaction

> Zalando ticket : ARUHA-1757

## Description
Added log compaction functionality to Nakadi.

## Review
- [x] Tests
- [x] Documentation
- [x] CHANGELOG

## Deployment Notes
We need to setup separate kafka cluster and a storage in Nakadi for it. And we also need to define variable NAKADI_TIMELINES_STORAGE_COMPACTED for live environment to point to this new storage.


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -528,6 +554,16 @@ private void validateOrderingKeys(final Schema schema, final EventTypeBase event
         }
     }
 
+    private void validatePartitionCompactionKeys(final Schema schema, final EventTypeBase eventType)
+            throws InvalidEventTypeException, JSONException, SchemaException {
+        final List<String> absentFields = eventType.getPartitionCompactionKeys().stream()
+                .filter(field -> !schema.definesProperty(convertToJSONPointer(field)))
```
v-stepanov: @rcillo I didn't find a fast way to validate that there is a required field exists in schema. So here I only validate that property exists in schema (we actually do the same for other ..._key_fields properties). If you have an idea regarding that - please let me know and I will change it.

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
```diff
@@ -66,8 +68,10 @@ public TimelinesController(final TimelineService timelineService) {
         return Responses.create(MoreStatus.UNPROCESSABLE_ENTITY, ex.getMessage(), request);
     }
 
-    @ExceptionHandler(NotFoundException.class)
-    public ResponseEntity<Problem> notFound(final NotFoundException ex, final NativeWebRequest request) {
+    @ExceptionHandler({
+            NotFoundException.class,
+            TimelinesNotSupportedException.class})
```
rcillo: Wouldn't it make more sense to reply with 422? It actually found the event type but could not go on with the request.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -184,7 +187,7 @@ public boolean topicExists(final String topic) throws TopicRepositoryException {
             final ProducerRecord<String, String> kafkaRecord = new ProducerRecord<>(
                     topicId,
                     KafkaCursor.toKafkaPartition(item.getPartition()),
-                    item.getPartition(),
+                    item.getEventKey(),
```
rcillo: Do you know why we used to publish the partition as a key? Is it somehow change anything?

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
v-stepanov: Oh yes, sorry. I just added it to wrong handler accidentally. Will fix.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
v-stepanov: I think we were using partition as a key accidentally. In general if you directly specify partition to publish to - it will just ignore the key, it is stated like that in documentation. So I don't expect anything to be changed.
Or do you suggest to keep using partition as a key for cleanup_policy=delete? (not to break anything)

src/main/java/org/zalando/nakadi/service/EventPublisher.java
```diff
@@ -99,6 +99,7 @@ EventPublishResult publishInternal(final String events,
 
             validate(batch, eventType);
             partition(batch, eventType);
+            compact(batch, eventType);
```
rcillo: There is an alternative solution. When building the effective Schema, you can make this field mandatory and leverage the feature of validation, so that it doesn't come this point. I know it somehow couples validation with compaction, but I guess that's already the case for almost all aspects of event publishing.

More specifically, extending https://github.com/zalando/nakadi/blob/167c0c1206b83128f5ec16b5808d54ec20c75c82/src/main/java/org/zalando/nakadi/validation/JsonSchemaEnrichment.java#L36 by adding another `if` and then modifying the schema by marking `partition_compaction_key` to be `required` and to be longer than 0 character. The nice thing about doing that is that it provides an error with the complete path to the missing field.

Then the error would also be reported as a problem during validation, which I think makes more sense than "partitioning".

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: I think we can test it, no problem. Good that you already checked the docs and they say it should not change anything. We just need to confirm 👍

src/main/java/org/zalando/nakadi/repository/TopicRepository.java
```diff
@@ -38,7 +39,8 @@ public String getPartition() {
         }
     }
 
-    String createTopic(int partitionCount, Long retentionTimeMs) throws TopicCreationException;
+    String createTopic(int partitionCount, Long retentionTimeMs, CleanupPolicy cleanupPolicy)
```
rcillo: I'ld say this interface has a problem. Every time we need to tweak some configuration in the topic creation you need to change this interface. Maybe we could pass an object with all the configs? Then if we want to tweak clean up minimum latency in the future it becomes easier, just add a property to the object and then map it properly in the implementation, but the interface goes unchanged.

But I honestly cannot think about many other configuration options.

rcillo: Just thought about one: max message size, for example.

rcillo: But I guess we could also do this refactoring when editing it the next time.

src/main/java/org/zalando/nakadi/service/EventPublisher.java
v-stepanov: Yes, makes sense. Changed https://github.com/zalando/nakadi/pull/907/commits/7ab73d2309358778495a30321d864422a1bc4a68

src/main/java/org/zalando/nakadi/controller/TimelinesController.java
v-stepanov: fixed https://github.com/zalando/nakadi/pull/907/commits/939733792a424c68b893a490349e03a6bf343b62

src/main/java/org/zalando/nakadi/repository/TopicRepository.java
v-stepanov: I think it's not that bad yet. We have some places with 15 parameters :)
I think we should start this "Introduce Parameter Object" refactoring from the most problematic places we have.

v-stepanov:
@rcillo I addressed your comments

src/main/java/org/zalando/nakadi/validation/JsonSchemaEnrichment.java
```diff
@@ -164,15 +171,25 @@ private static JSONObject addMetadata(final JSONObject schema, final EventType e
         metadataProperties.put("parent_eids", arrayOfUUIDs);
         metadataProperties.put("flow_id", string);
         metadataProperties.put("partition", string);
+        metadataProperties.put("partition_compaction_key", string);
```
rcillo: I guess you are putting it twice. Here and in line 181.

v-stepanov: 🤦‍♂️ 
Fixed https://github.com/zalando/nakadi/pull/907/commits/5363d3ff43bbfeea7dad99c638ff585d1d118e9f

rcillo:
👍

v-stepanov:
👍

v-stepanov:
deploy validation please

src/main/java/org/zalando/nakadi/service/EventPublisher.java
adyach: we introduced another loop here, do you think it is time to combine it and do every step in one loop?

v-stepanov:
deploy validation please

rcillo:
👍

src/main/java/org/zalando/nakadi/service/EventPublisher.java
rcillo: @adyach in the API we always point out which step failed. So if something fails during erichment, for example, it means that everything is fine regarding validation. So if we did as you suggest, it could happen that for the first event it validates fine but then enrichment breaks. The user fix the payload, send again and this time if fails validation for the next event. So I think that even though it would be possible to do what you suggest, I think it would be more work then it looks like to not change the current behaviour.

adyach:
👍

PR #909, opened 2018-07-11
Aruha 1757 additional topic properties

v-stepanov:
# Additional properties for compacted topics;

> Zalando ticket : ARUHA-1757

## Description
This enables setting custom properties for compacted topics:
"segment.ms", "segment.bytes", "min.compaction.lag.ms"

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
-

### DIFF
--- a/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicConfigBuilder.java
+++ b/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicConfigBuilder.java
@@ -1,0 +1,84 @@
+package org.zalando.nakadi.repository.kafka;
+
+import kafka.admin.RackAwareMode;
+
+import java.util.Optional;
+
+public final class KafkaTopicConfigBuilder {
+
+    private String topicName;
+    private int partitionCount;
+    private int replicaFactor;
+    private String cleanupPolicy;
+    private long segmentMs;
+    private Long retentionMs;
+    private Long segmentBytes;
+    private Long minCompactionLagMs;
+    private RackAwareMode rackAwareMode;
+
+    private KafkaTopicConfigBuilder() {
+    }
+
+    public static KafkaTopicConfigBuilder builder() {
+        return new KafkaTopicConfigBuilder();
+    }
+
+    public KafkaTopicConfigBuilder withTopicName(final String topicName) {
+        this.topicName = topicName;
+        return this;
+    }
+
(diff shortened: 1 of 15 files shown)


src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicConfigBuilder.java
```diff
@@ -0,0 +1,84 @@
+package org.zalando.nakadi.repository.kafka;
+
+import kafka.admin.RackAwareMode;
+
+import java.util.Optional;
+
+public final class KafkaTopicConfigBuilder {
```
adyach: this is usually created as static class in class which is gonna be build

adyach:
👍

src/main/resources/application.yml
```diff
@@ -48,6 +48,10 @@ nakadi:
       replicaFactor: 1
       retentionMs: 172800000 # 2 days
       rotationMs: 86400000 # 1 day
+    compacted:
+      rotationMs: 3600000 # 1 hour
```
rcillo: I would set Kafka defaults at the project level by now. If we get some feedback we could change it to some other value.

rcillo:
👍

src/main/resources/application.yml
v-stepanov: Ok, makes sense. Will change it.
What about rotationMs ? Kafka default is 7 days, but in Nakadi we have 1 day. What should we define for compacted topics? 1 day?

rcillo: I mean, maybe use what we already used, 1 day. @v-stepanov let's make the min lag 3 hours as well? Like we decided in the previous alignment meeting?

v-stepanov: yes, sorry, I mean I changed to 1 day (not hour) :)

And I have set min lag to 3 hours now as you asked. I think it's a good thing as this is the easiest way to fix archiver problem.

v-stepanov: But also I think if we decided to set min lag to 3 hours - it can make sense to set segment.ms <= 3 hours. So that it can always compact after 3 hours. Because I think it will be difficult to explain users this 1day/3hour/1GB thing :) If we remove 1day it will be a bit easier as there will be only 3hour/1GB thing.

rcillo:
👍

v-stepanov:
👍

PR #913, opened 2018-07-18
Add Log compaction feature toggle

rcillo:
In order to unblock releases, we are disabling this feature, since it's
not production ready yet.

## Deployment Notes
In order to enable the feature, please change the value of the toggle.


lmontrieux:
👍

rcillo:
👍

PR #917, opened 2018-07-26
Don't allow schema root object type to be array

ferbncode:
Don't allow the root object type in schema for business and data category to be of type array

> Zalando ticket : ARUHA-1715 

## Description
Presently, it is allowed to specify the type of root object in schema to be of type `array`. That is now restricted.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Removed partition key fields check
 
-### Removed
-- Removed the ability to specify schema root object type to be of type array.
+### Fixed
+- Fixed and added check for schema root object type to be only of type object.
 
 ## [2.8.2] - 2018-07-31
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - Removed partition key fields check
 
+### Removed
+- Removed the ability to specify schema root object type to be of type array.
+
 ## [2.8.2] - 2018-07-31
 
 ### Removed
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -501,6 +501,11 @@ public class EventTypeService {
             final JSONObject schemaAsJson = new JSONObject(eventTypeSchema);
             final Schema schema = SchemaLoader.load(schemaAsJson);
 
+            if (schemaAsJson.has("type") && schemaAsJson.getString("type").equals("array")
+                    && eventType.getCategory() != EventCategory.UNDEFINED) {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Removed
```
lmontrieux: I think it is more of a 'Fixed' rather than 'Removed', since you are fixing a bug here

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -502,6 +502,11 @@ private void validateSchema(final EventTypeBase eventType) throws InvalidEventTy
             final JSONObject schemaAsJson = new JSONObject(eventTypeSchema);
             final Schema schema = SchemaLoader.load(schemaAsJson);
 
+            if (schemaAsJson.has("type") && schemaAsJson.getString("type").equals("array")
+                    && eventType.getCategory() != EventCategory.UNDEFINED) {
```
lmontrieux: I think that even undefined events must be objects

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -502,6 +502,11 @@ private void validateSchema(final EventTypeBase eventType) throws InvalidEventTy
...
+                throw new InvalidEventTypeException("\"type\" of root element in schema cannot be \"array\"");
```
lmontrieux: Actually it can't be anything else than object

lmontrieux:
👍

ferbncode:
deploy validation please

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -502,6 +502,10 @@ private void validateSchema(final EventTypeBase eventType) throws InvalidEventTy
             final JSONObject schemaAsJson = new JSONObject(eventTypeSchema);
             final Schema schema = SchemaLoader.load(schemaAsJson);
 
+            if (schemaAsJson.has("type") && !schemaAsJson.getString("type").equals("object")) {
```
antban: There are several problems here: 
1. It makes no sense in previous call to SchemaLoader.load(schemaAsJson) if schema type is not "object", so it's better to move this call down
2. In case if "type" exists and is null, then there will be JSONException triggered, probably it is not what was expected, it is better to write something like `schemaAsJson.has("type") && Objects.equals("object", schemaAsJson.getString("type"))`

antban: @ferbncode

antban:
:+1:

antban:
deploy validation please

ferbncode:
:+1:

antban:
:+1:

ferbncode:
:+1:

PR #919, opened 2018-07-30
Fix ISE when parsing number larger than max long

ferbncode:
Nakadi replies with status code 500 when the number being parsed is greater than max long.

> Zalando ticket : ARUHA-1784

## Description
When a user publishes an event with a number greater than max long, the strict json parser fails, and nakadi replies with 500.
## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

### DIFF
--- a/src/test/java/org/zalando/nakadi/domain/BatchFactoryTest.java
+++ b/src/test/java/org/zalando/nakadi/domain/BatchFactoryTest.java
@@ -133,12 +133,9 @@ public class BatchFactoryTest {
         } catch (JSONException e) {}
     }
 
-    @Test
+    @Test(expected = JSONException.class)
     public void testNumberLargerThanMaxLongInEvents() {
         final String events = "[{\"number\": 9223372036854775808 }]";
-        try {
-            BatchFactory.from(events);
-            fail();
-        } catch (JSONException e) {}
+        BatchFactory.from(events);
     }
 }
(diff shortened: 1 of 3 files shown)


rcillo:
👍

ferbncode:
:+1:

ferbncode:
deploy validation please

src/test/java/org/zalando/nakadi/domain/BatchFactoryTest.java
```diff
@@ -132,4 +132,13 @@ public void testGarbageBetweenEvents() {
...
         } catch (JSONException e) {}
     }
+
+    @Test
+    public void testNumberLargerThanMaxLongInEvents() {
+        final String events = "[{\"number\": 9223372036854775808 }]";
+        try {
+            BatchFactory.from(events);
+            fail();
+        } catch (JSONException e) {}
```
v-stepanov: @ferbncode 
You can do that in @Test annotation like `@Test(expected = JSONException.class)` instead of catching exception and calling `fail()`. 
Another thing that will improve - currently if `BatchFactory.from(events);` throws some other exception - your test will be green (but it should be red in such case)

v-stepanov:
👍

ferbncode:
:+1:

ferbncode:
deploy validation please

ferbncode:
:+1:

antban:
:+1:

PR #921, opened 2018-07-30
Extend logging and timeout for topic creation

rcillo:
As the number of topics grow, the time required for Kafka to create a topic grows.


rcillo:
deploy validation please

rcillo:
Well, no one seems to think this is important.

v-stepanov:
@rcillo I actually faced exactly this issue when creating event-types on kafka stack with 63 instances. Sometimes creation of ET failed because it was reaching this 5 second timeout.

PR #925, opened 2018-08-02
Enrichment for log-compacted event types

ferbncode:
Enrich events in log-compacted topics by including the log compaction key at the root of the event

> Zalando ticket : ARUHA-1812.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG


src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
```diff
@@ -21,6 +22,9 @@ public void enrich(final BatchItem batchItem, final EventType eventType) throws
             setPartition(metadata, batchItem);
             setVersion(metadata, eventType);
             batchItem.inject(BatchItem.Injection.METADATA, metadata.toString());
+            if (eventType.getCleanupPolicy() == CleanupPolicy.COMPACT) {
```
v-stepanov: @ferbncode @lmontrieux @rcillo For me it's strange that we do this enrichment for all compacted event-types. In fact we need it only for joins/filtering.

v-stepanov: I mean, as a user interested only in log-compaction it will feel strange for me that the partition key will be taken out of metadata and put to the place where I have business data.

lmontrieux: I get your point, it's weird for people not using stream processing. But if we only do this on demand, we have another issue:
If someone has a log-compacted topic without this enrichment, and decides to use stream processing in the future, he'll be stuck, unless he re-publishes an event for every key that is stored.

v-stepanov: Ok, I get it. Maybe we should then have some additional property for event-type? Something like `joins_prepared = true/false`?

PR #928, opened 2018-08-07
API for timelines without creating a new topic

rcillo:
This API change makes it possible to connect an event type to an
existing topic.

I considered that the timeline is the structure that acts as an interface to a low level storage and not the event type.

Adding a topic to the event type would not be enough because we would also need to specify the "storage". The only entity that is connected to a storage is the timeline so for me it was the logical place to do that. Also, attaching an event type to a topic eliminates the possibility to later on "attach" other topics to an event type.

So I think that allowing for this in the timeline is the most flexible place.

### DIFF
--- a/src/acceptance-test/java/org/zalando/nakadi/webservice/TimelinesControllerAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/webservice/TimelinesControllerAT.java
@@ -1,59 +1,117 @@
 package org.zalando.nakadi.webservice;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.google.common.collect.ImmutableList;
 import com.jayway.restassured.RestAssured;
 import com.jayway.restassured.http.ContentType;
 import org.apache.http.HttpStatus;
 import org.hamcrest.Matchers;
 import org.joda.time.DateTime;
+import org.json.JSONObject;
 import org.junit.Test;
+import org.zalando.nakadi.domain.CleanupPolicy;
+import org.zalando.nakadi.domain.EnrichmentStrategyDescriptor;
+import org.zalando.nakadi.domain.EventCategory;
 import org.zalando.nakadi.domain.EventType;
 import org.zalando.nakadi.domain.EventTypeOptions;
 import org.zalando.nakadi.utils.EventTypeTestBuilder;
+import org.zalando.nakadi.view.TimelineView;
 import org.zalando.nakadi.webservice.utils.NakadiTestUtils;
 
 import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
 
+import static com.jayway.restassured.http.ContentType.JSON;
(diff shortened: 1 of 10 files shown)


adyach:
👍

src/acceptance-test/java/org/zalando/nakadi/webservice/TimelinesControllerAT.java
```diff
@@ -1,59 +1,117 @@
...
+                .body("[1].event_type", equalTo(eventType.getName()))
                 .body("[1].order", Matchers.is(2))
-                .body("[1].storage_id", Matchers.equalTo("default"))
+                .body("[1].storage_id", equalTo("default"))
                 .body("[1].switched_at", dateWithin(1, MINUTES, new DateTime()))
                 .body("[1].cleaned_up_at", nullValue());
     }
 
+    @Test
+    public void testCreatTimelineOnExistentTopic() throws Exception {
```
lmontrieux: Existing topic :)

src/acceptance-test/java/org/zalando/nakadi/webservice/TimelinesControllerAT.java
```diff
@@ -1,59 +1,117 @@
...
+        final EventType eventType = EventTypeTestBuilder.builder()
+                .cleanupPolicy(CleanupPolicy.COMPACT)
+                .category(EventCategory.BUSINESS)
+                .enrichmentStrategies(ImmutableList.of(EnrichmentStrategyDescriptor.METADATA_ENRICHMENT))
+                .build();
+
+        // This event type is created only so that it creates a topic which we will use as an existent topic for the
```
lmontrieux: existing

src/acceptance-test/java/org/zalando/nakadi/webservice/TimelinesControllerAT.java
```diff
@@ -1,59 +1,117 @@
...
+        // new timeline. That's not how this feature will work on production, since topics are not shared between
```
lmontrieux: in production

src/acceptance-test/java/org/zalando/nakadi/webservice/TimelinesControllerAT.java
```diff
@@ -1,59 +1,117 @@
...
+        // KafkaRepository to create a dummy topic.
+        final EventType dummyEventType = EventTypeTestBuilder.builder()
+                .cleanupPolicy(CleanupPolicy.COMPACT)
+                .category(EventCategory.BUSINESS)
+                .enrichmentStrategies(ImmutableList.of(EnrichmentStrategyDescriptor.METADATA_ENRICHMENT))
+                .build();
+
+        NakadiTestUtils.createEventTypeInNakadi(eventType);
+        NakadiTestUtils.createEventTypeInNakadi(dummyEventType);
+
```
lmontrieux: Why are there 2 blank lines here?

rcillo:
We are no longer merging these changes since they are not necessary. We are not going to use KSQL anymore, so that we regain control over the topic creation.

PR #936, opened 2018-08-23
Improved test process by combining tests in a job

adyach:
Improved test process by combining tests in one job, in that way each if checkstyle failed, which is the first and fastest task, the whole build is failed and it is not required to wait 15 minutes until other jobs are fully executed.
The job itself became faster, because unnecessary builds are removed and docker images cache added.


.travis.yml
```diff
@@ -13,41 +13,44 @@ addons:
...
   global:
     - DOCKER_COMPOSE_VERSION="1.7.1"
-  matrix:
-    - TEST_SUITE="./gradlew checkstyle --stacktrace"                  # checkstyle
-    - TEST_SUITE="travis_wait ./gradlew testWithReport --stacktrace"  # unit tests with coverage report
-    - TEST_SUITE="./gradlew fullAcceptanceTest --stacktrace"          # acceptances tests
 
 # run test suits in parallel
-script: "$TEST_SUITE"
+#script: "$TEST_SUITE"
```
lmontrieux: Is this line necessary?

PR #938, opened 2018-08-24
Issue 902: remove network-mode: host

viartemev:
# One-line summary

> Zalando ticket: ARUHA-280 (only if appropriate)

## Description
A few sentences describing the overall goals of the pull request's
commits.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
These should highlight any DB migrations, feature toggles, etc.


adyach:
@viartemev thank you for contribution! the build on travis fails, looks like docker-compose version should be bumped:
```
> Task :nakadi:testInitNakadi FAILED
Version in "./docker-compose.yml" is unsupported. You might be seeing this error because you're using the wrong Compose file version. Either specify a version of "2" (or "2.0") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```
https://travis-ci.org/zalando/nakadi/builds/420017401?utm_source=github_status&utm_medium=notification

viartemev:
@adyach yes, I have fixed it

adyach:
@viartemev 
```
Waiting for Nakadi to start up
Running command: curl http://localhost:8080/health
curl: (7) Failed to connect to localhost port 8080: Connection refused
...
```
already for 15 mins

docker-compose.yml
```diff
@@ -36,27 +34,32 @@ services:
...
+    environment:
+      ZOO_MY_ID: 1
+    volumes:
+    - ./zk-kafka/zookeeper/data:/data
+    - ./zk-kafka/zookeeper/datalog:/datalog
 
   kafka:
-    image: wurstmeister/kafka:1.1.0
-    network_mode: "host"
+    image: confluentinc/cp-kafka:5.0.0
```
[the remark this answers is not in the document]
lmontrieux: Hi @viartemev , thanks a lot for your contribution. I agree with @adyach : Nakadi uses Apache Kafka, so I would like to keep using it as well - even though in practice I don't know that the difference is that much. What you could do is update the version number to 1.1.1 or even 2.0.0, but that's not strictly necessary.

src/acceptance-test/java/org/zalando/nakadi/webservice/StorageControllerAT.java
```diff
@@ -10,6 +11,7 @@
 
 public class StorageControllerAT extends BaseAT {
 
+    @Ignore("Temporary ignore this test (zk_address should be passed as an env argument)")
```
lmontrieux: Wouldn't it be better to fix this test? Quite often, "temporary" becomes too easily "permanent". For example, we have a few tests in this project that have been "temporarily" skipped for several years now :)

docker-compose.yml
```diff
@@ -36,27 +34,32 @@ services:
...
+      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: LISTENER_DOCKER_INTERNAL:PLAINTEXT,LISTENER_DOCKER_EXTERNAL:PLAINTEXT
+      KAFKA_INTER_BROKER_LISTENER_NAME: LISTENER_DOCKER_INTERNAL
+      KAFKA_ZOOKEEPER_CONNECT: "zookeeper:2181"
       KAFKA_BROKER_ID: 0
     volumes:
-      - /var/run/docker.sock:/var/run/docker.sock
\ No newline at end of file
+    - /var/run/docker.sock:/var/run/docker.sock
+    depends_on:
+    - zookeeper
```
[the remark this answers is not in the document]
viartemev: @adyach I think the best option is to add support of EditorConfig. It will solve this kind of problems. What do you think about it? I can create a issue.

src/acceptance-test/java/org/zalando/nakadi/webservice/StorageControllerAT.java
viartemev: @lmontrieux yes, I think so too. For this test, I need to pass a zookeeper docker container address. I have 2 options:
- set up the ip address to some env variable and use it in the test
- add testcontainers support

What do you think?

docker-compose.yml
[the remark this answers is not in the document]
adyach: It makes sense to file an issue for that definitely! For now we can go with the simple fix.

src/acceptance-test/java/org/zalando/nakadi/webservice/StorageControllerAT.java
adyach: Although, I think testcontainers solves different issues (probably bring others), it should be a change not for single test, but to the project in general, and for that it would be nice to understand the advantages(disadvantages). Maybe you can file a ticket for that as well?

regarding ip address, isn't enough `nakadi.zookeeper.brokers` env var ?

bsideup: Did you know that we (testcontainers) have KafkaContainer, so that you don't have to deal with Docker Compose at all? ;)
https://github.com/testcontainers/testcontainers-java/blob/5d3904e1580f581dc42beaaaecd4f5ba3e5bc05d/modules/kafka/src/test/java/org/testcontainers/containers/KafkaContainerTest.java

But if you really want to use Docker Compose, there is DockerComposeContainer which solves your use case without any (known) issues :)

Just fyi :)

adyach: Nice to see a main testcontainers contribute in our repo :)
Thank you for the info, yes, I did know it and it could be a way to go for us. This particular ticket was opened to solve one specific issue, and we moved to testcontainers in discussion. I suggested to open another ticket to clarify details for the whole team then it could be simpler for us to decide.

bsideup: @adyach feel free to mention / ping me in the conversation :)

lmontrieux:
Also fixed in #980

PR #941, opened 2018-08-29
ARUHA-1854 Introduce ordering_instance_ids.

antban:
ARUHA-1854 Introduce ordering_instance_ids.

The field was implemented because of user request.


docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2635,6 +2635,22 @@ definitions:
...
 
+      ordering_instance_ids:
+        type: array
+        description: |
+          This is only an informational field.
+
+          Specifies array of fields that are used to be event identifier.
+
+          This field can be modified at any moment, but event type owners are expected to notify consumer in advance
+          about the change.
```
rcillo: @antban what do you think about further elaborating why someone would use this field? I thought something like: "This field differs from `ordering_key_fields` in the sense that it is useful for comparing two events related to a single entity, but not useful for comparing any two arbitrary events. For streams of Data Capture Change events that are generated by publishing events every time an entity is updated, the usage of this field can explicitly provide a mean for the consumer to reorder the event related to a single entity."

@tfrauenstein What do you think about the proposed explanation above? Please, let us know if you would like to make further amendments.

tfrauenstein: I have edited a more comprehensive definition of these attributes --> see 
https://docs.google.com/document/d/1NdaorNkwqYriHz1eZWynfTuEK-7XOo1Gaw0jAaku1dk/edit

I propose to agree on the proposal and than take it consistently for the PR for Nakadi Event Bus API and the Guidelines, thx.

antban:
:+1:

rcillo:
👍

antban:
:+1:

PR #945, opened 2018-08-30
Add a link to the OpenAPI interactive doc in the README

cbornet:
This is a nice trick to be able to use the Nakadi API quickly by leveraging the swagger-ui hosted by swagger.io

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,8 @@ To stop the running Nakadi:
 
 Please read the [manual](https://zalando.github.io/nakadi/manual.html) for the full API usage details.
 
+If your Nakadi server is started, you can also use the interactive OpenAPI documentation by opening http://petstore.swagger.io/?url=http://localhost:8080/api
+
 ### Creating Event Types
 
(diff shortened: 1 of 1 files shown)


cbornet:
Actually there's a small issue with this : the OAS forces an HTTPS scheme so the requests will fail for `localhost:8080` since there's no SSL support on this path.
It's still useful since you can build your request with all the support of swagger-ui, execute the request, copy/paste the curl command and remove the "s" in "https" but it would be even better if the http scheme were also added to the OAS so the user can choose which scheme to use.

README.md
```diff
@@ -169,6 +169,8 @@ To stop the running Nakadi:
 
 Please read the [manual](https://zalando.github.io/nakadi/manual.html) for the full API usage details.
 
+If your Nakadi server is started, you can also use the interactive OpenAPI documentation by opening http://petstore.swagger.io/?url=http://localhost:8080/api
```
adyach: `https` va `http`

PR #952, opened 2018-09-12
Put nakadi-event-bus-api.yaml under the api/ folder in docker

cbornet:
# One-line summary

Put nakadi-event-bus-api.yaml under the api/ folder in docker

## Description

When starting the server using the IDE or bootRun, the server currently fails to start because the default setup is to look for `nakadi-event-bus-api.yaml` in the root dir whereas it is in the `api/` folder.
This PR sets it to `api/` by default and makes the appropriate change in the Dockerfile.

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER Team Aruha, team-aruha@zalando.de
 
 WORKDIR /
 ADD build/libs/nakadi.jar nakadi.jar
-ADD api/nakadi-event-bus-api.yaml nakadi-event-bus-api.yaml
+ADD api/nakadi-event-bus-api.yaml api/nakadi-event-bus-api.yaml
 
 EXPOSE 8080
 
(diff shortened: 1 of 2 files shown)


cbornet:
With this PR the `test` profile has the same settings than the `default` so there's no need to have a specific section about it in `application.yml`. The beans annotated with `@Profile("!test")` will still not be created if the Spring profile `test` is used.
So unless I've missed something, the tests using the `test` profile should still work (and they seem to work in CI). Do you have something else in mind ?

adyach:
I find it strange to use `test` profile, if there is no test profile :) 
https://github.com/zalando/nakadi/blob/master/src/test/java/org/zalando/nakadi/security/AuthenticationTest.java#L81

cbornet:
There is a `test` profile. It's just not explicit in `application.yml`. I can keep an empty section if you prefer.

adyach:
yes, please, leave an empty section

PR #957, opened 2018-09-21
Use Kafka to retreive broker metadata instead of Zookeeper

### LINKED ISSUE #956: Use Kafka to retreive broker metadata instead of Zookeeper
Currently, Nakadi gets the list of Kafka brokers IP/port and the list of topics from Zookeeper. Since Kafka 0.9 the recommended way is to get these info directly from Kafka and not interact with Zookeeper.
Doing it in Nakadi would have the benefit of uncoupling the ZK used by Nakadi for its own configuration data from the one used by Kafka. You could for instance have 2 ZK clusters : 1 for Kafka and 1 for Nakadi with different configuration.

cbornet:
Fix #956

## Review
- [ ] Tests
- [ ] Documentation
- [ ] CHANGELOG

## Deployment Notes
Need to configure the `bootstrap-servers` property

### DIFF
--- a/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
+++ b/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
@@ -8,7 +8,6 @@ import org.json.JSONException;
 import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.util.StringUtils;
 import org.zalando.nakadi.repository.zookeeper.ZooKeeperHolder;
 
@@ -16,6 +15,9 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
 
 public class KafkaLocationManager {
 
@@ -23,22 +25,23 @@ public class KafkaLocationManager {
     private static final String BROKERS_IDS_PATH = "/brokers/ids";
 
     private final ZooKeeperHolder zkFactory;
-    private final Properties kafkaProperties;
+    private final Properties kafkaProperties = new Properties();
     private final KafkaSettings kafkaSettings;
 
     public KafkaLocationManager(final ZooKeeperHolder zkFactory, final KafkaSettings kafkaSettings) {
         this.zkFactory = zkFactory;
+        this.kafkaSettings = kafkaSettings;
         if (StringUtils.isEmpty(kafkaSettings.getBootstrapServers())) {
-            this.kafkaProperties = buildKafkaProperties(fetchBrokers());
+            updateBrokers();
+            final ScheduledExecutorService executor = Executors.newScheduledThreadPool(1);
+            executor.scheduleWithFixedDelay(this::updateBrokers,30, 30, TimeUnit.SECONDS);
         } else {
-            this.kafkaProperties = new Properties();
-            this.kafkaProperties.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG,
+            kafkaProperties.setProperty(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG,
                     kafkaSettings.getBootstrapServers());
         }
-        this.kafkaSettings = kafkaSettings;
     }
 
-    static class Broker {
+    private static class Broker {
         final String host;
         final Integer port;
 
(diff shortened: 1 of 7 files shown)


src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -1,101 +1,21 @@
...
-import org.slf4j.LoggerFactory;
-import org.springframework.scheduling.annotation.Scheduled;
-import org.zalando.nakadi.repository.zookeeper.ZooKeeperHolder;
 
-import java.io.UnsupportedEncodingException;
-import java.util.ArrayList;
-import java.util.List;
 import java.util.Properties;
 
 public class KafkaLocationManager {
```
cbornet: The remaining methods in this class could probably be moved to `KafkaFactory` and this class removed.

cbornet: Just tell me if I do it.

cbornet:
Note : if it's desirable to keep the reading of Kafka bootstrap servers from ZK for backward-compatibility or to reduce the number of configuration, we could also have a mechanism where if `nakadi.kafka.bootstrap-servers` is empty, then we query ZK.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -73,19 +80,19 @@ public String toString() {
...
         final Properties props = new Properties();
-        props.put("bootstrap.servers", buildBootstrapServers(brokers));
+        props.put(CommonClientConfigs.BOOTSTRAP_SERVERS_CONFIG, buildBootstrapServers(brokers));
         return props;
     }
 
     @Scheduled(fixedDelay = 30000)
     private void updateBrokers() {
-        if (kafkaProperties != null) {
+        if (StringUtils.isEmpty(kafkaSettings.getBootstrapServers()) && kafkaProperties != null) {
```
adyach: it means that you run a thread every 30 seconds even though the condition won't be fulfilled ever. could we change  it to have a scheduled action only if we need it?

cbornet: I've thought about `@ConditionalOnProperty` but I don't know if it works there. I'll give it a shot.

cbornet: Are you sure the scheduled task gets called with the current code ? KafkaLocationManager is not a bean (it used to) so it will not be scanned. I checked by adding a breakpoint and I'm never called. This looks like a bug...

adyach: Yes, this is definitely a bug, `@Component` was gone during refactoring, when we started to using it without injections. could you please fix it (using scheduled thread pool will be fine with me) ? it will also help us to avoid unnecessary starts of thread when we start with bootstrap servers.
Thank you for catching it up !

adyach:
Checkstyle fails: https://travis-ci.org/zalando/nakadi/jobs/432844850
`/home/travis/build/zalando/nakadi/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java:39: Line is longer than 120 characters (found 123). [LineLength]`

cbornet:
Fixed style

adyach:
👍

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -96,11 +96,11 @@ public KafkaTopicRepository(final ZooKeeperHolder zkFactory,
...
     }
 
-    public List<String> listTopics() throws TopicRepositoryException {
+    public Set<String> listTopics() throws TopicRepositoryException {
         try {
-            return zkFactory.get()
-                    .getChildren()
-                    .forPath("/brokers/topics");
+            return kafkaFactory.getConsumer()
+                    .listTopics()
```
rcillo: I would not stop merging this PR since it introduces a new relevant feature, but this change is not strictly related and I'm not confident about the performance impact. Is listing from ZK faster or slower than getting new consumer, and fetching all metadata? It could be faster. It could be slower.

rcillo:
Thank you for the contribution @cbornet it's really appreciated and of high quality.

rcillo:
👍

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: You are right. To be honest, I went through the code and only found usages in tests. So, besides the already existent bad smell, no problem here.

cbornet:
I just realized something is completely wrong with my PR : it only works with 1 Kafka and will fail if there are multiple Kafka/Storages/timelines. So bootstrap-servers should be a `Storage` property.

PR #970, opened 2018-10-29
Remove logging for something that should not be logged

antban:
Remove logging for something that should not be logged


antban:
:+1:

adyach:
👍

PR #975, opened 2018-11-02
Added updated_at to subscription

Kunal-Jha:
# Track subscription updated_at attribute and expose it in API  

> Zalando ticket : ARUHA-1954

Added the field updated_At and updated it at the time of updation


v-stepanov:
@Kunal-Jha 
I think there are several issues here to work on.
1) What value the new field "updated_at" will have for already existing subscriptions?
2) What value the new field "updated_at" will have for newly created subscriptions (that was never updated)?
3) As you added new field - it would be nice to write a test for that. I think extending SubscriptionAT would work.
4) Please also update the [API documentation file](https://github.com/zalando/nakadi/blob/master/docs/_data/nakadi-event-bus-api.yaml).

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2423,6 +2424,14 @@ definitions:
           specified when creating subscription and sending it may result in a client error.
         format: RFC 3339 date-time
         example: '1996-12-19T16:39:57-08:00'
+      updated_at:
+        type: string
+        readOnly: true
+        description: |
+          Timestamp of last updation of the subscription. This is generated by Nakadi. It should not be
```
v-stepanov: I'm not sure if "updation" is a correct word. 
@lmontrieux?

src/main/java/org/zalando/nakadi/domain/Subscription.java
```diff
@@ -23,8 +17,18 @@ public Subscription(final String id, final DateTime createdAt, final Subscriptio
         super(subscriptionBase);
         this.id = id;
         this.createdAt = createdAt;
+        this.updatedAt=createdAt;
```
v-stepanov: I understand that it will do the trick, but initializing it here like that will introduce a hidden behavior that will cause problems at some point. I think its initialization should be more explicit.

src/test/java/org/zalando/nakadi/utils/RandomSubscriptionBuilder.java
```diff
@@ -83,6 +83,7 @@ public Subscription build() {
         final Subscription subscription = new Subscription();
         subscription.setId(id);
         subscription.setCreatedAt(createdAt);
+        subscription.setUpdatedAt(createdAt);
```
v-stepanov: and here as well. It introduces some hidden trick which may bring problems in the future.

docs/_data/nakadi-event-bus-api.yaml
lmontrieux: Definitely not a word :)
You're looking for `update`

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -144,6 +146,7 @@ public Subscription updateSubscription(final String subscriptionId, final Subscr
 
         subscriptionValidationService.validateSubscriptionChange(old, newValue);
         old.mergeFrom(newValue);
+        old.setUpdatedAt(new DateTime());
```
v-stepanov: `DateTimeZone.UTC`

docs/_data/nakadi-event-bus-api.yaml
v-stepanov: @Kunal-Jha please address this problem.

src/main/java/org/zalando/nakadi/domain/Subscription.java
v-stepanov: @Kunal-Jha I checked usage of this constructor. It's called from `SubscriptionDbRepository.createSubscription(...)` + 3 usage in tests. For me it would make sense to add one more parameter to this constructor and pass the value from the code that calls it. That would be "more natural" and will not introduce hidden behavior.

v-stepanov:
@Kunal-Jha just one thing we forgot - can you please create update script for DB to set this value for all existing subscriptions? We put such scripts to `database/migration`

database/migration/aruha-1954-subscription-add-field.sql
```diff
@@ -0,0 +1,3 @@
+SET ROLE zalando_nakadi_data_owner;
+
+UPDATE zn_data.subscription SET s_subscription_object = jsonb_set(s_subscription_object,'{updated_at}',concat('"',s_subscription_object->>'created_at','"' )::jsonb, true);
```
v-stepanov: please don't forget about code formatting in this query.
and also endline at the end of the file.

v-stepanov:
👍

Kunal-Jha:
👍

Kunal-Jha:
👍

v-stepanov:
👍

Kunal-Jha:
deploy validation please

rcillo:
deploy validation please

PR #977, opened 2018-11-09
Fix error when access to timelines is forbidden

lmontrieux:
# Fix error when access to timelines is forbidden

## Description
Getting the list of timelines requires nakadi.read access, but the error message mentions admin on the event type, which is incorrect.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
+++ b/src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
@@ -15,8 +15,8 @@ import org.zalando.nakadi.domain.CleanupPolicy;
 import org.zalando.nakadi.domain.DefaultStorage;
 import org.zalando.nakadi.domain.EventType;
 import org.zalando.nakadi.domain.EventTypeBase;
-import org.zalando.nakadi.domain.EventTypeResource;
 import org.zalando.nakadi.domain.NakadiCursor;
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -337,8 +337,8 @@ private void finishTimelineUpdate(final String eventTypeName) throws TimelineExc
     public List<Timeline> getTimelines(final String eventTypeName)
             throws AccessDeniedException, UnableProcessException, TimelineException, NotFoundException {
         if (!adminService.isAdmin(AuthorizationService.Operation.READ)) {
-            throw new AccessDeniedException(AuthorizationService.Operation.ADMIN,
-                    new EventTypeResource(eventTypeName, null));
+            throw new AccessDeniedException(AuthorizationService.Operation.READ,
+                    new NakadiResource("nakadi", null));
```
adyach:
```suggestion
                    new NakadiResource(eventTypeName, null));
```

adyach: or timelines

adyach:
@lmontrieux what is the status here?

antban:
Unfortunately this PR is outdated, closing

Issue #951, opened 2018-09-12
Replace AuthMode.OFF by AuthMode.NONE


cbornet:
When using application.yml to configure `nakadi.oauth2.mode` YAML parser translates `OFF` to `false` if you don't put quotes and the Spring property binding fails with 
```
Caused by: java.lang.IllegalStateException: Cannot convert value of type 'java.lang.String' to required type 'org.zalando.nakadi.config.SecuritySettings$AuthMode': no matching editors or conversion strategy found
	at org.springframework.beans.TypeConverterDelegate.convertIfNecessary(TypeConverterDelegate.java:307)
	at org.springframework.beans.TypeConverterDelegate.convertIfNecessary(TypeConverterDelegate.java:109)
	at org.springframework.beans.TypeConverterSupport.doConvert(TypeConverterSupport.java:64)
	... 98 more
```
The workaround is to add quotes like `nakadi.oauth2.mode: "OFF"` but it is a bit awkward and you can miss it when running your service.

Adding a NONE value in the enum that has the same effect as OFF would prevent that and you could keep OFF for backward compatibility.

lmontrieux:
Thank you @cbornet for your suggestion. This looks like a nice improvement, and could be a good first issue for someone looking into contributing to Nakadi.

adyach:
> The workaround is to add quotes like nakadi.oauth2.mode: "OFF" but it is a bit awkward and you can miss it when running your service.
@lmontrieux @cbornet could you please elaborate on that? for me it is not clear what is the issue having quotes.

PR #978, opened 2018-11-09
#951 Replace AuthMode.OFF by AuthMode.NONE

fbrns:
# AuthMode.NONE is added as an addition to AuthMode.OFF

## Description
When using application.yml to configure nakadi.oauth2.mode YAML parser translates OFF to false if it isn't put into quotes. Adding a NONE value in the enum that has the same effect as OFF prevents that and OFF can be kept for backward compatibility.

## Review
- [ ] Tests
- [ ] Documentation


adyach:
@fbrns thank you for contribution!
can we just put `OFF` in quotes? I also see that this behaviour is changed for yaml 1.2, which interprets it as not a boolean. we have never observed such behaviour, most probably, the reason is that we use Nakadi with docker compose.

fbrns:
Putting `OFF` into quotes could a quick but a bit awkward workaround (@cbornet, https://github.com/zalando/nakadi/issues/951). However with yaml 1.2 the situation will change somewhere in the future, but then a new state would not harm as well.

The behavior can be reproduced by removing `NAKADI_OAUTH2_MODE=OFF` from `docker-compose.yml` and adding `OFF` to `application.yml`.

adyach:
@fbrns I do not understand why it is a problem to put it in quotes. From the other hand adding another identical auth mode looks like duplication and creates complication in understanding what purpose it serves. This state was created specifically for testing/debugging purposes thus it is used in conjunction with docker compose.

fbrns:
The functionality is not needed, therefore I close the pull-request.

PR #983, opened 2018-11-21
Access log Bug on unauthorized subscription (WIP)

Kunal-Jha:
> Zalando ticket: ARUHA-1391 

## Description
access log filter check added in order to mitigate discrepancy between user message and access logs.


ferbncode:
I approved this by mistake. I'll try to fix this :)

adyach:
@Kunal-Jha it looks like in that case we will see 102 in the access log, but 403 is expected
`GET "/subscriptions/1b856cad-9156-4157-badf-a70266bbbd12/events" "curl/7.54.0" "-" 102 12ms "-" "-" 0B`

adyach:
@Kunal-Jha could we close it?

PR #985, opened 2018-11-29
Instance based subscription cache for cursor service

adyach:
# One-line summary
Instance based subscription cache for cursor service, it is not distributed caching solution rather simple improvement to tolerate database failures and unnecessary calls to database.
The lazing caching is used, invalidation happens only if deletion of subscription happens on the same instance or 5 minutes interval after last access by key.


src/main/java/org/zalando/nakadi/service/SubscriptionCache.java
```diff
@@ -0,0 +1,68 @@
...
+import org.zalando.nakadi.domain.Subscription;
+import org.zalando.nakadi.exceptions.runtime.NoSuchSubscriptionException;
+import org.zalando.nakadi.exceptions.runtime.ServiceTemporarilyUnavailableException;
+import org.zalando.nakadi.repository.db.SubscriptionDbRepository;
+
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * The caching works only inside the instance of the application, it does not sync cahce across the instances, that's
```
rcillo: `not sync cahce/not sync cache`

src/main/java/org/zalando/nakadi/service/SubscriptionCache.java
```diff
@@ -0,0 +1,68 @@
...
+     * @throws ServiceTemporarilyUnavailableException
+     */
+    public Subscription getSubscription(final String subscriptionId)
+            throws NoSuchSubscriptionException, ServiceTemporarilyUnavailableException {
+        try {
+            return subscriptionsCache.get(subscriptionId);
+        } catch (final UncheckedExecutionException e) {
+            final Throwable cause = e.getCause();
+            if (cause instanceof NoSuchSubscriptionException) {
+                throw new NoSuchSubscriptionException(e.getCause().getMessage());
```
antban: Why not to throw cause ?

src/main/java/org/zalando/nakadi/service/SubscriptionCache.java
```diff
@@ -0,0 +1,68 @@
...
+            }
+            throw new ServiceTemporarilyUnavailableException(e.getCause().getMessage());
```
antban: Why not to add cause?

src/main/java/org/zalando/nakadi/service/SubscriptionCache.java
```diff
@@ -0,0 +1,68 @@
...
+        } catch (final ExecutionException e) {
+            throw new ServiceTemporarilyUnavailableException(e.getCause().getMessage());
```
antban: The same question

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -145,6 +148,7 @@ public Subscription updateSubscription(final String subscriptionId, final Subscr
         subscriptionValidationService.validateSubscriptionChange(old, newValue);
         old.mergeFrom(newValue);
         subscriptionRepository.updateSubscription(old);
+        subscriptionCache.invalidateSubscription(subscriptionId);
```
antban: I believe this call is useless when we have 150 instances

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -213,6 +217,7 @@ public void deleteSubscription(final String subscriptionId)
         nakadiKpiPublisher.publish(subLogEventType, () -> new JSONObject()
                 .put("subscription_id", subscriptionId)
                 .put("status", "deleted"));
+        subscriptionCache.invalidateSubscription(subscriptionId);
```
antban: I believe this call is useless when we have 150 instances

rcillo:
👍 this is a great addition of value. I would even say that it's not necessary to invalidate on the instance. Rather, add a sentence to the API saying that subscription deletion is eventually consistent and cached.

Do you know what's the behaviour of a commit made to a subscription that has been deleted but the cache did not expire yet? Does it fail with 500?

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
adyach: but what if you have 1 instance?

rcillo:
@andrey if it would not be too much work to gracefully handle this 500, that would be ideal.

rcillo:
@adyach I'm even fine with adding a ticket for that, if it's too much work, I would approve this PR.

adyach:
@rcillo it is quite hard to avoid it, because cache is not distributed. I have left cache only for commits, so there will be no 500s, but I will have 422 due to invalid session. In order to avoid 422, I will have to add additional call to zk, when session is invalid, I decided not to do it at this point

rcillo:
👍

adyach:
👍

adyach:
👍

rcillo:
👍

adyach:
👍

rcillo:
👍

PR #989, opened 2018-12-21
Replace Kafka producer on timeout

adyach:
## Description
During rolling restart of Kafka cluster it was noticed, that Kafka client fails to request metadata in 60 seconds, after broker was restarted. Only fix is to restart Nakadi cluster.
```
org.apache.kafka.common.errors.TimeoutException: Failed to update metadata after 60000 ms.
```


rcillo:
👍

PR #994, opened 2019-01-04
Log 403 for blacklisted in ResponseEntity with body outputStream

ferbncode:
# One-line summary
> Zalando ticket : ARUHA1391- (https://jira.zalando.net/browse/ARUHA-1391)
## Description
Fixes Access Log status to `403` for blacklisted consumption. One idea to fix status code for other similar problems (like log for `404`) would be to maintain status in `StreamingContext`.

Moved checks for conflicting partitions, no slots available, no subscription and authorization/blacklist check to `SubscriptionValidationService`.


antban:
To be honest, I do not like solution here. 
The reason why different conditions are verified during async request handling is the fact that there are some locks that should be done and so on. I thought that it makes sense to do it differently in ```LoggingFilter.java```:
```
if (resp.isAsyncStarted()) {
    resp.getAsyncContext().addListener(**asyncLoggingListenerInstance**); // This part is missing
} else {
   log(req, resp);
}
```
Right now there are 2 places to figure out if it possible to check if it is allowed to connect (i mean 409)
One of them is used just for logging, second - is the real check with zk lock.
With this approach we won't be able to figure out if 200 is 200 or 409.

PS: I do not know if this approach is working, but I thought that it should.

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -34,47 +36,117 @@ public LoggingFilter(final NakadiKpiPublisher nakadiKpiPublisher,
...
+        private long startTime;
+
+        private AsyncRequestListener(final HttpServletRequest request, final HttpServletResponse response,
+                                    final long startTime) {
+            this.request = request;
+            this.response = response;
+            this.startTime = startTime;
+
+            final RequestLogInfo requestLogInfo = new RequestLogInfo(request);
+            ACCESS_LOGGER.info("{} \"{}{}\" \"{}\" \"{}\" \"{}\" \"{}\"",
```
ferbncode: Does it make more sense to log the stream parameters in the state machine? This will log twice for each request to '/subscriptions/{}/events'. I logged it here as we might need the user who made the request.

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -34,47 +36,117 @@ public LoggingFilter(final NakadiKpiPublisher nakadiKpiPublisher,
...
+
+        }
+    }
+
     @Override
     protected void doFilterInternal(final HttpServletRequest request,
                                     final HttpServletResponse response, final FilterChain filterChain)
             throws ServletException, IOException {
         final long start = System.currentTimeMillis();
+        boolean willBeLogged = false;
```
adyach: can it be exchanged with `request.isAsyncStarted()` ?

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
adyach: I do not follow why you logged it here, the user is obtained in `RequestLogInfo`, so it will be logged in callback, right?

ferbncode: Yes, it will be logged in the callback. We see the user who requested the stream when the stream is closed (in the case of 200). I wanted to ask is it important to see the user when the request is initiated. Or is it fine to log just the stream parameters in the state machine?

adyach:
👍

adyach:
deploy validation please

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -34,47 +38,123 @@ public LoggingFilter(final NakadiKpiPublisher nakadiKpiPublisher,
...
+                    HttpStatus.PROCESSING.value(),
+                    0,
+                    requestLogInfo.contentEncoding,
+                    requestLogInfo.acceptEncoding,
+                    requestLogInfo.contentLength);
+        }
+
+        @Override
+        public void onComplete(final AsyncEvent event) {
+            FlowIdUtils.push(this.flowId);
```
adyach: you also need to clean up after that, because the flowid will be left on the thread which calls AsyncRequestListener

adyach:
deploy validation please

adyach:
👍

ferbncode:
:+1:

ferbncode:
:+1:

ferbncode:
:+1:

adyach:
👍

PR #998, opened 2019-01-10
ARUHA-2114 Comparison operation for schema changes should always work on non-null values

antban:
ARUHA-2114 Comparison operation for schema changes should always work on non-null values

### DIFF
--- a/src/main/java/org/zalando/nakadi/domain/SchemaChange.java
+++ b/src/main/java/org/zalando/nakadi/domain/SchemaChange.java
@@ -2,26 +2,42 @@ package org.zalando.nakadi.domain;
 
 public class SchemaChange {
     public enum Type {
-        ID_CHANGED,
-        DESCRIPTION_CHANGED,
-        TITLE_CHANGED,
-        PROPERTIES_ADDED,
-        SCHEMA_REMOVED,
-        TYPE_CHANGED,
-        NUMBER_OF_ITEMS_CHANGED,
-        PROPERTY_REMOVED,
-        DEPENDENCY_ARRAY_CHANGED,
-        DEPENDENCY_SCHEMA_CHANGED,
-        COMPOSITION_METHOD_CHANGED,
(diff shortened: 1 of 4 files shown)


src/main/java/org/zalando/nakadi/domain/SchemaChange.java
```diff
@@ -2,26 +2,42 @@
...
+        PROPERTY_REMOVED(Version.Level.MAJOR),
+        DEPENDENCY_ARRAY_CHANGED(Version.Level.MAJOR),
+        DEPENDENCY_SCHEMA_CHANGED(Version.Level.MAJOR),
+        COMPOSITION_METHOD_CHANGED(Version.Level.MAJOR),
+        ATTRIBUTE_VALUE_CHANGED(Version.Level.MAJOR),
+        ENUM_ARRAY_CHANGED(Version.Level.MAJOR),
+        SUB_SCHEMA_CHANGED(Version.Level.MAJOR),
+        DEPENDENCY_SCHEMA_REMOVED(Version.Level.MAJOR),
+        REQUIRED_ARRAY_CHANGED(Version.Level.MAJOR),
+        REQUIRED_ARRAY_EXTENDED(Version.Level.MINOR, Version.Level.MAJOR),
```
rcillo: `REQUIRED_ARRAY_EXTENDED` used to be `MINOR` for `forward` but `major` for `compatible`.

antban: @rcillo Is it changed?

antban: I mean it is exactly as it was: MINOR for forward and MAJOR for compatible.

rcillo: 😞 my bad, I did not read it through.

rcillo:
I did not actually understood where the problem was. Was there a change type missing for one of the compatibility modes?

antban:
:+1:

PR #999, opened 2019-01-11
ARUHA-2110 Do not close streaming connections in case of kafka address change

antban:
ARUHA-2110 Do not close streaming connections in case of kafka address change


adyach:
👍

antban:
:+1:

PR #1003, opened 2019-01-15
Kafka producer crutch

adyach:
# One-line summary
Producer has to be recreated in case of Kafka broker ip was changed, otherwise it reports that broker with old ip is not available.

### DIFF
--- a/src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
+++ b/src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
@@ -14,6 +14,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
+import java.io.Closeable;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
@@ -208,7 +209,7 @@ public class KafkaFactory {
         }
     }
 
-    private class KafkaCrutch {
+    private class KafkaCrutch implements Closeable {
 
         private final KafkaLocationManager kafkaLocationManager;
         private final Runnable brokerIpAddressChangeListener;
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -198,6 +198,7 @@ public boolean topicExists(final String topic) throws TopicRepositoryException {
             item.updateStatusAndDetail(EventPublishingStatus.FAILED, "internal error");
             throw new EventPublishingException("Error publishing message to kafka", e);
         } catch (final RuntimeException e) {
+            kafkaFactory.terminateProducer(producer);
```
antban: I thought that we should do that when kafka reports about exception, not (which it detected by listener in producer.send())

antban: Sorry, ignore this comment, I was wrong

src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -183,4 +184,47 @@ public KafkaCrutchException(final String message) {
...
+        }
+
+        @Override
+        public void close() {
+            kafkaCrutch.close();
+            super.close();
+        }
+    }
+
+    private class KafkaCrutch {
```
antban: Can you make it Closeable?

antban:
I like everything, there is only one suggestion

antban:
:+1:

antban:
:+1:

adyach:
👍

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
v-stepanov: @antban can you provide details regarding this? Because it was also not clear for me why we terminate producer by getting any RuntimeException.

PR #1004, opened 2019-01-17
ARUHA-2124 Upgrade plugin API version and refactor

lmontrieux:
# Upgrade plugin API version and refactor some come

> Zalando ticket : ARUHA-2124

## Description
This PR upgrades plugin API to 3.0.1. There is an additional filter(..) method in the plugin API interface, which by default returns the same list as the input.
Classes implementing Resource were also refactored

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes
Requires new version of any authorization plugin


src/main/java/org/zalando/nakadi/domain/Permission.java
```diff
@@ -43,11 +43,26 @@ public boolean equals(final Object o) {
         final Permission that = (Permission) o;
         return Objects.equals(resource, that.resource) &&
                 Objects.equals(operation, that.operation) &&
-                Objects.equals(authorizationAttribute, that.authorizationAttribute);
+                Objects.equals(authorizationAttribute.getDataType(), that.authorizationAttribute.getDataType()) &&
+                Objects.equals(authorizationAttribute.getValue(), that.authorizationAttribute.getDataType());
```
Kunal-Jha: I guess the value should be compared to  value 
```suggestion
                Objects.equals(authorizationAttribute.getValue(), that.authorizationAttribute.getValue());
```

src/main/java/org/zalando/nakadi/domain/Permission.java
```diff
@@ -43,11 +43,26 @@ public boolean equals(final Object o) {
...
+    public String toString() {
+        final StringBuilder sb = new StringBuilder();
+        sb.append("Permission {");
+        sb.append("resource: ");
+        sb.append(resource);
+        sb.append("\tOperation: ");
+        sb.append(operation.toString());
+        sb.append("\tAuthorization attribute: ");
+        sb.append(authorizationAttribute.toString());
+        sb.append("}.");
```
v-stepanov: why not use chain calls here? `.append(...).append(...).append(...)`

src/main/java/org/zalando/nakadi/domain/ResourceImpl.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Optional;
+
+public class ResourceImpl<T> implements Resource {
+
+    public static final String ALL_DATA_ACCESS_RESOURCE = "all-data-access";
+    public static final String ADMIN_RESOURCE = "nakadi";
+    public static final String EVENT_TYPE_RESOURCE = "event-type";
+    public static final String SUBSCRIPTION_RESOURCE = "subscription";
+
+    private final T resource;
```
v-stepanov: How this field is used? I could not find usages in Nakadi, so probably it is used in a plugin. Can you please share the link to appropriate code?

lmontrieux: @v-stepanov it would be used to filter resources according to some criteria implemented in a plugin: https://github.com/zalando-nakadi/nakadi-plugin-api/pull/4

src/main/java/org/zalando/nakadi/domain/Permission.java
lmontrieux: I guess I must have been distracted. Will fix this right away

lmontrieux: Fixed in 0c0d075

v-stepanov:
👍

lmontrieux:
👍

PR #1005, opened 2019-01-17
retries for kafka client

adyach:
# One-line summary
Retries for producer and retries for getting stats/partitions.

### DIFF
--- a/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
+++ b/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
@@ -120,7 +120,8 @@ public class KafkaLocationManager {
         producerProps.put(ProducerConfig.COMPRESSION_TYPE_CONFIG, "lz4");
         producerProps.put(ProducerConfig.MAX_REQUEST_SIZE_CONFIG, kafkaSettings.getMaxRequestSize());
         producerProps.put(ProducerConfig.DELIVERY_TIMEOUT_MS_CONFIG, kafkaSettings.getDeliveryTimeoutMs());
-        producerProps.put(ProducerConfig.RETRIES_CONFIG, 0);
+        producerProps.put(ProducerConfig.RETRIES_CONFIG, 3);
+        producerProps.put(ProducerConfig.MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION, 1);
         producerProps.put(ProducerConfig.MAX_BLOCK_MS_CONFIG, kafkaSettings.getMaxBlockMs());
         return producerProps;
     }
--- a/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -463,7 +463,7 @@ public class KafkaTopicRepository implements TopicRepository {
         }
     }
 
-        @Override
+    @Override
     public EventConsumer.LowLevelConsumer createEventConsumer(
             @Nullable final String clientId, final List<NakadiCursor> cursors)
             throws ServiceTemporarilyUnavailableException, InvalidCursorException {
--- a/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -15,6 +15,7 @@ import org.apache.kafka.common.errors.NotLeaderForPartitionException;
 import org.apache.kafka.common.errors.TopicExistsException;
 import org.apache.kafka.common.errors.UnknownServerException;
 import org.apache.kafka.common.errors.UnknownTopicOrPartitionException;
+import org.echocat.jomon.runtime.concurrent.RetryForSpecifiedCountStrategy;
 import org.echocat.jomon.runtime.concurrent.RetryForSpecifiedTimeStrategy;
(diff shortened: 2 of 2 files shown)


src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -96,6 +97,65 @@ public KafkaTopicRepository(final ZooKeeperHolder zkFactory,
         this.circuitBreakers = new ConcurrentHashMap<>();
     }
 
+    private static CompletableFuture<Exception> publishItem(
```
adyach: reformatting

src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -120,7 +120,8 @@ public Properties getKafkaProducerProperties() {
         producerProps.put(ProducerConfig.COMPRESSION_TYPE_CONFIG, "lz4");
         producerProps.put(ProducerConfig.MAX_REQUEST_SIZE_CONFIG, kafkaSettings.getMaxRequestSize());
         producerProps.put(ProducerConfig.DELIVERY_TIMEOUT_MS_CONFIG, kafkaSettings.getDeliveryTimeoutMs());
-        producerProps.put(ProducerConfig.RETRIES_CONFIG, 0);
+        producerProps.put(ProducerConfig.RETRIES_CONFIG, 3);
+        producerProps.put(ProducerConfig.MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION, 1);
```
antban: Why?
As far as I know, we have 1 producer. 1 request per connection is a strange thing.

antban: Also, I see in configuration that this property is not prerequisite for retries.

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -305,6 +306,20 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
...
     public List<Optional<PartitionStatistics>> loadPartitionStatistics(
             final Collection<TimelinePartition> partitions) throws ServiceTemporarilyUnavailableException {
+        try {
+            return Retryer.executeWithRetry(() -> {
+                        return loadPartitionStatisticsInternal(partitions);
+                    },
+                    new RetryForSpecifiedCountStrategy(3)
+                            .withWaitBetweenEachTry(5000)
+                            .withExceptionsThatForceRetry(org.apache.kafka.common.errors.TimeoutException.class));
+        } catch (final Exception e) {
```
antban: RuntimeException?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
adyach: this is due to `retries` configuration:
```
Allowing retries without setting max.in.flight.requests.per.connection to 1 will potentially change the ordering of records because if two batches are sent to a single partition, and the first fails and is retried but the second succeeds, then the records in the second batch may appear first.
```
https://kafka.apache.org/documentation/#producerconfigs

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -344,14 +359,25 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
...
     public List<PartitionStatistics> loadTopicStatistics(final Collection<Timeline> timelines)
             throws ServiceTemporarilyUnavailableException {
+        try {
+            return Retryer.executeWithRetry(() -> {
+                        return loadTopicStatisticsInternal(timelines);
+                    },
+                    new RetryForSpecifiedCountStrategy(3)
+                            .withWaitBetweenEachTry(5000)
+                            .withExceptionsThatForceRetry(org.apache.kafka.common.errors.TimeoutException.class));
+        } catch (final Exception e) {
```
antban: RuntimeException?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -375,14 +401,25 @@ private void failUnpublished(final List<BatchItem> batch, final String reason) {
...
     public List<PartitionEndStatistics> loadTopicEndStatistics(final Collection<Timeline> timelines)
             throws ServiceTemporarilyUnavailableException {
+        try {
+            return Retryer.executeWithRetry(() -> {
+                        return loadTopicEndStatisticsInternal(timelines);
+                    },
+                    new RetryForSpecifiedCountStrategy(3)
+                            .withWaitBetweenEachTry(5000)
+                            .withExceptionsThatForceRetry(org.apache.kafka.common.errors.TimeoutException.class));
+        } catch (final Exception e) {
```
antban: RuntimeException?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
antban: Yes, but for us it is working on a different level, cause we have many nakadi instances, so for us anyways real value = max.in.flight.requests.per.connection * amount of nakadi instances

antban:
:+1:

adyach:
deploy validation please

PR #1006, opened 2019-01-18
Get user from subject name in LoggingFilter

ferbncode:
# One-line summary

> Zalando ticket : ARUHA- (https://jira.zalando.net/browse/ARUHA-2122)

## Description
Added a filter for adding X-consumer header to context. [WIP]. I'll move code to plugin as it should be.


src/main/java/org/zalando/nakadi/filters/ConsumerHeaderFilter.java
```diff
@@ -0,0 +1,43 @@
...
+    @Override
+    protected void doFilterInternal(final HttpServletRequest request, final HttpServletResponse response,
+                                    final FilterChain filterChain) {
+
+        final Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
+        if (authentication instanceof OAuth2Authentication) {
+            final OAuth2Authentication oAuth2Authentication = (OAuth2Authentication) authentication;
+            final String eid = (String) authentication.getPrincipal();
+            final UsernamePasswordAuthenticationToken token = new UsernamePasswordAuthenticationToken(
+                    new Subject(eid, request.getHeader("X-Consumer")), null);
```
rcillo: Do we need this in the open source project?

ferbncode: You're right. I am moving this to the plugin. I will update the pull request.

rcillo:
@ferbncode there's a conflict with this one.

src/main/java/org/zalando/nakadi/plugin/auth/DefaultAuthorizationService.java
```diff
@@ -23,7 +25,8 @@ public boolean isAuthorizationAttributeValid(final AuthorizationAttribute author
 
     @Override
     public Subject getSubject() {
-        return () -> SecurityContextHolder.getContext().getAuthentication().getName();
+        return () -> Optional.ofNullable(
```
rcillo: Is it going to return `null` in case the request is not authenticated?

ferbncode: Yes. Do you think it should be different? (The correct user will be covered by the plugin for OAuth2 requests) but for local nakadi, the user would be `null` (as it is presently))

rcillo: There are potentially other endpoints that are not authenticated, right? I guess "metrics" or "health" as well. We try to avoid returning `null` in this project. I see that changing it to `Option` could lead to extra effort and releasing a new `api` version (which takes hours). Let me know how you want to proceed. I would say we better change the API and return `Optional` or we document it somehow with `@Nullable`.

rcillo:
👍

ferbncode:
:+1:

src/main/java/org/zalando/nakadi/security/ClientResolver.java
```diff
@@ -42,7 +48,8 @@ public Client resolveArgument(final MethodParameter parameter,
                                   final ModelAndViewContainer mavContainer,
                                   final NativeWebRequest request,
                                   final WebDataBinderFactory binderFactory) throws Exception {
-        final Optional<String> clientId = Optional.ofNullable(request.getUserPrincipal()).map(Principal::getName);
+        final Optional<String> clientId = Optional.ofNullable(authorizationService.getSubject().isPresent() ?
```
rcillo: If it always has a value it's no longer an `Optional`, right?

ferbncode: Right, I'll remove `Optional`

src/main/java/org/zalando/nakadi/plugin/auth/DefaultAuthorizationService.java
```diff
@@ -19,6 +24,14 @@ public boolean isAuthorizationAttributeValid(final AuthorizationAttribute author
         return true;
     }
 
+    @Override
+    @Nullable
+    public Optional<Subject> getSubject() {
+        return Optional.of(
+                () -> Optional.ofNullable(SecurityContextHolder.getContext().getAuthentication())
+                        .map(Principal::getName).orElse("unauthorized"));
```
rcillo: Shouldn't this just return Optional.empty()? which is then handled in other places?

ferbncode: Right, fixed in `3350ff9`. That makes sense. Thanks :)

rcillo:
👍

ferbncode:
:+1:

ferbncode:
Validation stack: https://nakadi-staging-validation-aruha2122v1.aruha-test.zalan.do/

PR #1010, opened 2019-01-22
ARUHA-2130 filter event types and subscriptions

lmontrieux:
# Filter lists of event types and subscriptions when calling get endpoints

> Zalando ticket : ARUHA-2130

## Description
Filters lists of event types and subscriptions when calling their get endpoints (GET /event-types and GET /subscriptions). The actual filtering should be implemented in a plugin. The default implementation of the plugin interface returns all event types and subscriptions all the time.

## Review
- [ ] Tests
- [ ] Documentation


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -126,10 +129,17 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
         this.etLogEventType = etLogEventType;
         this.eventTypeOptionsValidator = eventTypeOptionsValidator;
         this.adminService = adminService;
+        this.authorizationService = authorizationService;
     }
 
     public List<EventType> list() {
-        return eventTypeRepository.list();
+        return authorizationService.filter(eventTypeRepository.list()
```
rcillo: @lmontrieux I think that for event types this new filter interface adapted very well. But I'm thinking about subscriptions, which are paginated and more numerous. For example, if a user asks for a page with 10 results, then we fetch from the DB 10 resources, filter out, we are left with 2. Then we have to fetch the next bunch, until we potentially had to load the entire DB.

I thought about this problem for a while and I cannot think of an easy solution for that. How to apply the filter to the query itself? Only if we used some kind of ORM and we passed the query to the plugin and let it manipulate it 😨 

That said, I think we can live with this inefficiency.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -126,10 +129,17 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
...
+                .stream()
+                .map(EventType::asResource)
+                .collect(Collectors.toList()))
+                .stream()
+                .map(Resource<EventType>::get)
```
rcillo: I don't get this part. Why is it necessary to "unwrapp" the resource? I thought the API expected a `List<Resource<T>>`, no?

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -126,10 +129,17 @@ public EventTypeService(final EventTypeRepository eventTypeRepository,
...
         this.adminService = adminService;
+        this.authorizationService = authorizationService;
     }
 
     public List<EventType> list() {
-        return eventTypeRepository.list();
+        return authorizationService.filter(eventTypeRepository.list()
+                .stream()
+                .map(EventType::asResource)
+                .collect(Collectors.toList()))
```
rcillo: The indentation here is kind of trick to read. There is an additional `)` which mean the next call from the line bellow is on the result of the filter (the unwrap part). Maybe we can somehow change the indentation to make it more clear.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -180,7 +184,15 @@ public UriComponents getSubscriptionUri(final Subscription subscription) {
...
-                subscriptionRepository.listSubscriptions(eventTypesFilter, owningAppOption, offset, limit);
+                authorizationService
+                        .filter(subscriptionRepository
+                                .listSubscriptions(eventTypesFilter, owningAppOption, offset, limit)
+                                .stream()
+                                .map(Subscription::asResource)
+                                .collect(Collectors.toList()))
+                        .stream()
+                        .map(Resource<Subscription>::get)
+                        .collect(Collectors.toList());
```
rcillo: Maybe we need to run this several times until we accumulated the requested number of items in the page?

adyach:
@lmontrieux what is the status here?

lmontrieux:
@adyach apparently we cancelled this. Closing.

PR #1014, opened 2019-01-30
Added Authorization implementation according latest plugin

Kunal-Jha:
# One-line summary

> Zalando ticket : ARUHA-2132
Added Authorization implementation according latest plugin
## Description
Made Changes to Nakadi according to the new plugin api interfaces. All the `allAuthorizationAttributeValid` are changed except for admin authorization where permissions are used. 
## Review
- [ ] Tests
- [ ] Documentation


src/main/java/org/zalando/nakadi/domain/Subscription.java
```diff
@@ -17,7 +17,7 @@ public Subscription() {
 
     public Subscription(final String id, final DateTime createdAt, final DateTime updatedAt,
                         final SubscriptionBase subscriptionBase) {
-        super(subscriptionBase);
+        super(subscriptionBase, id);
```
rcillo: Why is this necessary?

Kunal-Jha: To create a resource from base itself.

src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
```diff
@@ -23,6 +24,7 @@
         CURSORS
     }
 
+    private String id;
```
v-stepanov: I don't exactly get it. Now both `Subscription` and `SubscriptionBase` have `id` field. What is the idea behind that?

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -86,18 +84,11 @@ private void checkAuthAttributesNoDuplicates(final Map<String, List<Authorizatio
...
-                    .flatMap(Collection::stream)
-                    .filter(attr -> !authorizationService.isAuthorizationAttributeValid(attr))
-                    .map(attr -> String.format("authorization attribute %s:%s is invalid",
-                            attr.getDataType(), attr.getValue()))
-                    .collect(Collectors.joining(", "));
-
-            if (!Strings.isNullOrEmpty(errorMessage)) {
-                throw new UnableProcessException(errorMessage);
+            if (authorizationService.isAuthorizationForResourceValid(resource)) {
+                throw new UnableProcessException("Authorization Attributes are not valid");
```
v-stepanov: I think this is a change that makes user experience worse. 

Before this change user was informed which attributes are invalid. Now the user will only be informed about a fact that they are invalid.

I think method `isAuthorizationForResourceValid` can't return just a boolean as `isAuthorizationAttributeValid` does. I think it can return a list of errors or just throw an exception that will contain information about specific attributes that are invalid.

src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
```diff
@@ -48,7 +50,8 @@
     public SubscriptionBase() {
     }
 
-    public SubscriptionBase(final SubscriptionBase subscriptionBase) {
+    public SubscriptionBase(final SubscriptionBase subscriptionBase, final String id) {
+        this.setId(id);
```
v-stepanov: the idea of `SubscriptionBase` is that it have only what user entered during creation of subscription. So it can't really have an id, as we generate it later.

src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
```diff
@@ -126,4 +127,8 @@ public boolean equals(final Object o) {
     public int hashCode() {
         return Objects.hash(owningApplication, eventTypes, consumerGroup, readFrom, initialCursors);
     }
+
+    public Resource<SubscriptionBase> asBaseResource() {
+        return new ResourceImpl<>(null, ResourceImpl.SUBSCRIPTION_RESOURCE, getAuthorization(), this);
```
v-stepanov: If resource name is supposed to be null - then it should be defined as `@Nullable`. 
But it would be strange to make it nullable as Resource interface uses Optional in another place and usually it's not good to use several approaches to represent no-value. 

So my suggestion is:
- make resource name optional
or
- for this specific case provide some predefined resource name like "new-subscription"

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionValidationService.java
```diff
@@ -100,7 +100,7 @@ public void validateSubscriptionChange(final Subscription old, final Subscriptio
         if (!Objects.equals(newValue.getInitialCursors(), old.getInitialCursors())) {
             throw new SubscriptionUpdateConflictException("Not allowed to change initial cursors");
         }
-        authorizationValidator.validateAuthorization(old.getAuthorization(), newValue.getAuthorization());
+        authorizationValidator.validateAuthorization(old.asResource(), newValue.asBaseResource());
```
v-stepanov: As you call `newValue.asBaseResource()` - it will create a resource without a name. I think it's fine to create a resource without a name for a subscription that is not created yet, but in this user modifies subscription that already exists so I think the id should be used as a resource name.

Kunal-Jha: Resolved

src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
Kunal-Jha: Resolved

src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
Kunal-Jha: added in the commit

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
Kunal-Jha: added in the commit

src/main/java/org/zalando/nakadi/domain/SubscriptionBase.java
Kunal-Jha: added in the commit

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -86,6 +87,21 @@ private void checkAuthAttributesNoDuplicates(final Map<String, List<Authorizatio
         }
     }
 
+    private void checkAuthorisationForResourceAreValid(final Resource resource)
+            throws UnableProcessException, ServiceTemporarilyUnavailableException, ForbiddenOperationException {
+
+        checkAuthAttributesAreValid(resource.getAuthorization());
```
rcillo: what's the difference between `checkAuthAttributesAreValid` and `areAllAuthorizationsForResourceValid`? I thought that the plan was to replace the former by the later, right? Since `areAllAuthorizationsForResourceValid` can perform all of the work that `checkAuthAttributesAreValid` performs. And in fact `checkAuthAttributesAreValid` only makes a number of calls to `isAuthorizationAttributeValid`. I think it's may be the source of potential security breaches if we maintain both methods and a user ends up thinking that it's safe to only `isAuthorizationAttributeValid` when in fact, `areAllAuthorizationsForResourceValid` should be also called.

Please, correct me if I missed something.

rcillo:
@Kunal-Jha please, check the comments once more. I think we have only some minor things and it's missing a merge from master.

Kunal-Jha:
@rcillo Apologies!!  somehow I missed that :( . . Added the new commits.

rcillo:
👍

Kunal-Jha:
👍

PR #1016, opened 2019-02-01
Check for VIEW on /event-types/* and /subscriptions/*

ferbncode:
Add functions to authorize view on resources

> Zalando ticket : ARUHA-2136


v-stepanov:
@ferbncode why did you create a branch in your repo?

src/main/java/org/zalando/nakadi/domain/SubscriptionAuthorization.java
```diff
@@ -43,6 +48,14 @@ public SubscriptionAuthorization(@JsonProperty("admins") final List<Authorizatio
         return readers;
     }
 
+    @JsonIgnore
+    public List<AuthorizationAttribute> getAll() {
+        return Stream.of(
+                Optional.ofNullable(readers).orElse(new ArrayList<>()),
+                Optional.ofNullable(admins).orElse(new ArrayList<>()))
```
v-stepanov: I think it doesn't make sense to check for null something that is not marked as `@Nullable`
In Nakadi we keep the not-nullable-by-default practice which allows to keep code clean from unnecessary null checks.

src/main/java/org/zalando/nakadi/domain/ResourceAuthorization.java
```diff
@@ -60,6 +64,15 @@ public ResourceAuthorization(@JsonProperty("admins") final List<AuthorizationAtt
         return writers;
     }
 
+    @JsonIgnore
+    public List<AuthorizationAttribute> getAll() {
+        return Stream.of(
+                Optional.ofNullable(this.readers).orElse(new ArrayList<>()),
+                Optional.ofNullable(this.writers).orElse(new ArrayList<>()),
+                Optional.ofNullable(this.admins).orElse(new ArrayList<>()))
```
v-stepanov: I think it doesn't make sense to check for null something that is not marked as @Nullable
In Nakadi we keep the not-nullable-by-default practice which allows to keep code clean from unnecessary null checks.

src/main/java/org/zalando/nakadi/domain/ResourceImpl.java
```diff
@@ -40,6 +40,9 @@ public String getType() {
     @Override
     public Optional<List<AuthorizationAttribute>> getAttributesForOperation(
             final AuthorizationService.Operation operation) {
+        if (null == authorization) {
```
v-stepanov: is it a real case that authorization is null?
In `ResourceImpl` constructor it is passed as one of parameters, so if we expect null as one of possible values - it should be marked as `@Nullable` in constructor parameter.

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -134,6 +134,28 @@ private void authorizeResourceAdmin(final Resource resource) throws AccessDenied
         }
     }
 
+    private void authorizeResourceView(final Resource resource) throws AccessDeniedException {
+        try {
+            if (!adminService.isAdmin(AuthorizationService.Operation.ADMIN)) {
+                if (!authorizationService.isAuthorized(AuthorizationService.Operation.VIEW, resource)) {
```
v-stepanov: why not to do it within single `if`?

v-stepanov: What will happen on plugun side when you call `authorizationService.isAuthorized(AuthorizationService.Operation.VIEW, resource)`? How will it decide that client has VIEW permission for a resource?

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -134,6 +134,28 @@ private void authorizeResourceAdmin(final Resource resource) throws AccessDenied
...
+            throw new ServiceTemporarilyUnavailableException("Error calling authorization plugin", e);
+        }
+    }
+
+    public void authorizeEventTypeView(final EventType eventType)
+            throws AccessDeniedException, ServiceTemporarilyUnavailableException {
+        authorizeResourceView(eventType.asResource());
+    }
+
+    public void authorizeSubscriptionView(final Subscription subscription) throws
```
v-stepanov: it's better to move `throws` to a next line, like you did for a method above.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -115,7 +119,7 @@ public void createTimeline(final String eventTypeName, final String storageId, f
         }
         try {
             final EventType eventType = eventTypeCache.getEventType(eventTypeName);
-
+            authorizationValidator.authorizeEventTypeView(eventType);
```
v-stepanov: Why did you decide to check it for timeline creation endpoint? It's anyway an admin endpoint that allows access only to Nakadi-admins.

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -356,6 +360,7 @@ private void finishTimelineUpdate(final String eventTypeName) throws TimelineExc
 
         try {
             final EventType eventType = eventTypeCache.getEventType(eventTypeName);
+            authorizationValidator.authorizeEventTypeView(eventType);
```
v-stepanov: and get-timelines as well. It's allowed only to Nakadi admins, so I think it doesn't make sense to check if event-type-view is allowed.

v-stepanov:
@ferbncode I think the following endpoints left uncovered:
```
GET /event-types/{name}/schemas (history of schemas for an event type)
GET /event-types/{name}/schemas/{version} (one version of the schema)
POST /subscriptions/{subscription_id}/cursors (commit cursors)
PATCH /subscriptions/{subscription_id}/cursors (reset cursors)
```

ferbncode:
@v-stepanov For these endpoints, I do the check for `view` in the `CursorsService.java` (in `resetCursors` and `commitCursors` function) and `EventTypeService.java` (`get` function). I'll fix the changes required after review and move the checks to first validations. Thanks for pointing it out. :)

v-stepanov:
```
For these endpoints, I do the check for view in the CursorsService.java (in resetCursors and commitCursors function)
```
@ferbncode sorry, I can't find this code. Can you please point me?

ferbncode:
* For schemas endpoint https://github.com/zalando/nakadi/blob/0b7b7ad83c4e7a3c35801f26ad6220c5533d608c/src/main/java/org/zalando/nakadi/service/EventTypeService.java#L461

* I didn't add the checks for `PUT` and `PATCH` on `/subscription/<id>/cursors`. I'll add them. You're right. Looks like I missed them.

On second thought, do you think I should put these checks in the controller?

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
ferbncode: As suggested, removed the check for both endpoints.

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
ferbncode: Fixed this.

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
ferbncode: Right. Fixed this. :)

src/main/java/org/zalando/nakadi/domain/ResourceImpl.java
ferbncode: Yes, `authorization` can be `null`, when "authorization" section is not specified in the request when creating an event-type or subscription. I added the annotation as you suggested to fix it.

src/main/java/org/zalando/nakadi/domain/ResourceAuthorization.java
ferbncode: Fixed in `de2a148`

src/main/java/org/zalando/nakadi/domain/SubscriptionAuthorization.java
ferbncode: You're right. Fixed in `de2a148`

src/main/java/org/zalando/nakadi/domain/ResourceImpl.java
v-stepanov: But I suggested to have annotation in ResourceImpl constructor.

src/main/java/org/zalando/nakadi/domain/EventTypeBase.java
```diff
@@ -63,6 +63,7 @@
     private EventTypeOptions options;
 
     @Valid
+    @Nullable
```
v-stepanov: no need to have annotation here as it is a private field. `@Nullable` annotation is usually used only for public methods. And getAuthorization() method already has this annotation.

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
v-stepanov: ok, this is answered already

src/main/java/org/zalando/nakadi/domain/ResourceImpl.java
```diff
@@ -17,6 +18,7 @@
     private final T resource;
     private final String name;
     private final String type;
+    @Nullable
```
v-stepanov: It doesn't make much sense to mark private values as `@Nullable`. It's better to mark it nullable in public methods (e.g. constructor parameter, getter)

ferbncode: Fixed this. :sweat_smile:

ferbncode:
:+1:

src/main/java/org/zalando/nakadi/service/AuthorizationValidator.java
```diff
@@ -134,6 +134,27 @@ private void authorizeResourceAdmin(final Resource resource) throws AccessDenied
...
+            if (!(adminService.isAdmin(AuthorizationService.Operation.ADMIN) ||
+                    authorizationService.isAuthorized(AuthorizationService.Operation.VIEW, resource))) {
+                throw new AccessDeniedException(AuthorizationService.Operation.VIEW, resource);
+            }
+        } catch (final PluginException e) {
+            throw new ServiceTemporarilyUnavailableException("Error calling authorization plugin", e);
+        }
+    }
+
+    public void authorizeEventTypeView(final EventType eventType)
```
Kunal-Jha: I don't understand why there are separate methods for event type view and subscription view.  IMO you can pass and authorize Resource directly since both these methods behave the same unless I am missing something.

ferbncode: I follow the same pattern as other authorization functions in `AuthorizationValidator.java`, for example, `authorizeEventTypeWrite()`, `authorizeEventTypeAdmin()`, `authorizeSubscriptionAdmin()`. I think it is good as it does not mention `Resource` entity in places we call them.

Kunal-Jha:
@ferbncode Everything looks good. For some reason Codacy fails though.

ferbncode:
@Kunal-Jha I am not sure how to fix Codacy. :sweat_smile: (Also, I am quite certain that it does not block merge)

Kunal-Jha:
👍

lmontrieux:
👍

ferbncode:
:+1:

PR #1017, opened 2019-02-04
Fixed check if keep alive limit is reached for indefinitely configured limit

m99coder:
# Interpret `0` as indefinitely stream “keep-alives”

> Zalando ticket : ARUHA-2159

## Description
If a stream is configured with `stream_keep_alive_limit` as `0` the related check `isKeepAliveLimitReached` always returned `true` instead of `false`.

## Review
- [x] Tests
- [ ] Documentation

## Deployment Notes
Nothing to add here

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/StreamParameters.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/StreamParameters.java
@@ -61,7 +61,7 @@ public class StreamParameters {
                         .filter(timeout -> timeout > 0 && timeout <= EventStreamConfig.MAX_STREAM_TIMEOUT)
                         .orElse((long) EventStreamConfig.generateDefaultStreamTimeout()));
         this.maxUncommittedMessages = userParameters.getMaxUncommittedEvents().orElse(10);
-        this.batchKeepAliveIterations = userParameters.getStreamKeepAliveLimit();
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/nakadi/service/subscription/StreamParameters.java
```diff
@@ -83,7 +83,7 @@ public boolean isStreamLimitReached(final long commitedEvents) {
     }
 
     public boolean isKeepAliveLimitReached(final IntStream keepAlive) {
-        return batchKeepAliveIterations.map(it -> keepAlive.allMatch(v -> v >= it)).orElse(false);
+        return batchKeepAliveIterations.map(it -> keepAlive.allMatch(v -> v >= it) && it > 0).orElse(false);
```
v-stepanov: @m99coder thank you for contribution. 
I think the better place to check that condition would be `StreamParameters` constructor https://github.com/zalando/nakadi/blob/67e95a474593ef785357edaac509689e88901a22/src/main/java/org/zalando/nakadi/service/subscription/StreamParameters.java#L64 as it's a place where we handle all those special values of stream parameters and default values.

v-stepanov:
Looks good

v-stepanov:
👍

adyach:
👍

v-stepanov:
@m99coder that you for your contribution to Nakadi. We will release this bugfix with the next Nakadi release.

m99coder:
👍 Glad having been able to help

PR #1018, opened 2019-02-05
Cursor commit, patch and get protected by blacklisting CONSUMER_APP

rcillo:
https://jira.zalando.net/browse/ARUHA-2161

In case a consumer application goes out of control and starts
hammering the cursors endpoints, make it possible to block it. 

It's important because these endpoints are fairly expensive since they hit a non scalable
component: Zookeeper.

We consider that an app which is forbidden from consumption of events
is also forbidden from commits, which is part of the stream-commit
consumption process.

New checks for validity of partition and offset fields were added, to improve API user experience.


src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -82,6 +93,11 @@ public CursorsController(final CursorsService cursorsService,
         if (cursors.isEmpty()) {
             throw new CursorsAreEmptyException();
         }
+
+        if (blacklistService.isSubscriptionConsumptionBlocked(subscriptionId, client.getClientId())) {
```
v-stepanov: This will add a call to DB for quite a busy endpoint. Let's optimize this call.

rcillo: Thanks for pointing out. I just replaced the call to `SubscriptionDbRepository` by a call to `SubscriptionCache`. This should also improve the performance when opening a stream to a subscription.

rcillo:
👍

v-stepanov:
Looks good!

v-stepanov:
👍

rcillo:
👍

v-stepanov:
👍

rcillo:
deploy validation please

adyach:
deploy validation please

rcillo:
👍

v-stepanov:
👍

antban:
:+1:

lmontrieux:
👍

PR #1021, opened 2019-02-07
set log level to INFO

adyach:
# One-line summary

set log level to INFO


v-stepanov:
I checked current DEBUG logs in live nakadi. 
> 99% of DEBUG logs for last 4 hours were from classes "KafkaCrutchConsumer" and "ClosedConnectionsCrutch". I think if you lower log level for these two classes we will achieve the same log volume but we will keep valuable debug info for numerous endpoints.

adyach:
As far as I understand debug level should be turned on for debugging purposes. which endpoints require debugging?

antban:
:+1:

v-stepanov:
@adyach 
422 for commit at least.
I think the work for this ticket should also include reviewing all places where we log with debug and deciding if this should be moved to info or not.

adyach:
> I checked current DEBUG logs in live nakadi.
> > 99% of DEBUG log lines for last 4 hours were from classes "KafkaCrutchConsumer" and "ClosedConnectionsCrutch". I think if you lower log level for these two classes we will achieve the same log volume but we will keep valuable debug info for numerous endpoints.

done e4d8a7e

v-stepanov:
👍

v-stepanov:
👍

adyach:
👍

adyach:
👍

v-stepanov:
👍

PR #980, opened 2018-11-14
Initialize kpis

rcillo:
As part of the development of end-2-end test, we are automating the setup of KPI event types.

This PR also changes the way that we use docker to run Nakadi locally. I believe I found a way to make it work. I'm opening this PR to see if it works on travis.

PR #1023, opened 2019-02-12
Change several variables to a single zookeeper connection string

antban:
Change several variables to a single zookeeper connection string
ARUHA-2138


database/migration/aruha-2181-2-cleanup-storage-zookeeper.sql
```diff
@@ -0,0 +1,4 @@
+-- this script should be used after 2181 is successfully released
+-- Remove all obsolete nodes
+update storage
+set st_configuration = concat('{"zk_connection_string": "',st_configuration ->> 'zk_connection_string','"}')::jsonb;
```
v-stepanov: emptyline missing

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '3.7'
+version: '3'
```
v-stepanov: was there a specific reason for that?

src/main/java/org/zalando/nakadi/service/NakadiAuditLogPublisher.java
```diff
@@ -22,33 +25,36 @@
...
-                        final String resourceId,
-                        final Optional<String> userOrNone) {
+                        final String resourceId) {
         try {
             if (!featureToggleService.isFeatureEnabled(FeatureToggleService.Feature.AUDIT_LOG_COLLECTION)) {
                 return;
             }
 
-            final String user = userOrNone.orElse("<anonymous>");
+            final String user = authorizationService.getSubject().map(Subject::getName)
```
v-stepanov: I don't really get why you also did modifications to audit log publisher

docker-compose.yml
antban: What is the reason for 3.7? I don't have it on machine and see no reason for that.

src/main/java/org/zalando/nakadi/service/NakadiAuditLogPublisher.java
antban: Omg.. Will remove this part from pr.

database/migration/aruha-2181-2-cleanup-storage-zookeeper.sql
antban: @v-stepanov Added

docker-compose.yml
v-stepanov: It was changed on purpose last November. I think it makes sense to ask the person who changed it instead of silently lowering the version.

src/test/java/org/zalando/nakadi/security/AuthenticationTest.java
```diff
@@ -183,10 +182,10 @@ public FeatureToggleService featureToggleService() {
...
         }
 
-        @Bean
-        public ZooKeeperHolder mockZKHolder() {
-            return mock(ZooKeeperHolder.class);
-        }
+//        @Bean
+//        public ZooKeeperHolder mockZKHolder() {
+//            return mock(ZooKeeperHolder.class);
+//        }
```
v-stepanov: :confused:

src/main/java/org/zalando/nakadi/repository/zookeeper/ZooKeeperHolder.java
```diff
@@ -11,61 +11,111 @@
...
+        framework.start();
+        return new ZooKeeperHolder(framework);
     }
 
-    public CuratorFramework get() {
-        return zooKeeper;
+    private static EnsembleProvider createEnsembleProvider(
+            final String type, final String serversAndPort, final String path) {
+        final EnsembleProvider result;
+        if (TYPE_EXHIBITOR.equalsIgnoreCase(type)) {
```
v-stepanov: why do we still need all this exhibitor stuff here? For migration period?

docker-compose.yml
antban: @v-stepanov In regards to this I've spent some time figuring out what and why happen.
The change was introduced by means of this commit: https://github.com/zalando/nakadi/commit/fc0c7e82c1d194d136cdf0e125965d7ee5d1e62b of this pr: https://github.com/zalando/nakadi/pull/980/files
I took a look here - https://docs.docker.com/compose/compose-file/compose-versioning/#version-3
There are no properties, that we are using and that were added in versions 3.1-3.7. 

In general the comment is - we are not using any features from 3.7, but insisting on it. If you have newest installation ever - our docker-compose should work. But from the other hand - why should we force people to use this latest version? 

I feel that the answer is: just because we can.

And the last part - I we want to use latest docker-compose+docker installation - let's use it. But don't enforce it through file format version.

src/main/java/org/zalando/nakadi/repository/zookeeper/ZooKeeperHolder.java
antban: Yep

src/test/java/org/zalando/nakadi/security/AuthenticationTest.java
antban: Removed

src/main/java/org/zalando/nakadi/repository/zookeeper/ZooKeeperHolder.java
antban: And in case if we will still use it

v-stepanov:
Only couple of issues

1) The name NAKADI_ZOOKEEPER_CONN is a bit strange for me. I think NAKADI_ZOOKEEPER_CONNECTION_STRING would be better as it doesn't require additional explanation.

2) I assume we will remove "zookeeper://" prefix when we finish migration, right?

But I will approve anyway, as those are minor things

v-stepanov:
👍

antban:
@v-stepanov Actually I don't think that 1. NAKADI_ZOOKEEPER_CONNECTION_STRING is better than NAKADI_ZOOKEEPER_CONN, because it is hard to remember this long line (and it's way too big, I read too slow)
After migration line exhibitor://blablabla will be replaced by zookeeper://blablabla

v-stepanov:
@antban 
1) That's the point - you don't need to remember it :)
2) But I mean when we will have only zookeeper - what's the point to have zookeeper:// prefix in connection string ?

PR #1027, opened 2019-03-01
ARUHA-2199: Added open tracing;

v-stepanov:
# Open tracing

> Zalando ticket : ARUHA-2199

## Description
Added tracing of batch publishing requests

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
+++ b/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
@@ -43,7 +43,7 @@ public class LoggingFilter extends OncePerRequestFilter {
         this.authorizationService = authorizationService;
     }
 
-    public class RequestLogInfo {
+    private class RequestLogInfo {
 
         private String userAgent;
         private String user;
@@ -55,7 +55,7 @@ public class LoggingFilter extends OncePerRequestFilter {
         private String acceptEncoding;
         private Long requestTime;
 
-        public RequestLogInfo(final HttpServletRequest request, final long requestTime) {
+        private RequestLogInfo(final HttpServletRequest request, final long requestTime) {
             this.userAgent = Optional.ofNullable(request.getHeader("User-Agent")).orElse("-");
             this.user = authorizationService.getSubject().map(Subject::getName).orElse("-");
             this.method = request.getMethod();
--- a/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
+++ b/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
@@ -1,12 +1,12 @@
 package org.zalando.nakadi.filters;
 
 import com.google.common.net.HttpHeaders;
+import io.opentracing.util.GlobalTracer;
 import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Component;
 import org.springframework.web.filter.OncePerRequestFilter;
 import org.zalando.nakadi.plugin.api.authz.AuthorizationService;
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -134,35 +122,74 @@ protected void doFilterInternal(final HttpServletRequest request,
...
+                requestLogInfo.contentEncoding,
+                requestLogInfo.acceptEncoding,
+                requestLogInfo.contentLength);
+    }
+
+    private void traceRequest(final RequestLogInfo requestLogInfo, final int statusCode) {
+        if ("POST".equals(requestLogInfo.method) && requestLogInfo.path.startsWith("/event-types/") &&
+                requestLogInfo.path.contains("/events")) {
+
+            final String eventType = requestLogInfo.path.substring(13, requestLogInfo.path.lastIndexOf("/events"));
```
lmontrieux: Can we move the magic 13 to a constant? I guess it's the number of characters in `/event-types/`

v-stepanov: Yes, makes sense.
https://github.com/zalando/nakadi/pull/1027/commits/c140b6f54ec1113fdf86f8f76a4710f506dcc637

lmontrieux:
👍

v-stepanov:
👍

v-stepanov:
deploy try tmp

v-stepanov:
deploy try tmp

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -76,27 +74,17 @@ private RequestLogInfo(final HttpServletRequest request, final long requestTime)
...
-                    requestLogInfo.path,
-                    requestLogInfo.query,
-                    requestLogInfo.userAgent,
-                    requestLogInfo.user,
-                    HttpStatus.PROCESSING.value(),
-                    0,
-                    requestLogInfo.contentEncoding,
-                    requestLogInfo.acceptEncoding,
-                    requestLogInfo.contentLength);
+            logToAccessLog(this.requestLogInfo, this.response.getStatus(), 0L);
```
rcillo: @v-stepanov isn't it changing the behavior? Before this change we used to log 102 but now it's not clear if the status in the response is going to be 102. I think this is a special log for calls to consumption API where we only know the real response status code afterwards.

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -134,35 +122,75 @@ protected void doFilterInternal(final HttpServletRequest request,
...
+                requestLogInfo.user,
+                statusCode,
+                timeSpentMs,
+                requestLogInfo.contentEncoding,
+                requestLogInfo.acceptEncoding,
+                requestLogInfo.contentLength);
+    }
+
+    private void traceRequest(final RequestLogInfo requestLogInfo, final int statusCode, final Long timeSpentMs) {
+        if ("POST".equals(requestLogInfo.method) && requestLogInfo.path.startsWith("/event-types/") &&
```
rcillo: Is `path` guaranteed not to be null?

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
v-stepanov: Oh crap :( I have not noticed that I changed the behavior. Good catch! Will fix it.

v-stepanov: Fixed https://github.com/zalando/nakadi/pull/1027/commits/dbae4615dc693b826bf4d8605c58161cadf62e51

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
v-stepanov: I can't guarantee that, so it makes sense to add a check: https://github.com/zalando/nakadi/pull/1027/commits/dbae4615dc693b826bf4d8605c58161cadf62e51

rcillo:
👍

v-stepanov:
👍

v-stepanov:
deploy try tmp

PR #1029, opened 2019-04-02
Log complete exception in case of IOException

ferbncode:
Log complete exception instead of the message as message is `null` before flush of batch in case client closed connection. (relevant commit: [d90d28d](https://github.com/zalando/nakadi/commit/d90d28d1f1dd41b4b339c7d5cf8c61f9d0adcf43))

## Review
- [ ] Tests
- [ ] Documentation


rcillo:
👍

ferbncode:
:+1:

PR #1032, opened 2019-04-04
Revert "Merge pull request #1023 from zalando/ARUHA-2138"

antban:
As this modification introduced some non-compatible changes to internal api (storages), it is not possible to easily migrate. 
Therefore, it is rolled back before everything will be able to work with new configuration type


antban:
:+1:

ferbncode:
:+1:

PR #1034, opened 2019-04-18
return 409 on request during cursor reset

adyach:
# One-line summary

Return 409 on request during cursor reset
https://nakadi.io/manual.html#/subscriptions/subscription_id/events_get

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
+++ b/src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
@@ -144,9 +144,7 @@ public class SubscriptionStreamController {
                 } catch (final IOException e) {
                     LOG.error("Failed to write exception to response", e);
                 }
-            } else
-
-            {
+            } else {
                 LOG.warn("Exception found while streaming, but no data could be provided to client", ex);
             }
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -114,26 +138,8 @@ public void onException(final Exception ex) {
...
-                                ex.getMessage()));
-                    } else if (ex instanceof InternalNakadiException) {
-                        writeProblemResponse(response, out, Problem.valueOf(INTERNAL_SERVER_ERROR, ex.getMessage()));
-                    } else if (ex instanceof NoStreamingSlotsAvailable) {
-                        writeProblemResponse(response, out, Problem.valueOf(CONFLICT, ex.getMessage()));
-                    } else {
-                        writeProblemResponse(response, out, Problem.valueOf(SERVICE_UNAVAILABLE,
-                                "Failed to continue streaming"));
-                    }
+                    writeProblemResponse(response, out, exceptionProblem.getOrDefault(ex.getClass(),
```
v-stepanov: I think there is a problem with this refactoring. 
`instanceof` checks if the class is instance of some class or its subclass. 
The way it is now will only check if it is instance of specific class.
So with this refactoring the behavior can change.

v-stepanov: ok, all these classes don't have child classes. So it will work fine.

v-stepanov:
👍

adyach:
👍

PR #1035, opened 2019-04-18
Upgrade library versions as much as we can

rcillo:
https://jira.zalando.net/browse/ARUHA-2262

Some dependencies could not be easily updated due to major version
incompatibility. Once we proceed with spring boot update we should
cover most of the remaining threats.


Kunal-Jha:
👍

rcillo:
👍

Issue #1037, opened 2019-04-22
KafkaTopicRepositoryTest uses shared CircuitBreaker


mneverov:
### Description
KafkaTopicRepositoryTest uses shared CircuitBreaker. HystrixKafkaCircuitBreaker gets circuit breaker from the HystrixCircuitBreakerFactory which gets it from the cache (static map)- see com.netflix.hystrix.HystrixCircuitBreaker.Factory#circuitBreakersByCommand.  
As the result tests in KafkaTopicRepositoryTest class depend on each other and on the order of tests execution. JUnit runs tests in default order which is by hashes of their names sorted in ascending order. 
Any test with a hash of the test name greater than `whenKafkaPublishTimeoutThenCircuitIsOpened` will inherit circuit breaker in `OPEN` state from it which may be unexpected.

### Proposals
- do nothing
- add explicit annotation `FixMethodOrder.DEFAULT` with a comment about shared CircuitBreaker
- refactor test to use mocked CircuitBreaker

antban:
Hystrix support was removed in #1357, closing

Issue #741, opened 2017-08-24
Build failure: whenKafkaPublishTimeoutThenCircuitIsOpened FAILED


peteraba:
I'm absolutely newbie here, I just tried to follow the README and the build seems to be broken:

```
org.zalando.nakadi.repository.kafka.KafkaTopicRepositoryTest > whenKafkaPublishTimeoutThenCircuitIsOpened FAILED
    java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:86)
        at org.junit.Assert.assertTrue(Assert.java:41)
        at org.junit.Assert.assertTrue(Assert.java:52)
        at org.zalando.nakadi.repository.kafka.KafkaTopicRepositoryTest.whenKafkaPublishTimeoutThenCircuitIsOpened(KafkaTopicRepositoryTest.java:355)
```

I'm not sure what information you might need so I'll just share the obvious.
```
➤ java -version
openjdk version "1.8.0_144"
OpenJDK Runtime Environment (build 1.8.0_144-b01)
OpenJDK 64-Bit Server VM (build 25.144-b01, mixed mode)

➤ gradle -version

------------------------------------------------------------
Gradle 4.0.2
------------------------------------------------------------

Build time:   2017-07-26 15:04:56 UTC
Revision:     108c593aa7b43852f39045337ee84ee1d87c87fd

Groovy:       2.4.11
Ant:          Apache Ant(TM) version 1.9.6 compiled on June 29 2015
JVM:          1.8.0_144 (Oracle Corporation 25.144-b01)
OS:           Linux 4.9.43-1-MANJARO amd64
```

Let me know if you need anything else.

adyach:
@peteraba good day, could you tell me how you run it? we have not tested it against gradle 4.0.2

peteraba:
@adyach I just installed gradle via pacman and ran './gradlew build' as mentioned by the README.

peteraba:
As discussed privately, this appears to be a flakey test, it passed 3 out 6 times for me, every time running the whole test suite.

lmontrieux:
Thanks for reporting this bug, @peteraba . We'll prioritise it and fix it as soon as possible.

peteraba:
@lmontrieux @adyach: good luck guys, ping me if you need more info!

PR #1038, opened 2019-04-22
Fix kafka circuit breaker to switch to open state multiple times

max-neverov:
# One-line summary
Fix kafka circuit breaker to switch to open state multiple times.

## Description
Once transit to `Open` state KafkaCircuitBreaker will preserve this state forever.  

There are two problems:
- `HystrixKafkaCircuitBreaker` doesn't update `circuitOpened` on fail. This causes `com.netflix.hystrix.HystrixCircuitBreaker.HystrixCircuitBreakerImpl#isAfterSleepWindow` always passes and `allowRequest` will always return true after the window.
Fixed by adding `circuitBreaker.markNonSuccess();`
- `KafkaTopicRepository` uses `allowRequest`. This is the reason why the circuit breaker will never transit from `Open` to `HalfOpen` state and from `HalfOpen` to `Close`. Fixed by using `attemptExecution` instead.

To reproduce the issue checkout the branch, revert changes in `HystrixKafkaCircuitBreaker` and `KafkaTopicRepository` and run `KafkaTopicRepositoryTest#whenKafkaPublishTimeoutThenCircuitIsOpened`. 

Relates to https://github.com/zalando/nakadi/issues/741:
It is better to use shrinked health check interval instead of thread sleep because current implementation takes 15 seconds to run tests. It can be fixed after merge of #741 to master.

Relates to #1037:
 I would prefer to write a separate test on this issue, but it is a lot of redundant and misleading code since `KafkaTopicRepositoryTest` uses a shared circuit. So I crammed the circuit breaker's state test to the existing test.

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes

### DIFF
--- a/src/main/java/org/zalando/nakadi/repository/kafka/HystrixKafkaCircuitBreaker.java
+++ b/src/main/java/org/zalando/nakadi/repository/kafka/HystrixKafkaCircuitBreaker.java
@@ -44,9 +44,9 @@ public class HystrixKafkaCircuitBreaker {
                 hystrixCommandMetrics);
         concurrentExecutionCount = new AtomicInteger();
     }
-
-    public boolean allowRequest() {
-        return circuitBreaker.allowRequest();
+    
+    public boolean attemptExecution() {
+        return circuitBreaker.attemptExecution();
     }
 
     public void markStart() {
@@ -66,6 +66,7 @@ public class HystrixKafkaCircuitBreaker {
         concurrentExecutionCount.decrementAndGet();
         HystrixThreadEventStream.getInstance()
                 .executionDone(ExecutionResult.from(HystrixEventType.FAILURE), commandKey, threadPoolKey);
+        circuitBreaker.markNonSuccess();
     }
 
     public String getMetrics() {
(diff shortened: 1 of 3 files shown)


adyach:
@max-neverov thank you for contribution! 
I am not sure why I have implemented it this way back then, it looks like a bug at the moment.
Please, fix checkstyle issue which travis reports.

max-neverov:
@adyach done.
The `attemptExecution` method was added to Hystrix in 2017, so at the time `HystrixKafkaCircuitBreaker` was implemented there was no such method.

adyach:
👍

max-neverov:
@adyach 
I merged master and added some more circuit breaker params to reduce the test run time (before ~15sec for the whole class, now ~3sec), ptal

adyach:
👍

lmontrieux:
@max-neverov I'm doing a bit of cleanup, and it looks like we failed to follow up on your PR. Really sorry about that :(.

Would you mind fixing the conflict, so we can merge this asap? Thanks a lot!

max-neverov:
Hi @lmontrieux, 

please check the attached patch for the test
[Merge_with_master.patch.txt](https://github.com/zalando/nakadi/files/3597440/Merge_with_master.patch.txt)

adyach:
closed in  #1112

PR #1041, opened 2019-04-24
Revert "Validation rdecillo"

rcillo:
Reverts zalando/nakadi#1040

This introduced a regression on consumption path which needs to be fixed. From the logs `org.springframework.web.util.NestedServletException: Handler dispatch failed; nested exception is java.lang.AbstractMethodError: com.fasterxml.jackson.databind.ser.std.ReferenceTypeSerializer.withResolved(Lcom/fasterxml/jackson/databind/BeanProperty;Lcom/fasterxml/jackson/databind/jsontype/TypeSerializer;Lcom/fasterxml/jackson/databind/JsonSerializer;Lcom/fasterxml/jackson/databind/util/NameTransformer;Lcom/fasterxml/jackson/annotation/JsonInclude$Include;)Lcom/fasterxml/jackson/databind/ser/std/ReferenceTypeSerializer;`

Funny thing is that it passed all tests and publishing works fine.

PR #1042, opened 2019-04-24
Revert "Revert" on dependencies upgrade

rcillo:
Reverts zalando/nakadi#1041

As it appears, nakadi-plugin-authz which is added to the classpath later on during the build phase was overriding jackson databind with a different version. We forced the same version in the plugin and the problem should be solved by now.

But we do need to improve how we build the final Nakadi version, since it's brittle and we should expect conflicting versions in the plugin in the future again.


ferbncode:
:+1:

rcillo:
👍

rcillo:
deploy validation please

PR #1045, opened 2019-04-29
Do not throw 500 on client connection close

adyach:
# One-line summary
Do not throw 500 on client connection close, during data transmit

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/advice/NakadiProblemExceptionHandler.java
+++ b/src/main/java/org/zalando/nakadi/controller/advice/NakadiProblemExceptionHandler.java
@@ -3,6 +3,7 @@ package org.zalando.nakadi.controller.advice;
 import com.fasterxml.jackson.databind.JsonMappingException;
 import com.google.common.base.CaseFormat;
 import org.apache.commons.lang3.RandomStringUtils;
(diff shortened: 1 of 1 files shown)


adyach:
👍

src/main/java/org/zalando/nakadi/controller/advice/NakadiProblemExceptionHandler.java
```diff
@@ -84,6 +86,11 @@ private String generateErrorTraceId() {
         return create(Problem.valueOf(BAD_REQUEST, message), request);
     }
 
+    @ExceptionHandler(EofException.class)
+    public void handleEofException() {
+        LOG.info("Client closed connection");
```
v-stepanov: I don't really see much value in this log as it doesn't provide any specific information like subscription id or session id. Is it possible to have this info?

v-stepanov:
👍

src/main/java/org/zalando/nakadi/controller/advice/NakadiProblemExceptionHandler.java
v-stepanov: Ok, you pointed out that FlowId will be there. Good for me.

PR #1046, opened 2019-04-29
Detect changes to ArraySchema

rcillo:
https://jira.zalando.net/browse/ARUHA-2282

This fixes a NPE changing an ArraySchema through generalization: an
array that used to define schemas for each item specifically now has a
single schema for all items.

We also added tests for the other way around: a schema that used to
have a single definition for all items now has a per item schema.


adyach:
👍

rcillo:
👍

PR #1052, opened 2019-06-04
log time taken to open the stream and rebalance

adyach:
# One-line summary

log time taken to open the stream and rebalance


v-stepanov:
👍

adyach:
👍

PR #1056, opened 2019-06-06
ARUHA-2330: Delete zk lock node when deleting subscription;

v-stepanov:
ARUHA-2330: Delete zk lock node when deleting subscription;


adyach:
👍

v-stepanov:
👍

PR #1058, opened 2019-06-07
lock using zookeeper

adyach:
# One-line summary
The implementation of the lock for subscription api using zookeeper official client


src/main/java/org/zalando/nakadi/repository/zookeeper/ZookeeperProxy.java
```diff
@@ -0,0 +1,121 @@
...
+        try {
+            zookeeper.close();
+        } catch (final InterruptedException e) {
+            Thread.currentThread().interrupt();
+            LOG.error("Interrupted while closing zookeeper client", e);
+        }
+    }
+
+
+    private void retryDelay(final int requestCounter) {
```
antban: Killer method!
In case of interruption - hang up for specified amount of seconds constantly knocking to zk

src/main/java/org/zalando/nakadi/repository/zookeeper/ZookeeperProxy.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(ZookeeperProxy.class);
+    private static final int RETRY_DELAY_MS = 250;
+    private static final int RETRY_MAX_COUNT = 10;
+    private final ZooKeeper zookeeper;
+
+    public ZookeeperProxy(final ZooKeeper zookeeper) {
+        this.zookeeper = zookeeper;
+        try {
+            zookeeper.create("/nakadi/locks", null, ZooDefs.Ids.OPEN_ACL_UNSAFE, CreateMode.PERSISTENT);
```
antban: if you are creating `/nakadi/locks` here, than path in lockAccuire should be relative to this path. I don't see it.

src/main/java/org/zalando/nakadi/repository/zookeeper/ZookeeperProxy.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+
+    private void processException(final Exception e) {
+        LOG.error("Failed working with lock", e);
+        if (e instanceof KeeperException.SessionExpiredException) {
+            throw new RuntimeException("Failed to acquire lock", e);
+        }
+
+        if (e instanceof InterruptedException) {
```
antban: I think that in case of interruption you should throw exception here as well (Runtime), cause otherwise you will again be in meaningless cycle.

src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
```diff
@@ -86,39 +88,24 @@ protected Logger getLog() {
...
-            return result;
         } catch (final NakadiRuntimeException | NakadiBaseException e) {
             throw e;
         } catch (final Exception e) {
             throw new NakadiRuntimeException(e);
         }
     }
 
+    private String getLockPath() {
+        return "/nakadi/locks/" + subscriptionId;
```
antban: Makes sense to avoid "/nakadi/locks" here.

adyach:
@antban I have address your comments

src/main/java/org/zalando/nakadi/repository/zookeeper/ZooKeeperHolder.java
```diff
@@ -71,6 +80,26 @@ public CuratorFramework get() {
         return zooKeeper;
     }
 
+
+    public Closeable newZookeeperLock(final String lockObject, final long timeoutMs) throws RuntimeException {
+        try {
+            final ZookeeperLock zookeeperLock = new ZookeeperLock(new ZooKeeper(connectionString,
+                    sessionTimeoutMs, new NakadiZookeeperWatcher()));
+            return zookeeperLock.tryLock(lockObject, timeoutMs);
+        } catch (final Exception e) {
```
antban: Seems like stack trace extender, but I'm fine with it

antban:
:+1:

adyach:
👍

PR #1059, opened 2019-06-10
ARUHA-2330: added closing of subscription connections before deleting subscription from ZK;

v-stepanov:
ARUHA-2330: added closing of subscription connections before deleting subscription from ZK;


src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
```diff
@@ -126,6 +130,9 @@ protected Logger getLog() {
     @Override
     public final void deleteSubscription() {
         try {
+            if (!closeSubscriptionConnections()) {
```
adyach: but it will take up to 1 minute to delete subscription, if you want to leave it that way, it should be documented in api

v-stepanov: I was thinking about that. The other option is to wait for connections close and deletion of zk data asynchronously. 
1. Delete from DB; 
2. launch ZK deletion async;
3. return response to user;
4. ZK deletion finishes.

PR #1062, opened 2019-06-12
ARUHA-2344 Do not check authorization while creating system ET

antban:
ARUHA-2344 Do not check authorization while creating system ET

I know that it's a shitty way - to create flags, but it is also the simplest way. 
As a bonus duplication and regexp replacements bugs fixed


rcillo:
👍

antban:
:+1:

PR #1066, opened 2019-06-14
Add default charset to UTF-8 for incoming messages

### LINKED ISSUE #1065: Unicode characters are received corrupted if a default "application/json" content type is set
I am sending unicode characters such, as ⌀ or ½, as part of Nakadi events. The payload is sent with the default Content-Type header of `application/json`. The server, however, accepts charset of ISO-8859-1 as default, set by Spring, see https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/converter/StringHttpMessageConverter.java#L63.
Conversion from UTF-8 to ISO-8859-1 corrupts unicode data. This part of the code is responsible for the default charset:
https://github.com/zalando/nakadi/blob/1643b2182c1b4bb257ad6d822f48971eeaf9f91f/src/main/java/org/zalando/nakadi/config/WebConfig.java#L71
With changing the default charset to UTF-8 unicode symbols in the payload would be supported by default and this is what I suggest to change

ovolynets:
Fixes #1065
# Unicode characters are received corrupted if a default "application/json" content type is set

## Description
The change to the default charset of UTF-8 should support unicode symbols in the incoming messages

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes
As per developer's recommendations

### DIFF
--- a/src/main/java/org/zalando/nakadi/config/WebConfig.java
+++ b/src/main/java/org/zalando/nakadi/config/WebConfig.java
@@ -23,6 +23,7 @@ import org.zalando.nakadi.util.FlowIdRequestFilter;
 import org.zalando.nakadi.util.GzipBodyRequestFilter;
 
 import javax.servlet.Filter;
+import java.nio.charset.StandardCharsets;
 import java.util.List;
 
 @Configuration
@@ -68,7 +69,7 @@ public class WebConfig extends WebMvcConfigurationSupport {
 
     @Override
     protected void configureMessageConverters(final List<HttpMessageConverter<?>> converters) {
-        final StringHttpMessageConverter stringConverter = new StringHttpMessageConverter();
+        final StringHttpMessageConverter stringConverter = new StringHttpMessageConverter(StandardCharsets.UTF_8);
         stringConverter.setWriteAcceptCharset(false);
 
         converters.add(new ByteArrayHttpMessageConverter());


v-stepanov:
@ovolynets 
Hi. Thanks for contribution, but after deeper examination I think your fix doesn't really influence anything. 

Nakadi uses StringHttpMessageConverter with default encoding "ISO-8859-1”. So it will try to convert the request body to string using “ISO-8859-1” if the Content-Type header doesn’t have any information about charset, which seems to be a problem as it should use UTF-8 by default. But, this situation will actually never happen, because there was a fix in spring-boot specifically for this case: https://github.com/spring-projects/spring-boot/commit/109c3a343963ed4c861d775ce571d49c5b0639f2
It will add UTF-8 charset to Content-Type header if no charset is specified. Which will make StringHttpMessageConverter to use UTF-8. 

So current Nakadi behavior is that it will decode body with UTF-8 if user didn’t specify charset in his request. So I don't think that Nakadi needs any fix regarding that.

(in the chat I also described the behavior of the code you provided on your side)

ovolynets:
Hi @v-stepanov . Indeed, it seems to be the problem of the client code that should be always sending UTF-8 text. Let's close this issue now without any changes to the Nakadi code. Thanks for trying to reproduce the problem so quickly.

PR #1071, opened 2019-06-20
create curator ensemble and close curator for short commit timeout

adyach:
# One-line summary
create curator ensemble and close curator for short commit timeout


src/main/java/org/zalando/nakadi/service/subscription/zk/NewZkSubscriptionClient.java
```diff
@@ -216,4 +217,8 @@ public void transfer(final String sessionId, final Collection<EventTypePartition
         }
     }
 
+    @Override
+    public void close() throws IOException {
+        getCloseableCuratorFramework().close();
+    }
```
antban: Very strange. Obtain client in AbstractClient, release - in NewZkClient.

adyach: this is a mistake, I moved to abstract class

antban:
:+1:

adyach:
👍

PR #1072, opened 2019-06-20
close zk session after unregistering session

adyach:
# One-line summary
close zk session after unregistering session

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/state/CleanupState.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/state/CleanupState.java
@@ -20,11 +20,8 @@ public class CleanupState extends State {
     public void onEnter() {
         try {
             getContext().unregisterAuthorizationUpdates();
-            getContext().getZkClient().close();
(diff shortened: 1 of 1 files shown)


adyach:
👍

src/main/java/org/zalando/nakadi/service/subscription/state/CleanupState.java
```diff
@@ -33,6 +30,9 @@ public void onEnter() {
         } finally {
             try {
                 getContext().unregisterSession();
+                getContext().getZkClient().close();
```
antban: I believe that in case of any problems with zk unregisterSession() will throw exception, and client won't be closed.

adyach: my thinking was that if there is a problem in `unregisterSession` it is problem with zookeeper client, which could imply it is closed, I moved the call to another try catch

antban:
:+1:

adyach:
👍

PR #1068, opened 2019-06-18
Feature to force an authorization section for creation and update of ET

lmontrieux:
In order to create or update an event type, one must provide an
authorization section, if the feature FORCE_EVENT_TYPE_AUTHZ is
enabled

PR #1075, opened 2019-06-25
Added authorization feature on creation and updating of subscription

Kunal-Jha:
# One-line summary

> Zalando ticket : ARUHA-2376
https://jira.zalando.net/browse/ARUHA-2376

## Description
In order to create or update a subscription, one must provide an
authorization section, if the feature FORCE_EVENT_TYPE_AUTHZ is
enabled


src/main/java/org/zalando/nakadi/exceptions/runtime/AuthorizationSectionException.java
```diff
@@ -0,0 +1,8 @@
+package org.zalando.nakadi.exceptions.runtime;
+
+public class AuthorizationSectionException extends NakadiBaseException {
```
ferbncode: Maybe the name of the exception could be more clear like `AuthorizationSectionAbsentException` or `AuthorizationAbsentException`? What do you think?

ferbncode:
Also, do you think it is good if we add a test for this?

src/main/java/org/zalando/nakadi/service/FeatureToggleService.java
```diff
@@ -39,7 +39,8 @@ default void checkFeatureOn(final Feature feature) {
         KPI_COLLECTION("kpi_collection"),
         AUDIT_LOG_COLLECTION("audit_log_collection"),
         DISABLE_DB_WRITE_OPERATIONS("disable_db_write_operations"),
-        DISABLE_LOG_COMPACTION("disable_log_compaction");
+        DISABLE_LOG_COMPACTION("disable_log_compaction"),
+        FORCE_SUBSCRIPTION_AUTHZ("force_subscription_authz");
```
lmontrieux: Could we have a name that's more like the one in https://github.com/zalando/nakadi/pull/1068 ? We should try to be consistent whenever possible.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -119,11 +122,10 @@ public SubscriptionService(final SubscriptionDbRepository subscriptionRepository
...
             NoSuchEventTypeException, InconsistentStateException, WrongInitialCursorsException,
-            DbWriteOperationsBlockedException, UnableProcessException, ServiceTemporarilyUnavailableException {
-        if (featureToggleService.isFeatureEnabled(FeatureToggleService.Feature.DISABLE_DB_WRITE_OPERATIONS)) {
-            throw new DbWriteOperationsBlockedException("Cannot create subscription: write operations on DB " +
-                    "are blocked by feature flag.");
-        }
+            DbWriteOperationsBlockedException, UnableProcessException,
+            AuthorizationSectionException, ServiceTemporarilyUnavailableException {
+
+        checkFeatureToggles(subscriptionBase);
```
lmontrieux: I like that you move the checks for feature toggles to their own private method, it removes clutter.
However, I would suggest to be more precise in the naming: you're actually checking feature toggles that are relevant to the creation and update of subscriptions specifically, and this should be reflected in the method name.

src/main/resources/application.yml
```diff
@@ -183,3 +183,4 @@ nakadi.features.defaultFeatures:
   REMOTE_TOKENINFO: true
   KPI_COLLECTION: true
   DISABLE_DB_WRITE_OPERATIONS: false
+  FORCE_SUBSCRIPTION_AUTHZ: true
```
lmontrieux: For local deployments I would not make this mandatory: it keeps things as simple as possible.

src/main/java/org/zalando/nakadi/service/subscription/SubscriptionService.java
```diff
@@ -143,21 +143,21 @@ public Subscription createSubscription(final SubscriptionBase subscriptionBase)
         return subscription;
     }
 
-    private void checkFeatureToggles(final SubscriptionBase subscription) {
+    private void checkFeatureTogglesForSubscription(final SubscriptionBase subscription) {
```
lmontrieux: Sorry, I still think we need a better name for this. With this name, someone may think that this method also applied to getExistingSubscription(..) or listSubscriptions(..).

I would suggest: `checkFeatureTogglesForCreationAndUpdate(..)`. It's a mouthful, but it's clear.

lmontrieux:
👍

Kunal-Jha:
👍

PR #1078, opened 2019-07-16
ARUHA-2414 Increase kafka buffer size

antban:
ARUHA-2414 Increase kafka buffer size

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -65,6 +65,10 @@ nakadi:
     send.timeoutMs: 5000
     batch.size: 5242880
     max.request.size: 2098152
+    buffer.memory: 83886080 # May be defined as X * Y  * Z, where:
+                            # X - avg incoming traffic per instance per second (here: 4Mb/s)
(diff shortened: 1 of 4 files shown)


antban:
:+1:

src/main/resources/application.yml
```diff
@@ -65,6 +65,10 @@ nakadi:
     send.timeoutMs: 5000
     batch.size: 5242880
     max.request.size: 2098152
+    buffer.memory: 83886080 # May be defined as X * Y  * Z, where:
```
rcillo: Is this memory per producer? Do we have a single producer per instance (sorry a cannot remember anymore)?

rcillo:
👍

src/main/resources/application.yml
antban: We have as many producers as active storages.
There is also a case when producers are dying, and it may be temporary more than 1 producer per active storage. So it should be fine.

antban: Btw, default kafka value is 32 Mb. I changed to 80 Mb

rcillo: Ok 👍  sounds good

PR #1080, opened 2019-08-07
Check old authorization attributes when updating

ferbncode:
Check that the attributes (and hence the admin `*`, `*` is checked for updates too).

## Description
Right now, it is not checked for updates on event-types and subscriptions that the admin doesn't contain wildcard.


Kunal-Jha:
👍

ferbncode:
:+1:

PR #1086, opened 2019-09-10
Remove unused variables

lmontrieux:
# Remove unused variables

I found some things that we declare, but never use. They should go away.


lmontrieux:
:+1:

ferbncode:
:+1:

PR #1088, opened 2019-09-12
Allow admins to delete event types

lmontrieux:
To delete an event type if EVENT_TYPE_DELETION_ONLY_ADMINS, the following criteria must be met:
- the caller is in the admins section (possibly through a wildcard)
- there is an admin section
- the only subscriptions attached to the event type are subscriptions defined in the settings nakadi.eventType.deletableSubscription.owningApplication and nakadi.eventType.deletableSubscription.consumerGroup

### DIFF
--- a/src/main/java/org/zalando/nakadi/config/NakadiSettings.java
+++ b/src/main/java/org/zalando/nakadi/config/NakadiSettings.java
@@ -23,6 +23,8 @@ public class NakadiSettings {
     private final AuthorizationAttribute defaultAdmin;
     private final String warnAllDataAccessMessage;
     private final String logCompactionWarnMessage;
+    private final String deletableSubscriptionOwningApplication;
+    private final String deletableSubscriptionConsumerGroup;
 
     @Autowired
     public NakadiSettings(@Value("${nakadi.topic.max.partitionNum}") final int maxTopicPartitionCount,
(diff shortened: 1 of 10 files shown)


src/main/java/org/zalando/nakadi/config/NakadiSettings.java
```diff
@@ -39,7 +41,11 @@ public NakadiSettings(@Value("${nakadi.topic.max.partitionNum}") final int maxTo
                           @Value("${nakadi.admin.default.dataType}") final String defaultAdminDataType,
                           @Value("${nakadi.admin.default.value}") final String defaultAdminValue,
                           @Value("${nakadi.authz.warnAllDataAccessMessage}") final String warnAllDataAccessMessage,
-                          @Value("${nakadi.topic.compacted.warnMessage}") final String logCompactionWarnMessage) {
+                          @Value("${nakadi.topic.compacted.warnMessage}") final String logCompactionWarnMessage,
+                          @Value("${nakadi.eventType.deletableSubscription.owningApplication}")
```
adyach: I do not know if you thought about it maybe it should be the list of applications, but I do not insist on that

src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -148,7 +148,11 @@ public void deleteSubscription(final String id)
         }
 
         queryBuilder.append(" ORDER BY s_subscription_object->>'created_at' DESC LIMIT ? OFFSET ? ");
-        params.add(limit);
```
adyach: the change is not clear, there is a default value passed of 20 and there is a check in SubscriptionService

adyach: ah, I see what you do down there

adyach: if response from DB will be too big, it will Nakadi, for me it makes sense to use pagination

lmontrieux: Fixed in 14d8fd6 and 89b80aa

adyach:
👍

Kunal-Jha:
👍

lmontrieux:
:+1:

lmontrieux:
:-1: 

Do not merge for now: I'd like to discuss what to do if an event type is the source of a Nakadi SQL query, first.

lmontrieux:
:+1:

PR #1089, opened 2019-09-17
Added ability to increase partitions

Kunal-Jha:
# Allow users to increase partitions

> Zalando ticket : ARUHA-2533 

## Description
In order to self increase the partitions, we need to allow users to increase the number of partitions with certain conditions. 
## Review
- [x] Tests

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -496,8 +496,15 @@ public class EventTypeService {
         if (existing != null && newStatistics == null) {
             return existing;
         }
-        if (!Objects.equals(existing, newStatistics)) {
-            throw new InvalidEventTypeException("default statistics must not be changed");
+        if ((newStatistics.getReadParallelism() > nakadiSettings.getMaxTopicPartitionCount())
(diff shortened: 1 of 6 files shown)


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -496,8 +496,15 @@ private EventTypeStatistics validateStatisticsUpdate(
         if (existing != null && newStatistics == null) {
             return existing;
         }
-        if (!Objects.equals(existing, newStatistics)) {
-            throw new InvalidEventTypeException("default statistics must not be changed");
+        if ((newStatistics.getReadParallelism() > nakadiSettings.getMaxTopicPartitionCount())
```
rcillo: This means that you can change it in a non meaningful way.

Previous: `read=2, write=4`. After: `read=3, write=4`. Do we want to prevent this?

Kunal-Jha: Since we take maximum of read and write parallilsm there is no change in partition count i.e., it will remain 4 however, the default statistics will still update read to 3.

Kunal-Jha:
Closing it here to open it on a separate branch

PR #1090, opened 2019-09-18
Allow users to increase partitions

Kunal-Jha:
# Allow users to increase partitions

> Zalando ticket : ARUHA-2533 
https://jira.zalando.net/browse/ARUHA-2534

## Description
In order to self increase the partitions, we need to allow users to increase the number of partitions with certain conditions. 
## Review
- [x] Tests

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -496,8 +496,8 @@ public class EventTypeService {
         if (existing != null && newStatistics == null) {
             return existing;
         }
-        if (existing == null && newStatistics == null) {
-            return null;
+        if ((existing == null && newStatistics == null) || (existing.equals(newStatistics))) {
+            return existing;
         }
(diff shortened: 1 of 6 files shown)


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -496,8 +496,28 @@ private EventTypeStatistics validateStatisticsUpdate(
         if (existing != null && newStatistics == null) {
             return existing;
         }
-        if (!Objects.equals(existing, newStatistics)) {
-            throw new InvalidEventTypeException("default statistics must not be changed");
+        if (existing == null && newStatistics == null) {
```
ferbncode: Maybe we can check if both are equal, if so, we don't need to validate again given it was validated before? So, the if would look like:
```
if ((existing == null && newStatistics == null) || existing.equals(newStatistics)) {
    return existing;
}
```

ferbncode:
:+1:

Kunal-Jha:
👍

PR #1092, opened 2019-09-23
ARUHA-2536: Add new partitions to topology for subscription

ferbncode:
> Zalando ticket : ARUHA-2536 (https://jira.zalando.net/browse/ARUHA-2536)

## Description
Edit Zookeeper topology for subscriptions if event type is re-partitioned.
## Review
- [x] Tests

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/zk/NewZkSubscriptionClient.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/zk/NewZkSubscriptionClient.java
@@ -3,6 +3,7 @@ package org.zalando.nakadi.service.subscription.zk;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
 import org.apache.zookeeper.KeeperException;
 import org.zalando.nakadi.domain.EventTypePartition;
 import org.zalando.nakadi.exceptions.runtime.NakadiRuntimeException;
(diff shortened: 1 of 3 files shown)


src/main/java/org/zalando/nakadi/service/subscription/zk/NewZkSubscriptionClient.java
```diff
@@ -216,4 +218,50 @@ public void transfer(final String sessionId, final Collection<EventTypePartition
...
+
+    @Override
+    public void repartitionTopology(final String eventTypeName, final int newPartitionsCount)
+            throws NakadiRuntimeException {
+        final Topology currentTopology = getTopology();
+        final List<Partition> partitionsList = Lists.newArrayList(currentTopology.getPartitions());
+        final List<Partition> newPartitionsList = new ArrayList<>();
+
+        // Set other event type's partition state to unassigned
+        for (final Partition partition : partitionsList) {
```
adyach: why do we set current partitions in use to unassigned state?

ferbncode: If I understood correctly, we should disconnect all the sessions using this subscription, and they reconnect again? (If this is the case, then maybe I should clean the `/subscriptions/<id>/sessions` node too.)

adyach: This will not help with reconnect, because you also should wait for commit to come until commit timeout. I would not go this way, because it changes internal structure at runtime, I would rather do it the same way it is done for reset cursor, but it is work for another pr.

ferbncode: Got it, thanks! :), fixed this in `d0df4e5`.

adyach:
👍

PR #1093, opened 2019-10-07
Aruha 2537 1

adyach:
# One-line summary
Rework of subscription reset cursor to apply action for subscription streams close. It is required in order to close streaming subscription after re-partitioning.


Kunal-Jha:
👍

PR #1094, opened 2019-10-07
Tracing for publishing 

Kunal-Jha:
# Added tracing for publishing

> Zalando ticket: ARUHA-2467

## Description
This PR adds tracing for the publishing of data to Nakadi. This would give the publisher the visibility of their events when publishing to Nakadi. 

https://jira.zalando.net/browse/ARUHA-2547


src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -52,8 +57,8 @@ public EventPublishingController(final EventPublisher publisher,
                                      final EventTypeMetricRegistry eventTypeMetricRegistry,
                                      final BlacklistService blacklistService,
                                      final NakadiKpiPublisher nakadiKpiPublisher,
-                                     @Value("${nakadi.kpi.event-types.nakadiBatchPublished}")
-                                         final String kpiBatchPublishedEventType) {
+                                     @Value("${nakadi.kpi.event-types.nakadiBatchPublished}") final
```
rcillo: I wonder if wouldn't it be possible to capture this request via a springboot filter and if you further need to add things to the `span` that are not available in the filter, you could still create the span in the filter and pass it to the controller as the last parameter. Something like the client resolver https://github.com/zalando/nakadi/blob/1643b2182c1b4bb257ad6d822f48971eeaf9f91f/src/main/java/org/zalando/nakadi/config/WebConfig.java#L85 if I'm not mistaken.

Also, how is this code coexisting with already configured operation to track publishing? Did it replace that? I talk about the one in https://github.com/zalando/nakadi/blob/41fd11ada6d513cf5c4fe8e9fb16e82c12cf326a/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java#L185

Kunal-Jha:
> To make the code less intrusive (not changing the interface for all the methods), did you consider doing something similar to the FlowidUtils
> 
> https://github.com/zalando/nakadi/blob/1643b2182c1b4bb257ad6d822f48971eeaf9f91f/src/main/java/org/zalando/nakadi/util/FlowIdUtils.java#L18
> 
> where we keep a stack per thread. You could have the "immediately open" span on top of the stack and whenever you want to produce a children out of it you can get it from the stack.

@rcillo I considered it. I think we had a discussion about that. Since NDC takes string input only we cannot use it as we cannot recreate a span just from the id/context. We will have to create a custom "smart" stack to store span objects to pass around the scope manager and creating this data structure seemed too complicated to me. We can discuss it further in person.

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -69,37 +75,52 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName,
             throws AccessDeniedException, BlockedException, ServiceTemporarilyUnavailableException,
             InternalNakadiException, EventTypeTimeoutException, NoSuchEventTypeException {
         LOG.trace("Received event {} for event type {}", eventsAsString, eventTypeName);
-
+        final HttpServletRequest httpServletRequest = request.getNativeRequest(HttpServletRequest.class);
+        final Scope publishingScope = TracingService
```
rcillo: Is it possible that a `span` fails to be instantiated?  I'm just asking to make sure that broken instrumentation would not cause Nakadi to fail.

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -180,22 +199,14 @@ private void traceRequest(final RequestLogInfo requestLogInfo, final int statusC
...
-                .setTag("content_length", requestLogInfo.contentLength)
-                .finish(TimeUnit.MILLISECONDS.toMicros(requestLogInfo.requestTime + timeSpentMs));
+        final String eventType = requestLogInfo.path.substring("/event-types/".length(),
+                requestLogInfo.path.lastIndexOf("/events"));
+        TracingService.setCustomTags(scope, ImmutableMap.<String, Object>builder()
+                .put("event_type", eventType)
+                .put("http.status_code", statusCode)
+                .put("slo_bucket", sloBucket)
+                .put("content_length", requestLogInfo.contentLength)
+                .put("error", statusCode == 207 || statusCode >= 500).build());
```
rcillo: Isn't `error` also set in the controller?

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -222,67 +241,91 @@ public boolean topicExists(final String topic) throws TopicRepositoryException {
...
+                final Map<BatchItem, CompletableFuture<Exception>> sendFutures = new HashMap<>();
+                for (final BatchItem item : batch) {
+                    item.setStep(EventPublishingStep.PUBLISHING);
+                    final HystrixKafkaCircuitBreaker circuitBreaker = circuitBreakers.computeIfAbsent(
+                            item.getBrokerId(), brokerId -> new HystrixKafkaCircuitBreaker(brokerId));
+                    if (circuitBreaker.allowRequest()) {
+                        sendFutures.put(item, publishItem(producer, topicId, item, circuitBreaker, publishingSpan));
+                    } else {
+                        shortCircuited++;
+                        item.updateStatusAndDetail(EventPublishingStatus.FAILED, "short circuited");
```
rcillo: At other places you replaced this by `handlePublishingException`. Any reason why not doing it here as well?

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
Kunal-Jha: I did not see any exception being declared in the opentracing definitions. AFAIK it suppresses failures.

src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
rcillo: Then I guess I missed something. Why is there a "filter" span and a "controller" span? Can't we have a single "parent" span that lasts throughout the entire publishing operation?

Kunal-Jha: I see your point. I was hoping to have a child span starting at the controller level but since it does not add much value alone we can combine it. Fixed in bceb690

src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
Kunal-Jha: fixed in b0cbd05

rcillo:
👍

Kunal-Jha:
👍

PR #1097, opened 2019-10-09
ARUHA-2538: repartitioning itself combined

adyach:
# One-line summary
Combine depended tasks interface to repartition event type

Add partitions to active timeline Kafka topic
Update timelines of the event type to have offsets for partitions added
Update event type cache
Close streaming connections for the subscriptions

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -72,6 +72,7 @@ import java.util.Date;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
@@ -576,7 +577,7 @@ public class EventTypeService {
             zkClient.repartitionTopology(eventType.getName(), partitions);
             zkClient.closeSubscriptionStreams(
                     () -> LOG.info("subscription streams were closed, after repartitioning"),
-                    nakadiSettings.getMaxCommitTimeout());
+                    TimeUnit.SECONDS.toMillis(nakadiSettings.getMaxCommitTimeout()));
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
```diff
@@ -172,6 +171,12 @@ public void updateTimeLineForRepartition(final EventType eventType, final int pa
             }
             timelineDbRepository.updateTimelime(timeline);
         }
+
+        try {
+            eventTypeCache.updated(eventType.getName());
+        } catch (Exception e) {
+            throw new NakadiBaseException(e.getMessage(), e);
```
lmontrieux: Is there a better exception we could throw?

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -537,22 +543,36 @@ private void repartitionEventType(
...
+    public void updateSubscriptionsForRepartitioning(final EventType eventType, final int partitions)
+            throws NakadiBaseException {
+        final List<Subscription> subscriptions = subscriptionRepository.listSubscriptions(
+                ImmutableSet.of(eventType.getName()), Optional.empty(), 0, 1000);
+        for (final Subscription subscription : subscriptions) {
+            final ZkSubscriptionClient zkClient = subscriptionClientFactory.createClient(subscription,
+                    LogPathBuilder.build(subscription.getId(), "repartition"));
+            zkClient.repartitionTopology(eventType.getName(), partitions);
+            zkClient.closeSubscriptionStreams(
+                    () -> LOG.info("subscription streams were closed, after repartitioning"), 60000);
```
lmontrieux: Can we use a constant (and also in line 567?)

lmontrieux:
@adyach I have finished the review

src/main/java/org/zalando/nakadi/service/timeline/TimelineService.java
adyach: I do not see a reason for that, at the end it is converted to internal server error in the calling method

src/main/java/org/zalando/nakadi/service/EventTypeService.java
adyach: done

lmontrieux:
:+1:

adyach:
👍

lmontrieux:
:+1:

PR #1099, opened 2019-10-11
Added tracing for subscription and commit end points

Kunal-Jha:
# Added tracing for subscription and commit endpoints

## Zalando ticket: ARUHA-2547
https://jira.zalando.net/browse/ARUHA-2547

## Description
This PR adds tracing for the consumption and committing of data to Nakadi. This would give the publisher the visibility of their events when publishing to Nakadi.

### DIFF
--- a/src/main/java/org/zalando/nakadi/controller/CursorsController.java
+++ b/src/main/java/org/zalando/nakadi/controller/CursorsController.java
@@ -1,5 +1,6 @@
 package org.zalando.nakadi.controller;
 
+import io.opentracing.Span;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.PathVariable;
@@ -28,6 +29,7 @@ import org.zalando.nakadi.view.CursorCommitResult;
 import org.zalando.nakadi.view.SubscriptionCursor;
 import org.zalando.nakadi.view.SubscriptionCursorWithoutToken;
 
+import javax.servlet.http.HttpServletRequest;
 import javax.validation.Valid;
 import javax.validation.constraints.NotNull;
 import java.util.ArrayList;
@@ -78,6 +80,7 @@ public class CursorsController {
     public ResponseEntity<?> commitCursors(@PathVariable("subscriptionId") final String subscriptionId,
                                            @Valid @RequestBody final ItemsWrapper<SubscriptionCursor> cursorsIn,
                                            @NotNull @RequestHeader("X-Nakadi-StreamId") final String streamId,
+                                           final HttpServletRequest request,
                                            final Client client)
             throws NoSuchEventTypeException,
             NoSuchSubscriptionException,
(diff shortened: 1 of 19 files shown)


src/main/java/org/zalando/nakadi/filters/TracingFilter.java
```diff
@@ -0,0 +1,226 @@
...
+            GlobalTracer.get().inject(baseSpan.context(),
+                    HTTP_HEADERS, new TextMapInjectAdapter(spanContextToInject));
+            response.setHeader(SPAN_CONTEXT, spanContextToInject.toString());
+            baseSpan.finish();
+        }
+    }
+
+
+    private RequestType getRequestType(final RequestInfo requestLogInfo) {
+        if (requestLogInfo.path != null && "POST".equals(requestLogInfo.method) &&
```
thomasabraham: Maybe we can add `if(requestLogInfo.path == null){ return RequestType.OTHER; }` above and remove the null check from individual if cases.

src/main/java/org/zalando/nakadi/service/TracingService.java
```diff
@@ -1,6 +1,7 @@
 package org.zalando.nakadi.service;
 
 import com.google.common.collect.ImmutableMap;
+import io.opentracing.References;
```
thomasabraham: Why `TracingService` is annotated with `@Configuration`?

src/main/java/org/zalando/nakadi/filters/TracingFilter.java
```diff
@@ -0,0 +1,244 @@
...
+        if (requestType.equals(RequestType.OTHER)) {
+            return;
+        }
+        final Map<String, Object> tags = new HashMap<String, Object>() {{
+            put("http.status_code", statusCode);
+            put("content_length", requestLogInfo.contentLength);
+        }};
+
+        if (requestType.equals(RequestType.PUBLISHING)) {
+            String sloBucket = "5K-50K";
```
thomasabraham: Is this duplicated in Publishing controller?

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -78,36 +78,34 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName,
         final HttpServletRequest httpServletRequest = request.getNativeRequest(HttpServletRequest.class);
         final Scope publishingScope = TracingService
                 .activateSpan((Span) httpServletRequest.getAttribute("span"), false);
+        String sloBucket = "5K-50K";
```
thomasabraham: Is this somehow duplicate?

thomasabraham:
👍

src/main/java/org/zalando/nakadi/controller/CursorsController.java
```diff
@@ -88,15 +93,16 @@ public CursorsController(final CursorsService cursorsService,
         if (cursors.isEmpty()) {
             throw new CursorsAreEmptyException();
         }
-
+        final Scope commitScope = TracingService.activateSpan((Span) request.getAttribute("span"), false);
```
adyach: can we have `TracingService.activateSpan` take `request`, and do all the casting within the method ?

src/main/java/org/zalando/nakadi/controller/EventPublishingController.java
```diff
@@ -78,36 +78,35 @@ public ResponseEntity postEvent(@PathVariable final String eventTypeName,
         final HttpServletRequest httpServletRequest = request.getNativeRequest(HttpServletRequest.class);
         final Scope publishingScope = TracingService
                 .activateSpan((Span) httpServletRequest.getAttribute("span"), false);
+        publishingScope.span().setOperationName("publish_events");
+        String sloBucket = "5K-50K";
```
adyach: make sense to extract to the method and move values to constants
`getSloBucket(httpServletRequest)`

adyach:
@Kunal-Jha in general I agree with @rcillo that having logic to identify the commit cursor endpoint in filter is not the best way to go, I would rather pass up on have a reference span. At the same time having refactoring the whole exceptions handing logic is also not an option otherwise it is not possible to finish spans with specific errors context which will have a lot more changes in the existing code, so I would say it is fine to go with this changes. Maybe @rcillo has something to comment about here.

adyach:
👍

Kunal-Jha:
👍

PR #1101, opened 2019-10-14
Validating informational fields against effective schema

thomasabraham:
# Validating informational fields (ordering_key_fields and ordering_instance_ids) against effective schema.

> Zalando ticket : [ARUHA-2048](https://jira.zalando.net/browse/ARUHA-2048)

## Description
Both ordering_key_fields and ordering_instance_ids are validated against the user-provided schema during event type creation, but the documentation (and the API guild) say that it should be validated against the effective schema, such that people could reference the metadata object, for example.

AC:
Both fields validated against the effective schema during event type creation.

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -27,6 +27,7 @@ import org.zalando.nakadi.domain.Timeline;
 import org.zalando.nakadi.enrichment.Enrichment;
 import org.zalando.nakadi.exceptions.runtime.AccessDeniedException;
 import org.zalando.nakadi.exceptions.runtime.AuthorizationSectionException;
+import org.zalando.nakadi.exceptions.runtime.CannotAddPartitionToTopicException;
 import org.zalando.nakadi.exceptions.runtime.ConflictException;
 import org.zalando.nakadi.exceptions.runtime.DbWriteOperationsBlockedException;
 import org.zalando.nakadi.exceptions.runtime.DuplicatedEventTypeNameException;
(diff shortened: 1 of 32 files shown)


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -101,6 +102,7 @@
     private final String etLogEventType;
     private final NakadiAuditLogPublisher nakadiAuditLogPublisher;
     private final EventTypeOptionsValidator eventTypeOptionsValidator;
+    private final JsonSchemaEnrichment jsonSchemaEnrichment = new JsonSchemaEnrichment();
```
Kunal-Jha: Isn't it better to add the initialization to the constructor?

src/test/java/org/zalando/nakadi/service/EventTypeServiceTest.java
```diff
@@ -215,6 +218,16 @@ public void testFeatureToggleForbidsDeleteEventTypeWithNonDeletableSubscription(
...
     }
 
+    @Test
+    public void testAllowCreatingEventTypeWithInformationalFieldsFromEffectiveSchema(){
+        final EventType et = EventTypeTestBuilder.builder()
+                .category(EventCategory.DATA)
+                .build();
+        et.setOrderingKeyFields(Collections.singletonList("metadata.occurred_at"));
+        et.setOrderingInstanceIds(Collections.singletonList("metadata.partition"));
+        eventTypeService.create(et, true);
```
Kunal-Jha: Shouldn't the test have some validation step?

Kunal-Jha:
@thomasabraham looks good. 👍  some minor comments. Also please change the description to reflect the things changed in the PR. Alos adding a link to the ticket is usually helpful

src/test/java/org/zalando/nakadi/service/EventTypeServiceTest.java
thomasabraham: Before the fix, the test will break with exception.

src/main/java/org/zalando/nakadi/service/EventTypeService.java
thomasabraham: Done

src/test/java/org/zalando/nakadi/service/EventTypeServiceTest.java
thomasabraham: Okay I got the point; fixed.

thomasabraham:
Screwed up the branch; will create a new branch.

PR #1102, opened 2019-10-14
Trigger repartitioning only if parallelism changes

ferbncode:
Change re-partitioning only if parallelism parameters of `default-statistics` change.
For this we return if they are equal.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -546,8 +546,7 @@ public class EventTypeService {
             if (getMaxPartitions(newStatistics) == 1) {
                 return;
             } else {
-                updateNumberOfPartitions(original, newEventType, Math.max(newStatistics.getReadParallelism(),
-                        newStatistics.getWriteParallelism()));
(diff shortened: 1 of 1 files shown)


src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -546,7 +546,8 @@ private void repartitionEventType(
                     newStatistics.getWriteParallelism()));
             return;
         }
-        if (existingStatistics.equals(newStatistics)) {
+        if (existingStatistics.getReadParallelism().equals(newStatistics.getReadParallelism())
```
adyach: you have the same check on line https://github.com/zalando/nakadi/pull/1102/files#diff-0cc0d4406e77f42ca8a2ee3e3ce445e1R569

also on line https://github.com/zalando/nakadi/pull/1102/files#diff-0cc0d4406e77f42ca8a2ee3e3ce445e1R544
is a mistake, it should not trigger repartition from 1 partition to 1

ferbncode: Right, fixed in `0ec119d`

ferbncode:
Closing as planned to be done in https://github.com/zalando/nakadi/pull/1104.

PR #1114, opened 2019-11-01
ARUHA-1714 Use different approach for caching event types information

antban:
ARUHA-1714 Use different approach for caching event types information


src/main/java/org/zalando/nakadi/cache/EventTypeDataProvider.java
```diff
@@ -0,0 +1,83 @@
...
+                        .filter(EventTypeDbRepository.EtChange::isDeleted)
+                        .map(EventTypeDbRepository.EtChange::getName)
+                        .collect(Collectors.toList())
+        );
+    }
+
+    private EventTypeProxy convert(final EventType et) {
+        return new EventTypeProxy(
+                et,
+                JsonUtils.serializeDateTime(objectMapper, et.getUpdatedAt())
```
rcillo: When we use `updatedAt` for versioning we are assuming that system clock is good enough as an indication of order, given concurrent changes are not very frequent, right?

src/main/java/org/zalando/nakadi/cache/SimpleCache.java
```diff
@@ -0,0 +1,99 @@
...
+
+    public SimpleCache(
+            final CacheDataProvider<Versioned, Version> dataProvider,
+            final Function<Versioned, T> rawDataConverter) {
+        this.dataProvider = dataProvider;
+        this.rawDataConverter = rawDataConverter;
+    }
+
+    private T loadItemLocked(final String key) {
+        // this operation is relatively long, therefore we are making it concurrent
```
rcillo: I didn't understand. How is this concurrent? You mean that `computeIfAbsent` doesn't block other concurrent calls to different keys?

src/main/java/org/zalando/nakadi/cache/SimpleCache.java
```diff
@@ -0,0 +1,99 @@
...
+        // 2 - Just remove data from cache.
+        // As massive deletion of event types is actually rare operation, we will use second approach (it's simpler)
+
+        // order of removal is important, btw.
+        synchronized (versionedDataLock) {
+            fullChangeList.getDeletedKeys().forEach(versionedData::remove);
+            fullChangeList.getModified().forEach(versionedData::remove);
+        }
+        // Once registered data was removed - we can remove it from cache.
+        // It's find if versionedData will contain values, while activeList - not.
```
rcillo: s/find/fine/

src/main/java/org/zalando/nakadi/cache/Cache.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Sometimes it is important to ensure that cache will be providing refreshed version of value.
+     * This method ensures that value will be refreshed.
+     *
+     * @param key key to invalidate
+     */
+    void invalidate(String key);
+
+    /**
+     * Adds listener for event type invalidation. It is guaranteed, that any event type that was loaded into cache
```
v-stepanov: Your comment mentions event-type but as I understand this cache interface is for general purpose, not specifically for event-type.

src/main/java/org/zalando/nakadi/cache/CacheChange.java
```diff
@@ -0,0 +1,21 @@
+package org.zalando.nakadi.cache;
+
+import java.util.List;
+
+public class CacheChange {
+    private final List<String> modified;
+    private final List<String> deletedKeys;
+
+    public CacheChange(final List<String> modified, final List<String> deletedKeys) {
```
v-stepanov: Does `modified` parameter also contain keys? If yes - it probably makes sense to name it to `modifiedKeys` to keep naming consistent with `deletedKeys`

src/main/java/org/zalando/nakadi/cache/CacheDataProvider.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Defines interface for cache data provider.
+ *
+ * @param <T> the type of cached entity
+ * @param <V> Type of version used for entities
+ */
+public interface CacheDataProvider<T extends VersionedEntity<V>, V extends Comparable> {
+    /**
+     * Loads value for cache. In case if value is not found method should return null. Several calls of this method may
```
v-stepanov: > In case if value is not found method should return null

`@Nullable` then?

src/main/java/org/zalando/nakadi/cache/EventTypeDataProvider.java
```diff
@@ -0,0 +1,83 @@
...
+        );
+    }
+
+    public static class EventTypeProxy implements VersionedEntity<String> {
```
v-stepanov: I'm a bit confused with a name `EventTypeProxy`. What do you think about `VersionedEventType` or something like that?

src/main/java/org/zalando/nakadi/util/JsonUtils.java
```diff
@@ -13,4 +16,14 @@ public static void checkEventTypeSchemaValid(final String jsonInString) throws I
             throw new InvalidEventTypeException("schema must be a valid json: " + jpe.getMessage());
         }
     }
+
+    public static String serializeDateTime(final ObjectMapper objectMapper, final DateTime dateTime) {
+        try {
+            final String serialized = objectMapper.writeValueAsString(dateTime);
+            return serialized.substring(1, serialized.length() - 1);
```
v-stepanov: what happens here?

src/main/java/org/zalando/nakadi/repository/db/EventTypeDbRepository.java
```diff
@@ -92,6 +96,69 @@ private void insertEventTypeSchema(final EventType eventType) throws JsonProcess
...
+                    ", deleted=" + deleted +
+                    '}';
+        }
+    }
+
+    public List<EtChange> getChangeset(final Map<String, String> listToGet) {
+        if (listToGet.isEmpty()) {
+            return Collections.emptyList();
+        }
+        // Yes, we know that bat symbols are not allowed in both event type name and updated at, but still filtering
```
v-stepanov: https://media.giphy.com/media/l396BoOTIFem9xqQU/giphy.gif

src/main/java/org/zalando/nakadi/cache/Cache.java
```diff
@@ -0,0 +1,42 @@
+package org.zalando.nakadi.cache;
+
+import java.util.function.Consumer;
+
+public interface Cache<T> {
+    /**
+     * Returns value associated with a key. If value is not found, returns null.
```
v-stepanov: `@Nullable`?

v-stepanov:
@antban can you please remind me what is inside ARUHA-1714 ?
AFAIR the main problem of old cache was that it relied on ZK watchers which sometimes stopped working, and that resulted in cache holding the old value and unhappy users. But as I see the new cache also relies on ZK watchers.
So it's not exactly clear for me which problems the new cache solves.

v-stepanov:
Ok, I see now. It has a force refresh.

src/main/java/org/zalando/nakadi/cache/EventTypeDataProvider.java
antban: Yep, changes are not very frequent + we are listening for any modification of the value, not just increase.

src/main/java/org/zalando/nakadi/cache/SimpleCache.java
antban: Yes, exactly

src/test/java/org/zalando/nakadi/cache/SimpleCacheTest.java
```diff
@@ -0,0 +1,158 @@
...
+        cache.invalidate("key1");
+        Assert.assertEquals(Arrays.asList("key1"), recordedChanges);
+        cache.get("key1"); // Let's put it again to cache
+
+        Mockito.when(dataProvider.getFullChangeList(Mockito.anySet())).thenReturn(
+                new CacheChange(Arrays.asList("key3"), Arrays.asList("key1")));
+        cache.refresh();
+        Assert.assertEquals(Arrays.asList("key1", "key1", "key3"), recordedChanges);
+    }
+}
```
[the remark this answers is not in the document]
antban: Sorry, I didn't get the joke (((

src/main/java/org/zalando/nakadi/cache/CacheDataProvider.java
antban: OMG. PR Police is here )) Will fix

src/main/java/org/zalando/nakadi/util/JsonUtils.java
antban: `"` symbols removed

src/test/java/org/zalando/nakadi/cache/SimpleCacheTest.java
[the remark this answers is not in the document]
v-stepanov: It's not visible here, but if you see it on "Files Changed" tab - you will see that there is no endline at the end of the file :)

rcillo:
@antban does this change cache strategy to lazy loading during boot time?

src/acceptance-test/java/org/zalando/nakadi/cache/ZookeeperNodeInvalidatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+        final List<ZookeeperNodeInvalidator> invalidators = caches
+                .stream()
+                .map(v -> new ZookeeperNodeInvalidator(v, CURATOR, "/xxx/yyy", TimeUnit.MINUTES.toMillis(10)))
+                .collect(Collectors.toList());
+
+        invalidators.forEach(ZookeeperNodeInvalidator::start);
+        try {
+            // Wait 100 ms for caches to start
+            ThreadUtils.sleep(200);
```
rcillo: wouldn't it be better to wait on a condition instead of waiting on time? This tends to produce brittle tests.

src/acceptance-test/java/org/zalando/nakadi/cache/ZookeeperNodeInvalidatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+            // Notify about updates several times and ensure that all caches are updated.
+            for (int idx = 0; idx < 5; ++idx) {
+                invalidators.get(idx % invalidators.size()).notifyUpdate();
+                ThreadUtils.sleep(200);
```
rcillo: The same applied for this. It would be better always that possible not to have tests which rely on `sleep`.

src/acceptance-test/java/org/zalando/nakadi/cache/ZookeeperNodeInvalidatorTest.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test(timeout = 1000)
+    public void cachePeriodicUpdateTriggered() throws InterruptedException {
+        final Cache cache = Mockito.mock(Cache.class);
+
+        final ZookeeperNodeInvalidator invalidator =
+                new ZookeeperNodeInvalidator(cache, CURATOR, "/xxx/yyy1", 100);
+        invalidator.start();
+        try {
+            ThreadUtils.sleep(600);
```
rcillo: We can leave this test like this because I cannot think on an easy way to verify that there is a `refresher` running, but if it fails due to timing, then I wouldn't see a problem in removing it.

rcillo:
@antban the code looks good and I don't have any constructive feedback here, besides minor comments on tests. From my understanding it improves the situation a lot by:

1. reducing amount of watchers to Zookeeper
2. reacting to missed notifications by comparing Zxid

It does change the dynamics of how the application works a little, and I cannot tell how it's going to behave on live. Lazy loading, calling a lot of listeners on changes and replacing battle tested Guava cache library by own implementation. But for that the best is to try and observe how it behaves.

src/acceptance-test/java/org/zalando/nakadi/cache/ZookeeperNodeInvalidatorTest.java
antban: @rcillo I found a way, thanks for pointing to the way how to fix it: https://github.com/zalando/nakadi/pull/1114/commits/97c6bab225c5bfab04b4dd13a0305ad2e8357673

rcillo:
👍

antban:
:+1:

rcillo:
👍

PR #1115, opened 2019-11-05
Add delivery.yaml

antban:
The file is added only for tests to figure out if it is possible to build image with cdp

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -4,18 +4,21 @@ pipeline:
   overlay: ci/java
   type: script
   commands:
-  - desc: checkstyle
+  - desc: Install dependencies
+    cmd: |
+      pip3 install docker-compose=1.24.1
+  - desc: Checkstyle
     cmd: |
       ./gradlew checkstyle --stacktrace
-  - desc: test
+  - desc: Test
     cmd: |
       ./gradlew testWithReport --stacktrace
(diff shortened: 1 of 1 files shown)


src/test/java/org/zalando/nakadi/security/AuthenticationTest.java
```diff
@@ -76,7 +76,7 @@
 import static org.zalando.nakadi.utils.TestUtils.randomUUID;
 
 @RunWith(SpringJUnit4ClassRunner.class)
-@SpringBootTest(classes = Application.class, webEnvironment = WebEnvironment.RANDOM_PORT)
+@SpringBootTest(classes = Application.class, webEnvironment = WebEnvironment.RANDOM_PORT, value = {"management.port=0"})
```
rcillo: @antban what's this change for?

antban: In case if unit tests are running in parallel, they are trying to use the same management port, defined in configuration. I just specified, that the port should be random (the same as WebEnvironment.RANDOM_PORT, but for management port)

rcillo:
👍

antban:
Rebased original pr

antban:
:+1:

rcillo:
👍

PR #1116, opened 2019-11-11
ARUHA-2608: moved repartitioning to partitions controller

adyach:
# One-line summary
Moved repartitioning to partitions controller


rcillo:
@adyach I rebased this PR on an up-to-date master after Nakadi modularization.

adyach:
@rcillo thank you! there were discussions that way: `/event-types/{name}/partitions` api is not restful, because GET method returns completely different structure. makig something like `/event-types/{name}/partitions/count` or `/event-types/{name}/partitions-number` should help. let me know wdyt?

rcillo:
@adyach it makes sense. If we use `PUT /event-types/{name}/partitions` then in the future we would not be able to implement something that modifies the the oldest and newest available offset for partitions. But thinking forward, what if we want to reduce number of partitions one day? Maybe it would make more sense to `DELETE /event-types/{name}/partitions/{number}`. So this makes a bit confused because `GET and DELETE` would be implemented directly on `/partitions` endpoint, but adding partitions is implemented in some different endpoint. So I think that having every operation on `/partitions` may not be restful in terms of object definition but it is semantically, in the sense that all operations are in fact changing making a CRUD on partitions.

adyach:
@rcillo i was not clear enough, it makes sense to have endpoint to put / patch on per field basis, like PUT `/event-types/{name}/default_statistic`(i used naming for clarity, of course it could be named differently), where you provide default statistic object with number of partitions, it will support delete scenario as well. what is not smooth with `/partitions` endpoint that it is not really partitions, but rather statistics about events in partitions.

rcillo:
@adyach understood 👍 I changed it here f2e1698dc

adyach:
👍

adyach:
@rcillo although pipeline is still failing

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -558,6 +561,18 @@ public EventType get(final String eventTypeName) throws NoSuchEventTypeException
         }
     }
 
+    private EventTypeStatistics validateStatisticsUpdate(
```
thomasabraham, line 564: I don't see why this method should return something. 
It could simply check `if (newStatistics is != null && !Objects.equals(existing, newStatistics))` 
and throw an exception in this case. 
And the calling method above can always use `existing`, when the validation is successful.

rcillo:
@adyach pipeline is broken on `pip install docker-compose==1.24.1`. I'm trying to check why. Maybe underlying ci/overlay changed.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -822,6 +822,46 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+  '/event-types/{name}/partitions-number':
```
thomasabraham, line 825: The name `partitions-number` is a confusing approximation for `number-of-partitions`.
I suggest one of `partition-count`, `partition-configs`, `partition-status` ....

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -822,6 +822,46 @@ paths:
...
+        - name: X-Flow-Id
+          in: header
+          description: |
+            The flow id of the request, which is written into the logs and passed to called services. Helpful
+            for operational troubleshooting and log analysis.
+          type: string
+        - name: partitions_number
+          in: body
+          schema:
+            $ref: '#/definitions/PartitionsNumber'
```
thomasabraham, line 846: `#/definitions/PartitionsNumber` is missing in definitions section.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
rcillo: Makes total sense a9014f384

docs/_data/nakadi-event-bus-api.yaml
rcillo: 91ea632b2 👍

docs/_data/nakadi-event-bus-api.yaml
rcillo: thank you, fixed here 91ea632b2 👍

rcillo:
👍

PR #1117, opened 2019-11-21
Introduce batch_timespan consumption parameter

rcillo:
This is an API extension proposal.

In order to improve the SLO from Nakadi archiver, we are introducing a
new parameter that makes slicing events in 5 minutes window possible.

As it's explained in the description of the parameter, using
`batch_flush_timeout` alone is not enough and it leads to the
accumulation of events produced throughout a longer timespan than the
desired 5 minutes window.

To further illustrate when this happens consider the following
sequence of events:

1. events are produced one per minute starting at minute 0
2. a stream to read these events is not openned until minute 10. This
   stream has batch_flush_timeout of 5 minutes.
3. at minute 15 a batch is flushed containing the events produced from
minute 0 to minute 15 (15 events).

This is not the desired behaviour for Nakadi archiver. Archiver would
like to get events from minute 0 to minute 5, another from minute 5 to minute
10 and finally a third from minute 10 to minute 15. Its goal is to slice
the topic by time.

This new parameter makes it possible.

# edit 1

Just o clarify, this doesn't mean that events would be deserialized on the consumption path. Instead of using `receive_at` we would use https://kafka.apache.org/0100/javadoc/org/apache/kafka/clients/consumer/ConsumerRecord.html#timestamp() which would have no noticeable performance impact.

# edit 2

This change would have another benefit: the batches produced are always the same regardless of the start time of stream. This would lead to fewer duplicates on Nakadi archiver, as failed attempts to process a batch would result in the exact same batch being streamed again, which would result in the same batch being upload to S3, e.g. Archiver becomes idempotent.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/StreamParameters.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/StreamParameters.java
@@ -21,6 +21,10 @@ public class StreamParameters {
      */
     private final Optional<Long> streamLimitEvents;
     /**
+     * Number of seconds to be batched based on Kafka record timestamp.
+     */
+    public final long batchTimespan;
+    /**
      * Timeout for collecting {@code batchLimitEvents} events. If not collected - send either not full batch
      * or keep alive message.
      */
@@ -55,6 +59,7 @@ public class StreamParameters {
             throw new WrongStreamParametersException("batch_limit can't be lower than 1");
         }
         this.streamLimitEvents = userParameters.getStreamLimit().filter(v -> v != 0);
+        this.batchTimespan = TimeUnit.SECONDS.toMillis(userParameters.getBatchTimespan().orElse(0L));
         this.batchTimeoutMillis = TimeUnit.SECONDS.toMillis(userParameters.getBatchFlushTimeout().orElse(30));
         this.streamTimeoutMillis = TimeUnit.SECONDS.toMillis(
                 userParameters.getStreamTimeout()
--- a/src/main/java/org/zalando/nakadi/service/subscription/StreamParameters.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/StreamParameters.java
@@ -21,7 +21,7 @@ public class StreamParameters {
      */
     private final Optional<Long> streamLimitEvents;
     /**
-     * Number of seconds to be batched based on Kafka record timestamp.
+     * Number of milliseconds to be batched based on Kafka record timestamp.
      */
     public final long batchTimespan;
     /**
--- a/src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
@@ -27,6 +27,7 @@ class PartitionData {
     private NakadiCursor commitOffset;
     private NakadiCursor sentOffset;
     private long lastSendMillis;
+    private long batchWindowStartTimestamp;
     private int keepAliveInARow;
     private long bytesInMemory;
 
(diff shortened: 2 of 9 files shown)


rcillo:
I'm presenting a POC implementation for this new parameter to show how simple it would be.

I believe that modifying the behaviour of existing parameters would break the API contract. Someone might rely on the existing parameters to continue exactly how they are, so I'm in favor of introducing a new backward compatible solution for slicing events over a tumbling window.

src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
```diff
@@ -52,25 +53,58 @@
...
     @Nullable
-    List<ConsumedEvent> takeEventsToStream(final long currentTimeMillis, final int batchSize,
-                                           final long batchTimeoutMillis, final boolean streamTimeoutReached) {
+    List<ConsumedEvent> takeEventsToStream(final long currentTimeMillis, final long batchTimespanMillis,
+                                           final int batchSize, final long batchTimeoutMillis,
+                                           final boolean streamTimeoutReached) {
         final boolean countReached = (nakadiEvents.size() >= batchSize) && batchSize > 0;
         final boolean timeReached = (currentTimeMillis - lastSendMillis) >= batchTimeoutMillis;
-        if (countReached || timeReached) {
+        final long lastRecordTimestamp = lastRecordTimestamp();
```
adyach: I think it makes sense to move `lastRecordTimestamp` and `batchWindowStartTimestamp` inside if clause, because it is used only there

src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
```diff
@@ -52,25 +53,58 @@
...
+        if (nakadiEvents.size() > 0) {
+            return nakadiEvents.get(nakadiEvents.size() - 1).getTimestamp();
+        } else {
+            return 0;
+        }
+    }
+
+    private List<ConsumedEvent> extractTimespan(final long batchWindowEndTimestamp) {
+        int count = 1;
+        while (count < nakadiEvents.size() && nakadiEvents.get(count).getTimestamp() < batchWindowEndTimestamp) {
```
adyach: I would really like to optimize, but having another sorted data structure would cost more than just having iterative approach

src/main/java/org/zalando/nakadi/service/subscription/StreamParameters.java
```diff
@@ -20,6 +20,10 @@
      * Maximum number of events that could be sent in session
      */
     private final Optional<Long> streamLimitEvents;
+    /**
+     * Number of seconds to be batched based on Kafka record timestamp.
+     */
+    public final long batchTimespan;
```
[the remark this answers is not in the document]
rcillo: True. I need to fix this.

src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
rcillo: I'm not sure about this. About `lastRecordTimestamp` it is used twice in the condition on line 71 and `batchWindowStartTimestamp` is part of the state of this object.

src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
rcillo: I understand your point. I thought about it and I considered it to not be a problem because this code is only invoked when the flush condition is met, so it's not run on every call to `takeEventsToStream`. There we always take a look at the last event that was loaded from Kafka `O(1)` and then once it matches the criteria, then we run a `O(n)` scan to find the element. I could even optimize it by doing some binary search, as I would expect `getTimestamp` to be more or less sorted (there could be some anomalies, but that would not a problem, as overall it's almost guaranteed to be sorted).

I could implement the binary search. But it would not improve the complexity, as copying the data takes `O(n)` anyway.

rcillo: I can also not rely on existing `extract` method and implement my own, that copies data as it checks it, avoiding going over the data twice.

src/main/java/org/zalando/nakadi/service/subscription/StreamParameters.java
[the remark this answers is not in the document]
rcillo: 7224a65a

src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
rcillo: e8e2069b

src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
rcillo: e8e2069b

adyach:
👍

rcillo:
👍

PR #1119, opened 2019-11-27
Broker list changed watcher is recreated on disconnects

adyach:
# One-line summary
Currently, broker list changed watcher is recreated on disconnects, which means a lot of watchers created for no reason and later on triggered for the same reason creating even more watchers and so on.
Although, it is clearly seen from the logs, this is hypothesis and the testing is required.


ferbncode:
:+1:

rcillo:
Shouldn't we recreate the watch in this situation as well? I see in the logs that this event is triggered because connection to Zookeeper is lost after 25 seconds of idle time (it says that it didn't hear from the server in 25 seconds). Maybe we should keep recreating the watchers but not so frequently. Maybe having some kind of exponential backoff as this is about "reconnecting" to a service.

The biggest problem I see is that I don't believe that Zookeeper failed to send keep-alives. I think that the client is not able to process the keep-alives for some reason (most probably saturated CPU).

If I'm not mistaken, this fix would completely stop recreating watchers which could ultimately lead to the broker list drifting away.

adyach:
@rcillo the watcher is triggered only on the node children change, if the watcher is triggered it has to be recreated. Other possible option to recreate the watcher on:
`None`
`NodeCreated`
`NodeDeleted`
`NodeDataChanged`
does not look like the reason for watcher to be triggered, because the node is not removed by anyone, and it is data is not changed, only list of children changes. `None` looks like for reporting the state of the zk session (according to the logs).

Anyway, the pr suggest to validate hypothesis to be sure about the behavior, I do not claim this will be correct, but for me it looks like that.

PR #1120, opened 2019-11-27
Handle exception in StreamingState if data is gone 

ferbncode:
> Zalando ticket : ARUHA-2618

Handle exception and switch state to `CleanupState` if the new offsets (even after updating expired ones from subscription) are invalid (This is possible only if the we lose some data as on the test environment). 

An alternative way to handle this is to check if the offsets in the subscription lie within the range of the available offsets for the event type partition (i.e after fetching both the newest and the oldest available offset) and then resetting to either the oldest available offset (if the committed offset is not in the range). This can result in some duplicate events.


src/main/java/org/zalando/nakadi/exceptions/runtime/InvalidCursorException.java
```diff
@@ -58,7 +58,8 @@ public String getMessage() {
             case PARTITION_NOT_FOUND:
                 return "non existing partition " + getPartition();
             case UNAVAILABLE:
-                return "offset " + getOffset() + " for partition " + getPartition() + " is unavailable";
+                return "offset " + getOffset() + " for partition " + getPartition() + " is unavailable. " +
+                        "Patch partition offset for subscription with available offset.";
```
adyach: I am not sure about call to action, because it is quite extraordinary situation, although, I understand your intention to reduce the operational load. could we rephrase it to something like: 
`offset 1 for partition 0 is unavailable due to data went over the retention time window"`
feel free to make wording better, but the idea is that offset is not available for some reason, and reason is not clear in the message.

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -573,15 +573,16 @@ private void reconfigureKafkaConsumer(final boolean forceSeek) {
...
                             offsets.get(pk).ensureDataAvailable(beforeFirstAvailable);
                             return offsets.get(pk).getSentOffset();
                         })
                         .collect(Collectors.toList());
-
                 eventConsumer.reassign(cursors);
             } catch (InvalidCursorException ex) {
-                throw new NakadiRuntimeException(ex);
+                getLog().error("Failed to reassign updated cursors for subscription", ex);
+                switchState(new CleanupState(ex));
```
adyach: this processing is already done on the level of `StreamingContext`, so you do not really change something :)

adyach:
@ferbncode it would be nice to see regression test for that specific case, could you add, please?

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -429,6 +441,9 @@ void reactOnTopologyChange() {
                 .toArray(Partition[]::new);
         addTask(() -> refreshTopologyUnlocked(assignedPartitions));
         trackIdleness(topology);
+        if (!streamInitialized) {
```
adyach: could you elaborate on why we want to do it? I see on the previous lines 
442: `addTask(() -> refreshTopologyUnlocked(assignedPartitions));`
so we have already scheduled for execution

ferbncode: Right, fixed this in `a95b29f`. Thanks for pointing out!

The idea was to force refreshing topology, once during the initialization (the partition assignment to this session is already scheduled by re-balance in the end of `StartingState` (as the session is registered)).

src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -89,10 +90,10 @@ public void onEnter() {
                 .collect(Collectors.toMap(et -> et, et -> new StreamKpiData()));
 
         idleStreamWatcher = new IdleStreamWatcher(getParameters().commitTimeoutMillis * 2);
-
         this.eventConsumer = getContext().getTimelineService().createEventConsumer(null);
 
         recreateTopologySubscription();
```
ferbncode: The exception is thrown later (after sending the headers), as the switch to the cleanup state is done by adding the task to the `taskQueue`. Hence, it executes after all other scheduled tasks are completed. For fixing this (handling switch to cleanup state as quickly as the exception occurs, I have two suggestions

1. We switch to cleanup state as soon as the exception occurs and don't schedule it using the task queue.
2. We schedule it using the task queue, but use a priority queue to set it's priority as maximum

ferbncode: I added the solution 1 as part of commit `4217a3d4`.

adyach:
👍

ferbncode:
:+1:

PR #1122, opened 2019-12-09
Switch state immediately to DEAD state if exception/do not schedule it

ferbncode:
Switch state to DEAD state in case of exception without scheduling it in the task queue. One cleaner way to achieve the same is using priority queue for tasks, but I do not see where we need to handle tasks with priority other than during an exception.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/state/CleanupState.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/state/CleanupState.java
@@ -32,11 +32,7 @@ public class CleanupState extends State {
                 getContext().unregisterSession();
 
             } finally {
-                if (null != exception) {
-                    getContext().switchStateImmediately(StreamingContext.DEAD_STATE);
-                } else{
-                    switchState(StreamingContext.DEAD_STATE);
-                }
+                switchState(StreamingContext.DEAD_STATE);
             }
 
             try {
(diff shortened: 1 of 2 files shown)


src/main/java/org/zalando/nakadi/service/subscription/state/CleanupState.java
```diff
@@ -32,7 +32,11 @@ public void onEnter() {
                 getContext().unregisterSession();
 
             } finally {
-                switchState(StreamingContext.DEAD_STATE);
+                if (null != exception) {
+                    getContext().switchStateImmediately(StreamingContext.DEAD_STATE);
```
adyach: it means that subscription will be switch to the dead state and then will be returned back here to finish clean up state. it brings behaviour when subscription is switched not from cleanup to dead state, but from inside cleanup state we go to the dead state + all the things in `taskQueue` for execution will be executed anyway.
does it make sense to clean the `taskQueue`, and add fresh task, if you need immediate switch?

ferbncode: Correct, that is much more cleaner and makes the code readable. Updated in `98eb8f52`

ferbncode: Correct! that is much more cleaner and makes the code better. Updated in `98eb8f52`. Thanks! :)

adyach: 👍 waiting for tests to pass

adyach:
👍

ferbncode:
:+1:

PR #1124, opened 2019-12-11
Improved exception handling on delete event-type

thomasabraham:
# Improved exception handling on delete event-type

> Zalando ticket : ARUHA-2636

## Description
Consider an event-type with subscriptions that cannot be deleted automatically. Delete operation of this event will result in an access denied exception. But the HTTP response of an access denied exception is not very useful. This PR adds the reason for access denied so that the user understands why access is denied. 

## Review
- [x] Tests
- [x] Documentation


src/main/java/org/zalando/nakadi/exceptions/runtime/AccessDeniedException.java
```diff
@@ -26,6 +35,7 @@ public Resource getResource() {
     }
 
     public String explain() {
-        return "Access on " + operation + " " + resource.getType() + ":" + resource.getName() + " denied";
+        return "Access on " + operation + " " + resource.getType() + ":" + resource.getName() + " denied"
```
ferbncode: Maybe we can improve the message further by checking if `null` == operation and show a different message in case it is. Something generic (like `Unable to complete operation on "event-type-name" + reason`. What do you think?

thomasabraham: Good point

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -258,8 +258,13 @@ public void delete(final String eventTypeName) throws EventTypeDeletionException
             authorizationValidator.authorizeEventTypeAdmin(eventType);
 
             if (featureToggleService.isFeatureEnabled(EVENT_TYPE_DELETION_ONLY_ADMINS)) {
-                if (eventType.getAuthorization() == null || hasNonDeletableSubscriptions(eventType.getName())) {
```
antban: I was looking at this section and break my brain.

I think it makes sense to replace whole section
```
            if (featureToggleService.isFeatureEnabled(EVENT_TYPE_DELETION_ONLY_ADMINS)) {
                if (eventType.getAuthorization() == null || hasNonDeletableSubscriptions(eventType.getName())) {
                    throw new AccessDeniedException(eventType.asResource());
                }
            }

            if (featureToggleService.isFeatureEnabled(DELETE_EVENT_TYPE_WITH_SUBSCRIPTIONS)
                    || featureToggleService.isFeatureEnabled(EVENT_TYPE_DELETION_ONLY_ADMINS)) {
                topicsToDelete = deleteEventTypeWithSubscriptions(eventTypeName);
            } else {
                topicsToDelete = deleteEventTypeIfNoSubscriptions(eventTypeName);
            }
```

to something like
```
    if (eventType.getAuthorization() == null) {
        throw new AccessDeniedException(AuthorizationService.Operation.ADMIN, eventType.asResource(), "Can't delete event type that have no authorization specified");
    }

    if (featureToggleService.isFeatureEnabled(EVENT_TYPE_DELETION_ONLY_ADMINS)) {
        if (hasNonDeletableSubscriptions(eventType.getName())) {
            throw new AccessDeniedException(AuthorizationService.Operation.ADMIN, eventType.asResource(), "This event-type have subscriptions that cannot be deleted automatically.");
        }
        topicsToDelete = deleteEventTypeWithSubscriptions(eventTypeName);
    } else if (featureToggleService.isFeatureEnabled(DELETE_EVENT_TYPE_WITH_SUBSCRIPTIONS)) {
        topicsToDelete = deleteEventTypeWithSubscriptions(eventTypeName);
    } else {
        topicsToDelete = deleteEventTypeIfNoSubscriptions(eventTypeName);
    }
```

thomasabraham: I agree, I'll do it.

antban:
:+1:

thomasabraham:
👍

PR #1125, opened 2019-12-11
ARUHA-2642 Temporary allow deletion of events

antban:
There are requests to delete a lot of events in a compacted event types, and right now nakadi basically lacks this functionality. 

I know that the code is not pretty, but in order to solve current needs of users, we can provide non-documented endpoint to allow deletion of events (the endpoint and service is not covered with unit tests), so my suggestion will be to merge this pr to master, tests it on data and then revert it, keeping this pr open in order to be able to rework it and allow this endpoint to everyone. 

The base idea is that publishing and deletion is almost the same - users have to publish the same events to different endpoint, and nakadi will publish null to kafka to remove events.

### DIFF
--- a/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
+++ b/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
@@ -218,7 +218,7 @@ public class KafkaRepositoryAT extends BaseAT {
             items.add(item);
         }
 
-        kafkaTopicRepository.syncPostBatch(topicId, items, null);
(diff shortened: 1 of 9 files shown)


rcillo:
@antban I like the pragmatic approach. Just to confirm if I understood the code correctly, to delete events we would have to post a batch like: `curl -X DELETE $NAKADI_URL/event-types/et-name/deleted-events -d  '[{"metadata":{"partition_compaction_key":"id-of-event-to-delete-1", "occurred_at":"valid-time", "eid":"uuid"}}]'`

antban:
@rcillo Almost like that, there should be also some fields set that are responsible for partition calculation (if they differ)

antban:
:+1:

rcillo:
👍

PR #1126, opened 2019-12-16
ARUHA-2637 Use transactions while creating event type

antban:
ARUHA-2637 Use transactions while creating event type

Also, As a part of the same problem - do not reset cache during transaction, run event type cache update only after successful transaction commit, as updating cache in the middle of transaction only leads to bugs.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -37,6 +37,7 @@ import org.zalando.nakadi.exceptions.runtime.FeatureNotAvailableException;
 import org.zalando.nakadi.exceptions.runtime.InconsistentStateException;
 import org.zalando.nakadi.exceptions.runtime.InternalNakadiException;
 import org.zalando.nakadi.exceptions.runtime.InvalidEventTypeException;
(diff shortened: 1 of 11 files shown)


antban:
:+1:

src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -176,18 +177,33 @@ public void create(final EventTypeBase eventType, final boolean checkAuth)
...
             }
-            throw e;
+            if (null != createdEventType.get()) {
+                try {
+                    eventTypeRepository.removeEventType(createdEventType.get().getName());
+                } catch (RuntimeException ex2) {
+                    LOG.warn("Failed to rollback event type creation (which is highly expected)", ex2);
+                }
+            }
+            throw ex;
```
adyach: maybe you want to wrap it in `InternalNakadiException` and give a good description

adyach:
Not all calls to event type update are handled with cache update:
`TimelineCleanupJob.markTimelineDeleted`
`TimelineService.updateTimeLineForRepartition`
etc

antban:
@adyach omg, thanks for catching it!

antban:
:+1:

adyach:
👍

PR #1127, opened 2019-12-16
Aruha 2642 - Continue event consumption if some of events were deleted 

antban:
Aruha 2642 - Continue event consumption if some of events were deleted 

There is a problem with consumption stack - in case if some events were deleted and log cleaner not yet compacted this events, then nakadi won't be able to consume it. 
This pr targets this issue - skip deleted events from the stream


adyach:
👍

antban:
:+1:

PR #1129, opened 2019-12-16
ARUHA-2640: timeout for lock

adyach:
# One-line summary
ARUHA-2640: timeout for lock

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
+++ b/src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
@@ -68,7 +68,7 @@ public class TimelineSyncImpl implements TimelineSync {
 
     @Autowired
     public TimelineSyncImpl(final ZooKeeperHolder zooKeeperHolder, final UUIDGenerator uuidGenerator)
-            throws InterruptedException {
+            throws InterruptedException, RuntimeException {
         this.nodeId = uuidGenerator.randomUUID().toString();
         this.zooKeeperHolder = zooKeeperHolder;
         this.initializeZkStructure();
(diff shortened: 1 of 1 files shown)


src/main/java/org/zalando/nakadi/service/timeline/TimelineSyncImpl.java
```diff
@@ -84,22 +85,28 @@ private String toZkPath(final String path) {
...
-                            .forPath(toZkPath("/nodes"), "[]".getBytes(Charsets.UTF_8));
-                } catch (final KeeperException.NodeExistsException ignore) {
-                }
+                // 2. Create locked event types structure
+                zooKeeperHolder.get().create().creatingParentsIfNeeded().withMode(CreateMode.PERSISTENT)
+                        .forPath(toZkPath("/locked_et"), "[]".getBytes(Charsets.UTF_8));
+                // 3. Create nodes root path
+                zooKeeperHolder.get().create().creatingParentsIfNeeded().withMode(CreateMode.PERSISTENT)
+                        .forPath(toZkPath("/nodes"), "[]".getBytes(Charsets.UTF_8));
+            } catch (final KeeperException.NodeExistsException ignore) {
```
ferbncode: If the creation of a node `/version` fails due to `NodeExistsException`, then the potential creation of other two persistent nodes (`/locked_et` and `/nodes`) will be skipped (and the same will happen for other newly launching instances)

I agree that this would happen only in case of a bad setup of a new zookeeper, and we can probably fix this if this scenario happens by deleting the (already) created nodes.

Although it's not related to changes in the PR, do you think it's a good idea to handle creation of node with possible exception `NodeExistsException` atomically for each path since `checkZkStructureInitied` checks for existence of all nodes?

adyach: a08135ef

ferbncode:
:+1:

adyach:
👍

PR #1130, opened 2019-12-18
Aruha 2637

antban:
Use transactions while creating event type, notify about updates only after transaction commit, do not rely on psql getTypeInfo call for null types, as connection may be in simple mode (no prepared statements)
ARUHA-2637


antban:
:+1:

antban:
:+1:

adyach:
👍

PR #1131, opened 2019-12-19
Batch timeout was used in the place of batch timespan

rcillo:
Accidentally batch timeout was offered in the place of batch timespan
which caused batch sizes to be reduced in size in some situations
adding latency to some consumers.


thomasabraham:
👍

rcillo:
👍

PR #1132, opened 2019-12-19
Avoid sending empty batches in case there are no events in the upcoming window

rcillo:
In case there are no events in the time window, but there a single
event outside the time window, this event would trigger extraction but
then no events would be sent, as the window is empty.

As no events would be sent, the value of the window would not be
updated.

This could cause low frequency streams to get stuck.


rcillo:
👍

rcillo:
👍

rcillo:
👍

antban:
:+1:

PR #1133, opened 2019-12-23
Add index on owning application for subscription table

ferbncode:
> Zalando ticket : ARUHA-2643

This is just a temporary solution to reduce the cost of queries that internal applications make for fetching subscriptions by owning application (while still using offset based pagination)


rcillo:
👍

ferbncode:
:+1:

PR #1134, opened 2019-12-23
instance termination plugin support

adyach:
# One-line summary
it should be possible to notify nakadi about coming instance termination, that it can release used resources and prepare itself for termination. A good example would be subscription connection termination in clean way, that consumer will not wait commit timeout

PR is not compilable due to plugin api change, which is expected here https://github.com/zalando-nakadi/nakadi-plugin-api/pull/10

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
@@ -11,6 +11,7 @@ import org.zalando.nakadi.domain.NakadiCursor;
 import org.zalando.nakadi.domain.Subscription;
 import org.zalando.nakadi.exceptions.runtime.AccessDeniedException;
 import org.zalando.nakadi.exceptions.runtime.NakadiRuntimeException;
+import org.zalando.nakadi.plugin.api.exceptions.PluginException;
 import org.zalando.nakadi.service.AuthorizationValidator;
 import org.zalando.nakadi.service.BlacklistService;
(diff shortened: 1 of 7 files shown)


src/main/java/org/zalando/nakadi/controller/HealthCheckController.java
```diff
@@ -12,8 +14,18 @@
...
+
+    @Autowired
+    public HealthCheckController(final TerminationService terminationService) {
+        this.terminationService = terminationService;
+    }
+
     @RequestMapping(method = GET)
     public ResponseEntity<String> healthCheck() {
+        if (terminationService.isTerminating()) {
+            return ResponseEntity.status(HttpStatus.IM_A_TEAPOT_418).build();
```
rcillo: Is this meant to be used by Nakadi Proxy? When maintaining the list of healthy Nakadi instances, those who report 418 during health check will not be added to the final result but at the same time not marked as unhealthy?

src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
```diff
@@ -493,11 +505,15 @@ public Builder setKpiCollectionFrequencyMs(final long kpiCollectionFrequencyMs)
             return this;
         }
 
+        public Builder setTerminationService(final TerminationService terminationService) {
```
rcillo: I see that this is set also in the constructor. Is this used or needed?

PR #1137, opened 2020-01-07
Extend API for discriminator in event-type definition

ferbncode:
### Description
Extend API for field discriminator in the event type
definition.
> Zalando ticket : ARUHA-2663 (only if appropriate)

### Review
- [x] Tests
- [x] Documentation


ferbncode:
@thomasabraham Thanks for pointing out to improve the definition and documentation around `discriminator` (Fixed in `6dcb96f`)

thomasabraham:
👍

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2858,6 +2861,26 @@ definitions:
...
+      Once created, the discrimator definition for an event-type can't be updated or removed.
+    properties:
+      path:
+        type: string
+        description: |
+          Path of the field in the event that contains the identifiers of the entities that can access the event.
+      type:
+        type: string
+        description: |
+          Type of the field specified by the path the contains the identifiers of entities.
```
ePaul: Why is the type needed here? Can't this be derived from the schema itself?

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2858,6 +2861,26 @@ definitions:
...
+      Event field parameters to differentiate consumption of events published on per-consumer level.
+      The discriminator can be used to specify a field to be used from the events published to this event type.
+      This field can then be used to implement logic using a plugin to whether an event would be provided
+      in a consumer's stream (`EventsStreamBatch` or `SubscriptionEventStreamBatch`).
+      Once created, the discrimator definition for an event-type can't be updated or removed.
+    properties:
+      path:
+        type: string
+        description: |
+          Path of the field in the event that contains the identifiers of the entities that can access the event.
```
ePaul: I don't think that this is the right approach.
Now the event producer needs to know the names of the services (or users) which are allowed to read each individual event? This makes the producer's job unnecessarily hard (it needs to be updated whenever a new consumer needs data) – and old data can't ever be read by new consumers –, and puts data into the event which does not belong there.

Maybe a better approach is that the event data contains some classifier (which is indicated by this property), which then can be mapped in the event type's authorization setting to a set of services/users.

For example, an order event could have a business partner ID, and each service then belongs to a business partner (or several of them) and can only read events for its own business partner ID.

This whole field could possibly then move to the authorization object (to make it obvious that the discrimiator's purpose is authorization). (Though you'll need to validate that the path points to an existing (potentially required) field in the schema, I guess?)

The authorization object could then look like this.

```yaml
  authorization:
     discriminator:
       path: 'data/partner_code'
     writers: [...] # as now
     admins: [...] # as now
     readers:
        # those can read everything
        - data_type: service
          value: order_archiver
        # those can read just the indicated class
        - data_type: service
          value: partner1-order-mapper
          discriminator_value: partner1
        - data_type: team
          value: team-partner1-support
          discriminator_value: partner1
```
(I guess the exact syntax in the `readers` section would depend on the authorization plugin – I can even imagine having some logical expression here, or require some OAuth scope.)

For alternative names: `classifier` would be another option.

ePaul: After reading the (Zalando-internal) design document, it looks like you already want to do something like this (possibly with the authorization also based on the OAuth token – that is part of the authorization plugin).

Then it just needs better documentation. 

```suggestion
          Path of the field in the event that contains a classifier, which will be used by an authorization plugin (together with definitions in the authorization section) to define who can read the event.
```

ferbncode: Agreed on all points. I've tried and updated the documentation to correctly reflect the purpose of the field as you suggested. We also decided to name the field as `event_auth_field` (aligning with other field specifiers like `ordering_key_fields`, etc). However, we don't make it part of authorization section as mainly, we want to keep the field immutable once defined.

docs/_data/nakadi-event-bus-api.yaml
ferbncode: Right now, for existing event types, we have some users wanting to implement per-event authorization on consumer end without changing the event type schema or introducing a field (required or not) for all publishers. So, as a result, we accept the type of the field too.

ePaul: Ah, then this should mention that the field does not need to be to exist in the event schema itself (though then it's not quite clear to me on what happens if the field does not exist, or has a different type).

ferbncode:
Fixing tests

docs/_data/nakadi-event-bus-api.yaml
ferbncode: Apologies, I misunderstood the `type` of the field. After clearing this up, I've updated this in `f22915a7f`. 
```
path: Path to a string field in the event that contains the a classifier, which can be used by authorization plugin (together with definitions in the authorization section) to define who can read the event. The field may not be present in the schema definition for the event type. In case the field is present, it will be available for reading for all authorized consumers.
type: Indicator of the type of the data (classifier) the field contains (eg: `retailer_id` or `application_name`).
```

I've updated the documentation to update what happens if the field is null/absent in `9158466b7`. Thanks for this pointing out. :)

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -541,6 +543,21 @@ private void validateAudience(final EventType original, final EventTypeBase even
         }
     }
 
+    private void validateEventAuthFieldUnchanged(final EventType original, final EventTypeBase eventTypeBase) throws
+            InvalidEventTypeException {
+        final EventAuthField originalEventAuthField = original.getEventAuthField();
+        final EventAuthField updatedEventAuthField = eventTypeBase.getEventAuthField();
+        if (updatedEventAuthField != null && originalEventAuthField != null) {
```
rcillo: What if `originalEventAuthField` is `null` and `updatedEventAuthField` is not?

rcillo: Or does it mean that admins can change this but normal users can only add it?

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2858,6 +2861,29 @@ definitions:
...
+      Parameters to define per-event authorization. This field can be used with an authorization plugin
+      to differentiate which authorized consumers for an event type can receive an event published to
+      the event type. In case the field is null or not present in the event, it will be allowed to be read
+      by all authorized consumers.
+      Once defined, the definition for EventAuthField for an event-type cannot be updated or removed.
+    properties:
+      path:
+        type: string
+        description: |
+          Path to a string field in the event that contains the a classifier, which can be used by an
```
rcillo: `the a`

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2858,6 +2861,29 @@ definitions:
...
+          authorization plugin (together with definitions in the authorization section) to define who can read the
+          event. The field may not be present in the schema definition for the event type. In case the field is not
```
rcillo: This is not very clear for me. It says that `The field may not be present in the schema definition for the event type` but then I think it doesn't make it clear that in case an `event` does not contain the specified `path` or its value is `null` then this event is not filtered and all authorized consumer can read it.

rcillo: And by `authorized consumers` it might be worth explaining that we are referring to those in the `authorization` section, which is a per-event-type authorization, in contrast to this one which is `per-event` authorization.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2858,6 +2861,29 @@ definitions:
...
+          present, it will be available for reading for all authorized consumers.
+      type:
+        type: string
+        description: |
+          Indicator of the type of the data (classifier) the field contains (eg: `retailer_id` or `application_name`).
```
rcillo: You can also use `example` field from `open-api`.

rcillo: It might also be worth explaining that its plugins responsibility to handle these values and this is installation specific.

docs/_documentation/using_per_event_authorization.md
```diff
@@ -0,0 +1,48 @@
+---
+title: Event-Based Authorization
+position: 13
+---
+
+## Event-Based Authorization using EventAuthField
+
+Nakadi provides an option to specify `event_auth_field` in an event type definition, which can be used to differentiate
```
rcillo: Let's be more straighforward. Something like `Nakadi provides per-event filtering, allowing event type owners to specify which users should be able to read an event`.

Then you can say that this is achieved through the field `event_auth_field`.

docs/_data/nakadi-event-bus-api.yaml
rcillo: Also it's important to define that a path is in dot notation. You can give an example as well to make it clearer.

docs/_documentation/using_per_event_authorization.md
```diff
@@ -0,0 +1,48 @@
...
+ consumption of events in an event type on a per-consumer level. The `event_auth_field` takes the path to a string field
+ and type of a field inside an event that can be then used by an authorization plugin to classify if a consumer
+ should receive an event published to the event type. It is optional for the field to be part of the event type's
```
rcillo: The fact that it's optional can be omitted in my opinion. This can be inferred from schema definition.

docs/_documentation/using_per_event_authorization.md
```diff
@@ -0,0 +1,48 @@
...
+ schema.
+
+ An `event_auth_field` field, hence, can be used to specify a classifier (which along with the authorization section)
```
rcillo: Remove `, hence` it's unnecessary.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
antban: > What if `originalEventAuthField` is `null` and `updatedEventAuthField` is not?

@rcillo I think that this case is perfectly valid - we need to set this field




> Or does it mean that admins can change this but normal users can only add it?

As I see right now - admins and normal users are in the same position. 
From the other hand I believe that nakadi admin should have extended capabilities, we can plan it for later

core-common/src/main/java/org/zalando/nakadi/domain/EventAuthField.java
```diff
@@ -0,0 +1,60 @@
...
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        if (!super.equals(o)) {
+            return false;
+        }
+        final EventAuthField that = (EventAuthField) o;
+        return (this.type.equals(that.type) && this.path.equals(that.path));
```
antban: I see that there is annotation `@NotNull`, but still, it is better to write something like 
```
Objects.equals(this.type, that.type) && Objects.equals(this.path, that.path)
```
The same goes to hash code - `Objects.hash(this.path)`

docs/_data/nakadi-event-bus-api.yaml
antban: I don't really get meaninng of this field - it seems that it's just for informational purposes...

core-common/src/main/java/org/zalando/nakadi/domain/EventTypeBase.java
```diff
@@ -74,6 +74,10 @@
     @JsonInclude(JsonInclude.Include.NON_NULL)
     private Audience audience;
 
+    @Nullable
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    private EventAuthField eventAuthField;
```
antban: You forgot `@Valid` annotation. Without it validation of `EventAuthField` won't be triggered

core-common/src/main/java/org/zalando/nakadi/domain/EventAuthField.java
```diff
@@ -0,0 +1,60 @@
+package org.zalando.nakadi.domain;
+
+import javax.validation.constraints.NotNull;
+
+public class EventAuthField {
+    @NotNull
+    private String type;
+
+    @NotNull
+    private String path;
```
antban: Maybe it makes sense to set validation min. length to 1, as empty field doesn't make sense (but I am not sure)

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
ferbncode: In that case, users will be able to add `eventAuthField` to an event type which already doesn't have the field.
Only `nakadi` admins are allowed to change the field in case of an incident/requirement change.

rcillo:
👍

docs/_documentation/using_per_event_authorization.md
```diff
@@ -0,0 +1,50 @@
...
+  "event_auth_field": {
+    "type": "teams",
+    "path": "security.exclusive_readers"
+  }
+  "category": "business",
+  ...
+}
+```
+
+An event may be published to the above event type, and the logic for authorization then, could only allows a single
```
adyach: s/allows/allow

rcillo:
👍

ferbncode:
:+1:

PR #1138, opened 2020-01-08
Session not cleaned if error before making zookeeper session change subscription

ferbncode:
In case of exception before creation of subscription for zookeeper changes to sessions list, the corresponding session registered in Zookeeper is not cleaned up, leading to a zombie state, where a stream is present, but not in `StreamingState`.

The bug that the consumption instance wouldn't clean up the session was because the session timed out while releasing the lock during registration of session (i.e before [switching](https://gist.github.com/ferbncode/5f97ee428d9124c869c51ccf0139ef42#file-exception-L66) to streaming state)

See exception here: https://gist.github.com/ferbncode/5f97ee428d9124c869c51ccf0139ef42

This can then only be fixed by terminating the instance where the exception occurred.


ferbncode:
:+1:

rcillo:
👍

adyach:
👍

ferbncode:
:+1:

PR #1140, opened 2020-01-13
Unregister session only if registered

ferbncode:
If there is a `409` response on opening a stream (that is no slots remaining), then un-registering the session (inside `CleanupState.onEnter()`) fails with an exception `NoNodeError`, which again loops the task to `CleanupState` calling `CleanupState.onEnter` indefinitely. This increases the load on zookeeper as the client indefinitely tries to delete a node not present. 

Another alternative implementation to solve the issue would be to handle task of Cleanup separately outside of task queue preventing the repeated `CleanupState.onEnter()` call.


ferbncode:
This repeated call to `CleanupState.onEnter()` in case of exception is definitely good for retry in case of a temporary exception, so I believe, keeping that behavior should be favorable. (and the current fix in this PR using a flag currently should be okay)

adyach:
👍

ferbncode:
:+1:

PR #1141, opened 2020-01-14
Check if node exists before attempt to delete

ferbncode:
Initially, I did set the flag (`sessionRegistered` back to false only after successfully un-registering a node. If an exception occurred after deletion of the node, it is happening that the variable is not unset and we loop through `CleanupState.onEntry()` logging too much

Added check if the node exists (it should not increase much over-head on zookeeper, since this is not a very frequent operation) before deletion as I don't want to set `sessionRegistered=false` if un-registering fails, which would cause the zombie session issue


adyach:
:+1:

ferbncode:
:+1:

PR #1142, opened 2020-01-14
Do unregister session as before

ferbncode:
Do unregister based on a flag as before. Right now, disk is still filling up if it fails with the `isActiveSession` call.

### DIFF
--- a/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
+++ b/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
@@ -76,6 +76,7 @@ public class StreamingContext implements SubscriptionStreamer {
     private State currentState = new DummyState();
     private ZkSubscription<List<String>> sessionListSubscription;
     private Closeable authorizationCheckSubscription;
+    private boolean sessionRegistered;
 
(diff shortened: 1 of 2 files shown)


ferbncode:
:+1:

ferbncode:
Before any changes to the subscription API (https://github.com/zalando/nakadi/pull/1123), unregistering the session was based on the `sessionListSubscription`.

It's creation is not moved to the `StreamingState.java`, hence, if an exception occurred between `StartingState` and `CleanupState`, the session is not cleaned up. 

```    public void unregisterSession() {
        log.info("Unregistering session {}", session);
        if (null != sessionListSubscription) {
            try {
                sessionListSubscription.close();
            } finally {
                this.sessionListSubscription = null;
                zkClient.unregisterSession(session);
            }
        }
    }
```
I changed this to check whether the node actually exists in zookeeper, however, in that case, I am not sure where the problem is, but if Nakadi fails to connect to zookeeper, it still fails to reconnect and check if the node exists and finally bring to `DummyState()`

In these changes, I resort back to checking via a boolean variable to check if a session should be tried to unregistered.

adyach:
👍

PR #1143, opened 2020-01-16
kafka header for event auth

adyach:
# One-line summary
Custom Kafka header extension to store/retrieve event authorization value. Apparently, I have not found a way to make it more explicit, because there is no mechanism in Kafka client to support header serialization.


core-common/src/main/java/org/zalando/nakadi/repository/kafka/EventAuthHeader.java
```diff
@@ -0,0 +1,27 @@
...
+import com.google.common.base.Charsets;
+import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.header.internals.RecordHeader;
+
+public class EventAuthHeader extends RecordHeader {
+
+    static final String HEADER_KEY = "X-Event-Auth";
+
+    public EventAuthHeader(final String value) {
+        super(HEADER_KEY, value.getBytes(Charsets.UTF_8));
```
ferbncode: If the `value` is `null`, there will be `NPE` here, but I guess, the value should be check before instantiating the header (before publishing), right?

adyach: I re read the api docs, it looks like it could be null
anyway, thanks 
ebf5b05

adyach:
👍

ferbncode:
:+1:

PR #1147, opened 2020-01-22
Publish header/authorization per event on publishing

ferbncode:
Publish header/authorization per event on publishing. (Handling `event_owner_selector` field in the event type.

- [x] Push tests/Fix tests


core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -211,14 +228,53 @@ private void setEventKey(final List<BatchItem> batch, final EventType eventType)
...
+                    final JsonPathAccess jsonPath = new JsonPathAccess(item.getEvent().getEventJson());
                     final String eventKey = jsonPath.get(partitionKeyField).toString();
-                    item.setEventKey(eventKey);
+                    item.setKey(eventKey);
                 }
             }
         }
     }
 
+    private void setHeaders(final List<BatchItem> batch, final EventType eventType) {
```
antban: Clearly this is wrong place for the method

ferbncode: On second thought, I see that, these should be set closer to publishing when `SUBMITTED`. I'll try and fix this. Thanks for pointing out! :)

core-common/src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -63,6 +68,7 @@ public static EmptyInjectionConfiguration build(final int position, final boolea
     private String brokerId;
     private String eventKey;
     private int eventSize;
+    private EventOwnerHeader owner;
```
adyach: `owner` is not the owner, it is EventOwnerHeader

core-common/src/main/java/org/zalando/nakadi/domain/EventPublishingStep.java
```diff
@@ -5,5 +5,6 @@
     VALIDATING,
     ENRICHING,
     PARTITIONING,
+    AUTHORIZING,
```
adyach: unfortunately, `EvnetPublishingStep` is Enum, changing will introduce breaking change
https://nakadi.io/manual.html#definition_BatchItemResponse*step

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -124,6 +133,8 @@ EventPublishResult processInternal(final String events,
             if (!delete) {
                 enrich(batch, eventType);
             }
+            eventOwnerExtractor.extract(batch, eventType);
```
adyach: I see that you do iteration over batch on this and next line. it looks like it could be combined without loosing readability, but reducing cpu cycles on publishing path.

core-common/src/main/java/org/zalando/nakadi/util/JsonPathAccess.java
```diff
@@ -27,12 +27,12 @@ public Object get(final String path) throws InvalidPartitionKeyFieldsException {
 
         while ((field = pathTokenizer.nextToken()) != null) {
             if (!(curr instanceof JSONObject)) {
```
adyach: it is not clear why you do check here, although you have JSONObject for sure from constructor and when getting the field.

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventOwnerExtractor.java
```diff
@@ -0,0 +1,72 @@
...
+    private final FeatureToggleService featureToggleService;
+
+    @Autowired
+    public EventOwnerExtractor(final FeatureToggleService featureToggleService) {
+        this.featureToggleService = featureToggleService;
+    }
+
+    public void extract(final Collection<BatchItem> items, final EventType eventType) {
+        final EventOwnerSelector selector = eventType.getEventOwnerSelector();
+        if (null == selector) {
```
adyach: shouldn't it be set to some default value instead? it looks like events without any header are accessible for everyone.

antban: @adyach It should, but it's part of other task. Here we are introducing header extraction

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
antban: @adyach 
The full code is 
```
validate(batch, eventType, parentSpan, delete);
partition(batch, eventType);
setEventKey(batch, eventType);
enrich(batch, eventType);
eventOwnerExtractor.extract(batch, eventType);
authorizeEventWrite(batch);
```
I think there it was done because of readability.

core-common/src/main/java/org/zalando/nakadi/domain/EventPublishingStep.java
antban: :((

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
adyach: I am talking only about your changes, and it is clearly can be improved

core-common/src/main/java/org/zalando/nakadi/domain/BatchItem.java
antban: I don't get the point. Class name is `EventOwnerHeader`, logical meaning is owner. 
I mean i'm not limiting myself to call all variables of class `String` to `string` only

adyach: then please rename the class name to represent logical meaning :)

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
antban: I thought that I answered why it is done like that. There are 2 things happening here - extraction of header and check of authorization. Logically they are different. 
In case if we are joining extraction with checks, then the method will be called
`extractAndCheckAuthorization`, that didn't seem right logically, but can reduce CPU usage because of failfast approach and iteration removal. 

To be honest, I do not have any preference between this 2 approaches (clear logic vs failfast in this case). 
Anyways, I am not the owner of this PR, and my commit came as a result of pairprogramming

core-common/src/main/java/org/zalando/nakadi/util/JsonPathAccess.java
ferbncode: While getting the field, we cast the `curr` object:
```
curr = ((JSONObject) curr).get(field);
```
However, the received `field`, (new current) may bot be instance of `JSONObject`.

core-common/src/main/java/org/zalando/nakadi/domain/EventPublishingStep.java
ferbncode: Thanks for pointing out. :+1: , We have a few options I believe:

1. a work-around would be to use an existing step (`VALIDATING` or `NONE`). Given the `event_owner_header` definition is part of event type definition (like `schema`), and it is checked against the actual event (like event against `schema`), I believe `VALIDATING` step could be a work around. (with update in documentation). What do you think/suggest? 

2. introduce breaking change: This might not be an option, but if existing clients don't use the step field, we could probably introduce this. I'll check this for major clients.

core-common/src/main/java/org/zalando/nakadi/util/JsonPathAccess.java
adyach: you are right, I missed that. just discard my comment.

core-common/src/main/java/org/zalando/nakadi/domain/EventPublishingStep.java
adyach: first option using `VALIDATING` step sounds good to me
the field `step` looks overwhelming, whatever it is, user has to go and read the detailed message, which Nakai provides. Processing based on that field can not be applied in automated way, it is done based on response codes from Nakadi.

ferbncode: Fixed this in https://github.com/zalando/nakadi/pull/1147/commits/dfb2c00f98ce6224e624ee6c28445612a6c82ccb

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
ferbncode: Tried to fix this in https://github.com/zalando/nakadi/pull/1147/commits/dfb2c00f98ce6224e624ee6c28445612a6c82ccb

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -133,8 +135,7 @@ EventPublishResult processInternal(final String events,
             if (!delete) {
                 enrich(batch, eventType);
             }
-            eventOwnerExtractor.extract(batch, eventType);
-            authorizeEventWrite(batch);
+            validateEventOwnership(eventType, batch);
```
adyach: the validation happens before partitioning and enrichment, it would make more sense to move `validateEventOwnership` close to `validate`

adyach:
👍

ferbncode:
:+1:

PR #1149, opened 2020-01-29
Create system event-types only if they are missing

thomasabraham:
Signed-off-by: Thomas Abraham <thomas.abraham@zalando.de>

# Create system event-types only if they are missing

> Zalando ticket : ARUHA-2704

## Description
When Nakadi starts up, it tries to create a few system event types.
When the event type creation fails, because they already exist, it logs the an error with complete stack trace. This PR introduces a `createIfMissing` function, which will check if the event exists first.

### DIFF
--- a/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -224,6 +224,24 @@ public class EventTypeService {
                 eventType.getName());
     }
 
+    public void createIfMissing(final EventTypeBase eventType, final boolean checkAuth)
+            throws AuthorizationSectionException,
+            TopicCreationException,
+            InternalNakadiException,
+            NoSuchPartitionStrategyException,
+            DuplicatedEventTypeNameException,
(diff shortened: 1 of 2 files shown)


ferbncode:
:+1:

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -224,6 +224,24 @@ public void create(final EventTypeBase eventType, final boolean checkAuth)
...
+            TopicCreationException,
+            InternalNakadiException,
+            NoSuchPartitionStrategyException,
+            DuplicatedEventTypeNameException,
+            InvalidEventTypeException,
+            DbWriteOperationsBlockedException,
+            EventTypeOptionsValidationException {
+        try {
+            eventTypeRepository.findByName(eventType.getName());
+            LOG.info("Event-type {} already exist", eventType.getName());
```
ferbncode: Minor nitpick (`s/exist/exists`)

thomasabraham:
👍

ferbncode:
:+1:

PR #1150, opened 2020-01-29
Don't check event authorization if the owner is not set

ferbncode:
# One-line summary
Don't check `authorization` per event if owner is not set.


adyach:
👍

ferbncode:
:+1:

PR #1151, opened 2020-01-29
Fix handling of `null` path as StrictJsonParser uses JSONObject.NULL

ferbncode:
StrictJsonParser.java returns `JSONObject.NULL` for fields where value is set as `null`. Thus, fixed the code to handle `nulls` as expected.

### DIFF
--- a/core-services/src/main/java/org/zalando/nakadi/service/publishing/EventOwnerExtractorFactory.java
+++ b/core-services/src/main/java/org/zalando/nakadi/service/publishing/EventOwnerExtractorFactory.java
@@ -44,8 +44,8 @@ public class EventOwnerExtractorFactory {
         return (batchItem) -> {
             try {
                 final JsonPathAccess jsonPath = new JsonPathAccess(batchItem);
-                final String value = jsonPath.get(selector.getValue()).toString();
-                return null == value ? null : new EventOwnerHeader(selector.getName(), value);
+                final Object value = jsonPath.get(selector.getValue());
+                return JSONObject.NULL == value ? null : new EventOwnerHeader(selector.getName(), value.toString());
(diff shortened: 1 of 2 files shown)


core-services/src/main/java/org/zalando/nakadi/service/publishing/EventOwnerExtractorFactory.java
```diff
@@ -44,8 +44,8 @@ public EventOwnerExtractorFactory(final FeatureToggleService featureToggleServic
         return (batchItem) -> {
             try {
                 final JsonPathAccess jsonPath = new JsonPathAccess(batchItem);
-                final String value = jsonPath.get(selector.getValue()).toString();
-                return null == value ? null : new EventOwnerHeader(selector.getName(), value);
+                final Object value = jsonPath.get(selector.getValue());
```
adyach: is it possible jsonPath.get returns null ?

adyach: it looks like it is gonna be exception if there is no such path

adyach:
👍

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventOwnerExtractorFactory.java
ferbncode: Right, jsonPath.get would throw `null` ideally. However, because we use the self-implemented `StrictJsonParser`, it returns `JsonObject.NULL` for `null` fields.

In case of no path, it would throw exception, and this extractor would return `null`

ferbncode:
:+1:

PR #1152, opened 2020-01-31
Support autocommit and filtering during consumption

antban:
Because of event ownership


api-consumption/test.sh
```diff
@@ -0,0 +1,39 @@
...
+10.154.144.19
+10.154.144.20
+10.154.144.21
+10.154.128.11
+10.154.128.12
+EOM
+
+for ip in $KAFKAIPS; do
+  piu request-access --odd-host  odd-eu-central-1-migration.aruha.zalan.do ${ip} "Check kafka open files" >/dev/null 2>/dev/null
+  ssh -tA dsorokin@odd-eu-central-1-migration.aruha.zalan.do ssh -o StrictHostKeyChecking=no dsorokin@10.154.144.11 2>/dev/null
```
adyach: there is a bug, it should be variable `${ip}` :D

api-consumption/src/main/java/org/zalando/nakadi/service/EventStream.java
```diff
@@ -81,16 +82,24 @@ public void streamEvents(final AtomicBoolean connectionReady, final Runnable che
...
 
                 checkAuthorization.run();
 
                 if (consumedEvents.isEmpty()) {
-                    // TODO: There are a lot of optimizations here, one can significantly improve code by processing
-                    // all events at the same time, instead of processing one by one.
-                    consumedEvents.addAll(eventConsumer.readEvents());
+                    final List<ConsumedEvent> eventsFromKafka = eventConsumer.readEvents();
+                    for (final ConsumedEvent evt: eventsFromKafka) {
+                        if (eventStreamChecks.isConsumptionBlocked(evt)) {
```
adyach: I expected to see this code in common place like NakadiKafkaConsumer, where the filtering could be done in one place instead of putting it in 2 places (lola, hila).

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/autocommit/PartitionState.java
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * There are several assumptions on behavior of this class.
+ * 1. Offsets to skip are only added to the end of skipped events queue.
+ * 2. Commits are added at any place, and on every commit we are removing queue of intervals to skip that are
+ * completely before committed offset.
+ * Thus, if we are about to do autocommit, we should check first interval only.
+ * 3. Committed offsets are only increasing (this is guaranteed by other part of the code)
+ */
+public class PartitionState {
+    private final CursorOperationsService cursorOperationsService;
```
adyach: PartitionState looks like a model at least from the name perspective, I would expect to have additional class which is able to work with it instead of having cursorOperationsService inside the data class.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -217,17 +227,21 @@ private void pollDataFromKafka() {
         events.forEach(this::rememberEvent);
         if (!events.isEmpty()) {
             addTask(this::streamToOutput);
+            addTask(this::pollDataFromKafka);
+        } else {
+            scheduleTask(this::pollDataFromKafka, POLL_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
```
adyach: I am not sure I understand the change

api-cursors/src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -254,33 +245,4 @@ private void validateCursorsBelongToSubscription(final Subscription subscription
             throw new UnableProcessException("Event type does not belong to subscription: " + wrongEventTypes);
         }
     }
-
-    private class SubscriptionCursorComparator implements Comparator<SubscriptionCursorWithoutToken> {
```
adyach: is it just clean up due to the fact that zero version cursor is not used anymore ?

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -109,6 +112,7 @@ public void onEnter() {
         addTask(this::initializeStream);
         addTask(this::pollDataFromKafka);
         scheduleTask(this::checkBatchTimeouts, getParameters().batchTimeoutMillis, TimeUnit.MILLISECONDS);
+        scheduleTask(this::autocommitPeriodically, AUTOCOMMIT_INTERVAL_SECONDS, TimeUnit.SECONDS);
```
adyach: why are we sure that 5 seconds is enough in terms of speed of consumption and resource usage ?

antban: That was my best guess, I have no other excuses

api-cursors/src/main/java/org/zalando/nakadi/service/CursorsService.java
antban: No, because this class is wrong and not needed

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
antban: I will roll it back, it was attempt to optimize something different, not related to the ticket.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/autocommit/PartitionState.java
antban: I already shared my opinion that essential part of OOP approach requires joining data (or object state) with methods of it's processing. 
In order to make it separate there should be some reasons. Here I don't really see this reasons.

But I agree that class name is somehow misleading. How do you think `PartitionSkippedCursorsOperator` name reflects properly the actions that are being executed by this class?

api-consumption/src/main/java/org/zalando/nakadi/service/EventStream.java
antban: Thanks for pointing out! Will work on it.

antban: After spending some time refactoring to this approach, it seems that it is not working, and is completely wrong. The reason is that we may be using several data providers (not only kafka), they all should be working, and we can not use it. 
Event thought we have only KAFKA inplace, organization of code enforces permissions check on a different level, as for example filtering on NakadiKafkaConsumer level will lead to timelines being broken for some cases.
Therefore I propose not to do this move.

antban: @adyach

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/autocommit/PartitionState.java
antban: @adyach  ?

adyach:
👍

adyach:
👍

antban:
:+1:

PR #1156, opened 2020-02-07
Use synced event type definition for checking updates

ferbncode:
Ticket Link: https://jira.zalando.net/browse/ARUHA-2665

The solution is to use definition of the event type stored in database for checking if an update should be made. This would work temporarily as the traffic for updating event types on production is not very high and this change should not overload the database. See current traffic here: https://gist.github.bus.zalan.do/sgarg/201ecb21498f06517d5f3775556c438a

In the long term (for many updates on event-types), however, this should be handled by shared metadata (like et version) about event-types stored in some shared cache. (or implementing other possible ways of fast synchronization between distributed cache of Nakadi instances. I will make sure to add that task to the backlog if this implementation is temporarily accepted.


antban:
I'm not against logic behind the change, but it seems that it is much better on current stage to remove EventTypeRepository interface and have EventTypeCache and EventTypeRepository (the one that have db in it's name now) separated. What do you think about it, @ferbncode ?

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -90,49 +85,6 @@ private void insertEventTypeSchema(final EventType eventType) throws JsonProcess
                 jsonMapper.writer().writeValueAsString(eventType.getSchema()));
     }
 
-    public static class EtChange {
```
antban: Thanks!

antban:
:+1:

ferbncode:
:+1:

PR #1160, opened 2020-02-12
Update Kafka Clients to 2.4.0

ferbncode:
Kafka Client removed `zkUtils` and `adminUtils` from client libraries. Implemented their functions from `KafkaZkClient` and `AdminZkClient`.

For creating a `ZookeeperClient` instance, the following new parameters: https://jaceklaskowski.gitbooks.io/apache-kafka/kafka-ZooKeeperClient.html

I kept the max-in-flight requests parameter to `1000`(not too high), for we'd be using this zookeeper connection/client for only creation/update and deletion of topics.

We used the `zkUtils` and `adminUtils` libraries only for creating topics/updating retention time and deleting topics

### DIFF
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaTestHelper.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaTestHelper.java
@@ -1,14 +1,16 @@
 package org.zalando.nakadi.repository.kafka;
 
-import kafka.admin.AdminUtils;
 import kafka.admin.RackAwareMode;
 import kafka.server.ConfigType;
-import kafka.utils.ZkUtils;
+import kafka.zk.AdminZkClient;
+import kafka.zk.KafkaZkClient;
+import kafka.zookeeper.ZooKeeperClient;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.kafka.clients.consumer.KafkaConsumer;
 import org.apache.kafka.clients.producer.KafkaProducer;
 import org.apache.kafka.clients.producer.ProducerRecord;
 import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.utils.Time;
 import org.zalando.nakadi.view.Cursor;
 
 import java.util.List;
@@ -98,15 +100,35 @@ public class KafkaTestHelper {
     }
 
     public void createTopic(final String topic, final String zkUrl) {
-        ZkUtils zkUtils = null;
+        KafkaZkClient zkClient = null;
         try {
-            zkUtils = ZkUtils.apply(zkUrl, 30000, 10000, false);
-            AdminUtils.createTopic(zkUtils, topic, 1, 1, new Properties(), RackAwareMode.Safe$.MODULE$);
+            zkClient = getZkClient(zkUrl);
+            final AdminZkClient adminZkClient = new AdminZkClient(zkClient);
+            adminZkClient.createTopic(topic, 1, 1,
+                    new Properties(), RackAwareMode.Safe$.MODULE$);
         } finally {
-            if (zkUtils != null) {
-                zkUtils.close();
+            if (zkClient != null) {
+                zkClient.close();
             }
         }
+    }
+
+    private static KafkaZkClient getZkClient(final String zkUrl) {
+        KafkaZkClient zkClient = null;
+        zkClient = new KafkaZkClient(
+                new ZooKeeperClient(
+                        zkUrl,
+                        30000,
+                        10000,
+                        1000,
+                        Time.SYSTEM,
+                        "dummyMetricGroup",
+                        "dummyMetricType"
+                ),
+                false,
+                Time.SYSTEM
+        );
+        return zkClient;
     }
 
     public static Long getTopicRetentionTime(final String topic, final String zkPath) {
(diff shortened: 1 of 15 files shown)


antban:
@ferbncode Just to ensure that we are on the same page here - kafka client 2.4 adds dependency on zookeeper client 3.5.*, so the dependency will be not only for ET cache, but for kafka client as well.

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaTestHelper.java
```diff
@@ -98,17 +100,37 @@ public void writeMultipleMessageToPartition(final String partition, final String
...
+                        10000,
+                        1000,
+                        Time.SYSTEM,
+                        "dummyMetricGroup",
+                        "dummyMetricType"
+                ),
+                false,
+                Time.SYSTEM
+        );
+        return zkClient;
```
antban: why not `return new KafkaZkClient(`?

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaTestHelper.java
```diff
@@ -98,17 +100,37 @@ public void writeMultipleMessageToPartition(final String partition, final String
...
+            zkClient = getZkClient(zkUrl);
+            final AdminZkClient adminZkClient = new AdminZkClient(zkClient);
+            adminZkClient.createTopic(topic, 1, 1,
+                    new Properties(), RackAwareMode.Safe$.MODULE$);
         } finally {
-            if (zkUtils != null) {
-                zkUtils.close();
+            if (zkClient != null) {
+                zkClient.close();
             }
```
antban: i do not understand why to punish yourself, make it simpler.
```
final KafkaZkClient zkClient = createZkClient(zkUrl)
try {
  ...
} finally {
  zkClient.close();
}

File: acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaTestHelper.java [hunk 3]
```diff
@@ -98,17 +100,37 @@ public void writeMultipleMessageToPartition(final String partition, final String
...
         }
     }
 
+    private static KafkaZkClient getZkClient(final String zkUrl) {
```
antban: I think that it is not `get`, it is `create`.

ferbncode:
That is correct. In case if we pose problems for Kafka cluster too, we might have to go back to `2.3` which still has `zkUtils` and `adminUtils` - deprecated however, They are removed only in `2.4`.

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaTestHelper.java
```diff
@@ -118,8 +140,16 @@ public static String getTopicCleanupPolicy(final String topic, final String zkPa
...
+        KafkaZkClient zkClient = null;
+        try {
+            zkClient = getZkClient(zkPath);
+            final AdminZkClient adminZkClient = new AdminZkClient(zkClient);
+            final Properties topicConfig = adminZkClient.fetchEntityConfig(ConfigType.Topic(), topic);
+            return topicConfig.getProperty(propertyName);
+        } finally {
+            if (zkClient != null) {
+                zkClient.close();
+            }
```
antban: The same as with previous - make it final, don't check for null (reduce amount of lines)

api-consumption/build.gradle
```diff
@@ -79,8 +80,8 @@ dependencies {
     compile 'io.opentracing:opentracing-api:0.31.0'
     compile 'io.opentracing:opentracing-util:0.31.0'
 
-    compile 'org.apache.kafka:kafka-clients:2.1.0'
-    compile('org.apache.kafka:kafka_2.12:2.1.0') {
+    compile 'org.apache.kafka:kafka-clients:2.4.0'
+    compile('org.apache.kafka:kafka_2.12:2.4.0') {
```
antban: I think it makes sense to move kafka client version to a top level (like with zookeeper/curator)

app/build.gradle
```diff
@@ -102,8 +102,8 @@ dependencies {
     compile 'org.echocat.jomon:runtime:1.6.3'
 
     // kafka & zookeeper
-    compile 'org.apache.kafka:kafka-clients:2.1.0'
-    compile('org.apache.kafka:kafka_2.12:2.1.0') {
+    compile 'org.apache.kafka:kafka-clients:2.4.0'
+    compile('org.apache.kafka:kafka_2.12:2.4.0') {
```
antban: App doesn't need kafka, lets remove kafka dependency from here. 
Kafka is needed by api-consumption, -publishing, -misc (I don't remember all of them), but app is just a holder.

ferbncode:
I'll try on staging regarding update of client and Kafka 2.4. I'll build the image, try on staging and revert my changes (to keep master clean) after review.

ferbncode:
@antban Thank you for pointing out the changes required :). Fixed them in https://github.com/zalando/nakadi/pull/1160/commits/b7f9d91e6209b81c2f1a56b56204b7f2ff776b42

antban:
:+1:

ferbncode:
:+1:

PR #1163, opened 2020-02-18
Aruha-2738 Fix cache nodes deletion behavior

antban:
Aruha-2738 Fix cache nodes deletion behavior

### DIFF
--- a/core-services/src/main/java/org/zalando/nakadi/cache/ChangeSet.java
+++ b/core-services/src/main/java/org/zalando/nakadi/cache/ChangeSet.java
@@ -62,7 +62,7 @@ public class ChangeSet {
             final List<Change> newChangesSorted = newEntry.getValue().stream()
                     .sorted(Comparator.comparing(this::getActualChangeDate).reversed())
                     .collect(Collectors.toList());
-            final long newestAge = System.currentTimeMillis() - newChangesSorted.get(0).getOccurredAt().getTime();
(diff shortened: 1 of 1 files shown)


core-services/src/main/java/org/zalando/nakadi/cache/ChangeSet.java
```diff
@@ -60,9 +61,9 @@ private Date getActualChangeDate(final Change change) {
         for (final Map.Entry<String, List<Change>> newEntry : newChangeMap.entrySet()) {
             // figure out proper dates of the changes
             final List<Change> newChangesSorted = newEntry.getValue().stream()
-                    .sorted(Comparator.comparing(this::getActualChangeDate).reversed())
+                    .sorted(Comparator.comparing(this::getActualChangeDate).thenComparing(Change::getId).reversed())
                     .collect(Collectors.toList());
-            final long newestAge = System.currentTimeMillis() - newChangesSorted.get(0).getOccurredAt().getTime();
+            final long newestAge = System.currentTimeMillis() - getActualChangeDate(newChangesSorted.get(0)).getTime();
```
antban: The fix is : 
`newChangesSorted.get(0).getOccurredAt()` -> `getActualChangeDate(newChangesSorted.get(0))`

antban: Everything else is to guarantee the same behavior on all the nodes

antban:
:+1:

thomasabraham:
👍

PR #1165, opened 2020-02-18
ARUHA-2738 Fix deletion of zk nodes.

antban:
Finally bug was found - I was not calling method that supposed to remember previous state


antban:
:+1:

antban:
:+1:

ferbncode:
:+1:

PR #1172, opened 2020-02-27
Allow to modify partition_key_fields

rcillo:
https://jira.zalando.net/browse/ARUHA-2648

For a long time changing partition_key_fields was an operation
performed only by admins via a database hack. This used to be case in
order to protect consumers from unexpected changes on how events are
hashed, which could lead to a number of issues on the consumer side.

But as the number of such requests grew, we decided to let the users
decided whether this field should or not be changed. We give them the
responsibility over this operation.


ferbncode:
:+1:

rcillo:
👍

PR #1177, opened 2020-03-09
ARUHA-2746 reconfigure content negotiation

antban:
Sometimes users tend to create event types with names that have extensions like ones defined here: 
https://github.com/spring-projects/spring-framework/blob/master/spring-context-support/src/main/resources/org/springframework/mail/javamail/mime.types

This leads to 406 status code while getting event type. This fix is removing content type negotiation in this case.


antban:
:+1:

adyach:
👍

antban:
:+1:

lmontrieux:
:+1:

PR #1178, opened 2020-03-11
ARUHA-2755 Fix problem with the fact that BEGIN is not lexicographicaly comparable

antban:
The problem was introduced as a part of refactoring eliminating the need of additional dependencies. 1 corner case was forgotten


thomasabraham:
👍

antban:
:+1:

antban:
:+1:

thomasabraham:
👍

PR #1179, opened 2020-03-20
ARUHA-2714 Refactor Authz usage not to return 400 status code

antban:
ARUHA-2714 
The refactoring is made to be able to use several token info endpoints
one after another in case of error and the behavior was tested
intensively.

### DIFF
--- a/app/src/main/java/org/zalando/nakadi/config/AuthenticationConfig.java
+++ b/app/src/main/java/org/zalando/nakadi/config/AuthenticationConfig.java
@@ -1,25 +1,20 @@
 package org.zalando.nakadi.config;
 
 import com.codahale.metrics.MetricRegistry;
-import com.codahale.metrics.Timer;
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.RequestConfig;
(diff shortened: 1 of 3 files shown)


antban:
:+1:

adyach:
https://aws.amazon.com/builders-library/avoiding-fallback-in-distributed-systems/

rcillo:
👍

rcillo:
@adyach very interesting article. I understand that this code might introduce a latent issue. For example, if all of a sudden, all side cars stop working, we could easily bring down the entire IAM by calling the remote endpoint.

What the authors of the article point out as desirable altertive is:
- let the client handle it. This means, our clients should retry. I kind of agree with it.
- improve the availability of sidecar. This is tough because we would have to handle containers (nakadi and token info) launching and terminating in different order. I'm not even sure it's possible.

In fact, maybe this is a good risk to take. I think this happens very often (more than 2k occurences in the last 4 hours) so it's not like it's rarely exercised code. It'll be hit quite often and I think for this reason it has lesser of a chance of becoming a latent problem.

PR #1182, opened 2020-03-25
Reduce logging by conditionally disabling ACCESS_LOG

rcillo:
ACCESS_LOG lines represent 33% of all the logs from Nakadi. This
information is already available via internal event type
`nakadi.access.log` archived in the  datalake and many other important
operations are instrumented via open tracing, making this log line
redundant.

Since there might be a need to switch access logs back, we are adding
a feature toggle to enable access log. This toggle is turned off by
default meaning that new deployments will have access log turned off
by default.

The feature toggle is checked inside the method and not on the caller code because this method is used in multiple places.


rcillo:
👍

ferbncode:
:+1:

ferbncode:
@rcillo Do you think we should specify the default feature toggle too in `application.yaml`?

PR #1183, opened 2020-03-25
Log 5xx even when access log is disabled

rcillo:
Not logging 5xx would make it difficult to catch bugs as the stack
trace would be lost.

This change also makes successful publishing request to be logged, so turning access log on should be done with care and only when needed, otherwise the log volume could be even bigger.

Do you think I should add another condition on top of these so that successful publishing requests are never logged?

### DIFF
--- a/app/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
+++ b/app/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
@@ -81,7 +81,10 @@ public class LoggingFilter extends OncePerRequestFilter {
             this.response = response;
             this.flowId = flowId;
             this.requestLogInfo = new RequestLogInfo(request, startTime);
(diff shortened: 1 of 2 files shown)


ferbncode:
I think it might be good to not log successful publishing even with the toggle turned on. There might be a huge increase in logs if we were to leave access logs open for an evening or weekend for finding possible bug. What do you think?

rcillo:
@ferbncode I agree. It can see that happening very easily. I changed the conditional to also check that it's not a successful publishing request. 

So now we have access log in the following situations:

- it's a 5xx, regardless of which endpoint or feature toggles are turned on
- feature toggle is enabled and it's not a publishing request

It's quite complicated but I think it's the behaviour we are looking for to preserve easiness of debugging and not over log accidentally.

rcillo:
@ferbncode I also turned on on test and local environment by default, so that when you are coding locally you can see access logs.

ferbncode:
:+1:

rcillo:
👍

PR #1184, opened 2020-03-26
Log 207's and small refactoring

rcillo:
This commit logs 207's which is considered a server side error and
one of the biggest mistakes in the API design, from my point of
view, given how many clients interpreted it as success in the last 4
years of operating Nakadi. It's such a mistake that even ourselves overlooked
it yesterday when we planned to only log server side errors.

The only type of publishing request that we would like to log are
those where the error is on the server side. Client side errors can be
checked through lightstep and explained in the FAQ.


adyach:
👍

rcillo:
👍

rcillo:
@adyach I fixed checkstyle.

adyach:
👍

PR #1185, opened 2020-03-30
Allow Nakadi admins to change compatibility mode in unsafe ways

rcillo:
https://jira.zalando.net/browse/ARUHA-2804

Users often make mistakes when editing their schemas. Sometimes it's a
typo on a properties name, others it's something more serious that causes events to be
rejected. The rules often do not allowed for such changes to be
undone, as that would mean breaking the compatibility mode (at least
in theory).

But this has led our team to make manual changes to the database
several times as a way to rescue users from dead ends.

Since hacking the database is less than ideal, we are making it
possible for admins (Nakadi admins, not event type admins) to downgrade the compatibility mode, so that users
can fix their schemas and then set the compatibility mode back to what
it was, which they can do on their own.

This is only made available for Nakadi admins, as it's a potentialy
harmful operation if users execute this without properly reflecting on
the downstream consequences of performing incompatible changes to schemas.


ferbncode:
:+1:

rcillo:
👍

PR #1186, opened 2020-04-01
Register new schema independently from changing other properties

rcillo:
The only way to update a schema so far has been to call `PUT
/event-types/{name}` with the complete definition of the event type in
the body (including the modified schema).

This way of changing schemas has proven to be problematic, as users
could accidentally modify other properties of the event type when
their intention would be to only change the schema.

One example in particular has been causing a lot of problems:
automated schema evolution performed by Nakadi SQL.

Nakadi SQL listens to schema changes made to input event types of
queries and automatically reflects those changes in the schema of the
output event type. It then calls `PUT /event-types/{name}` with the
new schema in order to register it. But this call fails due to
something unrelated to the schema change: the authorization section.

Every time an event type update is performed, all the properties are
validated, including authorization. But when users leave the company,
their identifiers are no longer valid, meaning that updating updating
event types fail if they are not removed from the event type
definition.

We consider that removing users from the authorization section is
responsibility of the team owning the event type and not job of an
automated task related with schema updating. So we are decoupling both
changes in order to solve the problem of automated schema change
independently from maintaining an up-to-date authorization section.

## Rejected alternative

Automatically removing invalid users from authorization section has been considered a potentially harmful operation, as it could result in empty authorization sections where it's no longer possible to keep track of who were the original maintainers of the event type.

### DIFF
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
@@ -25,7 +25,6 @@ import org.zalando.nakadi.service.EventTypeService;
 import org.zalando.nakadi.service.SchemaService;
 
 import javax.validation.Valid;
-import javax.validation.constraints.NotNull;
 
 import static org.springframework.http.ResponseEntity.status;
 
@@ -43,7 +42,7 @@ public class SchemaController {
 
     @RequestMapping(value = "/event-types/{name}/schemas", method = RequestMethod.POST)
     public ResponseEntity<?> create(@PathVariable("name") final String name,
-                                    @Valid @RequestBody @NotNull final EventTypeSchemaBase schema,
+                                    @Valid @RequestBody final EventTypeSchemaBase schema,
                                     final Errors errors) {
         if (errors.hasErrors()) {
             throw new ValidationException(errors);
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
@@ -3,21 +3,31 @@ package org.zalando.nakadi.controller;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.validation.Errors;
 import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.context.request.NativeWebRequest;
 import org.zalando.nakadi.domain.EventType;
 import org.zalando.nakadi.domain.EventTypeSchema;
+import org.zalando.nakadi.domain.EventTypeSchemaBase;
 import org.zalando.nakadi.domain.PaginationWrapper;
 import org.zalando.nakadi.exceptions.runtime.InternalNakadiException;
 import org.zalando.nakadi.exceptions.runtime.InvalidLimitException;
 import org.zalando.nakadi.exceptions.runtime.InvalidVersionNumberException;
 import org.zalando.nakadi.exceptions.runtime.NoSuchEventTypeException;
 import org.zalando.nakadi.exceptions.runtime.NoSuchSchemaException;
+import org.zalando.nakadi.exceptions.runtime.ValidationException;
 import org.zalando.nakadi.service.EventTypeService;
 import org.zalando.nakadi.service.SchemaService;
+
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+
+import static org.springframework.http.ResponseEntity.status;
 
 @RestController
 public class SchemaController {
@@ -31,7 +41,21 @@ public class SchemaController {
         this.eventTypeService = eventTypeService;
     }
 
-    @RequestMapping("/event-types/{name}/schemas")
+    @RequestMapping(value = "/event-types/{name}/schemas", method = RequestMethod.POST)
+    public ResponseEntity<?> create(@PathVariable("name") final String name,
+                                    @Valid @RequestBody @NotNull final EventTypeSchemaBase schema,
+                                    final Errors errors) {
+        if (errors.hasErrors()) {
+            throw new ValidationException(errors);
+        }
+
+        final EventType eventType = eventTypeService.get(name);
+        schemaService.addSchema(eventType, schema);
+
+        return status(HttpStatus.OK).build();
+    }
+
+    @RequestMapping(value = "/event-types/{name}/schemas", method = RequestMethod.GET)
     public ResponseEntity<?> getSchemas(
             @PathVariable("name") final String name,
             @RequestParam(value = "offset", required = false, defaultValue = "0") final int offset,
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
@@ -48,8 +48,7 @@ public class SchemaController {
             throw new ValidationException(errors);
         }
 
-        final EventType eventType = eventTypeService.get(name);
-        schemaService.addSchema(eventType, schema);
+        schemaService.addSchema(name, schema);
 
         return status(HttpStatus.OK).build();
     }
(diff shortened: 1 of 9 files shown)


antban:
I'm good with api, but I thought there will be actual change implemented (((

rcillo:
@antban sorry I wasn't clear in the description. I'm waiting for meeting later today for a final alignment if this is the way to go.

rcillo:
After we agreed on the approach and the API, I just added the implementation in my latest commit.

This implementation adds very little new code. It mostly moves to
SchemaService most of the code that is now used by both update event
type endpoint and the new add schema endpoint since both have to
perform schema validation.

The tests were adjusted to test the right layer and I made my best to
avoid circular dependencies among services, this is why there are so
many dependencies autowired.

The happy-path test case was implemented in the UserJorney. The
exceptional cases are covered indirectly by schema validation tests.

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -31,7 +40,21 @@ public SchemaController(final SchemaService schemaService, final EventTypeServic
         this.eventTypeService = eventTypeService;
     }
 
-    @RequestMapping("/event-types/{name}/schemas")
+    @RequestMapping(value = "/event-types/{name}/schemas", method = RequestMethod.POST)
```
rcillo, line 43: This is the new endpoint

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -582,83 +571,4 @@ private void validateEventOwnerSelectorUnchanged(final EventType original, final
             throw new InvalidEventTypeException("event_owner_selector can't be set back to null");
         }
     }
-
-    private void validateSchema(final EventTypeBase eventType) throws InvalidEventTypeException {
```
rcillo, line 586: I moved all this code into `SchemaService` since it's now used by both `SchemaController` and.`EventTypeController`. It's still possible to update the schema using `PUT /event-types/{name}`.

api-metastore/src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -1,33 +1,86 @@
...
         this.paginationService = paginationService;
+        this.jsonSchemaEnrichment = jsonSchemaEnrichment;
+        this.schemaEvolutionService = schemaEvolutionService;
+        this.eventTypeRepository = eventTypeRepository;
+        this.adminService = adminService;
+        this.authorizationValidator = authorizationValidator;
+        this.eventTypeCache = eventTypeCache;
+    }
+
+    public void addSchema(final EventType originalEventType, final EventTypeSchemaBase newSchema) {
```
rcillo, line 70: This is new code (well, sort of since I just follow the same pattern used in other services, with respect to authorization and how cache is used).

api-metastore/src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -55,4 +108,82 @@ public EventTypeSchema getSchemaVersion(final String name, final String version)
         final EventTypeSchema schema = schemaRepository.getSchemaVersion(name, version);
         return schema;
     }
-}
+
+    public void validateSchema(final EventTypeBase eventType) throws InvalidEventTypeException {
```
rcillo, line 112: This is existent code, all the way to the end of this file.

api-metastore/src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -239,17 +236,6 @@ public void whenPostWithNoSchemaSchemaThenReturn422() throws Exception {
         postETAndExpect422WithProblem(eventType, expectedProblem);
     }
 
-    @Test
```
rcillo, line 242: All of these tests are related to schema change and they were moved to the appropriate layer under `SchemaServiceTest`. I didn't add anything new here.

api-metastore/src/test/java/org/zalando/nakadi/service/EventTypeServiceTest.java
```diff
@@ -247,20 +243,6 @@ public void testFeatureToggleDisableLogCompaction() {
         eventTypeService.create(eventType, true);
     }
 
-    @Test
```
rcillo, line 250: Moved tests from this class to the new class responsible for this behaviour now `SchemaService`.

api-metastore/src/test/java/org/zalando/nakadi/service/SchemaServiceTest.java
```diff
@@ -1,30 +1,57 @@
...
+import org.zalando.nakadi.validation.JsonSchemaEnrichment;
 
+import java.io.IOException;
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.mockito.Matchers.any;
+import static org.zalando.nakadi.domain.EventCategory.BUSINESS;
 import static org.zalando.nakadi.utils.TestUtils.buildDefaultEventType;
 
 public class SchemaServiceTest {
```
rcillo, line 30: Most tests in this class are old and were just transported from other classes. I'll mark the ones that are new.

rcillo:
I added comments to help in the review as most code was just moved around and it would not be fair to review it as well as my goal here was not to rewrite Nakadi but just add this new endpoint using existent code.

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -31,7 +40,21 @@ public SchemaController(final SchemaService schemaService, final EventTypeServic
...
+    public ResponseEntity<?> create(@PathVariable("name") final String name,
+                                    @Valid @RequestBody final EventTypeSchemaBase schema,
+                                    final Errors errors) {
+        if (errors.hasErrors()) {
+            throw new ValidationException(errors);
+        }
+
+        final EventType eventType = eventTypeService.get(name);
+        schemaService.addSchema(eventType, schema);
```
antban, line 52: It makes sense to pass event type name to `addSchema` call, instead of event type, that is takes from cache, not from database (or find another solution)

rcillo:
@antban regarding

> There is a need to update real database values, not cached value, as cache may have problems (event type service uses repository instead of cache to get value for update)

Please check line 85 from `SchemaService`. That's exactly what is done. It updates the database to only later invalidate the cache.

> During schema update it is probably needed to take timelines lock, this way it will be guaraneed that schema update will be propagated to all the instances (timelineSync.takeLock), and what is more important - user updates will not affect timelines switch.

I don't think this is necessary. Our cache implementation was rewritten at least 3 times to make this problem less likely to happen. I did wrapped (in the latest commit) all database calls in a transaction to make sure that concurrent updates wouldn't introduce some weird state in the database.

> It makes sense to pass event type name to addSchema call, instead of event type, that is takes from cache, not from database (or find another solution)

done in 4fd1a87f5

antban:
@rcillo 
>> Please check line 85 from SchemaService. That's exactly what is done. It updates the database to only later invalidate the cache.

I was mentioning state of PR before your last commit. In PR `SchemaService` was using `EventType` that was taken from cache by `SchemaController` from `EventTypeService`, and cached value may be not in the latest state. Now the problem has gone, thanks for fixing.

>> I don't think this is necessary. Our cache implementation was rewritten at least 3 times to make this problem less likely to happen.

It's very nice that our cache implementation was mentioned here, but the comment was about completely different thing. Timelines switch requires lock for event type updates. You are doing writes to database without holding this lock, effectively ignoring ongoing timeline switch. There is nothing about cache. 
Maybe I wasn't clear while explaining. 
During timelines switch there is a sequence of actions to be executed on top of event type, and it is expected, that there will be no modifications of event type during this time. Even publishing is paused for some time (and timeline switch ensures that cache is invalidated on every instance, not relying on eventual consistency).
It may happen, that the fields that are modified here are not related to timeline switch and are not affected, but for me it makes sense to be on a safe side and say `timelineSync.workWithEventType`, in order to avoid problems (I wrote wrong method name in the first comment).

rcillo:
@antban about timeline sync, I don't think it would add much. This is just about schema. It's ok to change the schema in the middle of timeline switch. Besides, timelines is not supposed to be used anymore, unless we are in the middle of an incident, something that happened only once in more than 12 months.

rcillo:
@antban please let me know if there are any changes that you consider important in order to merge this PR.

antban:
@rcillo I already told that I consider informing timeline switch process about the fact that you are updating event type to be very important. 
I even would agree with your comment about updating schema only, but it is not true. You are updating event type completely (because there is not way to do it differently through our api). 
I think that this approach (ignoring lock) will lead to the problems, and can not approve it.

rcillo:
@antban I hope I addressed your concerns in the latest commit ac917d6fa

antban:
I would go with try-with-resource, but probably it's a matter of taste and simplicity of testing

antban:
:+1:

rcillo:
👍

PR #1188, opened 2020-04-01
Validate request body on POST /settings/features

rcillo:
This endpoint used to reply with 500 when the request body was
invalid.

Now it replies with 400 and some meaningful error that guides the user
towards a fix.


thomasabraham:
👍

rcillo:
👍

PR #1189, opened 2020-04-02
Block external references in schemas

rcillo:
Prior to this fix it would be possible to refer an external schema in
a schema definition. If the URL would be valid, it would fetch it over
the network. All sorts of problems could happen here:

- invalid URL
- timeout due to server being unresponsive
- other network issues

Besides, loading a schema from an external source could mean that
schema evolution rules could be worked around, by externalizing the
chunk of the schema that would change without any rules, as we do not
verify external schemas.

On top of all this problems, external schemas were never meant to be
supported and this feature made its way in the project since version 0
of Nakadi without the maintainers being aware of it.

After this fix, using external urls will result in 422 with a message
like:

`external url reference is not supported: /invalid/url`


antban:
:+1:

antban:
I checked all schemas on live - they all don't have this kind of problem

ferbncode:
:+1:

PR #1191, opened 2020-04-06
recreate watcher on brokers change list

adyach:
# One-line summary
the fix narrows down the triggering on brokers change watcher only for the case when list of broker id nodes changed

replicates the broken branch :) https://github.com/zalando/nakadi/pull/1119
```
Currently, broker list changed watcher is recreated on disconnects, which means a lot of watchers created for no reason and later on triggered for the same reason creating even more watchers and so on.
Although, it is clearly seen from the logs, this is hypothesis and the testing is required.
```


adyach:
👍

rcillo:
👍

PR #1193, opened 2020-04-07
Change retention time in documentation, remove obsolete answer

### LINKED ISSUE #1190: Retention time is 2 days by default but it should be 4 days
According to [the documentation](https://nakadi.io/manual.html#definition_EventTypeOptions) default value for the `retention_time` is 345600000 ms (4 days).
When an event is created without `options.retention_time` [the event retention_time is 2 days](https://github.com/zalando/nakadi/blob/master/app/src/main/resources/application.yml#L52), and also [db migration](https://github.com/zalando/nakadi/blob/master/database/migration/R2016_08_22/aruha-312-set-default-retention-time-when-it-was-not-provided.sql#L2)

[faq](https://github.com/zalando/nakadi/blob/master/docs/_documentation/faq.md#how-do-i-define-how-long-will-events-be-persisted-for) should be probably also updated since the following is no longer true:
```
At the moment, retention can't be defined via the API per event type.
```

antban:
The default value is 2 days for quite a while, but it was'n reflected everywhere in the documentation. 
This PR removes some obsolete answers and fixes default value in api specification. 
Fixes #1190

### DIFF
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -2866,7 +2866,7 @@ definitions:
       retention_time:
         type: integer
         format: int64
-        default: 345600000 # 4 days
+        default: 172800000 # 4 days
         description: |
           Number of milliseconds that Nakadi stores events published to this event type.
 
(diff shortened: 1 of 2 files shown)


antban:
:+1:

antban:
:+1:

rcillo:
👍

PR #1194, opened 2020-05-05
update scope info required for accessing API

riteshpuj2013:
## Description
For consuming or submtting events, uid scope is enough, we don't need `nakadi.event_stream.write` and `nakadi.event_stream.read`.

I have removed other parts from the PR template as they were not relevant to the changes I made.

Please review and let me know if more info is required.

### DIFF
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -90,10 +90,8 @@ securityDefinitions:
         Grants access for changing Nakadi configuration.
       nakadi.event_type.write: |
         Grants access for applications to define and update EventTypes.
-      nakadi.event_stream.write: |
-        Grants access for applications to submit Events.
-      nakadi.event_stream.read: |
-        Grants access for consuming Event streams.
+      uid: |
+        Grants access for applications to submit or consume Events.
 
 paths:
   /metrics:


adyach:
@riteshpuj2013 thank you for contributing! 
there are mentions of the scopes in this document and tests, we need to remove it as well.

riteshpuj2013:
@adyach Thanks for pointing it out. I will work on them.

riteshpuj2013:
@adyach Here is what I found after looking into the code:

1. I checked the code and I figured that we use BASIC mode of authentication by **default** as mentioned in the [config](https://github.com/riteshpuj2013/nakadi/blob/b8fa2fa34d9ec615ca812d5718cd73fe3ee583af/app/src/main/resources/application.yml#L86-L86) 
In Basic mode, only [uid scope is required](https://github.com/riteshpuj2013/nakadi/blob/b8fa2fa34d9ec615ca812d5718cd73fe3ee583af/app/src/main/java/org/zalando/nakadi/config/SecurityConfiguration.java#L111-L111) for all the Endpoints.
This means that I can replace all the scopes with `uid` in the API spec.

2. We have security configurations for [FULL](https://github.com/riteshpuj2013/nakadi/blob/b8fa2fa34d9ec615ca812d5718cd73fe3ee583af/app/src/main/java/org/zalando/nakadi/config/SecurityConfiguration.java#L84-L84) which uses the specific scopes like `nakadi.event_stream.write`, `nakadi.event_type.write`. `nakadi.config.write` and `nakadi.event_stream.read`. As we have tests for these configurations, we can't remove the scopes from the tests. If we want to remove the tests, then we will also have to remove the code for FULL authentication mode.

Question: Is my understanding correct? Can I go ahead with findings from Point 1 and skip removing tests due to reasons in Point 2?
Please let me know if I am missing something here.

adyach:
@riteshpuj2013 i think I did a mistake suggesting to remove it, it looks like there is still scope based authz in Nakadi. my apologies, we should not remove it from the project.

riteshpuj2013:
no worries, in that case, should I just update the API doc with `uid` scope everywhere as `BASIC` is the default mode that we use and it needs only `uid` scope for all APIs?
If we change the Auth mode in future, then we can update the API doc to reflect the same. What do you think?

adyach:
@riteshpuj2013 as I understand you want to achieve consistency in api definition and some application configuration. I would not do that. the proper way would be to extend api doc to describe different ways of authentication in nakadi, and I do not know any way how to do it in swagger api. I suggest to close this pr, since the purpose was different, and if you know how to improve api doc describing all possible authentication mechanisms, please, open another pr.

riteshpuj2013:
@adyach I see, thanks for the explanation. I close this PR.

PR #1195, opened 2020-05-06
Change schema created at time only if new schema

ferbncode:
Don't change schema created at time if nothing changes for schema when event type is updated.
> Zalando ticket : ARUHA-2837 (only if appropriate)

Even without no changes to the schema, event type updates would change the schema `created_at` time without version upgrade.


rcillo:
Could you please add some small test to cover this behavious?

thomasabraham:
👍

ferbncode:
:+1:

PR #1196, opened 2020-05-12
Add tracing span for fetch stats request

ferbncode:
Add span for request to fetch subscription statistics. Note that closing of the span is handled in [TracingFilter.java](https://github.com/zalando/nakadi/blob/0f8127af2272d106e1edb6fdba7f3708629097d5/core-common/src/main/java/org/zalando/nakadi/filters/TracingFilter.java#L134). Error code handling for the span is also handled [here](https://github.com/zalando/nakadi/blob/0f8127af2272d106e1edb6fdba7f3708629097d5/core-common/src/main/java/org/zalando/nakadi/filters/TracingFilter.java#L128).

### DIFF
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
@@ -81,7 +81,7 @@ public class SubscriptionController {
             NoSuchEventTypeException, NoSuchSubscriptionException, ServiceTemporarilyUnavailableException {
         final Span statsSpan = TracingService.extractSpan(request, "fetch_stats")
                 .setTag("subscription_id", subscriptionId)
-                .setTag("time_lag", showTimeLag);
+                .setTag("show_time_lag", showTimeLag);
         final StatsMode statsMode = showTimeLag ? StatsMode.TIMELAG : StatsMode.NORMAL;
         return subscriptionService.getSubscriptionStat(subscriptionId, statsMode, statsSpan);
     }
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
@@ -1,5 +1,6 @@
 package org.zalando.nakadi.controller;
 
+import io.opentracing.Span;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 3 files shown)


api-metastore/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -72,10 +75,14 @@ public SubscriptionController(final SubscriptionService subscriptionService) {
...
     public ItemsWrapper<SubscriptionEventTypeStats> getSubscriptionStats(
             @PathVariable("id") final String subscriptionId,
-            @RequestParam(value = "show_time_lag", required = false, defaultValue = "false") final boolean showTimeLag)
+            @RequestParam(value = "show_time_lag", required = false, defaultValue = "false") final boolean showTimeLag,
+            final HttpServletRequest request)
             throws InconsistentStateException,
             NoSuchEventTypeException, NoSuchSubscriptionException, ServiceTemporarilyUnavailableException {
+        final Span statsSpan = TracingService.extractSpan(request, "fetch_stats")
+                .setTag("subscription_id", subscriptionId)
+                .setTag("time_lag", showTimeLag);
```
antban, line 84: I think that `time_lag` should be renamed back to `show_time_lag`

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
```diff
@@ -245,14 +246,16 @@ public void deleteSubscription(final String subscriptionId)
...
+                                                                        final StatsMode statsMode,
+                                                                        final Span span)
             throws InconsistentStateException, NoSuchEventTypeException,
             NoSuchSubscriptionException, ServiceTemporarilyUnavailableException {
         final Subscription subscription;
         try {
             subscription = subscriptionRepository.getSubscription(subscriptionId);
             authorizationValidator.authorizeSubscriptionView(subscription);
         } catch (final ServiceTemporarilyUnavailableException ex) {
+            TracingService.logErrorInSpan(span, ex.getMessage());
```
antban, line 258: In case if ex.getMessage() is Null for some reason, then the error won't be logged (I know that everyone is using it like that), maybe it makes sense to add method `TracingService.logErrorInSpan(span, ex)` that will not have this problem.

ferbncode: From the docs for the log function:
``` 
key:value log fields. Tracer implementations should support String, numeric, and boolean values; some may also support arbitrary Objects
```
For fixing logging in the case that you suggested, I added a function that would log the exceptions' classname + optionally the localized message in https://github.com/zalando/nakadi/pull/1196/commits/e67181f6e343ceceda6b507e96e8e8f993f7c642

api-metastore/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
ferbncode: Fixed in https://github.com/zalando/nakadi/pull/1196/commits/2ee5b06be97b35973039e968b4dfb8246654e0d8

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
antban: Yep, that's almost exactly what I meant. I thought about `ex.getClass().getSimpleName() + ex.getMessage()`, but it's anyways good.

antban:
:+1:

ferbncode:
:+1:

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
ferbncode: Got it, thanks!

PR #1197, opened 2020-05-15
ARUHA-2882: PartitionData generated a lot short lived objects

adyach:
After having a look at heap dump, it is clear where the problem is:
VisualVM 1 4 4 2020-05-11 16-02-50
Eclipse Memory Analyzer 2020-05-11 16-39-19

Nakadi consumption uses TreeSet (PartitionData.allCursorsOrdered) to calculate the distance between cursors. Every time cursors are committed or a decision is made to send more data to the client a new subset is created, and once decision is made objects are thrown away, which happens all the time during streaming.

https://github.com/zalando/nakadi/blob/abbabfb0b5598352294bf488a4c852987e51a664/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java#L218

https://github.com/zalando/nakadi/blob/abbabfb0b5598352294bf488a4c852987e51a664/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java#L252

Changing a way to calculate distance will bring performance improvement in reduction of 30%-40% CPU usage on consumption stack as experiments showed.

**I would like first check the correctness, and then execute a performance test**

### DIFF
--- a/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
+++ b/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
@@ -72,6 +72,7 @@ public class StreamingContext implements SubscriptionStreamer {
     private final AutocommitSupport autocommitSupport;
     private final Span currentSpan;
     private final String kpiDataStreamedEventType;
+    private final CursorOperationsService cursorOperationsService;
 
     private final long kpiCollectionFrequencyMs;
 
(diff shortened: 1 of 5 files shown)


antban:
Everything looks good. 
It seems that it's also possible to refactor it not to put distance calculator there.
The only reason why to get commit count is to manage stream_limit, as far as I remember, so it is possible not to store in in PartitionData at all. 
But I think that we should do it later

adyach:
👍

antban:
:+1:

PR #1201, opened 2020-05-28
Handling SchemaValidation and SchemaEvolution exceptions correctly

thomasabraham:
# Handling SchemaValidation and SchemaEvolution exceptions correctly

> Zalando ticket : ARUHA-2898 

## Description
`SchemaValidationService` and `SchemaEvolutionService` were using `InvalidEventTypeException`, which is incorrect and is not handled by `SchemaExceptionHandler`. 
This PR corrects this by introducing `SchemaEvolutionException` and `SchemaValidationException` which are correctly handled by `SchemaExceptionHandler`

## Review
- [ ] Tests


api-metastore/src/main/java/org/zalando/nakadi/controller/advice/SchemaExceptionHandler.java
```diff
@@ -24,4 +27,20 @@
...
+    @ExceptionHandler(SchemaEvolutionException.class)
+    public ResponseEntity<Problem> handleSchemaEvolutionException(
+            final SchemaEvolutionException exception,
+            final NativeWebRequest request) {
+        AdviceTrait.LOG.debug(exception.getMessage());
+        return create(Problem.valueOf(UNPROCESSABLE_ENTITY, exception.getMessage()), request);
+    }
+
+    @ExceptionHandler(SchemaValidationException.class)
+    public ResponseEntity<Problem> handleSchemaEvolutionException(
```
rcillo, line 40: Shouldn't the name of this method be `handleSchemaValidationException`?

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -440,6 +446,12 @@ public void update(final String eventTypeName,
             LOG.error("Failed to wait for timeline switch", e);
             throw new ServiceTemporarilyUnavailableException(
                     "Event type is currently in maintenance, please repeat request", e);
+        } catch (final SchemaValidationException e) {
+            LOG.error("Schema validation failed {}", e.getMessage());
```
rcillo, line 450: I would argue this is not an error log. It's not like something broke internally. The user provided invalid data.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -440,6 +446,12 @@ public void update(final String eventTypeName,
...
+            throw new InvalidEventTypeException(e);
+        } catch (final SchemaEvolutionException e) {
+            LOG.error("Schema evolution failed {}", e.getMessage());
```
rcillo, line 453: The same for this.

api-metastore/src/main/java/org/zalando/nakadi/service/SchemaEvolutionService.java
```diff
@@ -95,7 +96,7 @@ private void collectErrorMessages(final ValidationException e,
         }
     }
 
-    public EventType evolve(final EventType original, final EventTypeBase eventType) throws InvalidEventTypeException {
+    public EventType evolve(final EventType original, final EventTypeBase eventType) {
```
rcillo, line 99: Even though the exception is a runtime exception we try to "document" that such methods can throw some exception. So it would be nice to see `throws SchemaEvolutionException` (if that's the case).

api-metastore/src/main/java/org/zalando/nakadi/controller/advice/SchemaExceptionHandler.java
thomasabraham: True, fixed.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
thomasabraham: Reduced to warn

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
thomasabraham: Reduced to warn

api-metastore/src/main/java/org/zalando/nakadi/service/SchemaEvolutionService.java
thomasabraham: Done

rcillo:
👍

thomasabraham:
👍

PR #1202, opened 2020-05-29
Allow for type narrowing under forward compatibility mode

rcillo:
Changing a schema from empty `{}` to something that is not empty
`{type: string}` should be seen as a type narrowing operation that
doesn't break consumers of forward compatible event types, given the set of events accepted by the
narrowed definition is a subset of the original events. That means
that such a change should be allowed in this direction.

It could be argued that other changes are also type narrowing, as for example `{type:[string,null]}` to `{type:[string]}`, but detecting all possible "type narrowing" changes is a complex problem as it could be posed in a more formal way as "the new schema accepts only a subset of events from the previous schema" which is a computationally difficult problem ( I found some papers years ago). So this is an approximation that handles a specific case that is of significant value for us and fixes a bug in Nakadi SQL automatic schema evolution.

For fully compatible mode that is still not allowed as it means that
consumers developed based on the latest schema would not be able to
read events published under the oldest schema, contradicting the
definition of fully compatible.

### DIFF
--- a/api-metastore/src/main/java/org/zalando/nakadi/validation/schema/diff/SchemaDiff.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/validation/schema/diff/SchemaDiff.java
@@ -99,7 +99,7 @@ public class SchemaDiff {
         });
     }
 
-    private static boolean typeNarrowed(Schema originalIn, Schema updateIn) {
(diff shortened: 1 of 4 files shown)


thomasabraham:
👍

rcillo:
👍

PR #1205, opened 2020-06-26
Log full message instead of unusable one for event validation

antban:
Log full message instead of unusable one for event validation


antban:
:+1:

rcillo:
👍

PR #1206, opened 2020-06-26
Upgrade to spring boot 2.3.1.RELEASE

rcillo:
# update 2020-07-07

We achieved an stable version that passes all tests and runs on staging. We'll merge it as is and proceed to check its stability under higher loads.

# original description

This PR is not supposed to be merged and is here just to share the
current state that I find myself right now.

We did not upgrade spring boot since ever. This lead to accumulating a
lot of breaking changes.

My strategy was to get the project to build and boot by removing
dependencies that I couldn't reconciliate the version (due to
transitive incompatible dependencies) and lack of support for spring
boot 2 or spring framework 5+.

By no means I plan to remove all these features so please be
patient. Now that the project builds, I can start to bring them back
one by one. This way is better than try to upgrade everything at once,
which proved to be almost impossible task as the project didn't even
compile for days of work.

List of fixes needed:

- [x] kafka client (I had problem instantiating zk config bean and only solved by using spring-kafka)
- [x] hystrix (we had to change it due to springframework. The interface changed and a few methods we used are gone. We have to figure out if we have to use it differently)
- [x] org.zalando.zmon:zmon-actuator:0.9.8
- [x] WebConfig#requestMappingHandlerMapping() no longer part of WebMvcConfigurationSupport interface
- [x] twintip incompatibility
- [x] disk usage job incompatibility
- [x] mockito incompatibility
- [x] problem with serialization and deserialization
- [x] ensureOffsetsSubscriptionsAreRefreshedAndClosed and ensureInitializationFailsWhenInvalidCursorsUsed null pointers
- [x] fix plugins build
- [x] fix renaming of jetty config `Property source 'systemEnvironment': Key: server.jetty.acceptors	Replacement: server.jetty.threads.acceptors`


rcillo:
👍

thomasabraham:
👍

PR #1209, opened 2020-07-06
Fix query params

thomasabraham:
With new spring boot, in web config RequestMappingHandlerMapping needs to be replaced with PathMatchConfigurer


rcillo:
👍

thomasabraham:
👍

PR #1211, opened 2020-07-06
Copying api to /zalando-apis directory as well

thomasabraham:
According to API guild: 

> An API is published by copying its Open API specification into the reserved /zalando-apis directory of the deployment artifact used to deploy the provisioning service. The directory must only contain self-contained YAML files that each describe one API (exception see MUST only use durable and immutable remote references). We prefer this deployment artifact-based method over the past (now legacy) .well-known/schema-discovery service endpoint-based publishing process, that we only support for backward compatibility reasons.

As we removed twintip, we are adding the API to `/zalando-apis`


adyach:
👍

thomasabraham:
👍

PR #1214, opened 2020-07-07
Remove conflicting dependency

rcillo:
When launching the app with plugins, this dependency clashes with
dependencies introduced by the plugins.


rcillo:
👍

thomasabraham:
👍

PR #1215, opened 2020-07-07
Replace Hikari by Tomcat pool

rcillo:
In upgrading to spring boot 2 the default db pool was changed from
tomcat to hikari.

This was a non planned change and in the past we had problems with
hikari, leading us to sticking to tomcat's
https://github.com/zalando/nakadi/commit/8557790b4b5d861f56a3c44e8efb6deac8365236

Since hikari was also not configured, we end up running out of
connections on staging so we need to teardown staging, release a few
connections and then deploy this fix.

This change restores tomcat pool.


rcillo:
👍

thomasabraham:
👍

PR #1216, opened 2020-07-08
Introducing ThreadPoolTaskExecutor for Async support

thomasabraham:
# Introducing ThreadPoolTaskExecutor for Async support

## Description
With spring boot upgrade, we get the following warning.

```
    !!!
    An Executor is required to handle java.util.concurrent.Callable return values.
    Please, configure a TaskExecutor in the MVC config under "async support".
    The SimpleAsyncTaskExecutor currently in use is not suitable under load.
    -------------------------------
    Request URI: '/subscriptions/a3e37570-817c-4b7d-9622-773695761d59/events'
    !!!
```
This commit is introducing a ThreadPoolTaskExecutor which is configured
to break under small load. The purpose is to observe the failure mode
before making a reasonable configuration for high load.

## Deployment Notes
This commit shall be used for validation, followed by proper configuration before deploying to production systems.


thomasabraham:
We Identified that the problem is not with SimpleAsyncTaskExecutor, and decided to ignore the warning for now.

PR #1219, opened 2020-07-14
Support for compact and delete cleanup policy

rcillo:
Since Kafka 0.10.1.0 log compaction and log deletion are no longer
multually exclusive cleanup policies. They can be used in a combined
way as described in the KIP used to design it
https://cwiki.apache.org/confluence/display/KAFKA/KIP-71:+Enable+log+compaction+and+deletion+to+co-exist

This PR adds support for such an option.

### DIFF
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/EventTypeController.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/EventTypeController.java
@@ -144,8 +144,7 @@ public class EventTypeController {
 
         final List<String> warnings = Lists.newArrayList(nakadiSettings.getWarnAllDataAccessMessage());
 
-        if (eventType.getCleanupPolicy().equals(CleanupPolicy.COMPACT) ||
-                eventType.getCleanupPolicy().equals(CleanupPolicy.COMPACT_AND_DELETE)) {
+        if (eventType.getCleanupPolicy().equals(CleanupPolicy.COMPACT)) {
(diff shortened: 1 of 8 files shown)


api-metastore/src/main/java/org/zalando/nakadi/controller/EventTypeController.java
```diff
@@ -144,7 +144,8 @@ private HttpHeaders generateWarningHeaders(final EventTypeBase eventType) {
 
         final List<String> warnings = Lists.newArrayList(nakadiSettings.getWarnAllDataAccessMessage());
 
-        if (eventType.getCleanupPolicy().equals(CleanupPolicy.COMPACT)) {
+        if (eventType.getCleanupPolicy().equals(CleanupPolicy.COMPACT) ||
+                eventType.getCleanupPolicy().equals(CleanupPolicy.COMPACT_AND_DELETE)) {
```
ferbncode, line 148: Do you think we can skip warning for `compact_and_delete` event-types? If I understand correctly, the data would be always deleted too for `compact,delete` event-types, following retention time policies.

Another approach here might be to just change the internal warn message.

ferbncode:
:+1:

api-metastore/src/main/java/org/zalando/nakadi/controller/EventTypeController.java
rcillo: true a667c1397

rcillo:
👍

ferbncode:
:+1:

PR #1220, opened 2020-07-17
Change the order of assertions to improve test stability

rcillo:
This tests has been failing too much.

I tried to improve it by changing the order of the assertions. The
purpose of this test is to verify that committing already committed
offset results in 200.

But since it commits offset 25, which does not exist, given the topic
is empty, this may cause the stream to close depending if the commit
arrives before or after the stream code reaches a certain stage.

I kept all the conditions (testing for 200 and 204) but in a way that
we don't have to commit inexistent offset 25. I run it dozens of times
and it doesnt break anymore.


ferbncode:
:+1:

rcillo:
👍

PR #1221, opened 2020-07-17
Fix failing test due to eventual consistency of default storage update

rcillo:
When updating the default storage one has to wait for the zk watcher
to be processed. This may take some time. In fact, we may have a bug
here when the watcher for some reason is missed (for example lost
connectivity to zk and recreation of watcher). But I'm not fixing
Nakadi, instead my goal is to first have more stable tests.

Also using teardown to avoid the failure of this particular test
cascade interfere with other tests which makes it harder to find the
"root cause" of the failures.

### DIFF
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/StorageControllerAT.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/StorageControllerAT.java
@@ -1,12 +1,18 @@
 package org.zalando.nakadi.webservice;
 
+import org.apache.http.HttpStatus;
+import org.junit.After;
 import org.junit.Assert;
 import org.junit.Test;
+import org.zalando.nakadi.domain.EventType;
 import org.zalando.nakadi.utils.EventTypeTestBuilder;
 import org.zalando.nakadi.webservice.utils.NakadiTestUtils;
 
 import static com.jayway.restassured.RestAssured.given;
 import static com.jayway.restassured.http.ContentType.JSON;
+import static org.junit.Assert.fail;
+import static org.zalando.nakadi.utils.TestUtils.waitFor;
+import static org.zalando.nakadi.webservice.utils.NakadiTestUtils.createEventType;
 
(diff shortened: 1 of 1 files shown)


acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/StorageControllerAT.java
```diff
@@ -21,18 +27,32 @@ public void shouldChangeDefaultStorageWhenRequested() throws Exception {
...
-        NakadiTestUtils.createEventTypeInNakadi(EventTypeTestBuilder.builder().name("event_b").build());
-        storageId = (String) NakadiTestUtils.listTimelines("event_b").get(0).get("storage_id");
-        Assert.assertEquals("default-test", storageId);
+        given().contentType(JSON).put("/storages/default/default-test")
+                .then()
+                .statusCode(HttpStatus.SC_OK);
 
-        // cleanup
+        // change of default storage is eventually consistent and relies on zookeeper watcher being processed
+        waitFor(() -> {
```
lmontrieux, line 43: I think this could wait forever?

rcillo: @lmontrieux it has a default value. I'm not sure if I have to annotate the test though, to make sure it waits at least for this amount of time `waitFor(runnable, 10000L, 500);`.

lmontrieux:
👍

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/StorageControllerAT.java
lmontrieux: Since it has a default value I guess it should be fine

lmontrieux: I imagine the value is at least a few seconds

rcillo:
👍

PR #1222, opened 2020-07-21
Documented deletion of log compacted event-type

thomasabraham:
# Documented deletion of log compacted event-type

> Zalando ticket : ARUHA-2982 

## Description
Endpoint to delete an event from a log compacted event-type is missing documentation. This is added in this PR.

## Review
- [ ] Tests
- [x] Documentation

### DIFF
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -946,12 +946,14 @@ paths:
       security:
         - oauth2: ['nakadi.event_stream.write']
       description: |
-        Once an event is published to a logcompacted event-type in nakadi, it
+        Once an event is published to a log compacted event-type in nakadi, it
         will stay there until another event with the same key is published.
-        In order to delete an event from a logcompacted event-type, it can be
+        In order to delete an event from a log compacted event-type, it can be
         posted to the deleted-events endpoint. The content of the event is not
         validated, but the information required to generate key and select
-        partition should be present.
+        partition should be present. On next log compaction, which could be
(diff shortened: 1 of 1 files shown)


docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -939,6 +939,71 @@ paths:
...
             $ref: '#/definitions/Problem'
 
+  '/event-types/{name}/deleted-events':
+    post:
+      tags:
+        - stream-api
+      security:
+        - oauth2: ['nakadi.event_stream.write']
+      description: |
+        Once an event is published to a logcompacted event-type in nakadi, it
```
rcillo, line 949: please add space as in `log compacted`

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -939,6 +939,71 @@ paths:
...
+        will stay there until another event with the same key is published.
+        In order to delete an event from a logcompacted event-type, it can be
+        posted to the deleted-events endpoint. The content of the event is not
+        validated, but the information required to generate key and select
+        partition should be present.
```
rcillo, line 954: I'm not sure it belongs to the API definition but users might be interested in knowing when their events are going to be actually deleted. From the docs it seems that right after publishing the events would be deleted but in fact there's a number of factors that are better described in the FAQ. Following the organization motto of being user driven, even if this is an open source project, I would consider linking to the FAQ entry.

rcillo:
👍

thomasabraham:
👍

PR #1224, opened 2020-07-23
Revert change to jakarta websocket api

rcillo:
We are going to wait for jetty upgrade and not force the upgrade of
one of its dependencies.


rcillo:
👍

antban:
:+1:

PR #1226, opened 2020-08-04
Allow event type admins and nakadi admins to edit schema compatibility

rcillo:
Users often want to make incompatible changes to schema by removing
required properties or any other sort of change that has the potential
of causing consumers to break.

Often they are the sole consumer of that data or they have already
aligned with downstream service owners that such change will be
handled gracefully.

They have the confidence that making such changes is the least
disruptive way of delivering something and want to do it.

Our team ends up acting a bottleneck.

This change grants them with such freedom to change.


lmontrieux:
👍

rcillo:
👍

PR #1228, opened 2020-08-06
ARUHA-2225: Publish to Audit log and Event Type log for POST /event-types/{et}/schemas

ferbncode:
# One-line summary

> Zalando ticket : ARUHA-2225 

The new endpoint for addition of a schema (`POST /event-types/{et}/schemas`) to an event-type doesn't publish to audit log or the event-type log event-types. The changes ensure that in case if the new schema version is different from the previous, an event is published to the event-type log and audit log event types


thomasabraham:
👍

ferbncode:
:+1:

PR #1230, opened 2020-08-11
Upgrade Kafka Client to 2.6.0

rcillo:
Nakadi runs on ancient Kafka client 2.3.1 released years ago. Kafka is
preparing to replace Zookeeper by its own Raft based consensus
protocol and as part of this major step (expected to come with kafka 3
sometime in the future) they started to remove direct dependencies to
zookeeper, as for example topic creation, configuration and deletion.

The version of client that we use has an Admin that connected
directly to ZK which is expected to stop to work sometime in the
future and worse than that exposed some internals from Kafka which
were never mean to be exposed.

The new admin client uses a new API added to kafka brokers and is
going to become the default one. It's also more powerful in the sense
that it allows to change broker configurations on the fly (even though
we don't need it for now).

This also removed scala 2.12 dependency which has a couple of CVEs.

Besides these Zookeeper related changes, the bump in version should
address several bug fixes, performance issues and unlock some more
features on the client.

On top of it all, by using spring to manage kafka client version, we
are more likely to upgrade it in the future, since spring manages its
version for us.


adyach:
👍

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -273,15 +273,15 @@ public void repartition(final String topic, final int partitionsNumber) throws C
...
 
         final KafkaTopicConfig kafkaTopicConfig = kafkaTopicConfigFactory.createKafkaTopicConfig(nakadiTopicConfig);
-        try (KafkaZkClient client = createZkClient()) {
-            final AdminZkClient adminZkClient = new AdminZkClient(client);
-            adminZkClient.createTopic(
-                    kafkaTopicConfig.getTopicName(),
-                    kafkaTopicConfig.getPartitionCount(),
-                    kafkaTopicConfig.getReplicaFactor(),
-                    kafkaTopicConfigFactory.createKafkaTopicLevelProperties(kafkaTopicConfig),
-                    kafkaTopicConfig.getRackAwareMode()
```
adyach, line 283: this property is not set, how does that work ?

rcillo: If I understood it right (and I will test) this is also a broker property. So we don't need to specify it on the client. The new client does not have an algorithm for assigning partitions. It just tells the broker the number of partitions and the brokers does it. It does have an API for explicitly assigning partitions. If my assumption is wrong, I'll have to write some code that loads the brokers and the AZs and generates the assignment. But I think this is unnecessary. I couldn't find docs about it though so I have to test.

adyach:
👍

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
rcillo: @adyach I confirm that on the broker side there is logic for rack awareness and it uses it in case all brokers have rack's defined (which we do).

Kafka algorithm https://github.com/apache/kafka/blob/3e6dcb14dd9716f07d1ba96b1b06b78e55e92de1/core/src/main/scala/kafka/admin/AdminUtils.scala#L31-L123

Rack config on broker side https://github.com/zalando-nakadi/bubuku/blob/30086c0fb97d3e4d684ef2e34b135f1eaef5de83/bubuku/daemon.py#L52-L54

rcillo:
@adyach I have one self criticism to make here: we do not use any of the features from spring kafka and it's only used here to control the version of the client. I think maybe I should just go straight for the client and use the latest version with all the bug fixes and features. I'll change the pr once more, sorry.

rcillo:
@adyach https://github.com/zalando/nakadi/pull/1230/commits/3d6e8e42f5b23cc79ceb37cb6b5b3554048f1a57

adyach:
👍

adyach:
👍

rcillo:
👍

rcillo:
👍

adyach:
👍

PR #1233, opened 2020-08-27
Use token based instead of offset based approach for subscriptions listing

antban:
In case if there are a lot of subscriptions in the database, using offset
based approach for pagination is not really something that works at scale.
Therefore - there is an attempt to replace offset approach with token
approach, that is much more widely adopted and may actually solve the
problem with db overusage.

Token in this case is a string, representing subscription_id and direction
of iteration.
In case if this approach will be accepted by the nakadi team - tests
will be written in order to actually test the behavior (there is a bug
right now with backward iteration to the first page in case if
subscription was removed during iteration, but I am on vacation and we
are treating it as external contribution)


core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -25,11 +25,15 @@
 import org.zalando.nakadi.exceptions.runtime.ServiceTemporarilyUnavailableException;
 import org.zalando.nakadi.util.HashGenerator;
 import org.zalando.nakadi.util.UUIDGenerator;
+import sun.java2d.pipe.DrawImage;
```
thomasabraham, line 28: Interesting

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -121,8 +125,88 @@ public void deleteSubscription(final String id)
...
+            return "Token{" +
+                    "subscriptionId='" + subscriptionId + '\'' +
+                    ", tokenType=" + tokenType +
+                    '}';
+        }
+    }
+
     public List<Subscription> listSubscriptions(final Set<String> eventTypes, final Optional<String> owningApplication,
-                                                final int offset, final int limit)
+                                                final int offset, final int limit, @Nullable final Token token)
```
thomasabraham, line 209: This function is very long and complex by now. 
It will be nice if we can extract a `QueryBuilder` class with the `StringBuilder queryBuilder`, `List<String> clauses`, and `List<Object> params`.

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -142,16 +226,43 @@ public void deleteSubscription(final String id)
...
-
-        queryBuilder.append(" ORDER BY s_subscription_object->>'created_at' DESC LIMIT ? OFFSET ? ");
-        params.add(limit);
-        params.add(offset);
+        queryBuilder.append(order);
         try {
-            return jdbcTemplate.query(queryBuilder.toString(), params.toArray(), rowMapper);
+            final List<Subscription> result = jdbcTemplate.query(queryBuilder.toString(), params.toArray(), rowMapper);
+            if (null != token && token.getTokenType() != TokenType.FORWARD) {
+                Collections.reverse(result);
```
thomasabraham, line 263: Why reverse when there is ` ORDER BY s_id DESC LIMIT ?`

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
```diff
@@ -0,0 +1,97 @@
...
+import org.zalando.nakadi.domain.Subscription;
+import org.zalando.nakadi.repository.db.SubscriptionDbRepository;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+public class SubscriptionsUriHelper {
+
+    public static PaginationLinks createSubscriptionPaginationLinks(
```
thomasabraham, line 14: I prefer this functionality to be placed in `core-metastore` module along with `SubscriptionDbRepository`.
The reason being the logic in here and the token based query generation in `SubscriptionDbRepository` are working together to provide pagination.

thomasabraham: Also this function is too big to understand in one go.
It will be nice we we can extract `buildPrevLink` and `buildNextLink`.

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
antban: oops ((

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
antban: agree, changed to another approach

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
antban: I rewrote code in a way that it is specifically written why

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
antban: Changed it back. 
But I would not move the code back, as it is totally wrong. Pagination link is part of view logic, and should not be located in `core-something` project.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1142,12 +1142,22 @@ paths:
...
+          description: page offset. Should be deprecated as may generate a lot of load for database.
           type: integer
           format: int64
           required: false
           default: 0
           minimum: 0
+        - name: token
+          in: query
+          description: |
+            Replacement for offset. Used as a base for detecting from where and in which direction to iterate over
```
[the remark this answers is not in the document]
antban: The point is that it should not be created and people should only use links from response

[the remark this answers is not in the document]
thomasabraham: I realized that the user should not be creating it only after I read the implementation.
It will be nice to start the description like 

> This is present in `next`/`prev` links in the response, and should not be generated by client.  Also it is not required for to get the very first request.

Maybe we can improve this in the next stage where there is no confusing "new" option.

thomasabraham:
👍

antban:
:+1:

thomasabraham:
👍

PR #1234, opened 2020-08-28
Default statistics backward compatible change

rcillo:
In the PR where we changed the repartitioning endpoint we also
accidentally introduced a change in an undocumented behaviour that
some clients relied on. It used to be possible to update an event type
without providing default statistics.

We believe this behaviour should not be modified because this field is
optional, so some clients might not even support it (Nakadi UI, for
example).

We are also fixing a problem where repartition could be triggered
concurrently with an event type update resulting in some race
condition on database and timeline changes in database.


core-services/src/main/java/org/zalando/nakadi/service/RepartitioningService.java
```diff
@@ -89,18 +99,44 @@ public void repartition(final EventType eventType, final int partitions)
...
+            Thread.currentThread().interrupt();
+            LOG.error("Failed to wait for timeline switch", e);
+            throw new EventTypeUnavailableException("Event type " + eventType.getName()
+                    + " is currently in maintenance, please repeat request");
+        } catch (final TimeoutException e) {
+            LOG.error("Failed to wait for timeline switch", e);
+            throw new EventTypeUnavailableException("Event type " + eventType.getName()
+                    + " is currently in maintenance, please repeat request");
+        } catch (final InternalNakadiException | ServiceTemporarilyUnavailableException e) {
+            LOG.error("Error deleting event type " + eventType.getName(), e);
```
adyach, line 130: do we need to process this execeptions here? if so could you adjust the message please?

rcillo: fixed 7b017fe0a

adyach:
👍

rcillo:
:+1:

adyach:
👍

PR #1235, opened 2020-09-18
Change cleanup policy from delete to compact_and_delete

rcillo:
This commit is just a work in progress. It allows to change the
cleanup policty from delete to compact_and_delete.

On the way to implement this, I had to read some code that I could not
avoid changing. But I'm not sure it's looking better either.

I removed one unused feature toggle.

We need to figure out what to do with retention time that is null or
different from the actual config in Kafka.


rcillo:
@ferbncode can we add one test case changing cleanup policy from delete to compact_and_delete, just to exercise the whole code? I missed that in my earlier commits.

ferbncode:
@rcillo Added an acceptance test for journey of changing the cleanup policy from `delete` to `compact_and_delete` in https://github.com/zalando/nakadi/pull/1235/commits/3742b078e5f8686400c96aacad0f23c740c9c41d, thanks! :)

rcillo:
:+1:

ferbncode:
:+1:

PR #1236, opened 2020-10-01
validate control characters while parsing events

adyach:
Added support for control characters in custom implementation of JSON parser. The commit provides validation of unicode control characters and throws exception whenever it can not parse incoming events. Accroding to JSON spec[1] only the following characters should be escaped: from '\u0000' to '\u001F'.

[1]: https://www.ietf.org/rfc/rfc4627.txt, chapter 2.5. Strings


antban:
:+1:

adyach:
👍

PR #1238, opened 2020-10-23
Adding published_by to metadata

thomasabraham:
# Adding published_by to metadata

> Zalando ticket : ARUHA-2896

## Description
Adding `published_by` field to metadata with the id of the publisher; 
the subject that published this specific event to Nakadi.

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes

### DIFF
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -2251,6 +2251,12 @@ definitions:
             Event identifier of the Event that caused the generation of this Event.
             Set by the producer.
           example: '105a76d8-db49-4144-ace7-e683e8f4ba46'
+      published_by:
+        description: |
+          ID of the subject that published this specific event to Nakadi.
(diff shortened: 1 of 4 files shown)


docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2251,6 +2251,10 @@ definitions:
             Event identifier of the Event that caused the generation of this Event.
             Set by the producer.
           example: '105a76d8-db49-4144-ace7-e683e8f4ba46'
+      published_by:
```
rcillo, line 2254: Maybe it's worth mentioning that this field is automatically enriched by nakadi and if provided by the user it will be overwritten.

thomasabraham: I agree, I added a note for this.

rcillo:
:+1:

thomasabraham:
👍

PR #1240, opened 2020-10-26
Aruha 2896 bugfix: adding metadata.published_by to effective schema

thomasabraham:
# bugfix: adding metadata.published_by to effective schema

> Zalando ticket : ARUHA-2896

## Description
Missing `metadata.published_by` in effective schema blocks clients that forward an event from one topic to another when Nakadi started enriching metadata with published_by. Adding the `published_by` to the schema of metadata should unblock such clients.

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes

### DIFF
--- a/core-services/src/main/resources/schema_metadata.json
+++ b/core-services/src/main/resources/schema_metadata.json
@@ -34,6 +34,9 @@
       },
       "flow_id": {
         "type": "string"
+      },
+      "published_by": {
+        "type": "string"
       }
     },
     "required": [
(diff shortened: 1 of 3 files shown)


core-services/src/main/resources/schema_metadata.json
```diff
@@ -34,6 +34,9 @@
       },
       "flow_id": {
         "type": "string"
+      },
+      "published_by": {
+        "type": "string"
```
ferbncode, line 39: I think it should be also included for `compact` and `compact_and_delete`

thomasabraham: done

antban:
:-1: 
The value is overriden in deployment repo

rcillo:
I believe you should still have merged this. Now no one can run an open source project without our private repo.

PR #1241, opened 2020-11-05
Using different image name for PR build

thomasabraham:
# Using different image name for PR build

> Zalando ticket : ARUHA-3086

## Description
Adding `-pr` to the image name of the docker images built from a pull request to identify them easily.

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes

### DIFF
--- a/delivery.yaml
+++ b/delivery.yaml
@@ -30,5 +30,8 @@ pipeline:
         cmd: |
           ./gradlew clean app:bootJar
           IMAGE="registry-write.opensource.zalan.do/aruha/nakadi-oss:${CDP_BUILD_VERSION}"
+          if [ "$CDP_PULL_REQUEST_NUMBER" ]; then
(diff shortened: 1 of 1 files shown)


delivery.yaml
```diff
@@ -30,5 +30,8 @@ pipeline:
         cmd: |
           ./gradlew clean app:bootJar
           IMAGE="registry-write.opensource.zalan.do/aruha/nakadi-oss:${CDP_BUILD_VERSION}"
```
rcillo, line 32: Maybe you always want to build "pr" version when it's merged to master because in the deployment of staging I believe you'll always deploy from pr image and you may want to deploy a "pr" image that is just like the "master" image but with a different name.

rcillo: you might want to name it "test" for this reason, because it's not always "pr" since it can be built from master but still be named "test" just for the sake of simplicity in the deployment part of the pipeline.

thomasabraham:
👍

rcillo:
:+1:

PR #1244, opened 2020-11-25
Remove zk watchers when streaming connection is closed

adyach:
Apache Zookeeper 3.5 brought possibility to remove watchers, which was not possible in the previous versions.
Removal of the watchers helps to avoid growing number of zk watchers on server and avoid OOM otherwise zk jvm process heap is saturated with watchers objects.

In regards to listeners removal - the result of test that creates 1 subscription consuming connection:
```
[2020-12-01 14:50:40.206Z] Listener 1477800410 for key /nakadi/subscriptions/7aebc8e9-b15b-4b86-93e6-0fcbaa05d426/topology registered
[2020-12-01 14:50:40.213Z] Listener 866834755 for key /nakadi/subscriptions/7aebc8e9-b15b-4b86-93e6-0fcbaa05d426/sessions registered
[2020-12-01 14:50:40.232Z] Listener 1263713668 for key /nakadi/subscriptions/7aebc8e9-b15b-4b86-93e6-0fcbaa05d426/offsets/XaEctfvZWoAvvwZEoDZqKurEkAIOLulnIbV.nGCeWWrufxLRJQVWvUPCMzvmOcby/0 registered
[2020-12-01 14:50:40.258Z] Failed to delete subscription 1477800410 for key /nakadi/subscriptions/7aebc8e9-b15b-4b86-93e6-0fcbaa05d426/topology, reason: null
[2020-12-01 14:50:40.258Z] Listener 943361685 for key /nakadi/subscriptions/7aebc8e9-b15b-4b86-93e6-0fcbaa05d426/topology registered
[2020-12-01 14:51:40.401Z] Successfully deleted subscription 943361685 for key /nakadi/subscriptions/7aebc8e9-b15b-4b86-93e6-0fcbaa05d426/topology
[2020-12-01 14:51:40.402Z] Successfully deleted subscription 1263713668 for key /nakadi/subscriptions/7aebc8e9-b15b-4b86-93e6-0fcbaa05d426/offsets/XaEctfvZWoAvvwZEoDZqKurEkAIOLulnIbV.nGCeWWrufxLRJQVWvUPCMzvmOcby/0
[2020-12-01 14:51:40.426Z] Successfully deleted subscription 866834755 for key /nakadi/subscriptions/7aebc8e9-b15b-4b86-93e6-0fcbaa05d426/sessions
```
It seems that it can't delete watcher on zk side, as zk already triggered watcher. Therefore there is 1 failure. But in general it works.


nakadi.sh
```diff
@@ -2,7 +2,7 @@
 
 function waitForNakadi() {
   echo "Waiting for Nakadi to start up"
-  while True; do
+  while :; do
```
adyach, line 5: does not look like it is related

a1exsh: That's true.  I've just spotted it in the log with the failing tests and was wondering if it could be somehow related.  It wasn't.

acceptance-test/build.gradle
```diff
@@ -37,7 +37,7 @@ dependencies {
     ext {
         dropwizardVersion = '3.1.3'
         curatorVersion = '4.2.0'
```
adyach, line 39: for note: the upgrade to 5.0 is not planned due to braking changes https://curator.apache.org/breaking-changes.html. the upgrade of curator library will be made as separate task.

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionImpl.java
```diff
@@ -68,6 +68,9 @@ public ReturnType getData() throws NakadiRuntimeException {
     @Override
     public void close() {
         listener = null;
+        if (null != curatorFramework && !curatorFramework.isZk34CompatibilityMode()) {
```
adyach, line 71: I am not sure I understand why `curatorFramework.isZk34CompatibilityMode()` used.
the doc says:
```
If mode is true, create a ZooKeeper 3.4.x compatible client. IMPORTANT: If the client
library used is ZooKeeper 3.4.x <code>zk34CompatibilityMode</code> is enabled by default.
```
I do not think we create curator framework instance in compatible mode and we use 3.6 zookeeper.

a1exsh: It is safer to check it explicitly, because the `remove()` call on the next line will throw an `IllegalStateException` if in compatibility mode.

If somebody starts curator in this mode (and it's as easy as reverting the zookeeper dependency version: no actual _code_ changes), this will start throwing exceptions.

adyach: in this case tests will show it, but we can also remove it once migrated to CF5.0.
thanks

a1exsh: Yes, but the trouble is that this is only caught by the acceptance test, and the problem is exposed indirectly: quite a number of tests fail with unexpected results and the root cause needs to be figured out from the server logs.

I see the following options:
- Have a test that checks this condition explicitly and remove it from production code.
- Make AT failure reason more apparent (see also next option).
- Check it at server start and refuse to run in compatibility mode altogether.

adyach: I am fine to have this check, because we will eventually remove it due to migration to curator framework >= 5.x, which does not have such method.

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionImpl.java
```diff
@@ -68,6 +68,9 @@ public ReturnType getData() throws NakadiRuntimeException {
...
+            curatorFramework.watches().remove(this);
```
antban, line 72: @adyach `.forPath(key)` should be added, or it won't execute anything

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionImpl.java
antban: @adyach `!= null` check should be removed (as it spoils all the attempts to use `@Nullable` annotations. Field is not nullable, therefore the check is not needed, and adds confusion

antban: I still do not understand who is the author of PR @a1exsh or @adyach

a1exsh: It was mostly me.  Does it matter? :)

antban: Regarding `!= null` - matters, regarding author - it's just hard to follow the discussion

a1exsh: Actually, it depends how you count.  The one liner code change was suggested by Andrey, and then we were trying to make it pass the tests ;-)

antban:
:+1:

a1exsh:
:+1:

PR #1248, opened 2020-11-26
Building and running on Java 11

rcillo:
https://jira.zalando.net/browse/ARUHA-3111

This PR makes the least changes necessary to build and run Nakadi on
Java 11.

- uses approved base image
https://cloud.docs.zalando.net/howtos/compliant-images/#use-of-allowed-docker-base-images
- upgraded gradle to 6.7.1 which introduced `toolchain`
https://docs.gradle.org/current/userguide/toolchains.html which helps
us make sure that the same java version is used everywhere
- included an extra dependency (unfortunatelly) to fix springboot
dependency which is no longer part of default java module
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-with-Java-9-and-above

This way the app builds and passes acceptance tests. I'm not sure this
is production ready yet as there could be changes in GC, for
example. Some work is still required.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.opensource.zalan.do/stups/openjdk:1.8.0-191-19
+FROM registry.opensource.zalan.do/library/openjdk-11-jre-slim:11-jre-slim-20201005
 
 MAINTAINER Team Aruha, team-aruha@zalando.de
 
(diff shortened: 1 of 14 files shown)


rcillo:
To get to a production ready state, I'm following this upgrade guide https://jvm.docs.zalando.net/java11_migration/

But it's not clear if we need to change everything in the guide at once.

a1exsh:
:+1:

a1exsh:
:+1:

thomasabraham:
👍

PR #1249, opened 2020-11-30
Remove watchers on close of zk subscription

antban:
# One-line summary

Clean up subscription watchers for subscriptions


antban:
closed, as there is already #1244

PR #1251, opened 2020-12-15
Removed exhibitor support, upgrading curator to 5.1.0

thomasabraham:
# Removed exhibitor support, upgrading curator to 5.1.0

> Zalando ticket : ARUHA-3237

## Description
Incompatible versions of zookeeper and curator causes some runtime exceptions (`java.lang.NoSuchFieldError: addr`), which are fixed by upgrading curator to latest compatible version. Exhibitor support had to be removed when curator is upgraded.

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes

### DIFF
--- a/acceptance-test/build.gradle
+++ b/acceptance-test/build.gradle
@@ -36,7 +36,7 @@ configurations {
 dependencies {
     ext {
         dropwizardVersion = '3.1.3'
-        curatorVersion = '4.2.0'
+        curatorVersion = '5.1.0'
         zookeeperVersion = '3.6.1'
         jacksonVersion = '2.9.8'
     }
--- a/acceptance-test/build.gradle
+++ b/acceptance-test/build.gradle
@@ -34,12 +34,6 @@ configurations {
 }
 
 dependencies {
-    ext {
-        dropwizardVersion = '3.1.3'
-        curatorVersion = '5.1.0'
-        zookeeperVersion = '3.6.1'
-        jacksonVersion = '2.9.8'
-    }
     // Override spring-boot BOM versions
     ext['json.version'] = '20180130'
     ext['json-path'] = '2.4.0'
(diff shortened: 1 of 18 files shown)


acceptance-test/build.gradle
```diff
@@ -36,7 +36,7 @@ configurations {
 dependencies {
     ext {
         dropwizardVersion = '3.1.3'
-        curatorVersion = '4.2.0'
+        curatorVersion = '5.1.0'
```
antban, line 39: It makes sense to use the same variable in root `build.gradle`, like `kafkaClientVersion`

core-common/src/main/java/org/zalando/nakadi/repository/zookeeper/ZooKeeperHolder.java
```diff
@@ -112,26 +107,7 @@ private CuratorFramework createCuratorFramework(final int sessionTimeoutMs,
...
-                        EXHIBITOR_POLLING_MS,
-                        new ExponentialBackoffRetry(EXHIBITOR_RETRY_TIME, EXHIBITOR_RETRY_MAX)) {
-                    @Override
-                    public String getConnectionString() {
-                        return super.getConnectionString() + conn.getPathPrepared();
-                    }
-                };
-                result.pollForInitialEnsemble();
-                return result;
+                throw new RuntimeException("Exhibitor support is removed on 2020-12-15; use zookeeper connection");
```
antban, line 110: I'm a bit concerned about this. In order not to flood - let's just remove this enum constant at all?

antban (changes requested):
Let's stop modifying all the files, and move curator version to parent build.gradle and remove all the core related to obsolete enum value

acceptance-test/build.gradle
thomasabraham: 👍

thomasabraham:
👍

antban:
:+1:

PR #1252, opened 2020-12-15
Use more optimal way of publishing events to internal event types, to reduce the chance of overload.

antban:
There is inoptimal thread usage for kpi and supplementary message publishing.
One or several threads may actually block whole the processing flow.
In order to avoid it intermediate batcher thread was introduced, with the only purpose of batching requests.
In case if publishing is stuck, assembled batches are dropped with error message.

### DIFF
--- a/core-services/src/main/java/org/zalando/nakadi/service/publishing/EventsProcessor.java
+++ b/core-services/src/main/java/org/zalando/nakadi/service/publishing/EventsProcessor.java
@@ -133,9 +133,17 @@ public class EventsProcessor {
         executorService.submit(r);
     }
 
+    /**
+     * Generates batches from {@link #eventsQueue} single events with the following constraints:
+     * <ul>
+     *     <li>Batch size is not more than {@link #maxBatchSize} events each</li>
+     *     <li>Eatch batch is assembled for at most {@link #batchCollectionTimeout} ms (or close to it)</li>
+     * </ul>
+     */
     private void dispatch() {
         final Map<String, BatchedRequest> batchesBeingAssembled = new HashMap<>();
         try {
+            // The time here is a moment in time at which batches should be evaluated - if they should be sent or not
             long nextTimeCheck = System.currentTimeMillis() + batchCollectionTimeout;
             while (true) {
                 long currentTime = System.currentTimeMillis();
@@ -143,12 +151,14 @@ public class EventsProcessor {
                         Math.max(nextTimeCheck - currentTime, 1), TimeUnit.MILLISECONDS);
                 currentTime = System.currentTimeMillis();
                 boolean batchWasSent = false;
+                // In case if data was taken - add it.
                 if (data != null) {
                     BatchedRequest batch = batchesBeingAssembled.get(data.eventType);
                     if (null == batch) {
                         batch = new BatchedRequest(data.eventType, currentTime + batchCollectionTimeout);
                         batchesBeingAssembled.put(data.eventType, batch);
                     }
+                    // In case if batch size is crossing maxBatchSize - send batch
                     if (batch.add(data.object) >= maxBatchSize) {
                         scheduleSendBatchedRequest(batch);
                         batchesBeingAssembled.remove(batch.eventType);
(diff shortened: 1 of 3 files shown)


core-services/src/main/java/org/zalando/nakadi/service/publishing/EventsProcessor.java
```diff
@@ -10,84 +10,183 @@
...
+    }
+
+    private void dispatch() {
+        final Map<String, BatchedRequest> batchesBeingAssembled = new HashMap<>();
+        try {
+            long nextTimeCheck = System.currentTimeMillis() + batchCollectionTimeout;
+            while (true) {
+                long currentTime = System.currentTimeMillis();
+                final EventToPublish data = eventsQueue.poll(
+                        Math.max(nextTimeCheck - currentTime, 1), TimeUnit.MILLISECONDS);
```
adyach, line 143: more effective would be not to wait even 1ms.

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventsProcessor.java
```diff
@@ -10,84 +10,183 @@
...
+                if (data != null) {
+                    BatchedRequest batch = batchesBeingAssembled.get(data.eventType);
+                    if (null == batch) {
+                        batch = new BatchedRequest(data.eventType, currentTime + batchCollectionTimeout);
+                        batchesBeingAssembled.put(data.eventType, batch);
+                    }
+                    if (batch.add(data.object) >= maxBatchSize) {
+                        scheduleSendBatchedRequest(batch);
+                        batchesBeingAssembled.remove(batch.eventType);
+                        currentTime = System.currentTimeMillis();
```
adyach, line 155: since it is current time, probably, should be the last operation in this if clause

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventsProcessor.java
```diff
@@ -10,84 +10,183 @@
...
+                        batchWasSent = true;
+                    }
+                }
+                if (batchWasSent || currentTime > nextTimeCheck) {
+                    nextTimeCheck = currentTime + batchCollectionTimeout;
+                    for (final Map.Entry<String, BatchedRequest> entry : batchesBeingAssembled.entrySet()) {
+                        if (entry.getValue().finishCollectionAt < currentTime) {
+                            scheduleSendBatchedRequest(entry.getValue());
+                            batchesBeingAssembled.remove(entry.getKey());
+                        } else if (nextTimeCheck > entry.getValue().finishCollectionAt) {
```
adyach, line 165: i do not understand the check here. `nextTimeCheck` is aways greater than `finishCollectionAt` except the case you checked on line 162, because it is calculated as `nextTimeCheck = currentTime + batchCollectionTimeout`.

antban:
:+1:

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventsProcessor.java
```diff
@@ -10,84 +10,203 @@
...
+            }
+        };
+        executorService.submit(r);
+    }
+
+    /**
+     * Generates batches from {@link #eventsQueue} single events with the following constraints:
+     * <ul>
+     *     <li>Batch size is not more than {@link #maxBatchSize} events each</li>
+     *     <li>Eatch batch is assembled for at most {@link #batchCollectionTimeout} ms (or close to it)</li>
```
adyach, line 140:
```suggestion
     *     <li>Each batch is assembled for at most {@link #batchCollectionTimeout} ms (or close to it)</li>
```

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventsProcessor.java
antban: discussed offline

adyach:
👍

antban:
:+1:

PR #1253, opened 2020-12-16
Fix watchers processing in nakadi default storage listener

antban:
It is not allowed to immediately react on notifications from zookeeper, as it blocks it's address updates, and when session is expired or connection lost - zookeeper becomes unusable.


thomasabraham:
👍

antban:
:+1:

PR #1257, opened 2021-01-25
A lease is not removed on connection loss

adyach:
# One-line summary
During ConnectionLoss event under certain conditions (unknown yet)
Curator does not clean acquired leases, which makes it impossible
for clients to acquire a lease.
New curator lock instance is intended to avoid such issue by
closing Zookeeper session, which will closure of associated
ephemeral znodes like leases.


a1exsh:
:+1:

adyach:
👍

PR #1259, opened 2021-02-09
Bugfix: using expireAfterWrite to avoid caching forever

thomasabraham:
# Fixing cursor commit after authorization section update.

> Zalando ticket : team-aruha/issues/470

## Description
When a subscription is created without the read permissions for the client, client is not able commit cursors as expected. 
But the cursor commit is forbidden even after the permission is granted.

### Cause
- The subscription object is fetched using SubscriptionCache.getSubscription method.
- The SubscriptionCache uses a guava LoadingCache with expireAfterAccess configured.
  - The default value for expireAfterAccess is 300000ms, that is 5min.
- This means that the value is evicted from cache if it is not accessed for 5min.
  - This also means that the value can stay in cache forever, if there is no 5min window without any access.
     - The bug is probably facing this scenario as the client will keep retrying to commit.
- The above scenario alone cannot be the cause of the bug as the cache is invalidated explicitly as well.
  - The SubscriptionCache.invalidateSubscription invalidates the cache for a given subscription id.
- The invalidateSubscription method is called from only one place, CursorsService.validateStreamId.
  - And the validateStreamId is called only from CursorsService.commitCursor, after successful authorization!.
  - And the invalidateSubscription is called when streamId is not an active session in the Zookeeper.
  - This mean cache invalidation is not happening either.

### Fix
Replacing `expireAfterAccess` with `expireAfterWrite` to evict subscriptions from cache after 5 minutes.

### DIFF
--- a/core-metastore/src/main/java/org/zalando/nakadi/cache/SubscriptionCache.java
+++ b/core-metastore/src/main/java/org/zalando/nakadi/cache/SubscriptionCache.java
@@ -27,9 +27,9 @@ public class SubscriptionCache {
     @Autowired
     public SubscriptionCache(
             final SubscriptionDbRepository subscriptionRepository,
-            @Value("${nakadi.cache.subscription.expireAfterAccessMs:300000}") final long expireAfterAccessMs) {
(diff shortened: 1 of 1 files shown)


adyach:
👍

thomasabraham:
👍

PR #1260, opened 2021-02-09
Respond with status 400 for invalid requests handled by spring security firewall

burimshala:
# Spring firewall returns 500 when RequestRejectedException is thrown. The correct status code is 400.
This is going to be addressed by https://github.com/spring-projects/spring-security/issues/7568

> Zalando ticket : team-aruha/521

## Review
- [x] Implementation
- [x] Tests

### DIFF
--- a/app/src/main/java/org/zalando/nakadi/filters/RequestRejectedFilter.java
+++ b/app/src/main/java/org/zalando/nakadi/filters/RequestRejectedFilter.java
@@ -1,23 +1,30 @@
 package org.zalando.nakadi.filters;
 
-import org.aspectj.lang.ProceedingJoinPoint;
-import org.aspectj.lang.annotation.Around;
-import org.aspectj.lang.annotation.Aspect;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
 import org.springframework.security.web.firewall.RequestRejectedException;
 import org.springframework.stereotype.Component;
+import org.springframework.web.filter.GenericFilterBean;
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
 
-@Aspect
+
 @Component
-public class RequestRejectedFilter {
+@Order(Ordered.HIGHEST_PRECEDENCE)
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/zalando/nakadi/filters/RequestRejectedFilter.java
```diff
@@ -0,0 +1,23 @@
+package org.zalando.nakadi.filters;
+
+import org.aspectj.lang.ProceedingJoinPoint;
+import org.aspectj.lang.annotation.Around;
+import org.aspectj.lang.annotation.Aspect;
+import org.springframework.security.web.firewall.RequestRejectedException;
+import org.springframework.stereotype.Component;
+import javax.servlet.http.HttpServletResponse;
+
+@Aspect
```
adyach, line 10: i am not against aspects, but i am not really comfortable with it and it is not used in the project. maybe we could try to avoid using it.

burimshala: @adyach I have addressed your comment. I kept the same approach by intercepting doFilter method, but now the implementation is a Bean.

app/src/main/java/org/zalando/nakadi/filters/RequestRejectedFilter.java
```diff
@@ -0,0 +1,30 @@
...
+public class RequestRejectedFilter extends GenericFilterBean {
+
+    @Override
+    public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {
+        try {
+            chain.doFilter(req, res);
+        } catch (RequestRejectedException e) {
+            HttpServletResponse response = (HttpServletResponse) res;
+
+            response.sendError(HttpServletResponse.SC_NOT_FOUND);
```
a1exsh, line 27: Should it be `SC_BAD_REQUEST`?

burimshala: Fixed it

app/src/main/java/org/zalando/nakadi/filters/RequestRejectedFilter.java
adyach: @burimshala nice ! i like it :)

adyach:
👍

a1exsh:
:+1:

PR #1262, opened 2021-02-15
Use latest base docker image

a1exsh:
That makes it easier to stay on top of the latest security release by just re-triggering the build pipeline.


adyach:
👍

a1exsh:
:+1:

PR #1263, opened 2021-02-17
part I: rotate curator client on locks

adyach:
It was identified that apache curator and apache zookeeper clients does not have api to remove a single watcher from ther apache zookeeper server. the possibility either to remove all watchers from znode or remove local client watchers. such behaviour makes it hard to operator zookeeper because a lot of wathcers accumulated on server side until client session is closed, threats are: a) zookeeper server will not be able to move watcher to another server while current one went down b)watchers diestributed not evenly among zookeeper servers which disbalance resource usage such as cpu.

This commit brings ability to rotate apache curator client. Rotating framework keeps at most 2 alive clients, but in most of the time it should have only 1 alive client. The client is rotated by time. Once rotated it will be recycled after there is no usage of it.


core-common/src/main/java/org/zalando/nakadi/repository/zookeeper/RotatingCuratorFramework.java
```diff
@@ -0,0 +1,138 @@
...
+                activeCurator.counter = new AtomicLong();
+                retiringCurator.curator = tmpCurator;
+                retiringCurator.counter = tmpCounter;
+            } finally {
+                lock.writeLock().unlock();
+            }
+            activeCuratorRotatedAt = System.currentTimeMillis();
+            LOG.info("curator client rotated, current usage {}",
+                    retiringCurator.counter);
+        }
```
a1exsh, line 109: Let's add an `else` here and log the refcount of retiring curator just in case.

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/lock/CuratorNakadiLock.java
```diff
@@ -0,0 +1,73 @@
...
+            LOG.error("error while acquiring semaphore mutex lock", e);
+            return false;
+        }
+    }
+
+    @Override
+    public void unlock() {
+        try {
+            lock.release();
+            rotatingCuratorFramework.returnCuratorFramework(
```
a1exsh, line 67: Let's move this to `finally`.

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/lock/CuratorNakadiLock.java
```diff
@@ -0,0 +1,73 @@
...
+        this.currentCuratorFramework = null;
+    }
+
+    @Override
+    public boolean lock() {
+        try {
+            final CuratorFramework curatorFramework =
+                    rotatingCuratorFramework.takeCuratorFramework();
+            if (currentCuratorFramework != curatorFramework) {
+                lock = new InterProcessSemaphoreMutex(
```
a1exsh, line 44: What if this throws?

core-common/src/test/java/org/zalando/nakadi/repository/zookeeper/RotatingCuratorFrameworkTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+    @Test
+    public void testShouldReplaceCurator() throws InterruptedException {
+        final ZooKeeperHolder zkh = Mockito.mock(ZooKeeperHolder.class);
+        final CuratorFramework cf1 = Mockito.mock(CuratorFramework.class);
+        final CuratorFramework cf2 = Mockito.mock(CuratorFramework.class);
+
+        Mockito.when(zkh.newCuratorFramework()).thenReturn(cf1);
+
+        final RotatingCuratorFramework rotatingCuratorFramework = new RotatingCuratorFramework(zkh, 1000);
```
a1exsh, line 18:
```suggestion
        final RotatingCuratorFramework rotatingCuratorFramework = new RotatingCuratorFramework(zkh, 10);
```

core-common/src/test/java/org/zalando/nakadi/repository/zookeeper/RotatingCuratorFrameworkTest.java
```diff
@@ -0,0 +1,73 @@
...
+
+        CuratorFramework cfTmp = rotatingCuratorFramework.takeCuratorFramework();
+        Assert.assertEquals(cf1, cfTmp);
+        rotatingCuratorFramework.returnCuratorFramework(cfTmp);
+
+        Thread.sleep(1000);
```
a1exsh, line 24:
```suggestion
        Thread.sleep(10);
```

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/lock/CuratorNakadiLock.java
```diff
@@ -0,0 +1,87 @@
...
+            }
+            return acquired;
+        } catch (final Exception e) {
+            LOG.error("error while acquiring semaphore mutex lock", e);
+            try {
+                rotatingCuratorFramework.returnCuratorFramework(
+                        curatorToRelease);
+                curatorToRelease = null;
+            } catch (final RuntimeException re) {
+                LOG.error("error while returning curator framework", e);
```
a1exsh, line 65:
```suggestion
                LOG.error("error while returning curator framework", re);
```
;)

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/lock/CuratorNakadiLock.java
```diff
@@ -0,0 +1,87 @@
...
+                    TimeUnit.SECONDS);
+            if (!acquired) {
+                LOG.warn("failed to acquire semaphore mutex within {} seconds",
+                        SECONDS_TO_WAIT_FOR_LOCK);
+            }
+            return acquired;
+        } catch (final Exception e) {
+            LOG.error("error while acquiring semaphore mutex lock", e);
+            try {
+                rotatingCuratorFramework.returnCuratorFramework(
```
a1exsh, line 61: There is a possibility that we don't have `curatorToRelease` at this point (it's `null`).

antban: agree

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/lock/CuratorNakadiLock.java
```diff
@@ -0,0 +1,87 @@
...
+    @Override
+    public boolean lock() {
+        try {
+            curatorToRelease = rotatingCuratorFramework
+                    .takeCuratorFramework();
+            // the check is to not recreate semaphore every time
+            if (curatorToRelease != curatorLastUsed) {
+                curatorLastUsed = curatorToRelease;
+                semaphore = new InterProcessSemaphoreMutex(
+                        curatorLastUsed, lockPath);
```
antban, line 47: On unlock curator is released, therefore semaphore will have wrong reference to curator. I suggest not to cache semaphore as a first step

antban (commented):
I actually thought that the rotating curator will be used for offsets as well, and take/release methods will be applied globally for streaming connection to simplify migration/development.

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/lock/CuratorNakadiLock.java
adyach: this is fine, rotating curator know how to work with it

adyach: i do not see how that will happen: lock and unlock methods are used by single thread only.

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/lock/CuratorNakadiLock.java
adyach: you are right, but semaphore is not used outside of the class and access to it is defined by 2 methods. the next call to lock will replace it. so it will not be used with the same instance of curator.

I have just checked the code of InterProcessSemaphoreMutex, and it does execute any heavy init like creating watchers / znodes. the heavy stuff happens during acquire / release. I think it is good to go with a new instance every time.

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/lock/CuratorNakadiLock.java
a1exsh: It can happen if `takeCuratorFramework()` throws an exception here: https://github.com/zalando/nakadi/pull/1263/files#diff-bc88c93ce2c4b4ac1bb3d022553af0a9da93ddc3be9fb0be5c0ab54dd972453cR41

adyach:
> I actually thought that the rotating curator will be used for offsets as well, and take/release methods will be applied globally for streaming connection to simplify migration/development.

@antban as i mentioned offline it is `part I`. lease acquisition generates significant (it is around half of all watchers) amount of watchers. if successful, then `part II` will address streaming.

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/lock/CuratorNakadiLock.java
adyach: I decided not to do it and keep it like that for now :)

antban:
:+1:

a1exsh:
:+1:

PR #1264, opened 2021-02-19
always call unlock on nakadi lock

adyach:
in the situation when lock method returns false (lock was not acquired) exception is thrown in client code and unlock is not call (so undrelining used structures are not released)


adyach:
👍

a1exsh:
:+1:

a1exsh:
:+1:

adyach:
👍

PR #1266, opened 2021-03-04
Use critical-large VMs to speed up build

lmontrieux:
Use critical-large VMs to speed up build


lmontrieux:
Almost cuts build time in half.

lmontrieux:
👍

adyach:
👍

PR #1267, opened 2021-03-19
return 422 status when consumed offset is nonsense

adyach:
atm nakadi will return 503 http status code if consumed_offset argument is unknown to nakadi. instead it should return 422 unprocessable entity.


a1exsh:
:+1:

adyach:
👍

PR #1268, opened 2021-03-19
Reject number formats for numbers that start with `.`

burimshala:
# One-line summary

> Zalando ticket : team-aruha#565

## Description
`StringJsonParser` does not complain for numbers that start with `.` and they are parsed to double by `Double.valueOf`

## Review
- [ ] Tests

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
+++ b/core-common/src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
@@ -139,6 +139,9 @@ public class StrictJsonParser {
             throw syntaxError("Unexpected symbol '" + value + "'", tokenizer);
         }
         final int start = tokenizer.getCurrentPosition() - 1;
+        if (!Character.toString(tokenizer.value.charAt(start)).matches("\\d|-")) {
+            throw syntaxError("Numbers cannot start with:'" + value + "'", tokenizer);
+        }
         while (tokenizer.hasNext()) {
             if (POSSIBLE_NUMBER_DIGITS.indexOf(tokenizer.next()) < 0) {
                 tokenizer.back();
--- a/core-common/src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
+++ b/core-common/src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
@@ -159,11 +159,15 @@ public class StrictJsonParser {
 
         if (stringNumber.indexOf('.') > -1 || stringNumber.indexOf('e') > -1
                 || stringNumber.indexOf('E') > -1 || "-0".equals(stringNumber)) {
-            final Double d = Double.valueOf(stringNumber);
-            if (!d.isInfinite() && !d.isNaN()) {
-                return d;
-            } else {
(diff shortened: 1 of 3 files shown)


core-common/src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
```diff
@@ -139,6 +139,9 @@ private static Object readNumberTillTheEnd(final char value, final StringTokeniz
             throw syntaxError("Unexpected symbol '" + value + "'", tokenizer);
         }
         final int start = tokenizer.getCurrentPosition() - 1;
+        if (!Character.toString(tokenizer.value.charAt(start)).matches("\\d|-")) {
```
a1exsh, line 142: Performance-wise regex looks like too heavy here.  We could just check with `>= '0' && <= '9'` or something like that.
Alternatively we might not even go into this function if we see it start with the dot.

burimshala: Would you still think that Regex that has a very small syntax tree and is applied against a single character value could be overwhelming for this part?

a1exsh: For one the regexp will be compiled every time, or am I missing something?

a1exsh:
:+1:

core-common/src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
burimshala: Yes, in this case, it does compile every time but there seem to be workarounds.

burimshala:
👍

core-common/src/main/java/org/zalando/nakadi/domain/StrictJsonParser.java
```diff
@@ -135,7 +136,7 @@ private static Object readTrueTillTheEnd(final StringTokenizer tokenizer) {
     }
 
     private static Object readNumberTillTheEnd(final char value, final StringTokenizer tokenizer) {
```
Jan-M, line 138: Should function name also change to Integer?

Jan-M:
I would also be interested if it cant just check for "."?

burimshala:
> I would also be interested if it cant just check for "."?

You mean that you would be up for further validation such as for the whole number spec?

```
number
    integer fraction exponent

integer
    digit
    onenine digits
    '-' digit
    '-' onenine digits

digits
    digit
    digit digits

digit
    '0'
    onenine

onenine
    '1' . '9'

fraction
    ""
    '.' digits

exponent
    ""
    'E' sign digits
    'e' sign digits

sign
    ""
    '+'
    '-'
```

Jan-M:
No I was more wondering due to the Issue/Error in the PR if the problem is only arriving from "." in the begging that we don't like.

And that something else is already doing the remaining part of the validation, so why care for more than the "."

a1exsh:
:+1:

burimshala:
👍

a1exsh:
:+1:

burimshala:
👍

PR #1269, opened 2021-03-19
return 409 http status if subscription exists

adyach:
when the same subscription was created by several requests at almost the same time the first request gets 201 http status code, but the second - 503 status code. the expected response status code in such case is 409 conflict.

### DIFF
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/advice/PostSubscriptionExceptionHandler.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/advice/PostSubscriptionExceptionHandler.java
@@ -6,6 +6,7 @@ import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.context.request.NativeWebRequest;
 import org.zalando.nakadi.controller.PostSubscriptionController;
 import org.zalando.nakadi.exceptions.runtime.AuthorizationNotPresentException;
+import org.zalando.nakadi.exceptions.runtime.DuplicatedSubscriptionException;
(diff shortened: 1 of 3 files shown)


api-metastore/src/main/java/org/zalando/nakadi/controller/advice/PostSubscriptionExceptionHandler.java
```diff
@@ -51,4 +53,13 @@
         AdviceTrait.LOG.debug(exception.getMessage());
         return create(Problem.valueOf(UNPROCESSABLE_ENTITY, exception.getMessage()), request);
     }
+
+    @ExceptionHandler(DuplicatedSubscriptionException.class)
+    public ResponseEntity<Problem> handleSubscriptionCreationDisabledException(
```
a1exsh, line 58:
```suggestion
    public ResponseEntity<Problem> handleDuplicatedSubscriptionException(
```

adyach: ups

a1exsh:
```
> Task :api-metastore:checkstyleMain FAILED
[ant:checkstyle] [ERROR] /workspace/api-metastore/src/main/java/org/zalando/nakadi/controller/PostSubscriptionController.java:16:8: Unused import - org.zalando.nakadi.exceptions.runtime.DuplicatedSubscriptionException. [UnusedImports]
```

a1exsh:
:+1:

adyach:
👍

PR #1274, opened 2021-04-21
Improve error message with cursor

antban:
While subscription is created and cursors are out of allowed range - the event type information is not returned to user

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/exceptions/runtime/InvalidCursorException.java
+++ b/core-common/src/main/java/org/zalando/nakadi/exceptions/runtime/InvalidCursorException.java
@@ -9,27 +9,41 @@ public class InvalidCursorException extends NakadiBaseException {
     private final CursorError error;
     private final Cursor cursor;
     private final NakadiCursor position;
+    private final String eventType;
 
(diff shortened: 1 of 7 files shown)


core-common/src/main/java/org/zalando/nakadi/exceptions/runtime/InvalidCursorException.java
```diff
@@ -9,29 +9,43 @@
...
         this.position = null;
+        this.eventType = eventType;
     }
 
     public CursorError getError() {
         return error;
     }
 
+    private String getEventType() {
+        if (null != position) {
```
adyach, line 40: is it possible the position is null ?

adyach:
👍

core-common/src/main/java/org/zalando/nakadi/exceptions/runtime/InvalidCursorException.java
antban: Yes, it is possible. 
I decided to make it small, and therefore did in the same way as `#getPartition` and `#getOffset` of this class

antban:
:+1:

PR #1276, opened 2021-04-28
API definition of occurred_at vs. received_at event attributes

tfrauenstein:
Minor editorial update to avoid misunderstandings often observed on event producer and consumer side.
Only update of field descriptions in API definition -- no semantic or code changes, no test needed, but re-deployment of API spec.


antban:
:+1:

tfrauenstein:
👍

PR #1277, opened 2021-05-06
use optimistic locking to upload subscription topology

adyach:
the commit changes approach on how to do distributed locking in favour of optimistic locking. the current approach with apache curator lock recipe brings issues like not all reassignments are done in time which introduces errors while openning subscriptions or doing client rebalancing. the new approach is leveraging zookeeper version to upload changes if version is changed it will read changed structure from zookeeper and if the change still requires it will retry with the new version.


a1exsh:
`Execution failed for task ':core-common:checkstyleMain'.` :(

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/ClosingState.java
```diff
@@ -159,7 +159,7 @@ private void freePartitions(final Collection<EventTypePartition> keys) {
                 }
             }
         }
-        getZk().runLocked(() -> getZk().transfer(getSessionId(), keys));
+        getZk().transfer(getSessionId(), keys);
```
antban, line 162: @adyach Either we need to have feature toggle to use/not use `runLocked` call here, or (as a first iteration) leave `runLocked` call, as it won't be possible to seamlessly migrate to lockless approach. The same goes to SteamingState call.
I propose to use featureToggles

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/NewZkSubscriptionClient.java
```diff
@@ -199,27 +202,55 @@ protected String getOffsetPath(final EventTypePartition etp) {
...
+                            changeSet.add(candidate.toState(
+                                    Partition.State.ASSIGNED,
+                                    candidate.getNextSession(),
+                                    null));
+                        }
+                    }
+
+                    if (!changeSet.isEmpty()) {
+                        // The list of sessions didn't change, therefore one should not update sessionsHash.
+                        final Topology newTopology = getTopology()
```
antban, line 234:
```suggestion
                        final Topology newTopology = topology
```

antban (changes requested):
Nice! Seems that soon we will be able to eliminate recipes usage!
There are 2 small things to fix and we are good to go

antban:
:+1:

PR #1280, opened 2021-05-10
rebalance uses optimistic locking

adyach:
the commit changes used locking mechanism of rebalance function. the change removes  usage of global curator lock in favor of using optimistic locking meaning: 1)read topology with zk node version 2)prepare change set 3)try to write to zk node with the read version number in step 1 if failed retry. the obsolete logic related to hash generation for sessions transfer was removed.

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/NewZkSubscriptionClient.java
+++ b/core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/NewZkSubscriptionClient.java
@@ -2,6 +2,7 @@ package org.zalando.nakadi.service.subscription.zk;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.function.Function;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/NewZkSubscriptionClient.java
```diff
@@ -2,6 +2,7 @@
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Function;
```
antban, line 5: Seems that there is a typo - probably usual java Function will be enough

antban (commented):
Only one small comment, everything else seems to be fine

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/NewZkSubscriptionClient.java
adyach: 👍

adyach:
👍

antban:
:+1:

PR #1281, opened 2021-05-12
remove nakadi zookeeper lock call

adyach:
methods which used `runLocked` method were migrated to optimistic locking. in this commit `runLocked] method is removed leaving only not locking methods to update subscription states

### DIFF
--- a/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StartingState.java
+++ b/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StartingState.java
@@ -25,7 +25,7 @@ public class StartingState extends State {
             switchState(new CleanupState(e));
             return;
         }
-        getZk().runLocked(this::registerSessionAndStartStreaming);
+        registerSessionAndStartStreaming();
     }
 
     /**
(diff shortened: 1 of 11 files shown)


api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StartingState.java
```diff
@@ -38,7 +38,7 @@ public void onEnter() {
      * 4. Switches to streaming state.
      */
     private void registerSessionAndStartStreaming() {
-        final boolean subscriptionJustInitialized = SubscriptionInitializer.initializeSubscriptionLocked(getZk(),
+        final boolean subscriptionJustInitialized = SubscriptionInitializer.initialize(getZk(),
                 getContext().getSubscription(), getContext().getTimelineService(), getContext().getCursorConverter());
         getContext().getCurrentSpan().setTag("session.id", getContext().getSessionId());
         if (!subscriptionJustInitialized) {
```
antban, line 44: I think that this flag is not needed at all, but I also understand that it is not a part of PR intention. Would be nice to have it removed.

adyach:
👍

antban:
:+1:

PR #1283, opened 2021-05-19
log less useless messages

antban:
The message to remove is never used, and is executed many times a second, flooding logs with info level (while it is more of a trace level). 
Instead of changing the level it was intentionally removed.


antban:
:+1:

adyach:
👍

PR #1284, opened 2021-05-20
better track inability to join streaming on consumption

adyach:
In order to synchronize behavior of several instances, there is an optimistic lock is in use.
Unfortunately one place was not handled properly - registration of the streaming session.
The problem is that streaming code is allowing to have more streaming sessions, than there are partitions, or have more than 1 streaming connections for manually assigned partition.
This causes some problems - after unsuccessful rebalance (obviously rebalance is saying that it can't stream 2 partitions to 3 connections) some of streaming connections are dying.

In order to avoid that, there should be 2 changes made:
1. Session should be registered in optimistic manner (check, register, check. If not - session must be removed, https://github.com/zalando/nakadi/blob/master/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StartingState.java#L46-L87)
2. It should be expected, that rebalance may predictably fail for a very short amount of time, and if it is - it should retry (usage of https://github.com/zalando/nakadi/blob/master/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/SubscriptionRebalancer.java).

### DIFF
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaRebalanceAT.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaRebalanceAT.java
@@ -327,7 +327,7 @@ public class HilaRebalanceAT extends BaseAT {
     }
 
     @Test(timeout = 10000)
-    public void testNoConflictOnTheSamePartitionRequest() throws Exception {
+    public void testAtLeastOneClientGets409OnTheSamePartitionRequest() throws Exception {
         final TestStreamingClient client1 = new TestStreamingClient(
                 URL, subscription.getId(), "batch_flush_timeout=1",
                 Optional.empty(),
@@ -341,8 +341,9 @@ public class HilaRebalanceAT extends BaseAT {
         client1.start();
         client2.start();
 
-        waitFor(() -> assertThat(client1.getResponseCode(), Matchers.is(HttpStatus.CONFLICT.value())));
-        waitFor(() -> assertThat(client2.getResponseCode(), Matchers.is(HttpStatus.CONFLICT.value())));
+        waitFor(() -> assertThat("at least once client should get 409 conflict",
+                        client1.getResponseCode() != HttpStatus.CONFLICT.value() ||
+                        client1.getResponseCode() != HttpStatus.CONFLICT.value()));
     }
 
     public List<SubscriptionCursor> getLastCursorsForPartitions(final TestStreamingClient client,
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaRebalanceAT.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaRebalanceAT.java
@@ -3,6 +3,7 @@ package org.zalando.nakadi.webservice.hila;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import org.hamcrest.Matchers;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.springframework.http.HttpStatus;
@@ -323,6 +324,25 @@ public class HilaRebalanceAT extends BaseAT {
 
         // check that we get 204
         assertThat(commitStatusCode, Matchers.is(HttpStatus.NO_CONTENT.value()));
+    }
+
+    @Test(timeout = 10000)
+    public void testNoConflictOnTheSamePartitionRequest() throws Exception {
+        final TestStreamingClient client1 = new TestStreamingClient(
+                URL, subscription.getId(), "batch_flush_timeout=1",
+                Optional.empty(),
+                Optional.of("{\"partitions\":[" +
+                        "{\"event_type\":\"" + eventType.getName() + "\",\"partition\":\"0\"}]}"));
+        final TestStreamingClient client2 = new TestStreamingClient(
+                URL, subscription.getId(), "batch_flush_timeout=1",
+                Optional.empty(),
+                Optional.of("{\"partitions\":[" +
+                        "{\"event_type\":\"" + eventType.getName() + "\",\"partition\":\"0\"}]}"));
+        client1.start();
+        client2.start();
+
+        waitFor(() -> assertThat(client1.getResponseCode(), Matchers.is(HttpStatus.CONFLICT.value())));
+        waitFor(() -> assertThat(client2.getResponseCode(), Matchers.is(HttpStatus.CONFLICT.value())));
     }
 
     public List<SubscriptionCursor> getLastCursorsForPartitions(final TestStreamingClient client,
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaRebalanceAT.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaRebalanceAT.java
@@ -341,9 +341,9 @@ public class HilaRebalanceAT extends BaseAT {
         client1.start();
         client2.start();
 
-        waitFor(() -> assertThat("at least once client should get 409 conflict",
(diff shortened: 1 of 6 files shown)


acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaRebalanceAT.java
```diff
@@ -325,6 +326,25 @@ public void testCommitWhenDirectAssignment() throws Exception {
         assertThat(commitStatusCode, Matchers.is(HttpStatus.NO_CONTENT.value()));
     }
 
+    @Test(timeout = 10000)
+    public void testNoConflictOnTheSamePartitionRequest() throws Exception {
```
adyach, line 330: that's unstable, because sometimes one stream takes partition, but the other one is too late, which results in 102 and 409
i appreciate ideas on test

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/SubscriptionRebalancer.java
```diff
@@ -36,11 +40,15 @@
...
+                final Optional<Partition> partitionOpt = partitionsLeft.stream()
                         .filter(p -> p.getKey().equals(requestedPartition))
-                        .findFirst()
-                        .orElseThrow(() -> new RebalanceConflictException(
-                                "Two existing sessions request the same partition: " + requestedPartition));
+                        .findFirst();
+                if (!partitionOpt.isPresent()) {
+                    LOG.warn("Two existing sessions request the same partition: " + requestedPartition);
+                    return new Partition[0];
+                }
```
antban, line 49: This covers only first part of problem - whenever clients have requested partitions. But when there are just more sessions than partitions - It is a different situation - there is a 
```
throw new IllegalArgumentException("Can not rebalance " + itemCount + " onto " + weigths.length);
```
exception later. 

I suggest a different approach do not modify this part of code, but - change `IllegalArgumentException` to `RebalanceConflictException`, For every RebalanceConflictException - do not do rebalance. 

This will allow to have tests for different rebalance problems and will still solve the issue that we have with concurrent sessions creation.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StartingState.java
```diff
@@ -41,38 +41,9 @@ private void registerSessionAndStartStreaming() {
...
-                    .flatMap(s -> s.getRequestedPartitions().stream())
-                    .filter(requestedPartitions::contains)
-                    .collect(Collectors.toList());
-            if (!conflictPartitions.isEmpty()) {
-                TracingService.logStreamCloseReason(getContext().getCurrentSpan(),
-                        "Partition already taken by other stream of the subscription");
-                switchState(new CleanupState(SubscriptionPartitionConflictException.of(conflictPartitions)));
-                return;
-            }
+        if (!subscriptionJustInitialized &&
```
antban, line 44: I think that `subscriptionJustInitialized` variable should be removed at all, as it works only once right after first initialization. Also the variable has questionable meaning right now, as it may be returned twice `true` for different concurrent sessions.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StartingState.java
```diff
@@ -85,11 +56,54 @@ private void registerSessionAndStartStreaming() {
...
         } catch (Exception ex) {
             switchState(new CleanupState(ex));
             return;
         }
+
         switchState(new StreamingState());
     }
 
+    private boolean areStreamingSlotsAvailable(final Collection<Session> sessions,
+                                               final String excludeSessionId) {
```
antban, line 72: Would not it be better if the sessions list will be filtered before the method?
For example, the calls will be like this:
```
1. areStreamingSlotsAvailable(getZk().listSessions())
2. areStreamingSlotsAvailable(getZk().listSessions().stream().filter(v -> !v.getId().equals(getSessionId()))
```
It will make the method more straightforward and bug-prone

antban (changes requested):
I think there some improvements could be made

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaRebalanceAT.java
```diff
@@ -325,6 +326,25 @@ public void testCommitWhenDirectAssignment() throws Exception {
...
+        final TestStreamingClient client2 = new TestStreamingClient(
+                URL, subscription.getId(), "batch_flush_timeout=1",
+                Optional.empty(),
+                Optional.of("{\"partitions\":[" +
+                        "{\"event_type\":\"" + eventType.getName() + "\",\"partition\":\"0\"}]}"));
+        client1.start();
+        client2.start();
+
+        waitFor(() -> assertThat(client1.getResponseCode(), Matchers.is(HttpStatus.CONFLICT.value())));
+        waitFor(() -> assertThat(client2.getResponseCode(), Matchers.is(HttpStatus.CONFLICT.value())));
```
a1exsh, line 345: Do we really want _both_ clients to get a conflict response?

adyach: that is possible when clients open connections in very close moment. session registration is not synchronized, so i can not tell which one should be dropped, so both.

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaRebalanceAT.java
antban: Maybe check that at least one of them will should have conflict?

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
```diff
@@ -313,11 +314,15 @@ private void rebalance() {
...
-            );
+            try {
+                zkClient.updateTopology(topology ->
+                        rebalancer.apply(
+                                zkClient.listSessions(),
+                                topology.getPartitions())
+                );
+            } catch (final RebalanceConflictException e) {
+                log.warn("failed to rebalance partitions: {}", e.getMessage(), e);
+            }
```
antban, line 325: Would not it be better to handle exception where it occurs?  I mean:
```
            zkClient.updateTopology(topology ->
                try {
                        return rebalancer.apply(
                                zkClient.listSessions(),
                                topology.getPartitions())
                } catch (final RebalanceConflictException e) {
                    log.warn("failed to rebalance partitions: {}", e.getMessage(), e);
                   return []; // Or how it is written in java
                }
            );
```

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StartingState.java
```diff
@@ -85,11 +57,52 @@ private void registerSessionAndStartStreaming() {
...
+            }
         } catch (Exception ex) {
             switchState(new CleanupState(ex));
             return;
         }
+
         switchState(new StreamingState());
     }
 
+    private boolean areStreamingSlotsAvailable(final Collection<Session> sessions) {
```
antban, line 74: `nitpick`: While looking closer to this method, It seems to be not the best name for the method `areStreamingSlotsAvailable` is right now responsible for not only checking, but also for switching state of streaming to a CleanupState (with reason). 

Maybe it makes sense either to rename the method or to move switchToState somehow to a caller part (don't know what is the best name here). Overall - there is no bug, but looks suspicious, feel free to leave it as it is.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
antban: I think it eliminates the need to figure out if the exception is actually thrown in non-modified form through the unknown code hidden by interface.

antban:
:+1:

adyach:
:

adyach:
👍

antban:
:+1:

PR #1285, opened 2021-05-25
Make example in api definition better-looking

antban:
While generating example of api definition the values that are used to provide several examples are leading to a weirdly-looking value with ` ,` in the middle. The better approach to avoid it - use single example


docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2718,7 +2718,7 @@ definitions:
           structure is not enforced. For example a team name and data type can be used such as
```
a1exsh, line 2718: Maybe also remove "team name can be used"?  Let's not recommend that (or change the wording to the opposite).

antban: :+1:

adyach:
👍

PR #1287, opened 2021-06-01
Update: API definition of occurred_at vs. received_at event attributes

tfrauenstein:
Minor editorial update to make it consistent with API Guideline definition (see https://github.com/zalando/restful-api-guidelines/pull/659) and to avoid misunderstandings still observed in production.

Just an update of field descriptions in API definition -- no semantic or code changes, no test needed, but re-deployment of API spec.

### DIFF
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -2227,19 +2227,24 @@ definitions:
         example: 'pennybags.payment-business-event'
       occurred_at:
         description: |
-          Timestamp of when the event was created by the producer application. 
-          (Typically, it differs from when the event is posted by the producer and received by Nakadi via the API.)
+         Technical timestamp of when the event object was created during processing of the
+         business event by the producer application. Note, it may differ from the timestamp 
+         when the related real-world business event happened (e.g. when the packet was handed 
+         over to the customer), which should be passed separately via an event type specific 
+         attribute. 
+         Depending on the producer implementation, the timestamp is typically some milliseconds  
+         earlier than when the event is published and received by the API event post endpoint 
+         server -- see below.
         type: string
         format: RFC 3339 date-time
         example: '1996-12-19T16:39:57-08:00'
       received_at:
         type: string
         readOnly: true
-        description: |
-          Timestamp of when the event was received by Nakadi via the event post endpoints. 
-          It is automatically enriched, and events will be rejected if set by the Event producer.
-        format: RFC 3339 date-time
-        example: '1996-12-19T16:39:57-08:00'
+      description: |
+        Timestamp of when the event was received via the API event post endpoints. 
+        It is automatically enriched, and events will be rejected if set by the event producer.
+      example: '1996-12-19T16:39:57-08:00'
       version:
         type: string
         readOnly: true
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -2227,24 +2227,25 @@ definitions:
         example: 'pennybags.payment-business-event'
       occurred_at:
         description: |
-         Technical timestamp of when the event object was created during processing of the
-         business event by the producer application. Note, it may differ from the timestamp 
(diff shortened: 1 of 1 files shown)


tfrauenstein:
👍

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2227,19 +2227,24 @@ definitions:
...
         readOnly: true
-        description: |
-          Timestamp of when the event was received by Nakadi via the event post endpoints. 
-          It is automatically enriched, and events will be rejected if set by the Event producer.
-        format: RFC 3339 date-time
-        example: '1996-12-19T16:39:57-08:00'
+      description: |
+        Timestamp of when the event was received via the API event post endpoints. 
+        It is automatically enriched, and events will be rejected if set by the event producer.
+      example: '1996-12-19T16:39:57-08:00'
```
antban, line 2247: @tfrauenstein Seems that this is not the intended change (move 2 spaces left the description of `received_at`

a1exsh:
I was not aware of this change in the API guidelines, and I believe there is some lack of transparency about these (upcoming) changes.

docs/_data/nakadi-event-bus-api.yaml
tfrauenstein: @antban yes, this minor change is intended  (lower case of 'event producer').

tfrauenstein:
@a1exsh It basically is not an API guideline change. It is about addressing uncertainties around semantics of `occurred_at`in the Nakadi API definition. (We just 1st did the update in the API Guidelines where the affected Nakadi API payload aspects are also defined.)  We observed that `occurred_at` uncertainties were around a) clear distinction between real-world business event vs. Nakadi event object creation and b) that the timestamp of event object creation is different from the timestamp of event publishing and event receiving via Nakadi endpoint. The new description is more clear here (and aligned with multiple Nakadi stakeholders) -- it is important esp. for (event) data analytics.

a1exsh:
@tfrauenstein it all makes sense to me after I was thinking about the distinction in the back of my head for a few days now, but it would be nice to have the reasoning like that written down somewhere for future reference.  Maybe I just haven't found it yet?

tfrauenstein:
@a1exsh fine. Some context in given in the [Guideline PR](https://github.com/zalando/restful-api-guidelines/pull/659) discussion and linked chat thread. (Next time we will start with the Nakadi API change -- since mastership should be here.)

docs/_data/nakadi-event-bus-api.yaml
a1exsh: Github does not allow me to add a multi-line suggestion to the lines that already have comments.  I think @antban is pointing out that `description` and `example` were unintentionally moved two spaces to the left, so they are no longer belonging to `received_at` description.  Also not sure if removal of `format` was intentional.

a1exsh:
@tfrauenstein I've made sure to read all available communication wrt. `occurred_at` naming, but I haven't found detailed reasoning.  I do agree with the proposed changes, don't get me wrong.  I'd like though the reasoning to be properly captured and be available for the next person to stumble upon it.

The name of this attribute is not possible to change, but a phrasing like "a (business) event occurred at" does invoke certain picture, at least in my mind, and it's not about a timestamp that is automatically assigned by the application when generating the event payload (`generated_at` might have been a better name, but again, it's too late to reconsider that).  I have more thoughts about this, but not sure where to express them best.  Comments to this PR are probably not the best place. ;)

docs/_data/nakadi-event-bus-api.yaml
tfrauenstein: +1 -- fixed: indention and format definition

a1exsh:
:+1:

tfrauenstein:
👍

tfrauenstein:
@a1exsh Yes, we cannot change the name, and PR comments are not the best place for discussions. 
I'm happy to discuss your thoughts -- just directly contact me via email / chat.
Can you meanwhile merge the PR (I have no write access), thx!

PR #1288, opened 2021-06-02
part II: rotate curator client on streams

adyach:
as a follow up of having rotating curator in place(#1263) and not used in the zookeeper locks anymore(#1284): 1)cleaned up code which worked with locks and it is not used anymore 2)use rotating curator for subscription streaming.


core-common/src/main/java/org/zalando/nakadi/repository/zookeeper/ZooKeeperHolder.java
```diff
@@ -85,13 +85,15 @@ public CuratorFramework getCuratorFramework() {
 
     public static class StaticCuratorFramework extends CloseableCuratorFramework {
```
antban, line 86: Please, change name from `StaticCuratorFramework` to `DeafultSubscriptionCurator` or something else more appropriate.

antban:
:+1:

adyach:
👍

PR #1289, opened 2021-06-03
use scheduled executor for curator rotation thread

adyach:
after depromting rotating curator to normal class (not Spring component https://github.com/zalando/nakadi/pull/1288) it was not able to execute @Schedule method.


antban:
:+1:

adyach:
👍

PR #1290, opened 2021-06-04
close zk client when it was taken

adyach:
at the moment when zk client is taken it is not always closed which violaties a good practice to release whatever you acquired. it also prevents rotating curator framework from working properly because it is taken, but never returned.


antban:
:+1:

adyach:
👍

PR #1291, opened 2021-06-08
remove unused session hash

adyach:
since migration to optimistic locking(https://github.com/zalando/nakadi/pull/1280) there is no need to have session hash to perform rebalance


adyach:
👍

antban:
:+1:

PR #1292, opened 2021-06-15
Filter subscriptions and event types by authroziation

burimshala:
# One-line summary

> Zalando ticket : GHE-747

## Description
Extend API to support GET /event-types?writers=service,user and GET /subscriptions?readers=service,user

## Review
- [x] Tests
- [x] Documentation

### DIFF
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
@@ -104,10 +104,10 @@ public class EventTypeDbRepositoryTest extends AbstractDbRepositoryTest {
     }
 
     @Test
-    public void whenEventExistsFindByAuthorizationReturnsSomething() throws Exception {
+    public void whenEventTypeExistsFindByAuthorizationReturnsSomething() throws Exception {
         final EventType eventType1 = buildDefaultEventType();
         final AuthorizationAttributeQueryParser auth = new AuthorizationAttributeQueryParser();
-        auth.setAsText("service:stups_test-app-2");
+        auth.setAsText("service:stups_test-app");
 
         eventType1.setAuthorization(new ResourceAuthorization(
                 Collections.emptyList(),
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
@@ -8,17 +8,22 @@ import org.springframework.jdbc.support.rowset.SqlRowSet;
 import org.zalando.nakadi.domain.EventCategory;
 import org.zalando.nakadi.domain.EventType;
 import org.zalando.nakadi.domain.EventTypeSchema;
+import org.zalando.nakadi.domain.ResourceAuthorization;
 import org.zalando.nakadi.domain.Version;
 import org.zalando.nakadi.exceptions.runtime.DuplicatedEventTypeNameException;
 import org.zalando.nakadi.exceptions.runtime.NoSuchEventTypeException;
+import org.zalando.nakadi.model.AuthorizationAttributeQueryParser;
+import org.zalando.nakadi.plugin.api.authz.AuthorizationAttribute;
 import org.zalando.nakadi.utils.TestUtils;
 
 import java.io.IOException;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
-import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.CoreMatchers.*;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasProperty;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.core.IsEqual.equalTo;
 import static org.zalando.nakadi.utils.TestUtils.buildDefaultEventType;
@@ -95,6 +100,25 @@ public class EventTypeDbRepositoryTest extends AbstractDbRepositoryTest {
         final EventType persistedEventType = repository.findByName(eventType2.getName());
 
         assertThat(persistedEventType, notNullValue());
+    }
+
+    @Test
+    public void whenEventExistsFindByAuthorizationReturnsSomething() throws Exception {
+        final EventType eventType1 = buildDefaultEventType();
+        final AuthorizationAttributeQueryParser auth = new AuthorizationAttributeQueryParser();
+        auth.setAsText("service:stups_test-app");
+
+        eventType1.setAuthorization(new ResourceAuthorization(
+                Collections.emptyList(),
+                Collections.emptyList(),
+                List.of((AuthorizationAttribute) auth.getValue())
+        ));
+
+        insertEventType(eventType1);
+
+        final List<EventType> persistedEventType = repository.list((AuthorizationAttribute) auth.getValue());
+
+        assertThat(persistedEventType, hasItem(hasProperty("name", is(eventType1.getName()))));
     }
 
     @Test(expected = NoSuchEventTypeException.class)
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
@@ -27,8 +27,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.hasProperty;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.core.IsEqual.equalTo;
-import static org.zalando.nakadi.utils.TestUtils.buildDefaultEventType;
-import static org.zalando.nakadi.utils.TestUtils.randomString;
+import static org.zalando.nakadi.utils.TestUtils.*;
 
 public class EventTypeDbRepositoryTest extends AbstractDbRepositoryTest {
 
@@ -107,7 +106,7 @@ public class EventTypeDbRepositoryTest extends AbstractDbRepositoryTest {
     @Test
     public void whenEventTypeExistsFindByAuthorizationReturnsSomething() throws Exception {
         final EventType eventType1 = buildDefaultEventType();
-        final ResourceAuthorizationAttribute auth = new ResourceAuthorizationAttribute("service", "stups_test-app");
+        final ResourceAuthorizationAttribute auth = new ResourceAuthorizationAttribute("service", "stups_test"+randomTextString());
 
         eventType1.setAuthorization(new ResourceAuthorization(
                 Collections.emptyList(),
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
@@ -116,9 +116,9 @@ public class EventTypeDbRepositoryTest extends AbstractDbRepositoryTest {
 
         insertEventType(eventType1);
 
-        final List<EventType> persistedEventType = repository.list((AuthorizationAttribute) auth.getValue());
+        final List<EventType> persistedEventTypes = repository.list((AuthorizationAttribute) auth.getValue());
 
-        assertThat(persistedEventType, hasItem(hasProperty("name", is(eventType1.getName()))));
+        assertThat(persistedEventTypes, hasItem(hasProperty("name", is(eventType1.getName()))));
     }
 
     @Test(expected = NoSuchEventTypeException.class)
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
@@ -27,6 +27,7 @@ import static org.hamcrest.Matchers.hasProperty;
 import static org.hamcrest.Matchers.hasSize;
 import static org.hamcrest.core.IsEqual.equalTo;
 import static org.zalando.nakadi.utils.TestUtils.buildDefaultEventType;
+import static org.zalando.nakadi.utils.TestUtils.randomString;
 
 public class EventTypeDbRepositoryTest extends AbstractDbRepositoryTest {
 
@@ -106,7 +107,7 @@ public class EventTypeDbRepositoryTest extends AbstractDbRepositoryTest {
     public void whenEventExistsFindByAuthorizationReturnsSomething() throws Exception {
         final EventType eventType1 = buildDefaultEventType();
         final AuthorizationAttributeQueryParser auth = new AuthorizationAttributeQueryParser();
-        auth.setAsText("service:stups_test-app");
+        auth.setAsText("service:stups_test-app-2");
 
         eventType1.setAuthorization(new ResourceAuthorization(
                 Collections.emptyList(),
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
@@ -9,6 +9,7 @@ import org.zalando.nakadi.domain.EventCategory;
 import org.zalando.nakadi.domain.EventType;
 import org.zalando.nakadi.domain.EventTypeSchema;
 import org.zalando.nakadi.domain.ResourceAuthorization;
+import org.zalando.nakadi.domain.ResourceAuthorizationAttribute;
 import org.zalando.nakadi.domain.Version;
 import org.zalando.nakadi.exceptions.runtime.DuplicatedEventTypeNameException;
 import org.zalando.nakadi.exceptions.runtime.NoSuchEventTypeException;
@@ -106,18 +107,17 @@ public class EventTypeDbRepositoryTest extends AbstractDbRepositoryTest {
     @Test
     public void whenEventTypeExistsFindByAuthorizationReturnsSomething() throws Exception {
         final EventType eventType1 = buildDefaultEventType();
-        final AuthorizationAttributeQueryParser auth = new AuthorizationAttributeQueryParser();
-        auth.setAsText("service:stups_test-app");
+        final ResourceAuthorizationAttribute auth = new ResourceAuthorizationAttribute("service", "stups_test-app");
 
         eventType1.setAuthorization(new ResourceAuthorization(
                 Collections.emptyList(),
                 Collections.emptyList(),
-                List.of((AuthorizationAttribute) auth.getValue())
+                List.of(auth)
         ));
 
         insertEventType(eventType1);
 
-        final List<EventType> persistedEventTypes = repository.list((AuthorizationAttribute) auth.getValue());
+        final List<EventType> persistedEventTypes = repository.list(auth);
 
         assertThat(persistedEventTypes, hasItem(hasProperty("name", is(eventType1.getName()))));
     }
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
@@ -129,7 +129,7 @@ public class SubscriptionDbRepositoryTest extends AbstractDbRepositoryTest {
                 .collect(toList());
 
         final List<Subscription> subscriptions = repository.listSubscriptions(ImmutableSet.of("et1"),
-                Optional.of(owningApp), 0, 10);
+                Optional.of(owningApp), emptySet(), 0, 10);
         assertThat(subscriptions, equalTo(expectedSubscriptions));
     }
 
@@ -155,7 +155,7 @@ public class SubscriptionDbRepositoryTest extends AbstractDbRepositoryTest {
                 .collect(toList());
 
         final List<Subscription> subscriptions = repository.listSubscriptions(ImmutableSet.of(et1, et2),
-                Optional.empty(), 0, 10);
+                Optional.empty(), emptySet(), 0, 10);
         assertThat(subscriptions, equalTo(expectedSubscriptions));
     }
 
@@ -170,7 +170,7 @@ public class SubscriptionDbRepositoryTest extends AbstractDbRepositoryTest {
         testSubscriptions.subList(3, testSubscriptions.size()).clear();
 
         final List<Subscription> subscriptions = repository.listSubscriptions(
-                emptySet(), Optional.of(owningApp), 2, 3);
+                emptySet(), Optional.of(owningApp), emptySet(), 2, 3);
         assertThat(subscriptions, equalTo(testSubscriptions));
     }
 
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
@@ -6,11 +6,14 @@ import com.google.common.collect.ImmutableSet;
 import org.junit.Before;
 import org.junit.Test;
 import org.zalando.nakadi.config.JsonConfig;
+import org.zalando.nakadi.domain.AuthorizationAttributeProxy;
 import org.zalando.nakadi.domain.Subscription;
 import org.zalando.nakadi.domain.SubscriptionBase;
 import org.zalando.nakadi.exceptions.runtime.DuplicatedSubscriptionException;
 import org.zalando.nakadi.exceptions.runtime.NoSuchSubscriptionException;
 import org.zalando.nakadi.exceptions.runtime.ServiceTemporarilyUnavailableException;
+import org.zalando.nakadi.model.AuthorizationAttributeQueryParser;
+import org.zalando.nakadi.plugin.api.authz.AuthorizationAttribute;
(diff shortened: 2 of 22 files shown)


api-metastore/src/main/java/org/zalando/nakadi/controller/EventTypeController.java
```diff
@@ -68,10 +70,10 @@ public EventTypeController(final EventTypeService eventTypeService,
     }
 
     @RequestMapping(method = RequestMethod.GET)
-    public ResponseEntity<?> list() {
-        final List<EventType> eventTypes = eventTypeService.list();
+    public ResponseEntity<?> list(@RequestParam final Optional<String[]> writers) {
```
a1exsh, line 73: Why not use `Set<>` like we do for `readers` (and `event_type`) in `SubscriptionController`?

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1168,6 +1173,11 @@ paths:
           description: show subscription status
           type: boolean
           default: false
+        - name: readers
+          in: query
+          description: |
+            The username or service name that has explicit read authorization rights.
+          type: string
```
a1exsh, line 1180:
```suggestion
          collectionFormat: multi
          type: array
          items:
            type: string
```

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
```diff
@@ -226,26 +228,30 @@ public UriComponents getSubscriptionUri(final Subscription subscription) {
...
                     eventTypesFilter,
                     owningAppOption,
+                    readersFilter,
                     offset,
                     limit);
             final Optional<PaginationLinks.Link> prev = Optional.of(offset).filter(v -> v > 0)
                     .map(o -> createSubscriptionListLink(
-                            owningAppOption, eventTypesFilter, Math.max(0, o - limit), Optional.empty(), limit, 
-                            showStatus));
+                            owningAppOption, eventTypesFilter, Collections.emptySet(),
```
a1exsh, line 236:
```suggestion
                            owningAppOption, eventTypesFilter, readersFilter,
```

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
```diff
@@ -226,26 +228,30 @@ public UriComponents getSubscriptionUri(final Subscription subscription) {
...
+                            Math.max(0, o - limit), Optional.empty(), limit, showStatus));
             final Optional<PaginationLinks.Link> next = Optional.of(subscriptions.size()).filter(v -> v >= limit)
                     .map(size -> createSubscriptionListLink(
-                            owningAppOption, eventTypesFilter, offset + size, Optional.empty(), limit, showStatus));
+                            owningAppOption, eventTypesFilter, Collections.emptySet(),
```
a1exsh, line 240:
```suggestion
                            owningAppOption, eventTypesFilter, readersFilter,
```

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
```diff
@@ -226,26 +228,30 @@ public UriComponents getSubscriptionUri(final Subscription subscription) {
...
 
             paginationWrapper = new PaginationWrapper<>(subscriptions, new PaginationLinks(prev, next));
         } else {
             final SubscriptionTokenLister.ListResult listResult = subscriptionTokenLister.listSubscriptions(
-                    eventTypesFilter, owningAppOption, tokenObj, limit);
+                    eventTypesFilter, owningAppOption, readersFilter, tokenObj, limit);
             final Optional<PaginationLinks.Link> prev = Optional.ofNullable(listResult.getPrev())
                     .map(t -> createSubscriptionListLink(
-                            owningAppOption, eventTypesFilter, 0, Optional.of(t), limit, showStatus));
+                            owningAppOption, eventTypesFilter, Collections.emptySet(),
```
a1exsh, line 249:
```suggestion
                            owningAppOption, eventTypesFilter, readersFilter,
```

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
```diff
@@ -226,26 +228,30 @@ public UriComponents getSubscriptionUri(final Subscription subscription) {
...
+                            0, Optional.of(t), limit, showStatus));
             final Optional<PaginationLinks.Link> next = Optional.ofNullable(listResult.getNext())
                     .map(t -> createSubscriptionListLink(
-                            owningAppOption, eventTypesFilter, 0, Optional.of(t), limit, showStatus));
+                            owningAppOption, eventTypesFilter, Collections.emptySet(),
```
a1exsh, line 253:
```suggestion
                            owningAppOption, eventTypesFilter, readersFilter,
```

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -102,6 +103,18 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
                 new EventTypeMapper());
     }
 
+    public List<EventType> list(final String[] writers){
+        return jdbcTemplate.query(
+                String.format("SELECT et_event_type_object\n" +
```
a1exsh, line 108: We should either use `SELECT DISTINCT` or find a way around using set-returning `jsonb_to_recordset()`.

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -102,6 +103,18 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
...
+                        "FROM zn_data.event_type," +
+                        "jsonb_to_recordset(et_event_type_object->'authorization'->'writers')" +
+                        "as writers(value text)\n" +
```
a1exsh, line 111: I also think we should match both `data_type` and `value`.  The query string can look like `?writer=user:jdoe&writer=service:appname` — this is consistent with the error reports on wrong authorization attributes at least.

api-metastore/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -52,10 +52,11 @@ public SubscriptionController(final SubscriptionService subscriptionService) {
             @RequestParam(value = "limit", required = false, defaultValue = "20") final int limit,
             @RequestParam(value = "offset", required = false, defaultValue = "0") final int offset,
             @RequestParam(value = "token", required = false) final String token,
+            @RequestParam(value = "readers", required = false) final Set<String> readers,
```
a1exsh, line 55:
```suggestion
            @Nullable @RequestParam(value = "readers", required = false) final Set<String> readers,
```
See the declaration of `eventTypes` parameter a few lines up: I think we have the same requirements here.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1168,6 +1173,11 @@ paths:
           description: show subscription status
           type: boolean
           default: false
+        - name: readers
+          in: query
+          description: |
+            The username or service name that has explicit read authorization rights.
```
a1exsh, line 1179: This should be extended to specify semantics: `OR` vs. `AND`.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1168,6 +1173,11 @@ paths:
           description: show subscription status
           type: boolean
           default: false
+        - name: readers
```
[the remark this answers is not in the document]
burimshala: What if we change the type of this item from string to array. The plural makes more sense for me. Also, it is bounded to GET /subscription**s** which is plural.

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
burimshala: This would impact the convenience on both ends. Users will have to follow a format that we have to describe and we'd need to do a bit of parsing. I was considering get with body, but post is already taken off.

docs/_data/nakadi-event-bus-api.yaml
burimshala: Is array more appropriate to describe a list of readers?

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
a1exsh: But without specifying user or service it is ambiguous and we can avoid that by using `type:value` format: that's not too much parsing for being future-proof, IMO.

docs/_data/nakadi-event-bus-api.yaml
a1exsh: I've copied it over from the definition of `event_type` param: https://github.com/zalando/nakadi/blob/master/docs/_data/nakadi-event-bus-api.yaml#L1127-L1137

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
burimshala: I added group by clause to prevent dupes

burimshala:
> Should it be `Set<String>`?

We do want to allow users to provide a list of writers, no?

api-metastore/src/main/java/org/zalando/nakadi/controller/EventTypeController.java
```diff
@@ -68,10 +71,12 @@ public EventTypeController(final EventTypeService eventTypeService,
     }
 
     @RequestMapping(method = RequestMethod.GET)
-    public ResponseEntity<?> list() {
-        final List<EventType> eventTypes = eventTypeService.list();
+    public ResponseEntity<?> list(@Nullable @RequestParam final Set<String[]> writers) {
```
a1exsh, line 74: What I mean is this is a set of arrays now, we just want a set or array.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -122,6 +122,14 @@ paths:
             The flow id of the request, which is written into the logs and passed to called services. Helpful
             for operational troubleshooting and log analysis.
           type: string
+        - name: writers
+          in: query
+          description: |
+            Will query all subscriptions that provided user or service can write to based on authorization section. Service or user
+            should be listed explicitly on authorization section.
+          type: string
```
adyach, line 130: @burimshala do you think it is enough to have a single writer? could we extend it to array like 
`?writer_data_type=type1,type2&writer_data_value=value1,value2`. I also suggest to add readers right away or in the following PR.

burimshala: We already were going in that direction, but together with @a1exsh we decided not to because YAGNI and more problems were popping out.

api-metastore/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
```diff
@@ -52,10 +61,11 @@ public SubscriptionController(final SubscriptionService subscriptionService) {
             @RequestParam(value = "limit", required = false, defaultValue = "20") final int limit,
             @RequestParam(value = "offset", required = false, defaultValue = "0") final int offset,
             @RequestParam(value = "token", required = false) final String token,
+            @Nullable @RequestParam(value = "reader", required = false) final AuthorizationAttribute reader,
```
adyach, line 64: AuthorizationAttribute became a view. usually, view model is done via mappings.

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
```diff
@@ -10,13 +12,17 @@
...
-            final Optional<String> owningApplication, final Set<String> eventTypes, final int offset,
+            final Optional<String> owningApplication, final Set<String> eventTypes,
+            final Optional<AuthorizationAttribute> reader, final int offset,
             final Optional<SubscriptionTokenLister.Token> token, final int limit, final boolean showStatus) {
 
         final UriComponentsBuilder urlBuilder = UriComponentsBuilder.fromPath("/subscriptions");
         if (!eventTypes.isEmpty()) {
             urlBuilder.queryParam("event_type", eventTypes.toArray());
         }
+        reader.ifPresent(_reader -> {
```
adyach, line 23: `_reader ` is `_` conventional notation for lambda vars?

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
```diff
@@ -10,13 +12,17 @@
...
+            urlBuilder.queryParam("reader", AuthorizationAttributeQueryParser.getQuery(_reader));
```
adyach, line 24: and again, that is a good place where model -> view mapping would be good, so you do not need to integrate static method into AuthorizationAttributeQueryParser

docs/_data/nakadi-event-bus-api.yaml
adyach: is the intended request gonna happen with single user all the time or you will do several requests for the same subscription?

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
burimshala: Nope, I just didn't have a better choice.

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
burimshala: Can you share an example? I decided to encapsulate this on the parser class so the logic if parsing is kept in one place only.

api-metastore/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
burimshala: is this considered mapping https://github.com/zalando/nakadi/pull/1292/files#diff-7348bf9e0c2e92e73a523b0b0cab97c411ab6243f153a9a02783638fb9632de2R51 ?

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
adyach, line 23:
```suggestion
        readerOpt.ifPresent(reader -> {
```

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
```diff
@@ -10,13 +12,17 @@
 public class SubscriptionsUriHelper {
 
     public static PaginationLinks.Link createSubscriptionListLink(
-            final Optional<String> owningApplication, final Set<String> eventTypes, final int offset,
+            final Optional<String> owningApplication, final Set<String> eventTypes,
+            final Optional<AuthorizationAttribute> reader, final int offset,
```
adyach, line 16:
```suggestion
            final Optional<AuthorizationAttribute> readerOpt, final int offset,
```

docs/_data/nakadi-event-bus-api.yaml
burimshala: The intended logic is to have only 1 user or service to query subscriptions.

api-metastore/src/main/java/org/zalando/nakadi/controller/SubscriptionController.java
adyach: @InitBinde magic is magic, it is not clear how that works. Instead of having view for the user request we are fine to prepare view as a model. In general it can be done, but that AuthorizationAttribute has no relation to `reader` except it has the same fields that's why I prefer to have a separate view and map it to model. I will not argue for that just keep it as it is as you see fit :)

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
adyach: static method looks unnatural in that class. maybe there is the same @InitBinder approach, but to map it back.

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
adyach: in the project we use lambdas without underscore, i suggest we keep it consistent

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
```diff
@@ -97,6 +103,25 @@ public void whenEventExistsFindByNameReturnsSomething() throws Exception {
         assertThat(persistedEventType, notNullValue());
     }
 
+    @Test
+    public void whenEventExistsFindByAuthorizationReturnsSomething() throws Exception {
```
a1exsh, line 107:
```suggestion
    public void whenEventTypeExistsFindByAuthorizationReturnsSomething() throws Exception {
```

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
```diff
@@ -170,7 +173,7 @@ public void whenListSubscriptionsLimitAndOffsetAreRespected() throws ServiceTemp
         testSubscriptions.subList(3, testSubscriptions.size()).clear();
 
         final List<Subscription> subscriptions = repository.listSubscriptions(
-                emptySet(), Optional.of(owningApp), 2, 3);
+                emptySet(), Optional.of(owningApp), Optional.empty(), 2, 3);
```
a1exsh, line 176: All tests here are using empty value for the newly introduced parameter.  Do we want an actual test that passes some reader filter?

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionTokenListerTest.java
```diff
@@ -175,7 +180,7 @@ public void whenListSubscriptionsByMultipleEventTypesThenOk() throws ServiceTemp
                 .collect(toList());
 
         final List<Subscription> subscriptions = subscriptionTokenLister.listSubscriptions(ImmutableSet.of(et1, et2),
-                Optional.empty(), null, 10).getItems();
+                Optional.empty(), Optional.empty(), null, 10).getItems();
```
a1exsh, line 183: Same here: only empty values used for the new parameter.

api-metastore/src/main/java/org/zalando/nakadi/model/AuthorizationAttributeQueryParser.java
```diff
@@ -0,0 +1,29 @@
...
+public class AuthorizationAttributeQueryParser extends PropertyEditorSupport {
+
+    @Override
+    public void setAsText(final String text) throws IllegalArgumentException {
+        if (StringUtils.isEmpty(text)) {
+            throw new IllegalArgumentException("Authorization query is empty");
+        } else {
+            final String[] authorizationQuery = text.split(":");
+            if (authorizationQuery.length != 2) {
+                throw new IllegalArgumentException("Authorization format is incorrect. Should be data_type:data_value");
```
a1exsh, line 18:
```suggestion
                throw new IllegalArgumentException("Authorization format is incorrect. Should be data_type:value");
```
Matches the API definition better.

api-metastore/src/test/java/org/zalando/nakadi/controller/SubscriptionControllerTest.java
```diff
@@ -309,10 +314,12 @@ private ResultActions getSubscriptions() throws Exception {
         return mockMvc.perform(get("/subscriptions"));
     }
 
-    private ResultActions getSubscriptions(final Set<String> eventTypes, final String owningApp, final int offset,
-                                           final int limit) throws Exception {
+    private ResultActions getSubscriptions(final Set<String> eventTypes, final String owningApp,
+                                           final Optional<AuthorizationAttribute> readers,
```
a1exsh, line 318:
```suggestion
                                           final Optional<AuthorizationAttribute> reader,
```
singular for now

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -102,6 +104,17 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
                 new EventTypeMapper());
     }
 
+    public List<EventType> list(final AuthorizationAttribute writers) {
```
a1exsh, line 107:
```suggestion
    public List<EventType> list(final AuthorizationAttribute writer) {
```

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -167,6 +174,11 @@ static void applyFilter(
                     .map(et -> format("\"{0}\"", et))
                     .forEach(params::add);
         }
+        if(reader.isPresent()){
```
a1exsh, line 177:
```suggestion
        if (reader.isPresent()) {
```
please ;)

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -122,6 +122,14 @@ paths:
             The flow id of the request, which is written into the logs and passed to called services. Helpful
             for operational troubleshooting and log analysis.
           type: string
+        - name: writer
+          in: query
+          description: |
+            Will query all subscriptions that the given user or service can write to based on authorization section. Service or user
```
a1exsh, line 128:
```suggestion
            Will query all event types that the given user or service can write to based on authorization section. Service or user
```

a1exsh:
Is there any chance to have an Acceptance Test for the new functionality?

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionTokenListerTest.java
burimshala: Would an acceptance test justify few unit tests?

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
burimshala: @InitBinder does not work here unfortuantely.

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionTokenListerTest.java
a1exsh: I would happily exchange an acceptance test for any number of mocked tests.  A unit test doesn't hurt, but if it requires mocking, let's go for acceptance test instead. ;)

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -63,8 +64,8 @@
 
 import java.io.Closeable;
 import java.io.IOException;
-import java.util.Date;
```
a1exsh, line 66: nitpick: this just got moved one line below, for no good reason

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
```diff
@@ -10,13 +12,17 @@
...
             final Optional<SubscriptionTokenLister.Token> token, final int limit, final boolean showStatus) {
 
         final UriComponentsBuilder urlBuilder = UriComponentsBuilder.fromPath("/subscriptions");
         if (!eventTypes.isEmpty()) {
             urlBuilder.queryParam("event_type", eventTypes.toArray());
         }
+        reader.ifPresent(readerOpt -> {
```
a1exsh, line 23: Not sure I really follow the naming here: `readerOpt` is no longer optional, as we know it's present, but OK.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
burimshala: My IDE is grouping dependencies by package. If not line 66, it would be 67.

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionsUriHelper.java
burimshala: Actually, it is optional because on SubscriptionRepository I am not doing a method overload as in EventTypeRepository, hence passing a nullable all the way down.

The reason for that is the implementation complexity would increase.

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -102,6 +104,17 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
...
     }
 
+    public List<EventType> list(final AuthorizationAttribute writer) {
+        return jdbcTemplate.query(
+                "SELECT et_event_type_object " +
+                        "FROM zn_data.event_type," +
+                        "jsonb_to_recordset(et_event_type_object->'authorization'->'writers') " +
+                        "AS writers(data_type text, value text) " +
+                        "WHERE writers.data_type = ? AND writers.value = ?",
+                Set.of(writer.getDataType(), writer.getValue()).toArray(),
```
a1exsh, line 114: Is the order well-defied here as we pull from the set to an array?..

burimshala: Very good point

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
a1exsh: Well, at least it was alphabetically sorted within the `java.util` group ;)

a1exsh:
:+1:

burimshala:
👍

PR #1295, opened 2021-06-28
503 when token info down

thomasabraham:
#  Differentiating tokenInfo endpoint problems from an invalid token. 

> Zalando ticket : https://github.bus.zalan.do/aruha/team-aruha/issues/82

## Description
The DefaultTokenInfoRequestExecutor treats not being able to validate a
token, probably because the tokenInfo endpoint is not working, as the
token not being valid. Adding NakadiTokenInfoRequestExecutor to
differentiate the problems with tokenInfo endpoint and throw a
ServiceTemporarilyUnavailableException exception when the tokenInfo is
not working.


a1exsh:
:+1:

thomasabraham:
:+1:

PR #1296, opened 2021-06-30
Leaderless partitions should not fail publishing

burimshala:
# One-line summary

> Zalando ticket : 764

## Description
At the time when Nakadi requests partition information from Kafka Producer, it can happen that a partition can have no broker as a leader. In this case, chances are that the event will not get published by Kafka Producer.

Currently, Nakadi throws a null pointer exception that fails the entire batch.

## Review
- [X] Tests
- [ ] Documentation

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -351,7 +351,8 @@ public class KafkaTopicRepository implements TopicRepository {
             final Map<BatchItem, CompletableFuture<Exception>> sendFutures = new HashMap<>();
             for (final BatchItem item : batch) {
                 if (item.getBrokerId() == null) {
-                    item.updateStatusAndDetail(EventPublishingStatus.FAILED, String.format("No leader for partition."));
+                    item.updateStatusAndDetail(EventPublishingStatus.FAILED, String.format("No leader for partition: %s, topic: %s.", item.getPartition(), topicId));
+                    LOG.warn("Failed to publish. No leader fo partition [{}], topic [{topicId}].", item.getPartition(), topicId);
                     continue;
                 }
                 item.setStep(EventPublishingStep.PUBLISHING);
--- a/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -353,8 +353,8 @@ public class KafkaTopicRepository implements TopicRepository {
                 if (item.getBrokerId() == null) {
                     item.updateStatusAndDetail(EventPublishingStatus.FAILED,
                             String.format("No leader for partition: %s, topic: %s.", item.getPartition(), topicId));
-                    LOG.warn("Failed to publish to kafka. No leader for partition [{}], topic [{topicId}].",
-                            item.getPartition(), topicId);
+                    LOG.error("Failed to publish to kafka. No leader for ({}:{}).",
+                            topicId, item.getPartition());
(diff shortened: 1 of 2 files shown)


core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -348,6 +350,11 @@ public void syncPostBatch(
             int shortCircuited = 0;
             final Map<BatchItem, CompletableFuture<Exception>> sendFutures = new HashMap<>();
             for (final BatchItem item : batch) {
+                if (item.getBrokerId() == null) {
+                    item.updateStatusAndDetail(EventPublishingStatus.FAILED, String.format("No leader for partition: %s, topic: %s.", item.getPartition(), topicId));
+                    LOG.warn("Failed to publish. No leader fo partition [{}], topic [{topicId}].", item.getPartition(), topicId);
```
adyach, line 355:
```suggestion
                    LOG.warn("Failed to publish to Kafka. No leader for partition [{}], topic [{topicId}].", item.getPartition(), topicId);
```

adyach: and probably it will be a lot of messages, maybe we want to cut it somehow

adyach: maybe the same way as for short circuiting

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -348,6 +350,13 @@ public void syncPostBatch(
             int shortCircuited = 0;
             final Map<BatchItem, CompletableFuture<Exception>> sendFutures = new HashMap<>();
             for (final BatchItem item : batch) {
+                if (item.getBrokerId() == null) {
+                    item.updateStatusAndDetail(EventPublishingStatus.FAILED,
+                            String.format("No leader for partition: %s, topic: %s.", item.getPartition(), topicId));
+                    LOG.warn("Failed to publish to kafka. No leader for partition [{}], topic [{topicId}].",
```
adyach, line 356: in Kafka it is pair (topic, partition) it will be more readable in the logs. also for consistency I would use either `{}` or both named `{topicId}` and `{partitionId}`

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
burimshala: https://github.com/zalando/nakadi/blob/master/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java#L367 This doesn't look too short.

This event usually doesn't happen, but if it does wouldn't it be better to have more logs?

adyach: I mean log less lines. compact them in some way, because most probably there will be a lot of same lines which we can avoid logging and the same lines do not bring any value.

burimshala: I prefer to see logs as long as the issue is happening. You can aggregate log lines to see what is the severity and once the logline stops appearing you know that the problem is gone.

adyach: my suggestion does not exclude logging while issues is happening. you will still be able to see logs, just logs you need and not repeating once. as we know the issue can persist for hours eating resources while logging.

adyach:
👍

burimshala:
👍

adyach:
👍

PR #1300, opened 2021-07-06
Added labels and annotations to the Nakadi API

thomasabraham:
# Added labels and annotations to the Nakadi API

> Zalando ticket : team-aruha/issues/752

## Description
This is the first step in introducing labels and annotations to the Nakadi API.
Actual implementation will follow the API definition.

### DIFF
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -127,7 +127,8 @@ paths:
           description: |
             Will query all event types that the given user or service can write to based on authorization section. Service or user
             should be listed explicitly on authorization section.
-            example: "service:stups_nakadi-ui"
+
+            Example: "service:stups_nakadi-ui"
           type: string
           required: false
       responses:
@@ -1183,7 +1184,8 @@ paths:
           description: |
             Will query all subscriptions that the given user or service can read from, based on authorization section. Service or user
             should be listed explicitly on authorization section.
-            example: "service:stups_nakadi-ui"
+
+            Example: "service:stups_nakadi-ui"
           type: string
           required: false
       responses:
@@ -3125,24 +3127,20 @@ definitions:
     description: >
       # [Work in progress] Annotations of the Nakadi resource.
 
-      A Nakadi event-type or a subscription is considered as a Nakadi resource.
-      A Nakadi resource can have a set of annotations associated with it.
-      They follow the same rules of annotations in Kubernetes resources.
-        For keys:
-      > Annotations are key/value pairs. Valid annotation keys have two segments:
-      an optional prefix and name, separated by a slash (`/`). The name segment is
-      required and must be 63 characters or less, beginning and ending with an
-      alphanumeric character (`[a-z0-9A-Z]`) with dashes (`-`), underscores (`_`),
-      dots (`.`), and alphanumerics between. The prefix is optional. If specified,
-      the prefix must be a DNS subdomain: a series of DNS labels separated by dots
-      (`.`), not longer than 253 characters in total, followed by a slash (`/`).
+      A Nakadi event-type or a subscription is considered as a Nakadi resource. A Nakadi resource can have a set of
+      annotations associated with it. The syntax of annotations is adopted from Kubernetes.
 
-        For values:
-      Kubernetes don't have any restrictions for the values of annotations.
-      But Nakadi limits this to 1000 bytes at the moment, this can be relaxed in the future.
+        From the Kubernetes documentation about annotation keys:
+      > Annotations are key/value pairs. Valid annotation keys have two segments: an optional prefix and name, separated
+      by a slash (`/`). The name segment is required and must be 63 characters or less, beginning and ending with an
+      alphanumeric character (`[a-z0-9A-Z]`) with dashes (`-`), underscores (`_`), dots (`.`), and alphanumerics
+      between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels
+      separated by dots (`.`), not longer than 253 characters in total, followed by a slash (`/`).
 
+        Values of the annotations are currently limited to 1000 bytes. This can be relaxed in the future.
     example:
-      nakadi-sql/query-input: "true"
+      nakadi.io/internal-event-type: "true"
+      criticality: "low"
     additionalProperties:
       type: string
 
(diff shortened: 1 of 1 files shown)


docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -3110,6 +3120,58 @@ definitions:
...
+      (`.`), not longer than 253 characters in total, followed by a slash (`/`).
+
+        For values:
+      Kubernetes don't have any restrictions for the values of annotations.
+      But Nakadi limits this to 1000 bytes at the moment, this can be relaxed in the future.
+
+    example:
+      nakadi-sql/query-input: "true"
+    additionalProperties:
+      type: string
```
a1exsh, line 3147: I thought it's only a boolean flag, like:
```suggestion
    additionalProperties: true
```
Or am I missing something?

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -127,9 +127,9 @@ paths:
           description: |
             Will query all event types that the given user or service can write to based on authorization section. Service or user
             should be listed explicitly on authorization section.
+            example: "service:stups_nakadi-ui"
```
adyach, line 130:
```suggestion
          example: "service:stups_nakadi-ui"
```

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -1181,9 +1183,9 @@ paths:
           description: |
             Will query all subscriptions that the given user or service can read from, based on authorization section. Service or user
             should be listed explicitly on authorization section.
+            example: "service:stups_nakadi-ui"
```
adyach, line 1186: why did you move example into description ?

thomasabraham: Not supported by Swagger 2.0.
We are using OpenAPI 3.0 syntax without actually adopting it completely.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -3110,6 +3120,58 @@ definitions:
...
 
 
+  Annotations:
+    type: object
+    description: >
+      # [Work in progress] Annotations of the Nakadi resource.
+
+      A Nakadi event-type or a subscription is considered as a Nakadi resource.
+      A Nakadi resource can have a set of annotations associated with it.
+      They follow the same rules of annotations in Kubernetes resources.
```
adyach, line 3130: can we avoid using Kubernetes in Nakadi API doc? in general there is no direct correlation between K8s and Nakadi annotations even though they can have a similar or the same set of rules.

docs/_data/nakadi-event-bus-api.yaml
thomasabraham: We can also specify the type of values allowed for additional properties.

docs/_data/nakadi-event-bus-api.yaml
thomasabraham: Good point.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -3110,6 +3120,58 @@ definitions:
...
+      dots (`.`), and alphanumerics between. The prefix is optional. If specified,
+      the prefix must be a DNS subdomain: a series of DNS labels separated by dots
+      (`.`), not longer than 253 characters in total, followed by a slash (`/`).
+
+        For values:
+      Kubernetes don't have any restrictions for the values of annotations.
+      But Nakadi limits this to 1000 bytes at the moment, this can be relaxed in the future.
+
+    example:
+      nakadi-sql/query-input: "true"
```
adyach, line 3145: I would use more understandable example for the user who is not aware of Nakadi SQL. and this is how I thought it will look like e.g.: 
```
{
  "key1": "value1", 
  "key2": "value2"
}
```

docs/_data/nakadi-event-bus-api.yaml
adyach: can we maybe make it from the capital letter and add a new line that it will be separated from the description ?

docs/_data/nakadi-event-bus-api.yaml
a1exsh: Ah, OK.  But do we really want to restrict it to `string` only?  I think in K8s structured annotation values are allowed (but not for labels).

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -3110,6 +3120,58 @@ definitions:
...
+    type: object
+    description: >
+      # [Work in progress] Annotations of the Nakadi resource.
+
+      A Nakadi event-type or a subscription is considered as a Nakadi resource.
+      A Nakadi resource can have a set of annotations associated with it.
+      They follow the same rules of annotations in Kubernetes resources.
+        For keys:
+      > Annotations are key/value pairs. Valid annotation keys have two segments:
+      an optional prefix and name, separated by a slash (`/`). The name segment is
```
adyach, line 3133: I understand we want to impose as many restrictions as possible because it is easier to relax them than adding them later, but at the same time I am not sure we need be that strict about format for the key: `Valid annotation keys have two segments: an optional prefix and name, separated by a slash (`/`).` The slash could be a part of the allowed symbols, that simplifies annotations / labels definition, but at the same type flexible enough and restrictive

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -3110,6 +3120,58 @@ definitions:
...
+      required and must be 63 characters or less, beginning and ending with an
+      alphanumeric character (`[a-z0-9A-Z]`) with dashes (`-`), underscores (`_`),
+      dots (`.`), and alphanumerics between. The prefix is optional. If specified,
```
adyach, line 3136: > If specified, the prefix must be a DNS subdomain

it is not clear what relation to Nakadi annotation / labels has DNS subdomain

docs/_data/nakadi-event-bus-api.yaml
thomasabraham: I'll change the example to a more generic one.

docs/_data/nakadi-event-bus-api.yaml
thomasabraham: I see them as JSON encoded as string.

docs/_data/nakadi-event-bus-api.yaml
adyach: aligned with Thomas offline. the K8s rules to define annotation / labels format is optional, I misread. 
it would be nice to clear separate that Nakadi would like to have annotations / labels defined as K8s formatted annotations / labels with prefixe, name and backslash and the strict requirement on the allowed symbols and length.

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -3110,6 +3122,52 @@ definitions:
...
+      annotations associated with it. The syntax of annotations is adopted from Kubernetes.
+
+        From the Kubernetes documentation about annotation keys:
+      > Annotations are key/value pairs. Valid annotation keys have two segments: an optional prefix and name, separated
+      by a slash (`/`). The name segment is required and must be 63 characters or less, beginning and ending with an
+      alphanumeric character (`[a-z0-9A-Z]`) with dashes (`-`), underscores (`_`), dots (`.`), and alphanumerics
+      between. The prefix is optional. If specified, the prefix must be a DNS subdomain: a series of DNS labels
+      separated by dots (`.`), not longer than 253 characters in total, followed by a slash (`/`).
+
+        Values of the annotations are currently limited to 1000 bytes. This can be relaxed in the future.
```
a1exsh, line 3140:
```suggestion
      Values of the annotations are currently limited to 1000 bytes. This can be relaxed in the future.
```
Feels like this should be indented at the same depth as the first paragraph above.  See also "From the K8s docs..."

a1exsh:
:+1:

docs/_data/nakadi-event-bus-api.yaml
thomasabraham: This is an intentional to make it render properly in the API portal.

thomasabraham:
:+1:

docs/_data/nakadi-event-bus-api.yaml
a1exsh: Ah, where can I see it?

PR #1299, opened 2021-07-06
Always clean streaming session in ZooKeeper

a1exsh:
# One-line summary

Internal ticket: 788

## Review
- [x] Tests
- [ ] Documentation

PR #1301, opened 2021-07-06
Safety check in unregister session method

a1exsh:
It should not throw exception if the node was already gone (or wasn't there in
the first place).  Otherwise our cleanup procedure will handle the exception
and try to unregister again and again.

Follow-up on #1299 

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
+++ b/core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
@@ -173,6 +173,8 @@ public abstract class AbstractZkSubscriptionClient implements ZkSubscriptionClie
     public final void unregisterSession(final Session session) {
         try {
             getCurator().delete().guaranteed().forPath(getSubscriptionPath("/sessions/" + session.getId()));
+        } catch (final KeeperException.NoNodeException ke) {
+            // It's OK that the session node was not there: all we want is to make sure it's deleted.
         } catch (final Exception e) {
             throw new NakadiRuntimeException(e);
         }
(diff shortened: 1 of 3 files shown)


core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
```diff
@@ -173,6 +173,8 @@ public final void registerSession(final Session session) {
     public final void unregisterSession(final Session session) {
         try {
             getCurator().delete().guaranteed().forPath(getSubscriptionPath("/sessions/" + session.getId()));
```
antban, line 175: @a1exsh What does it mean - guaranteed() ?

a1exsh: Well, from the docs: `Solves edge cases where an operation may succeed on the server but connection failure occurs before a response can be successfully returned to the client.`
https://curator.apache.org/apidocs/org/apache/curator/framework/api/Guaranteeable.html

Are you hinting that it should already include ignoring `NoNodeException`?

adyach: it looks like that doc is more precise https://curator.apache.org/apidocs/org/apache/curator/framework/api/GuaranteeableDeletable.html

adyach:
👍

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/AbstractZkSubscriptionClient.java
a1exsh: I only see "wreak havoc" :D

a1exsh:
:+1:

PR #1304, opened 2021-07-07
Fine-tune unregisterSession

a1exsh:
# One-line summary

In case ZooKeeper client has a problem we might not actually delete the
session's node, so let it try another time if needed by not clearing the flag
just yet.

Zalando ticket : 788


antban:
:+1:

a1exsh:
:+1:

PR #1305, opened 2021-07-07
Stop relying on Curator's guarantees

a1exsh:
The background attempts to retry the delete operations were causing more harm
than good.

We do not rely on the deletes to always work as the nodes are ephemeral and
will be removed eventually within the next ZooKeeper client rotation interval.

## Review
- [ ] Tests
- [ ] Documentation


antban:
:+1:

a1exsh:
:+1:

PR #1306, opened 2021-07-07
Less noisy logging in StreamingState

a1exsh:
The 'failed to write data to output' state is known to happen often, but it's
OK.  Don't log the full exception trace in this case.


adyach:
👍

a1exsh:
:+1:

PR #1307, opened 2021-07-09
publish nakadi audit log directly to kafka

adyach:
separate nakadi.audit.log from nakadi.access.log publishing to reduce amounts of events dropped due to event processor internal queue overload


a1exsh:
The related test is still failing, though.

a1exsh:
:+1:

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventsProcessor.java
```diff
@@ -215,16 +205,19 @@ private void sendLeftovesOnShutdown(final Map<String, BatchedRequest> batchesBei
...
-                .put("flow_id", FlowIdUtils.peek());
-        event.put("metadata", metadata);
-
+    public void queueEvent(final String etName, final JSONObject event) {
         if (!eventsQueue.offer(new EventToPublish(etName, event))) {
             LOG.warn("Rejecting events to be queued for {} due to queue overload", etName);
         }
     }
 
+    public void sendEventsDisabledAuthz(final String events, final String eventType) {
```
antban, line 214: Why not `JsonArray`? 
send(string, string) is not bug-prone.

PR #1308, opened 2021-07-09
Validate owning application, if this functionality is provided by plugins.

antban:
Right now event type evolution includes modification of different fields. And nakadi may have some plugins, that are allowing this evolution to proceed. 

Unfortunately, the plugins where not actually used on event-type/subscription creation/update. This PR reintroduces owning_application validation for event types and subscriptions. 

In order to have flawless roll out, the both checks are guarded with feature toggles.

Also, some tests were cleaned up


antban:
:+1:

api-metastore/src/main/java/org/zalando/nakadi/controller/EventTypeController.java
```diff
@@ -90,11 +90,10 @@ public void initBinder(final WebDataBinder binder) {
 
     @RequestMapping(method = RequestMethod.POST)
     public ResponseEntity<?> create(@Valid @RequestBody final EventTypeBase eventType,
-                                    final Errors errors,
-                                    final NativeWebRequest request)
+                                    final Errors errors)
             throws TopicCreationException, InternalNakadiException, NoSuchPartitionStrategyException,
             DuplicatedEventTypeNameException, InvalidEventTypeException, ValidationException,
-            FeatureNotAvailableException {
+            FeatureNotAvailableException, WrongOwningApplicationException {
```
a1exsh, line 96: Can we name it `InvalidOwningApplicationException` instead?  I believe "wrong" can be interpreted as valid, but not the expected one, though we can't have an expectation here: we don't know better than the user.

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
```diff
@@ -166,7 +168,7 @@ public Subscription updateSubscription(final String subscriptionId, final Subscr
 
         authorizationValidator.authorizeSubscriptionAdmin(old);
 
-        subscriptionValidationService.validateSubscriptionChange(old, newValue);
+        subscriptionValidationService.validateSubscriptionOnChange(old, newValue);
```
a1exsh, line 171:
```suggestion
        subscriptionValidationService.validateSubscriptionOnUpdate(old, newValue);
```

If it's called `CRUD`...

api-metastore/src/test/java/org/zalando/nakadi/service/EventTypeServiceTest.java
```diff
@@ -311,7 +327,55 @@ public void whenEventTypeUpdatedThenKPIEventSubmitted() throws Exception {
     }
 
     @Test
-    public void whenEventTypeDeletedThenKPIEventSubmitted() throws Exception {
+    public void whenEventTypeOwningApplicationUpdatedThanItIsValidated() {
```
a1exsh, line 330:
```suggestion
    public void whenEventTypeOwningApplicationUpdatedThenItIsValidated() {
```

a1exsh: See also the few next test names.

a1exsh:
`<=` https://github.com/zalando/nakadi/pull/1309

a1exsh:
:+1:

thomasabraham:
I'm resolving the conflicts.

a1exsh:
:+1:

thomasabraham:
:+1:

antban:
:-1:

antban:
:+1:

api-consumption/src/main/java/org/zalando/nakadi/controller/advice/SubscriptionStreamExceptionHandler.java
```diff
@@ -19,9 +19,10 @@
 @ControllerAdvice(assignableTypes = SubscriptionStreamController.class)
 public class SubscriptionStreamExceptionHandler implements AdviceTrait {
 
-    @ExceptionHandler(WrongStreamParametersException.class)
-    public ResponseEntity<Problem> handleWrongStreamParametersException(final WrongStreamParametersException exception,
-                                                                        final NativeWebRequest request) {
+    @ExceptionHandler(InvalidStreamParametersException.class)
+    public ResponseEntity<Problem> handleWrongStreamParametersException(
```
a1exsh, line 23:
```suggestion
    public ResponseEntity<Problem> handleInvalidStreamParametersException(
```

a1exsh:
:+1:

antban:
:+1:

PR #1312, opened 2021-07-15
Bugfix: Do not overwrite annotations and labels when missing.

thomasabraham:
# Bugfix: Do not overwrite annotations and labels when missing.

> Zalando ticket : team-aruha/issues/752

## Description
When annotations or labels is missing in the input, the existing value
should be preserved. This fix remove creation of empty annotation and
label objects when they are missing in the imput, thereby preserving the
existing values.

### DIFF
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/EventTypeAT.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/EventTypeAT.java
@@ -1,6 +1,8 @@
 package org.zalando.nakadi.webservice;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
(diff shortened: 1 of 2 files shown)


a1exsh:
Maybe we could have started with a test for this bug before fixing it?

thomasabraham:
We do have a test, but the test uses Java Objects on the client side when serialize to null rather than missing attribute in JSON.

a1exsh:
Mainly the question boils down to: how do we know if this change fixes the problem and that further changes will not re-introduce it, if we don't have a test for this specific interaction that can catch it?

thomasabraham:
Good point about further change reintroducing it, I'll add a test

a1exsh:
:+1:

thomasabraham:
:+1:

PR #1322, opened 2021-07-23
Fix partition list accepts cursor

samizzy:
Added support for using cursors on list partitions endpoint for PartitionController


api-metastore/src/main/java/org/zalando/nakadi/controller/PartitionsController.java
```diff
@@ -36,19 +42,20 @@
...
+import static java.util.Collections.emptyList;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
-import static org.springframework.http.ResponseEntity.noContent;
-import static org.springframework.http.ResponseEntity.ok;
-
+@Validated
```
thomasabraham, line 58: Please confirm that `@Validated` annotation is required here.

api-metastore/src/main/java/org/zalando/nakadi/controller/util/CursorUtil.java
```diff
@@ -0,0 +1,55 @@
...
+import org.zalando.nakadi.view.CursorLag;
+
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class CursorUtil {
+
+    public static Stream<CursorLag> toCursorLagStream(final List<Cursor> cursorList, final String eventTypeName,
```
thomasabraham, line 20: I think this method can be added to `CursorOperationsService` as a non-static one.

thomasabraham: Also adding a `CursorConverter` instance to `CursorOperationsService` will simplify the code.

api-metastore/src/main/java/org/zalando/nakadi/controller/PartitionsController.java
samizzy: Yea not needed, forgot to remove it :)

api-metastore/src/main/java/org/zalando/nakadi/controller/util/CursorUtil.java
samizzy: I refactored `CursorOperationsService` to include these methods but I have still kept `CursorConverter` as method argument because its only needed for this one method.

thomasabraham:
:+1:

api-metastore/src/main/java/org/zalando/nakadi/controller/PartitionsController.java
```diff
@@ -103,59 +111,174 @@ private static NakadiCursor selectLast(final List<Timeline> activeTimelines, fin
...
+                        cursorConverter)
+                .map(cl -> new EventTypePartitionView(
+                                cl.getPartition(),
+                                cl.getOldestAvailableOffset(),
+                                cl.getNewestAvailableOffset(),
+                                cl.getUnconsumedEvents())
+                ).collect(Collectors.toList());
+
+        if (cursorLags.size() != cursorList.size()){
+            throw new NakadiBaseException(String.format("Problem with cursors '%s'", cursorLags));
```
a1exsh, line 239: Can we provide a more specific message than just "problem with cursors"?

api-metastore/src/main/java/org/zalando/nakadi/controller/PartitionsController.java
```diff
@@ -103,59 +111,174 @@ private static NakadiCursor selectLast(final List<Timeline> activeTimelines, fin
...
+        } else {
+            lastStats = timelineService.
+                    getTopicRepository(timelines.get(timelines.size() - 1))
+                    .loadTopicStatistics(singletonList(timelines.get(timelines.size() - 1)));
+        }
+        return lastStats;
+    }
+
+    private List<EventTypePartitionView> getViewsWithUnconsumedEvents
+            (final String eventTypeName,
```
a1exsh, line 222: Why the line break before argument list?

samizzy: I think I was concerned about the 120 char limit for a line, but now i see that keeping the first argument on the same line does not exceed it.

api-metastore/src/main/java/org/zalando/nakadi/controller/PartitionsController.java
samizzy: Sure, earlier there was just an empty exception being thrown. Will come up with better message.

samizzy: done

api-metastore/src/main/java/org/zalando/nakadi/controller/PartitionsController.java
```diff
@@ -103,59 +111,174 @@ private static NakadiCursor selectLast(final List<Timeline> activeTimelines, fin
...
+        } else {
+            lastStats = timelineService.
+                    getTopicRepository(timelines.get(timelines.size() - 1))
+                    .loadTopicStatistics(singletonList(timelines.get(timelines.size() - 1)));
+        }
+        return lastStats;
+    }
+
+    private List<EventTypePartitionView> getViewsWithUnconsumedEvents(  final String eventTypeName,
+                                                                        final List<Cursor> cursorList) {
```
a1exsh, line 222:
```suggestion
    private List<EventTypePartitionView> getViewsWithUnconsumedEvents(final String eventTypeName,
                                                                      final List<Cursor> cursorList) {
```

api-metastore/src/main/java/org/zalando/nakadi/controller/PartitionsController.java
```diff
@@ -103,59 +111,174 @@ private static NakadiCursor selectLast(final List<Timeline> activeTimelines, fin
...
+                                cl.getPartition(),
+                                cl.getOldestAvailableOffset(),
+                                cl.getNewestAvailableOffset(),
+                                cl.getUnconsumedEvents())
+                ).collect(Collectors.toList());
+
+        if (cursorLags.size() != cursorList.size()){
+            throw new NakadiBaseException(String.format("Problem with cursors, input cursors (%s) " +
+                    "and output cursors (%s) length mismatch for input '%s'"
+                    , cursorList.size(), cursorLags.size(), cursorLags));
```
a1exsh, line 239:
```suggestion
            throw new NakadiBaseException(String.format("Input cursors (%d) " +
                    "and output cursors (%d) length mismatch for input '%s'"
                    , cursorList.size(), cursorLags.size(), cursorList));
```

samizzy: done

a1exsh:
:+1:

samizzy:
👍🏼

thomasabraham:
:+1:

PR #1323, opened 2021-08-09
[WiP] Use tracing ScopeManager to set the active span

a1exsh:
# One-line summary

This is the recommended way to enable discovery of the current active span
instead of passing the "parent" span explicitly everywhere.


api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/CleanupState.java
```diff
@@ -40,7 +40,6 @@ public void onEnter() {
             } catch (final IOException e) {
                 getLog().error("Unexpected fail to release zk connection", e);
             }
-            getContext().getCurrentSpan().finish();
```
antban, line 43: Seems that it was a wrong place to call `finish()`, but the place in `streamInternal()` is also not a good one (who puts this span there?).

a1exsh:
Obsoleted by #1324

PR #1324, opened 2021-08-10
Activate tracing spans

a1exsh:
# One-line summary

Start and finish spans in the same place, activate spans before usage.

## Deployment notes

Some span operation names were changed for consistency:
- `stream_events_request` -> `stream_events`
- `commit_events` -> `commit_cursors`
- `all_requests` -> `generic_request`

This may result in the need to update monitoring (streams, checks, alerts?).


core-common/src/main/java/org/zalando/nakadi/filters/TracingFilter.java
```diff
@@ -41,110 +41,146 @@ public TracingFilter(final AuthorizationService authorizationService) {
...
-            logOnEvent();
+            finish();
+        }
+
+        private void finish() {
+            traceResponse(TracingService.getActiveSpan(), response);
+        }
+    }
+
+    private class AsyncContextTracingWrapper extends AsyncContextWrapper {
```
a1exsh, line 73: Hm... maybe `TracingAsyncContextWrapper` is a better way to name it.

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
```diff
@@ -177,6 +179,7 @@ private ResponseEntity response(final EventPublishResult result) {
             case ABORTED:
                 return status(HttpStatus.UNPROCESSABLE_ENTITY).body(result.getResponses());
             default:
+                //TracingService.getActiveSpan().setTag("error", true);
```
a1exsh, line 182: The error tag gets overwritten in the TracingFilter...

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -283,29 +284,36 @@ private void validate(final List<BatchItem> batch, final EventType eventType, fi
...
         final String topic = activeTimeline.getTopic();
-        final Span publishSpan = TracingService.getNewSpanWithParent(parentSpan, "publishing_to_kafka")
-                .setTag(Tags.MESSAGE_BUS_DESTINATION.getKey(), topic);
-        try {
-            timelineService.getTopicRepository(eventType).syncPostBatch(topic, batch, eventType.getName(), delete);
+
+        final Span publishingSpan = TracingService.buildNewSpan("publishing_to_kafka")
+                .withTag(Tags.MESSAGE_BUS_DESTINATION.getKey(), topic)
+                .start();
+        try (Closeable ignored = TracingService.activateSpan(publishingSpan)) {
```
antban, line 306: Makes sense to utilize `TracingService.withActiveSpan()`, and log through `TracingService.log()` without providing a context

antban (changes requested):
Really nice improvement!
I am concerned about asyncListener in TracingFilter, seems that is should be done in a better way (async request processing is remove, therefore status codes are not shown)

api-consumption/build.gradle
```diff
@@ -62,8 +62,8 @@ dependencies {
     compile 'org.slf4j:slf4j-log4j12'
 
     // open tracing
-    compile 'io.opentracing:opentracing-api:0.33.0'
-    compile 'io.opentracing:opentracing-util:0.33.0'
+    compile "io.opentracing:opentracing-api:$opentracingVersion"
+    compile "io.opentracing:opentracing-util:$opentracingVersion"
```
antban, line 66: :+1:

api-consumption/src/main/java/org/zalando/nakadi/controller/SubscriptionStreamController.java
```diff
@@ -208,10 +206,10 @@ private StreamingResponseBody stream(final String subscriptionId,
                                          final Client client,
                                          final StreamParameters streamParameters) {
 
-        final Span requestSpan = TracingService.getActiveSpan();
-        requestSpan.setOperationName("stream_events")
+        TracingService.setOperationName("stream_events")
                 .setTag("subscription.id", subscriptionId);
 
+        final Span requestSpan = TracingService.getActiveSpan();
```
antban, line 212: Nice! 
Just wanted to mention, that from these 3 lines it is very much clear what is happening - we are settings properties for current span and then storing it for referencing later. Much much less time is needed to follow the code.

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
```diff
@@ -133,6 +131,10 @@ private ResponseEntity postEventInternal(final String eventTypeName,
             reportMetrics(eventTypeMetrics, result, totalSizeBytes, eventCount);
             reportSLOs(startingNanos, totalSizeBytes, eventCount, result, eventTypeName, client);
 
+            if (result.getStatus() == EventPublishingStatus.FAILED) {
```
antban, line 134: Was it because of custom 207 processing in TracingFilter?

core-common/src/main/java/org/zalando/nakadi/filters/TracingFilter.java
```diff
@@ -67,16 +110,39 @@ protected void doFilterInternal(final HttpServletRequest request,
...
-                    .setTag(Tags.ERROR, statusCode >= 400);
-        } finally {
-            span.finish();
+            if (!request.isAsyncStarted()) {
+                try {
+                    traceRequest(span, request, response);
+                } finally {
+                    span.finish();
+                }
+            }
```
antban, line 130:
```suggestion
            } else {
                request.getAsyncContext().addListener(new AsyncRequestSpanFinalizer(span, request, response));
            }
```

core-common/src/main/java/org/zalando/nakadi/filters/TracingFilter.java
```diff
@@ -67,16 +110,39 @@ protected void doFilterInternal(final HttpServletRequest request,
         try (Closeable ignored = TracingService.activateSpan(span)) {
             span.setTag("client_id", authorizationService.getSubject().map(Subject::getName).orElse("-"));
 
-            //execute request
             filterChain.doFilter(request, response);
 
+            if (request.isAsyncStarted()) {
+                request.getAsyncContext().addListener(new AsyncRequestSpanFinalizer(span, request, response));
+            }
```
antban, line 117: I suggest to move it below (see suggestion), so the finalization process will not be spread all over the method, but will concentrate in a single place.

antban (commented):
I see only 1 small improvement in tracing filter. everything else seems to be fine.

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
a1exsh: Yes, this should result in marking `207` responses with the `error` boolean flag.

antban:
:+1:

a1exsh:
:+1:

PR #1330, opened 2021-10-06
Add two more publishing metrics based on batch size (in bytes)

a1exsh:
Spikes in traffic are not always correlated in increas of number of events
published.  These new metrics should provide more direct correlation, as they
are based on batch size in bytes, not in the number of events.


adyach:
👍

a1exsh:
:+1:

PR #1341, opened 2021-11-08
Delete FeatureToggleService.java

a1exsh:
It was split into multiple files/projects, but the original one was not removed.


antban:
:+1:

a1exsh:
:+1:

PR #1343, opened 2021-11-08
Order returned Event Types by name

a1exsh:
Better late than never.

### DIFF
--- a/core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
+++ b/core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
@@ -99,7 +99,7 @@ public class EventTypeRepository extends AbstractDbRepository {
 
     public List<EventType> list() {
         return jdbcTemplate.query(
-                "SELECT et_event_type_object FROM zn_data.event_type",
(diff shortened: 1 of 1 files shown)


core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -109,7 +109,8 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
                         "FROM zn_data.event_type," +
                         "jsonb_to_recordset(et_event_type_object->'authorization'->'writers') " +
                         "AS writers(data_type text, value text) " +
-                        "WHERE writers.data_type = ? AND writers.value = ?",
+                        "WHERE writers.data_type = ? AND writers.value = ? " +
+                        "ORDER BY et_name",
```
adyach, line 113: can we have it as an option in the api? `sort_by=X`, X is event type name, owning_application etc

a1exsh: Not sure that brings a lot of value.  Filter by `owning_application` would be more appreciated, I guess.

adyach:
👍

antban:
:+1:

PR #1344, opened 2021-11-09
Fix orphan subscription

samizzy:
# One-line summary
Fix orphan subscription problem which occurs due to concurrent delete of event type and creation of a subscription of that event type.

### DIFF
--- a/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -398,27 +398,34 @@ public class EventTypeService {
 
     private Multimap<TopicRepository, String> deleteEventTypeWithSubscriptions(final String eventType) {
         try {
-            return transactionTemplate.execute(action -> {
-                eventTypeRepository.lockTable(EventTypeRepository.TableLock.ROW_EXCLUSIVE);
-                SubscriptionTokenLister.ListResult listResult = subscriptionTokenLister.listSubscriptions(
-                        ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), null, 100);
-                while (null != listResult) {
-                    listResult.getItems().forEach(s -> {
-                        try {
-                            subscriptionRepository.deleteSubscription(s.getId());
-                        } catch (final NoSuchSubscriptionException e) {
-                            // should not happen as we are inside transaction
-                            throw new InconsistentStateException("Subscription to be deleted is not found", e);
+            return eventTypeRepository.lockingTable(EventTypeRepository.TableLock.SHARE, transactionTemplate,
+                    action -> {
+                        SubscriptionTokenLister.ListResult listResult = subscriptionTokenLister.listSubscriptions(
+                                ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), null, 100);
+                        while (null != listResult) {
+                            listResult.getItems().forEach(s -> {
+                                try {
+                                    subscriptionRepository.deleteSubscription(s.getId());
+                                } catch (final NoSuchSubscriptionException e) {
+                                    // should not happen as we are inside transaction
+                                    throw new InconsistentStateException("Subscription to be deleted is not found", e);
+                                }
+                            });
+                            listResult = getNextResult(eventType, listResult);
                         }
-                    });
-                    listResult = null == listResult.getNext() ? null : subscriptionTokenLister.listSubscriptions(
-                            ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), listResult.getNext(), 100);
-                }
-                return deleteEventType(eventType);
-            });
+                        return deleteEventType(eventType);
+                    }
+            );
+
         } catch (final TransactionException e) {
             throw new InconsistentStateException("Failed to delete event-type because of race condition in DB", e);
         }
+    }
+
+    private SubscriptionTokenLister.ListResult getNextResult(final String eventType,
+                                                             final SubscriptionTokenLister.ListResult listResult) {
+        return null == listResult.getNext() ? null : subscriptionTokenLister.listSubscriptions(
+                ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), listResult.getNext(), 100);
     }
 
 
--- a/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -398,7 +398,8 @@ public class EventTypeService {
 
     private Multimap<TopicRepository, String> deleteEventTypeWithSubscriptions(final String eventType) {
         try {
-            return eventTypeRepository.lockingTable(EventTypeRepository.TableLockMode.SHARE, transactionTemplate,
+            return eventTypeRepository.lockingTable(EventTypeRepository.TableLockMode.ROW_EXCLUSIVE,
+                    transactionTemplate,
                     action -> {
                         SubscriptionTokenLister.ListResult listResult = listSubscriptions(eventType, null, 100);
                         while (null != listResult) {
@@ -410,7 +411,8 @@ public class EventTypeService {
                                     throw new InconsistentStateException("Subscription to be deleted is not found", e);
                                 }
                             });
-                            listResult = null == listResult.getNext() ? null : listSubscriptions(eventType, listResult.getNext(), 100);
+                            listResult = null == listResult.getNext() ? null :
+                                    listSubscriptions(eventType, listResult.getNext(), 100);
                         }
                         return deleteEventType(eventType);
                     }
--- a/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -399,6 +399,7 @@ public class EventTypeService {
     private Multimap<TopicRepository, String> deleteEventTypeWithSubscriptions(final String eventType) {
         try {
             return transactionTemplate.execute(action -> {
+                eventTypeRepository.lockTable(EventTypeRepository.TableLock.ROW_EXCLUSIVE);
                 SubscriptionTokenLister.ListResult listResult = subscriptionTokenLister.listSubscriptions(
                         ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), null, 100);
                 while (null != listResult) {
--- a/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -398,10 +398,9 @@ public class EventTypeService {
 
     private Multimap<TopicRepository, String> deleteEventTypeWithSubscriptions(final String eventType) {
         try {
-            return eventTypeRepository.lockingTable(EventTypeRepository.TableLock.SHARE, transactionTemplate,
+            return eventTypeRepository.lockingTable(EventTypeRepository.TableLockMode.SHARE, transactionTemplate,
                     action -> {
-                        SubscriptionTokenLister.ListResult listResult = subscriptionTokenLister.listSubscriptions(
-                                ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), null, 100);
+                        SubscriptionTokenLister.ListResult listResult = listSubscriptions(eventType, null, 100);
                         while (null != listResult) {
                             listResult.getItems().forEach(s -> {
                                 try {
@@ -411,7 +410,7 @@ public class EventTypeService {
                                     throw new InconsistentStateException("Subscription to be deleted is not found", e);
                                 }
                             });
-                            listResult = getNextResult(eventType, listResult);
+                            listResult = null == listResult.getNext() ? null : listSubscriptions(eventType, listResult.getNext(), 100);
                         }
                         return deleteEventType(eventType);
                     }
@@ -422,18 +421,17 @@ public class EventTypeService {
         }
     }
 
-    private SubscriptionTokenLister.ListResult getNextResult(final String eventType,
-                                                             final SubscriptionTokenLister.ListResult listResult) {
-        return null == listResult.getNext() ? null : subscriptionTokenLister.listSubscriptions(
-                ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), listResult.getNext(), 100);
+    private SubscriptionTokenLister.ListResult listSubscriptions(final String eventType,
+                                                                 final SubscriptionTokenLister.Token token,
+                                                                 final int limit) {
+        return subscriptionTokenLister.listSubscriptions(
+                ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), token, limit);
     }
-
 
     // TODO: This method should be fixed by creating proper db query.
     private boolean hasNonDeletableSubscriptions(final String eventTypeName) {
 
-        SubscriptionTokenLister.ListResult list = subscriptionTokenLister.listSubscriptions(
-                ImmutableSet.of(eventTypeName), Optional.empty(), Optional.empty(), null, 20);
+        SubscriptionTokenLister.ListResult list = listSubscriptions(eventTypeName, null, 20);
         while (null != list) {
             for (final Subscription sub : list.getItems()) {
                 if (!sub.getConsumerGroup().equals(nakadiSettings.getDeletableSubscriptionConsumerGroup())
@@ -442,8 +440,7 @@ public class EventTypeService {
                     return true;
                 }
             }
-            list = null == list.getNext() ? null : subscriptionTokenLister.listSubscriptions(
-                    ImmutableSet.of(eventTypeName), Optional.empty(), Optional.empty(), list.getNext(), 20);
+            list = null == list.getNext() ? null : listSubscriptions(eventTypeName, list.getNext(), 20);
         }
         return false;
     }
--- a/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
@@ -400,7 +400,7 @@ public class EventTypeService {
         try {
             return eventTypeRepository.lockingTable(EventTypeRepository.TableLockMode.ROW_EXCLUSIVE,
                     transactionTemplate,
-                    action -> {
+                    status -> {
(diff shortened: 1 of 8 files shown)


core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -103,6 +105,15 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
                 new EventTypeMapper());
     }
 
+    public List<EventType> listEventTypesWithRowLock(final Set<String> eventTypes) {
+        final String whereClause = "WHERE zn_data.event_type.et_name in ( "
+                + String.join(",", Collections.nCopies(eventTypes.size(), "?") ) +")";
+        return jdbcTemplate.query(
+                "SELECT et_event_type_object FROM zn_data.event_type " + whereClause + " FOR KEY SHARE",
```
a1exsh, line 112: Looks like a lot of string operations.  Could we make use `String.format()` instead of most concatenations here?

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
```diff
@@ -153,6 +164,29 @@ public Subscription createSubscription(final SubscriptionBase subscriptionBase)
...
+
+                    throw new NoSuchEventTypeException(String.format("Failed to create subscription, " +
+                            "event type(s) not found: '%s'", String.join("', '", missingEventTypes)));
+                }
+
+                return subscriptionRepository.createSubscription(subscriptionBase);
+            });
+        } catch (TransactionException e) {
+            LOGGER.error("Failed to create subscription, unable to obtain lock", e);
+            throw new InconsistentStateException("Failed to create subscription, unable to obtain lock");
```
a1exsh, line 186: Let's log or re-throw, but not both: the exception should be logged where it's caught ultimately.

Also not sure that "unable to obtain lock" is the only reason for `TransactionException` — should we really mention that?

a1exsh (commented):
I've thought about our approach a bit more and it looks like row-level lock is still inadequate to prevent the issue, so we might need to go with a table-level one.

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
samizzy: done

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -399,6 +399,7 @@ public void delete(final String eventTypeName) throws EventTypeDeletionException
     private Multimap<TopicRepository, String> deleteEventTypeWithSubscriptions(final String eventType) {
         try {
             return transactionTemplate.execute(action -> {
+                eventTypeRepository.lockTable(EventTypeRepository.TableLock.ROW_EXCLUSIVE);
```
a1exsh, line 402: I suggest we change it to something like the following:
```
eventTypeRepository.lockingTable(transactionTemplate, EventTypeRepository.TableLock.ROW_EXCLUSIVE,
    action -> {
        ...
    });
```

Otherwise we allow the caller to lock the table and walk away forgetting to unlock (by closing the transaction).

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -103,6 +119,21 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
                 new EventTypeMapper());
     }
 
+    public List<EventType> listEventTypes(final Set<String> eventTypes) {
+        final String query = String.format(
+                "SELECT et_event_type_object FROM zn_data.event_type WHERE zn_data.event_type.et_name in ( %s )",
```
a1exsh, line 124:
```suggestion
                "SELECT et_event_type_object FROM zn_data.event_type WHERE et_name IN ( %s )",
```

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
```diff
@@ -153,6 +164,30 @@ public Subscription createSubscription(final SubscriptionBase subscriptionBase)
         return subscription;
     }
 
+    private Subscription createSubscriptionWithEventTypeLock(final SubscriptionBase subscriptionBase) {
+        try {
+            return transactionTemplate.execute(action -> {
+                eventTypeRepository.lockTable(EventTypeRepository.TableLock.ROW_EXCLUSIVE);
```
a1exsh, line 170: On a second thought, a `SHARE` lock should suffice here, as long as the other transaction is taking the `ROW EXCLUSIVE` one.  This way concurrent requests to create subscriptions do not block each other (but block deletion of an event type).

a1exsh (changes requested):
Please see latest comments. :)

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -398,28 +398,36 @@ public void delete(final String eventTypeName) throws EventTypeDeletionException
...
-                    listResult.getItems().forEach(s -> {
-                        try {
-                            subscriptionRepository.deleteSubscription(s.getId());
-                        } catch (final NoSuchSubscriptionException e) {
-                            // should not happen as we are inside transaction
-                            throw new InconsistentStateException("Subscription to be deleted is not found", e);
+            return eventTypeRepository.lockingTable(EventTypeRepository.TableLock.SHARE, transactionTemplate,
+                    action -> {
+                        SubscriptionTokenLister.ListResult listResult = subscriptionTokenLister.listSubscriptions(
+                                ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), null, 100);
```
a1exsh, line 404:
```suggestion
                        SubscriptionTokenLister.ListResult listResult = getNextResult(eventType, null);
```

Assuming we change the 2nd parater of `getNextResult()` to whatever type `listResult.getNext()` has.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -398,28 +398,36 @@ public void delete(final String eventTypeName) throws EventTypeDeletionException
...
-                SubscriptionTokenLister.ListResult listResult = subscriptionTokenLister.listSubscriptions(
-                        ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), null, 100);
-                while (null != listResult) {
-                    listResult.getItems().forEach(s -> {
-                        try {
-                            subscriptionRepository.deleteSubscription(s.getId());
-                        } catch (final NoSuchSubscriptionException e) {
-                            // should not happen as we are inside transaction
-                            throw new InconsistentStateException("Subscription to be deleted is not found", e);
+            return eventTypeRepository.lockingTable(EventTypeRepository.TableLock.SHARE, transactionTemplate,
```
a1exsh, line 401: I think we should swap the lock types here and in the other transaction.

api-metastore/src/main/java/org/zalando/nakadi/service/SubscriptionService.java
```diff
@@ -153,6 +164,31 @@ public Subscription createSubscription(final SubscriptionBase subscriptionBase)
...
 
+    private Subscription createSubscriptionWithEventTypeLock(final SubscriptionBase subscriptionBase) {
+        try {
+            return eventTypeRepository.lockingTable(EventTypeRepository.TableLock.ROW_EXCLUSIVE, transactionTemplate,
+                    action -> {
+                        final Set<String> dbEventTypes = eventTypeRepository.
+                                listEventTypes(subscriptionBase.getEventTypes()).stream().
+                                map(EventType::getName).collect(Collectors.toSet());
+
+                        if (dbEventTypes.size() != subscriptionBase.getEventTypes().size()) {
```
a1exsh, line 175: If we start by collecting `missingEventTypes` first we could just check like this:
```java
if (!missingEventTypes.empty()) ...
```

api-metastore/src/test/java/org/zalando/nakadi/controller/EventTypeControllerTestCase.java
```diff
@@ -109,6 +109,11 @@ public void init() throws Exception {
         when(timelineService.getTopicRepository((Timeline) any())).thenReturn(topicRepository);
         when(timelineService.getTopicRepository((EventTypeBase) any())).thenReturn(topicRepository);
         when(authorizationService.getSubject()).thenReturn(Optional.empty());
+        when(eventTypeRepository.lockingTable(any(),any(),any())).thenAnswer(invocation -> {
+            final TransactionCallback callback = (TransactionCallback) invocation.getArguments()[2];
+            return callback.doInTransaction(null);
+
```
a1exsh, line 115:
```suggestion
```

api-metastore/src/test/java/org/zalando/nakadi/service/EventTypeServiceTest.java
```diff
@@ -113,6 +113,13 @@ public void setUp() {
                 authorizationValidator, timelineSync, transactionTemplate, nakadiSettings, nakadiKpiPublisher,
                 KPI_ET_LOG_EVENT_TYPE, nakadiAuditLogPublisher, eventTypeOptionsValidator,
                 eventTypeCache, schemaService, adminService, subscriptionTokenLister, applicationService);
+
+        when(eventTypeRepository.lockingTable(any(),any(),any())).thenAnswer(invocation -> {
+            final TransactionCallback callback = (TransactionCallback) invocation.getArguments()[2];
+            return callback.doInTransaction(null);
+
```
a1exsh, line 120:
```suggestion
```

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -22,12 +24,28 @@
...
+        private final String value;
+        TableLock(final String value){
+            this.value = value;
+        }
+
+        public String get(){
+            return value;
+        }
+
+    }
```
a1exsh, line 47:
```suggestion
        private final String value;

        TableLock(final String value) {
            this.value = value;
        }

        public String get() {
            return value;
        }
    }
```

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -103,6 +121,25 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
                 new EventTypeMapper());
     }
 
+    public List<EventType> listEventTypes(final Set<String> eventTypes) {
```
a1exsh, line 124:
```suggestion
    public List<EventType> listEventTypes(final Set<String> eventTypeNames) {
```
Maybe?

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -103,6 +121,25 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
...
+                "SELECT et_event_type_object FROM zn_data.event_type WHERE zn_data.event_type.et_name IN ( %s )",
+                String.join(",", Collections.nCopies(eventTypes.size(), "?"))
+        );
+        return jdbcTemplate.query(
+                query,
+                eventTypes.toArray(),
+                new EventTypeMapper());
+    }
+
+    public <T> T lockingTable(final TableLock lockTable, final TransactionTemplate transactionTemplate,
```
a1exsh, line 135:
```suggestion
    public <T> T lockingTable(final TableLock lockMode, final TransactionTemplate transactionTemplate,
```

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -22,12 +24,28 @@
...
 import java.sql.SQLException;
+import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 
 @DB
 @Component
 public class EventTypeRepository extends AbstractDbRepository {
 
+    public enum TableLock {
```
a1exsh, line 35:
```suggestion
    public enum TableLockMode {
```
Maybe?

a1exsh (changes requested):
Please see my comments.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
samizzy: The types are the same but passing null in 2nd parameter will result in null pointer exception.

a1exsh: Maybe let's remove `getNextResult()` altogether — I don't see it making code clearer or shorter for now.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -398,34 +398,42 @@ public void delete(final String eventTypeName) throws EventTypeDeletionException
...
     }
 
+    private SubscriptionTokenLister.ListResult listSubscriptions(final String eventType,
+                                                                 final SubscriptionTokenLister.Token token,
+                                                                 final int limit) {
+        return subscriptionTokenLister.listSubscriptions(
+                ImmutableSet.of(eventType), Optional.empty(), Optional.empty(), token, limit);
+    }
 
     // TODO: This method should be fixed by creating proper db query.
```
a1exsh, line 433: I've just noticed this TODO.  Should we check if we can fix it in the course of this change?

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -103,6 +121,25 @@ public EventType mapRow(final ResultSet rs, final int rowNum) throws SQLExceptio
...
+                query,
+                eventTypeNames.toArray(),
+                new EventTypeMapper());
+    }
+
+    public <T> T lockingTable(final TableLockMode lockMode, final TransactionTemplate transactionTemplate,
+                              final TransactionCallback<T> action) {
+        return transactionTemplate.execute(ac -> {
+            jdbcTemplate.execute("LOCK TABLE zn_data.event_type IN " + lockMode.get() + " MODE");
+            return action.doInTransaction(ac);
```
a1exsh, line 139: Why is this parameter called `ac`?  Should it be `status`?
https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/support/TransactionCallback.html

a1exsh (commented):
I have more comments, sorry :D

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
rishiraj88: Thanks,  @a2exsh, for the intuitive suggestion, but
`ac` is quite fine here as it is intended to mean an action usually. See the documentation:
[https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/support/TransactionTemplate.html#execute-org.springframework.transaction.support.TransactionCallback-](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/support/TransactionTemplate.html#execute-org.springframework.transaction.support.TransactionCallback-)

It says right in the beginning: `Execute the action specified by the given callback object within a transaction. `
It again denotes:
`    action - the callback object that specifies the transactional action`

a1exsh: @rishiraj88 
> `action - the callback object that specifies the transactional action`

Right.  But in our code this is what we pass to `execute`.  It implements `TransactionCallback` interface and `execute` calls our overridden `doInTransaction()` which takes a parameter `TransactionStatus status`.  That's why I think `ac` should be renamed to `status`.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
samizzy: Ill make a separate PR for this.

a1exsh: Why?  I think this is well in scope of the change with ensuring DB consistency.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -398,34 +398,42 @@ public void delete(final String eventTypeName) throws EventTypeDeletionException
...
-                while (null != listResult) {
-                    listResult.getItems().forEach(s -> {
-                        try {
-                            subscriptionRepository.deleteSubscription(s.getId());
-                        } catch (final NoSuchSubscriptionException e) {
-                            // should not happen as we are inside transaction
-                            throw new InconsistentStateException("Subscription to be deleted is not found", e);
+            return eventTypeRepository.lockingTable(EventTypeRepository.TableLockMode.ROW_EXCLUSIVE,
+                    transactionTemplate,
+                    action -> {
```
a1exsh, line 403: Then `action` at this point should also be named `status`, or not?

samizzy: forgot abt this correct, the whole Lambda is the action but the parameter is status.

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
samizzy: The ticket is about "Db consistency" correct but the problem statement is about orphan subscription. The todo deals with nakadi settings as far as i see. Also why nakadi settings should be fetched from DB is another question I have.

samizzy: I think I understood the problem (not related to fetch nakadi settings from db) now but would still prefer to make a separate PR

a1exsh: OK, I've had a closer look as well and I agree that this can be handled separately.  No dangling subscriptions can result from lack of transaction here (but some subscriptions may be deleted when they shouldn't be).

a1exsh (approved):

samizzy:
Not needed anymore as it was solved by this PR instead https://github.com/zalando/nakadi/pull/1352

PR #1345, opened 2021-11-12
Explicitly configure AsyncTaskExecutor

a1exsh:
The default `SimpleAsyncTaskExecutor` is not suitable under load (we even get a warning about it on startup):
```
An Executor is required to handle java.util.concurrent.Callable return values.
Please, configure a TaskExecutor in the MVC config under "async support".
The SimpleAsyncTaskExecutor currently in use is not suitable under load.
```


app/src/main/java/org/zalando/nakadi/config/NakadiConfig.java
```diff
@@ -30,11 +28,6 @@
 
     private static final Logger LOGGER = LoggerFactory.getLogger(NakadiConfig.class);
 
-    @Bean
-    public TaskExecutor taskExecutor() {
-        return new SimpleAsyncTaskExecutor();
```
adyach, line 35: what does Spring use instead by default to manage threads?

a1exsh: Good point — I wanted to check that, but got sidetracked :P

Looks like it uses the same one: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/context/request/async/WebAsyncManager.html#setTaskExecutor-org.springframework.core.task.AsyncTaskExecutor-

> By default a SimpleAsyncTaskExecutor instance is used.

a1exsh:
:+1:

app/src/main/java/org/zalando/nakadi/config/AsyncConfig.java
```diff
@@ -0,0 +1,18 @@
...
+import java.util.concurrent.Executors;
+
+@Configuration
+public class AsyncConfig {
+    @Bean
+    @Qualifier("mvcAsyncTaskExecutor")
+    public AsyncTaskExecutor taskExecutor() {
+        return new ConcurrentTaskExecutor(Executors.newCachedThreadPool());
+    }
+}
```
antban, line 18: @a1exsh 
Is there a reason why the configuration got to different class?

a1exsh: Not really.  I just want to make sure that we create the thread pool only once.  If there is a way to achieve that while keeping it in the `WebConfig` class, I'd take that happily.

a1exsh:
:+1:

PR #1346, opened 2021-11-15
Fix ordering of request filters wrt. client id availability

a1exsh:
Internal ticket: 959

### DIFF
--- a/app/src/main/java/org/zalando/nakadi/config/WebConfig.java
+++ b/app/src/main/java/org/zalando/nakadi/config/WebConfig.java
@@ -15,6 +15,7 @@ import org.springframework.http.converter.ResourceHttpMessageConverter;
 import org.springframework.http.converter.StringHttpMessageConverter;
 import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
 import org.springframework.http.converter.xml.SourceHttpMessageConverter;
+import org.springframework.security.web.context.AbstractSecurityWebApplicationInitializer;
 import org.springframework.web.context.request.async.TimeoutCallableProcessingInterceptor;
 import org.springframework.web.method.support.HandlerMethodArgumentResolver;
 import org.springframework.web.servlet.config.annotation.AsyncSupportConfigurer;
@@ -75,6 +76,12 @@ public class WebConfig extends WebMvcConfigurationSupport {
         return new TimeoutCallableProcessingInterceptor();
     }
 
+
+    // ==========================================================================================
+    // BEGIN FILTERS
+    //
+    // From the highest to the lowest precedence, in increments of 10 (SysV init-script style).
+    //
     @Bean
     public FilterRegistrationBean traceRequestFilter() {
         return createFilterRegistrationBean(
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/zalando/nakadi/config/WebConfig.java
```diff
@@ -83,35 +83,40 @@ public FilterRegistrationBean traceRequestFilter() {
...
+                new FlowIdRequestFilter(),
                 Ordered.HIGHEST_PRECEDENCE + 10);
     }
 
     @Bean
-    public FilterRegistrationBean flowIdRequestFilter() {
+    public FilterRegistrationBean requestRejectedFilter() {
         return createFilterRegistrationBean(
-                new FlowIdRequestFilter(),
+                new RequestRejectedFilter(),
```
adyach, line 95: the logging filter will not be called, in the previous version it was called. it could be important to log rejected requests

adyach: I meant the `LoggingFilter` is called after `RequestRejectedFilter`

a1exsh: Yeah, it is tricky.  We might have to split some of the filters to pre- and post- client id phase somehow.

a1exsh: Changed precedence of this filter to make sure rejected requests are logged as well.
This starts to look like too much magic, though.

a1exsh: And now it fails `InvalidRequestAT` due to unexpected error `500`, so handling of rejected exception is broken by this.

Looks like we will have to touch the plugin code as well to change precedence of filters there...

Ideally we would express dependencies and build a directed graph to determine the order of filters instead of messing with explicit ordering.

app/src/main/java/org/zalando/nakadi/config/WebConfig.java
```diff
@@ -83,35 +83,40 @@ public FilterRegistrationBean traceRequestFilter() {
...
+    //
+
     @Bean
-    public FilterRegistrationBean flowIdRequestFilter() {
+    public FilterRegistrationBean monitoringRequestFilter() {
         return createFilterRegistrationBean(
-                new FlowIdRequestFilter(),
-                Ordered.HIGHEST_PRECEDENCE + 20);
+                new MonitoringRequestFilter(metricRegistry, perPathMetricRegistry, authorizationService),
+                Ordered.LOWEST_PRECEDENCE - 50);
```
tor-vs-floki, line 103: Why are you invoking this filter with lower precedence than the lowest one in our constants? Could break integration with other filters in the future. Also why exactly 50?

tor-vs-floki: Would be nice to store all these values somewhere, and name them based on their impact.

a1exsh: See: https://github.com/zalando/nakadi/pull/1346#discussion_r749359591

app/src/main/java/org/zalando/nakadi/config/WebConfig.java
```diff
@@ -75,6 +76,12 @@ public TimeoutCallableProcessingInterceptor timeoutInterceptor() {
         return new TimeoutCallableProcessingInterceptor();
     }
 
+
+    // ==========================================================================================
```
adyach, line 80: does it mean that all the filters in the section are called before authentication token is validated?

adyach: I suggest to put every nakadi filter after spring security filters

a1exsh: > does it mean that all the filters in the section are called before authentication token is validated?

Yes, I understand it that way as well.

> I suggest to put every nakadi filter after spring security filters.

I think this is undesirable, as we might want to trace calls to the token info, and for that it is important that the tracing span is established earlier.

It also seems that `RequestRejectedFilter` needs a higher precedence to achieve the desired effect.

adyach: I understand, that it is current order of filters, then lets leave as it is

adyach:
👍

a1exsh:
:+1:

PR #1347, opened 2021-11-16
Optimize subscription query

samizzy:
Modified check and deletion of subscriptions during event type deletion to be inside a transaction.


core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -155,6 +162,33 @@ public void deleteSubscription(final String id)
         }
     }
 
+    public static class OffsetLimit {
```
a1exsh, line 165: I see multiple problems with this class.  But I wonder if we can sidestep all of it by actually deprecating use of `offset` parameter in the API first? https://nakadi.io/manual.html#/subscriptions_get*offset

samizzy:
:+1:

PR #1349, opened 2021-11-29
Move default storage information to database.

antban:
Right now storage information is located in the database. However, the `is_default` flag is located in zookeeper.

This causes a lot of problems:
 - there is a watcher that all the instances are listening to (not related, but leads to additional load)
 - refresh is using a separate thread on each nakadi instance
 - the information is not consistent.

In case if the flag is stored in the database - all the problems disappear.


core-services/src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -24,75 +23,58 @@
...
-            defaultStorageId = curator.getData()
-                    .usingWatcher(
-                            (Watcher) evt -> zkNotificationThread.submit(this::refreshDefaultStorageAndCreateWatcher))
-                    .forPath(ZK_TIMELINES_DEFAULT_STORAGE);
-        } catch (Exception ex) {
-            LOG.error("Failed to refresh default storage id, updates may be blocked", ex);
-            zkNotificationThread.schedule(this::refreshDefaultStorageAndCreateWatcher, 10, TimeUnit.SECONDS);
+    // TODO: the next method is completely useless in the real world, and makes sense only in test environment.
+    @PostConstruct
+    public void initializeDefaultStorage() {
```
antban, line 62: Moved from [here](https://github.com/zalando/nakadi/blob/38bb6a5f8c1765a6596a7a8e4d4a3dccd5ee4dcd/app/src/main/java/org/zalando/nakadi/config/NakadiConfig.java#L54)

api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -79,6 +80,7 @@
 import static org.zalando.nakadi.domain.Feature.FORCE_EVENT_TYPE_AUTHZ;
 
 @Component
+@DependsOn({"storageService"})
```
antban, line 83: The `DependsOn` probably should be tackled in another PR, as this one is already too big

antban:
:+1:

core-common/src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
```diff
@@ -57,10 +70,12 @@ public StorageDbRepository(final JdbcTemplate template, final ObjectMapper mappe
     public Storage createStorage(final Storage storage) throws DuplicatedStorageException, RepositoryProblemException {
         try {
             jdbcTemplate.update(
-                    "INSERT INTO zn_data.storage (st_id, st_type, st_configuration) VALUES (?, ?, ?::jsonb)",
+                    "INSERT INTO zn_data.storage (st_id, st_type, st_configuration, st_default) " +
```
thomasabraham, line 73: Can't we use `STORAGE_FIELDS` here as well?

core-common/src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
```diff
@@ -72,6 +87,16 @@ public Storage createStorage(final Storage storage) throws DuplicatedStorageExce
         }
     }
 
+    public void setDefaultStorage(final String storageId, final boolean isDefault) {
+        try {
+            jdbcTemplate.update("update zn_data.storage set st_default=? where st_id=?",
```
thomasabraham, line 92: Suggestion: follow the convention of using block letters for SQL keywords.

core-common/src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
```diff
@@ -41,11 +43,22 @@ public StorageDbRepository(final JdbcTemplate template, final ObjectMapper mappe
         return storages;
     }
 
+    public Optional<Storage> getDefaultStorage() throws RepositoryProblemException {
+        final List<Storage> storages = jdbcTemplate.query(
+                "SELECT " + STORAGE_FIELDS + " FROM zn_data.storage WHERE st_default=? order by st_id limit 1",
```
tor-vs-floki, line 48: Why do we store the default storage in the DB if it is defined by configuration?

core-services/src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -24,75 +23,58 @@
...
-        final byte[] defaultStorageId;
-        try {
-            defaultStorageId = curator.getData()
-                    .usingWatcher(
-                            (Watcher) evt -> zkNotificationThread.submit(this::refreshDefaultStorageAndCreateWatcher))
-                    .forPath(ZK_TIMELINES_DEFAULT_STORAGE);
-        } catch (Exception ex) {
-            LOG.error("Failed to refresh default storage id, updates may be blocked", ex);
-            zkNotificationThread.schedule(this::refreshDefaultStorageAndCreateWatcher, 10, TimeUnit.SECONDS);
+    // TODO: the next method is completely useless in the real world, and makes sense only in test environment.
```
thomasabraham, line 60: Suggestion: refer to the method by name rather than saying next method.
And it is better to have an action item in mind if we mark it with `TODO`, otherwise, this could be an informational comment without the `TODO`.

core-common/src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
antban: @tor-vs-floki 
Here you are touching the comment from `StorageService` init method. Default storage id was stored in zookeeper (and after this PR - in database), the configuration properties are working only for first initialization of the service (when there are no values present in database and zookeeper).

core-common/src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
antban: Everywhere except this line we are not relying on the order of the fields. The main criteria was to have the set of fields to select the same as in RowMapper below. For insert operation it makes sense to see right here what is being inserted and in which order.

core-services/src/main/java/org/zalando/nakadi/service/StorageService.java
antban: You are probably right, we wont fix this problem at all. I will remove the comment.

core-common/src/main/java/org/zalando/nakadi/repository/db/StorageDbRepository.java
tor-vs-floki: Could we store only storage on the database, having the default storage id only configurable from the environment, and then we could only query the database by storage ID to ensure it exists and get the required information from the DB?

antban: @tor-vs-floki 
We are able to create storage at runtime, use timeline to switch to it. Why setting storage as default should require redeployment. In other words - where is the benefit from what you are suggesting?

The second part - why I do not like to couple together environment vars and database configuration. Right now default storage is created if it was not created yet, and configuration is mystically taken somewhere from environment. It means, that we constantly provide configuration, that is:
- always not used
- most probably is outdated (totally incorrect values written there).

In these places (where configuration is stored in db), I would tend to remove configuration from environment at all, not to have 2 different places to maintain (and remove ambiguity between them).
The problem with it right now is that our tests and publishing to kpi event types are relying on existence of storage during application startup. If there would be more time, I would remove configuration from environment and would rely on external configuration only.

core-services/src/main/java/org/zalando/nakadi/service/StorageService.java
```diff
@@ -24,75 +23,57 @@
...
-        try {
-            defaultStorageId = curator.getData()
-                    .usingWatcher(
-                            (Watcher) evt -> zkNotificationThread.submit(this::refreshDefaultStorageAndCreateWatcher))
-                    .forPath(ZK_TIMELINES_DEFAULT_STORAGE);
-        } catch (Exception ex) {
-            LOG.error("Failed to refresh default storage id, updates may be blocked", ex);
-            zkNotificationThread.schedule(this::refreshDefaultStorageAndCreateWatcher, 10, TimeUnit.SECONDS);
+    @PostConstruct
+    public void initializeDefaultStorage() {
```
thomasabraham, line 61: Imagine the production system switched default storage to `storage-002` after some time. 
A new instance that start with the `nakadi.timelines.storage.default=storage-001` will try to create this default storage.
This is fine as long as the `storage-001` is not removed from the system in the first place; and DB actually checks for unique ID. 
But given that there is an option to delete a storage, I would prefer to check if there is a default storage before creating a new one.

antban: @thomasabraham 
The check is right after your comment (line 62). 
I removed the comment after previous review. This method is a crutch, and should be removed, as storages is a dynamic information and should be maintained in the database. This method should not exist at all, but it requires much more changes.

thomasabraham: Ahh, I missed it :+1:

thomasabraham:
:+1:

antban:
:+1:

tor-vs-floki:
👍

thomasabraham:
:+1:

PR #1350, opened 2021-12-01
docs: Create does not update a subscription

ePaul:
# One-line summary

Docs only: make it clear that POST /subscriptions will not update an existing subscription

## Description

In an [internal discussion](https://chat.google.com/room/AAAALJjUd9k/7gM7kMlsUa8) some confusion was noticed about the behavior
of the `POST /subscriptions` endpoint when a subscription with the same keys
already exists (specifically about the cursors).

This change should make this a bit clearer, adding a bit of text to the endpoint
and the subscription object definition.

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -2624,7 +2624,7 @@ definitions:
       number of clients by managing consumed event offsets and distributing load between instances.
       The key properties that identify subscription are 'owning_application', 'event_types' and 'consumer_group'.
       It's not possible to have two different subscriptions with these properties being the same.
-      Trying to create a subscription with these keys matching an existing one will just return that one, and
+      Creating a subscription with the keys matching existing subscription will neither create a new or change the existing subscription ignoring the remaining fields (e.g. `read_from`, `authorization`, etc.).
       ignore the remaining fields (e.g. `read_from`, `authorization`, etc.).
     properties:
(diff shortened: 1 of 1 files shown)


docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2624,6 +2624,8 @@ definitions:
       number of clients by managing consumed event offsets and distributing load between instances.
       The key properties that identify subscription are 'owning_application', 'event_types' and 'consumer_group'.
       It's not possible to have two different subscriptions with these properties being the same.
+      Trying to create a subscription with these keys matching an existing one will just return that one, and
```
adyach, line 2627: maybe it is a bit clearer that way 
```suggestion
      Creating a subscription with the keys matching existing subscription will neither create a new or change the existing subscription ignoring the remaining fields (e.g. `read_from`, `authorization`, etc.).
```

adyach:
👍

ePaul:
:+1:

docs/_data/nakadi-event-bus-api.yaml
ePaul: Thanks, I've applied this (with a tiny bit of grammar fix).

adyach:
👍

adyach:
@ePaul thank you for contributing !

PR #1351, opened 2021-12-02
docs: POST /subscriptions doesn't update (part 2)

ePaul:
# One-line summary

Docs only: make it clear that POST /subscriptions will not update an existing subscription

## Description

It turns out that I missed this file in #1350.
(This is for https://nakadi.io/manual.html#creating-subscriptions)

## Review
- [ ] Documentation

### DIFF
--- a/docs/_documentation/using_consuming-events-hila.md
+++ b/docs/_documentation/using_consuming-events-hila.md
@@ -66,6 +66,9 @@ Content-Type: application/json;charset=UTF-8
 }
 ```
 
+If there is already a subscription with same `owning_application`, `event_types` and `consumer_group`,
+it is just returned (and not updated, all other parts of the request body are then ignored).
+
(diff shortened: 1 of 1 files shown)


adyach:
👍

ePaul:
:+1:

docs/_documentation/using_consuming-events-hila.md
```diff
@@ -66,6 +66,9 @@ Content-Type: application/json;charset=UTF-8
 }
 ```
 
+If there is already a subscription with same `owning_application`, `event_types` and `consumer_group`,
+it is just returned (and not updated, all other parts of the request body are then ignored).
```
vsukharev, line 70: It would be nice to mention, whether the response code would be 200 OK

adyach:
@ePaul thanks for make it clearer !

PR #1352, opened 2022-01-03
Fix orphan subscription

samizzy:
## Description
Alternate PR for solving issue with orphan subscription caused due to deleting of event type. This PR uses row locks instead of table locks.


a1exsh:
:+1:

samizzy:
:+1:

PR #1353, opened 2022-01-04
Optimize list subscription query

samizzy:
Alternate PR for improving list subscription as pagination is not required for dealing with list of subscriptions internally.

### DIFF
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
@@ -129,7 +129,8 @@ public class SubscriptionDbRepositoryTest extends AbstractDbRepositoryTest {
                 .collect(toList());
 
         final List<Subscription> subscriptions = repository.listSubscriptions(ImmutableSet.of("et1"),
-                Optional.of(owningApp), Optional.empty(), 0, 10);
+                Optional.of(owningApp), Optional.empty(), Optional.of(new SubscriptionDbRepository.
+                        PaginationParameters(10, 0)));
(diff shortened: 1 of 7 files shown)


a1exsh:
:+1:

samizzy:
:+1:

adyach:
> Alternate PR for improving list subscription as pagination is not required for dealing with list of subscriptions internally

@samizzy could elaborate on the change please ?

samizzy:
> > Alternate PR for improving list subscription as pagination is not required for dealing with list of subscriptions internally
> 
> @samizzy could elaborate on the change please ?

1. The use of paginated querying of subscription was unnecessary in the the current code when used internally, hence subscription query generation code has been made to optionally accept Pagination params instead of mandatorily asking for it.
2. Currently, the check for listing subscriptions during deleting a eventtype is a 2 part process (involves calling select 2 times once when checking for subscription and other when deleting) which is not needed and could lead to inconsistent results. Now this is done inside a transaction.

a1exsh:
> Now querying is done once then deletion both of them done inside a transaction.

Yeah, actually it's 2 changes in one, and this is the more important part of it, IMO :D

adyach:
> > > Alternate PR for improving list subscription as pagination is not required for dealing with list of subscriptions internally
> > 
> > 
> > @samizzy could elaborate on the change please ?
> 
> 1. The use of paginated querying of subscription was unnecessary in the the current code when used internally, hence subscription query generation code has been made to optionally accept Pagination params instead of mandatorily asking for it.
> 2. Currently, the check for listing subscriptions during deleting a eventtype is a 2 part process (involves calling select 2 times once when checking for subscription and other when deleting) which is not needed and could lead to inconsistent results. Now querying is done once then deletion both of them done inside a transaction.

@samizzy thanks, please, add details in commit message when merging to master

PR #1354, opened 2022-01-06
Setting EventPublisher loglevel to Debug

samizzy:
# One-line summary
Setting EventPublisher loglevel to Debug as it helps in viewing event publishing validation errors


samizzy:
Not needed here

PR #1357, opened 2022-01-19
Removing Hystrix

thomasabraham:
# Removing hystrix circuit breaker between Nakadi and Kafka.

## Description
Hystrix library is no longer supported, and circuti breaker between
Nakadi and Kafka is not always useful. This commit removes Hystrix,
another circuit breaker can be added in future if required.

## Review
- [ ] Tests
- [ ] Documentation


a1exsh:
LGTM, but please check why the tests are failing.

thomasabraham:
> LGTM, but please check why the tests are failing.

Looks like there is some bad tests that will fail randomly on rare occasion, It was consistently passing locally and re triggering the pipeline fixed it in the build as well.

thomasabraham:
:+1:

PR #1358, opened 2022-01-20
use ByteArraySerializer for producer

adyach:
Nakadi needs to use byte serializers in order to work with binary data, the change is tech refactoring as precondition for binary formats in Nakadi

### DIFF
--- a/core-common/src/test/java/org/zalando/nakadi/domain/EventOwnerHeaderTest.java
+++ b/core-common/src/test/java/org/zalando/nakadi/domain/EventOwnerHeaderTest.java
@@ -6,6 +6,8 @@ import org.apache.kafka.clients.producer.ProducerRecord;
 import org.junit.Assert;
 import org.junit.Test;
 
+import java.nio.charset.StandardCharsets;
+
 public class EventOwnerHeaderTest {
 
     private static final String NAME = "teams";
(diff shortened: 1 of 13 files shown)


adyach:
👍

core-common/src/test/java/org/zalando/nakadi/domain/EventOwnerHeaderTest.java
```diff
@@ -14,7 +16,9 @@
     @Test
     public void testEventOwnerHeaderSerialization() {
         final EventOwnerHeader eventOwnerHeader = new EventOwnerHeader(NAME, VALUE);
-        final ProducerRecord<String, String> record = new ProducerRecord<>("topic", "value");
+        final ProducerRecord<byte[], byte[]> record = new ProducerRecord<>(
+                "topic",
```
a1exsh, line 20: How does it work here with "topic" not converted to `byte[]`?

adyach: conversion to bytes only for key / value, that should be non default producer constructor where key is null by default

adyach: yes, that is exactly the case 
```
    public ProducerRecord(String topic, V value) {
        this(topic, null, null, null, value, null);
    }
```

a1exsh: ah, I see.  the first argument is not key, but indeed a topic, which is expected to be of type `String`, regardless of the key and value types: https://kafka.apache.org/23/javadoc/org/apache/kafka/clients/producer/ProducerRecord.html#ProducerRecord-java.lang.String-V-
```
public ProducerRecord(String topic,
                      V value)
Create a record with no key
```

a1exsh:
:+1:

PR #1359, opened 2022-01-24
part 1: publishing to nakadi access log in avro

adyach:
# One-line summary
Publishing to Nakadi KPI evnet type nakadi.access.log in avro. 

<b id="docs-internal-guid-abe0f21b-7fff-c975-274d-69f8fc4f037c" style="font-style: normal; font-variant-caps: normal; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-weight: normal;"><h3 dir="ltr" style="line-height: 1.38; text-align: justify; margin-top: 16pt; margin-bottom: 4pt;"><span style="font-size: 13.999999999999998pt; font-family: Arial; color: rgb(67, 67, 67); background-color: transparent; font-weight: 400; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-variant-east-asian: normal; font-variant-position: normal; text-decoration: none; vertical-align: baseline; white-space: pre-wrap;">Envelope</span></h3><p dir="ltr" style="line-height: 1.38; text-align: justify; margin-top: 0pt; margin-bottom: 0pt;"><span style="font-size: 10pt; font-family: Arial; color: rgb(0, 0, 0); background-color: rgb(255, 255, 255); font-weight: 400; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-variant-east-asian: normal; font-variant-position: normal; text-decoration: none; vertical-align: baseline; white-space: pre-wrap;">Nakadi deserializes event’s metadata to identify Kafka partitions to publish, event compaction key and extends metadata with additional fields like event type, publisher uid, event receive time and JSON schema version. Avro is serialized to binary format. In order for Nakadi to deserialize Avro bytes it has to know the exact position of metadata. Hence, the envelope format has to be positional.</span></p><br><div dir="ltr" align="center" style="margin-left: 0pt;">

Field name | Metadata Version | Metadata length | Metadata | Payload length | Payload
-- | -- | -- | -- | -- | --
Size (bytes) | 1 | 4 | x | 4 | y
Description | Schema version to parse metadata | Metadata size in bytes | Message metadata | Payload size in bytes | User payload

</div><br><p dir="ltr" style="line-height: 1.38; text-align: center; margin-top: 0pt; margin-bottom: 0pt;"><span style="font-size: 10pt; font-family: Arial; color: rgb(0, 0, 0); background-color: rgb(255, 255, 255); font-weight: 400; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-variant-east-asian: normal; font-variant-position: normal; text-decoration: none; vertical-align: baseline; white-space: pre-wrap;">Table. 1</span></p><br><p dir="ltr" style="line-height: 1.38; text-align: justify; margin-top: 0pt; margin-bottom: 0pt;"><span style="font-size: 10pt; font-family: Arial; color: rgb(0, 0, 0); background-color: rgb(255, 255, 255); font-weight: 400; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-variant-east-asian: normal; font-variant-position: normal; text-decoration: none; vertical-align: baseline; white-space: pre-wrap;">The format works with the single messages as well as with an array of messages in a single request. The overall structure of the envelope does not depend on the format of the payload. This means, new formats can be used for payload in the future.</span></p></b><br class="Apple-interchange-newline" style="font-style: normal; font-variant-caps: normal; font-weight: normal; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none; caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0);"><br class="Apple-interchange-newline" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: normal; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-size-adjust: auto; -webkit-text-stroke-width: 0px; text-decoration: none;">

The next part will add consumption of messages in JSON.
The test will be added as the last part to test the whole path.


core-common/src/main/resources/metadata.avsc
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "metadata",
+  "type": "record",
+  "doc": "Event metadata defines data is not related to actual event payload data",
+  "fields": [
+    {
+      "name": "occurred_at",
+      "type": "string"
```
a1exsh, line 8:
```suggestion
      "type": "long",
      "logicalType": "timestamp"
```

core-services/src/main/java/org/zalando/nakadi/service/publishing/AvroEventPublisher.java
```diff
@@ -0,0 +1,155 @@
...
+        this.uuidGenerator = uuidGenerator;
+        this.avroSchema = avroSchema;
+    }
+
+    public void publishAvro(final String etName,
+                            final String publishedBy,
+                            final GenericRecord event) {
+        try {
+            final GenericRecord metadata = new GenericData.Record(avroSchema.getMetadataSchema());
+            metadata.put("occurred_at", new DateTime(DateTimeZone.UTC).toString());
```
a1exsh, line 68: Or was the `string` type here intentional?  I would expect `occurred_at` and `received_at` to have the same type.
Here, by the way, we could use the same timestamp instance: there is no distinct "received" moment.

core-services/src/main/java/org/zalando/nakadi/service/publishing/AvroEventPublisher.java
```diff
@@ -0,0 +1,155 @@
...
+        try (Closeable ignored = TracingService.activateSpan(publishingSpan)) {
+            final ByteArrayOutputStream eventOutputStream = new ByteArrayOutputStream();
+            final GenericDatumWriter genericDatumWriter = new GenericDatumWriter(eventMetadata.getSchema());
+            genericDatumWriter.setSchema(event.getSchema());
+            genericDatumWriter.write(event, EncoderFactory.get()
+                    .directBinaryEncoder(eventOutputStream, null));
+            final byte[] metadata = eventOutputStream.toByteArray();
+
+            genericDatumWriter.setSchema(event.getSchema());
+            genericDatumWriter.write(event, EncoderFactory.get()
```
a1exsh, line 126: I'm not sure that this is correct use of the output stream.  I'd sleep better if we've taken a separate instance here ;)

core-services/src/test/java/org/zalando/nakadi/service/publishing/AvroEventPublisherTest.java
```diff
@@ -0,0 +1,41 @@
...
+        final Resource accessLog = new DefaultResourceLoader().getResource("nakadi.access.log.avsc");
+        final AvroSchema avroSchema = new AvroSchema(metadata, accessLog);
+        final AvroEventPublisher eventPublisher = new AvroEventPublisher(timelineService,
+                cache, timelineSync, nakadiSettings, new UUIDGenerator(), avroSchema);
+        final EventType eventType = buildDefaultEventType();
+        Mockito.when(cache.getEventType(eventType.getName())).thenReturn(eventType);
+
+        final GenericRecord event = new GenericData.Record(avroSchema.getNakadiAccessLogSchema());
+        event.put("method", "POST");
+        event.put("path", "/evnet-types");
```
a1exsh, line 30:
```suggestion
        event.put("path", "/event-types");
```
Doesn't matter in tests, but doesn't look good either.

a1exsh (changes requested):
Please see comments.

core-services/src/main/java/org/zalando/nakadi/service/publishing/AvroEventPublisher.java
adyach: I do not see issue with reconfiguration

```
  public void setSchema(Schema root) {
    this.root = root;
  }

  public void write(D datum, Encoder out) throws IOException {
    Objects.requireNonNull(out, "Encoder cannot be null");
    write(root, datum, out);
  }
```

for your sleep Alex, I will create a new writer :)

core-services/src/main/java/org/zalando/nakadi/service/publishing/AvroEventPublisher.java
```diff
@@ -0,0 +1,156 @@
...
+                    metadata,
+                    payload.length,
+                    payload)
+            );
+
+            timelineService.getTopicRepository(eventType)
+                    .syncPostEvent(new NakadiRecord(
+                            eventTypeName,
+                            topic,
+                            null,
```
a1exsh, line 143: Do I get it right, that using `null` for partition results in random partitioning?

`If a valid partition number is specified that partition will be used when sending the record. If no partition is specified but a key is present a partition will be chosen using a hash of the key. If neither key nor partition is present a partition will be assigned in a round-robin fashion.`
https://kafka.apache.org/23/javadoc/org/apache/kafka/clients/producer/ProducerRecord.html#ProducerRecord-java.lang.String-java.lang.Integer-K-V-

adyach: yes, for the moment I chose that approach.

core-common/src/main/resources/metadata.avsc
```diff
@@ -1,45 +1,41 @@
...
-        "type": "string"
-      },
-      {
-        "name": "partition",
-        "type": "int"
-      }
-    ]
-  }
+  "type": "record",
+  "doc": "Event metadata defines data is not related to actual event payload data",
```
samizzy, line 4:
```suggestion
  "doc": "Event metadata defines data that is not related to actual event payload data",
```

core-services/src/main/java/org/zalando/nakadi/service/publishing/AvroEventPublisher.java
```diff
@@ -0,0 +1,155 @@
...
+                .withTag(Tags.MESSAGE_BUS_DESTINATION.getKey(), topic)
+                .withTag("event_type", eventTypeName)
+                .withTag("type", "avro")
+                .start();
+
+        try (Closeable ignored = TracingService.activateSpan(publishingSpan)) {
+            final ByteArrayOutputStream eventOutputStream = new ByteArrayOutputStream();
+            final GenericDatumWriter genericDatumWriter = new GenericDatumWriter(eventMetadata.getSchema());
+            genericDatumWriter.setSchema(event.getSchema());
+            genericDatumWriter.write(event, EncoderFactory.get()
```
samizzy: I think we want to get metadata encoded first
```suggestion
            genericDatumWriter.write(eventMetadata, EncoderFactory.get()
```

adyach: I'll address it in follow up PR for consumption

PR #1360, opened 2022-01-26
Consume Avro event type in JSON

adyach:
# One-line summary
Nakadi KPI event type nakadi.access.log is stored in Avro format. The PR makes consumption of that event type possible in JSON using Subscription API.


core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+                final Envelope envelop = Envelope.deserialize(data);
+                return deserializeAvro(envelop);
+            } catch (IOException e) {
+                throw new RuntimeException("failed to deserialize avro event", e);
+            }
+        }
+
+        throw new RuntimeException(String.format(
+                "event format is not defined, provided format: `%s`",
+                Arrays.toString(new byte[]{1, 2})));
```
a1exsh, line 48:
```suggestion
                Arrays.toString(eventFormat)));
```

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+                    .read(null, DecoderFactory.get()
+                            .binaryDecoder(envelop.getPayload(), null));
+            baos.reset();
+            baos.write("{\"metadata\":".getBytes(StandardCharsets.UTF_8));
+            baos.write(metadataRecord.toString().getBytes(StandardCharsets.UTF_8));
+            baos.write(',');
+            baos.write(accessLogRecord.toString().getBytes(StandardCharsets.UTF_8));
+            baos.write('}');
+
+            return baos.toByteArray();
```
a1exsh, line 68: To make this thread-safe let's just create the instance of `ByteArrayOutputStream` locally.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+            final GenericRecord metadataRecord = (GenericRecord) metadataDatumReader
+                    .read(null, DecoderFactory.get()
+                            .binaryDecoder(envelop.getMetadata(), null));
+
+            final GenericRecord accessLogRecord = (GenericRecord) nakadiAccessLogReader
+                    .read(null, DecoderFactory.get()
+                            .binaryDecoder(envelop.getPayload(), null));
+            baos.reset();
+            baos.write("{\"metadata\":".getBytes(StandardCharsets.UTF_8));
+            baos.write(metadataRecord.toString().getBytes(StandardCharsets.UTF_8));
```
a1exsh, line 63: Assumption here is that `toString()` produces a valid JSON document, right?  I don't see where this is defined in avro library, though.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+    public byte[] deserialize(final byte[] eventFormat, final byte[] data) {
+        if (eventFormat == null) {
+            // JSON
+            return data;
+        }
+
+        if (Arrays.equals(eventFormat, NakadiRecord.Format.AVRO.getFormat())) {
+            try {
+                final Envelope envelop = Envelope.deserialize(data);
+                return deserializeAvro(envelop);
```
a1exsh, line 40: Here we deserialize two times, but if feels like `Envelope` could just hold reference to the underlying byte array, and avoid copying the data.

core-services/src/main/java/org/zalando/nakadi/service/publishing/AvroEventPublisher.java
```diff
@@ -65,13 +63,13 @@ public void publishAvro(final String etName,
                             final GenericRecord event) {
         try {
             final GenericRecord metadata = new GenericData.Record(avroSchema.getMetadataSchema());
-            final DateTime dateTime = new DateTime(DateTimeZone.UTC);
-            metadata.put("occurred_at", dateTime);
+            final long now = System.currentTimeMillis();
+            metadata.put("occurred_at", now);
```
a1exsh, line 67: Wondering why wasn't it failing previously, if the data type didn't match?

core-services/src/main/java/org/zalando/nakadi/service/publishing/AvroEventPublisher.java
```diff
@@ -118,7 +116,7 @@ private void serializeAndSendToKafka(final String eventTypeName,
         try (Closeable ignored = TracingService.activateSpan(publishingSpan)) {
             final ByteArrayOutputStream eventOutputStream = new ByteArrayOutputStream();
             final GenericDatumWriter eventWriter = new GenericDatumWriter(eventMetadata.getSchema());
-            eventWriter.write(event, EncoderFactory.get()
+            eventWriter.write(eventMetadata, EncoderFactory.get()
```
a1exsh, line 119: Oops ;)

a1exsh (changes requested):
Please see comments.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
adyach: :D, thanks

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
adyach: that would require to have GenericDatumReader thread safe as well, but from the code it is not. in general, the class is used per Kafka Consumer, which is single threaded.

adyach: that would require to have GenericDatumReader thread safe as well, but from a quick look at the code I do not see that it is thread safe. in general, the class is used per Kafka Consumer, which is single threaded.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
adyach: https://avro.apache.org/docs/current/api/java/org/apache/avro/generic/GenericData.html#toString-java.lang.Object-

core-services/src/main/java/org/zalando/nakadi/service/publishing/AvroEventPublisher.java
adyach: it should be failing ...

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
adyach: Envelop deserializes message from Kafka bytes to byte array and keeps that array in Envelop. Then the byte array is desirialized to Avro. I am not sure I understand what you mean.

a1exsh: I mean in these two lines an instance of `Envelope` holds copy of the `data` bytes array, but strictly speaking, that's not necessary.  Could we avoid making that copy?

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
a1exsh: Can you help me connect the dots?  I'm still not sure about this.  Do we maybe already have a test that demonstrates that behavior?

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
```diff
@@ -0,0 +1,82 @@
...
+        try {
+            // fixme use schema registry later and cache
+            final GenericRecord metadataRecord = (GenericRecord) metadataDatumReader
+                    .read(null, DecoderFactory.get()
+                            .directBinaryDecoder(envelop.getMetadata(), null));
+
+            final GenericRecord accessLogRecord = (GenericRecord) nakadiAccessLogReader
+                    .read(null, DecoderFactory.get()
+                            .directBinaryDecoder(envelop.getPayload(), null));
+            baos.reset();
```
a1exsh, line 69: I understand that the intent here is to reduce allocations, but at the same time we potentially increase the total heap allocation as we hold to the peak required space for the buffer for up to one hour.  It would be interesting if this has a measurable effect.

All in all, I'd still prefer a thread-safe version.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
adyach: ill add the test

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
adyach: done

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
adyach: done

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
```diff
@@ -0,0 +1,72 @@
...
+            // JSON
+            return data;
+        }
+
+        if (Arrays.equals(eventFormat, NakadiRecord.Format.AVRO.getFormat())) {
+            try {
+                final EnvelopeHolder envelop = EnvelopeHolder.fromBytes(data);
+                if (envelop.getMetadataVersion() != AvroSchema.METADATA_VERSION) {
+                    throw new RuntimeException(String.format(
+                            "metadata version is not supported: `%d`",
```
a1exsh, line 37: Nitpick mode: seems like in Java world, first capital letter for an exception message is customary, so:
```suggestion
                            "Metadata version is not supported: `%d`",
```
And in other places as well.

adyach: I prefer to start from the lower case, it is nicer in the logs

PR #1361, opened 2022-01-26
provide Kafka Client compression type via env var

adyach:
at the moment it is not possible to set kafka client compression type using env var, it is hardcoded value of lz4


a1exsh:
:+1:

adyach:
👍

PR #1362, opened 2022-01-28
The first part in the series of changes to support binary formats in Nakadi

adyach:
The first part in the series of changes to support binary formats in Nakadi, in that particular reincarnation is Apache Avro.

Apache Avro™ is a data serialization system. Avro provides a compact, fast, binary data format and simple integration with dynamic languages. Avro supports code generation as an additional optimization, but it is only worth utilizing for statically typed languages. According to the research, it is faster in serialization and deserialization and around 10% more compact than JSON. 

1. Nakadi publishes Nakadi KPI event type nakadi.access.log in Avro to Kafka using the following format for Kafka record:
Nakadi deserializes event’s metadata to identify Kafka partitions to publish, event compaction key and extends metadata with additional fields like event type, publisher uid, event receive time and JSON schema version. Avro is serialized to binary format. In order for Nakadi to deserialize Avro bytes it has to know the exact position of metadata. Hence, the envelope format has to be positional.

Field name | Metadata Version | Metadata length | Metadata | Payload length | Payload
-- | -- | -- | -- | -- | --
Size (bytes) | 1 | 4 | x | 4 | y
Description | Schema version to parse metadata | Metadata size in bytes | Message metadata | Payload size in bytes | User payload

</div></b>

The format works with the single messages as well as with an array of messages in a single request. The overall structure of the envelope does not depend on the format of the payload. This means, new formats can be used for payload in the future.

2. Consumption of the Avro serialised events happens in JSON using Subscription API

3. At the moment Avro schemas for Nakadi metadata and event are stored in Nakadi source code. The schema registry will adopt it in the future.

The changes were reviewed in the following split PRs: https://github.com/zalando/nakadi/pull/1358, https://github.com/zalando/nakadi/pull/1359, https://github.com/zalando/nakadi/pull/1360


adyach:
👍

core-common/src/main/java/org/zalando/nakadi/domain/EnvelopeHolder.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    private static void writeIntToArray(
+            final byte[] data,
+            final int offset,
+            final int metadataLength) {
+        data[offset + 0] = (byte) (metadataLength >> 24);
+        data[offset + 1] = (byte) (metadataLength >> 16);
+        data[offset + 2] = (byte) (metadataLength >> 8);
+        data[offset + 3] = (byte) (metadataLength >> 0);
```
a1exsh, line 46:
```suggestion
            final int value) {
        data[offset + 0] = (byte) (value >> 24);
        data[offset + 1] = (byte) (value >> 16);
        data[offset + 2] = (byte) (value >> 8);
        data[offset + 3] = (byte) (value >> 0);
```

core-common/src/main/java/org/zalando/nakadi/domain/EnvelopeHolder.java
```diff
@@ -0,0 +1,91 @@
...
+    public static byte[] produceBytes(
+            final byte metadataVersion,
+            final EventWriter metadataWriter,
+            final EventWriter eventWriter) throws IOException {
+        final ByteArrayOutputStream eventOutputStream = new ByteArrayOutputStream();
+        eventOutputStream.write(metadataVersion);
+        eventOutputStream.write(new byte[4]);
+        metadataWriter.write(eventOutputStream);
+        final int metadataLength = eventOutputStream.size() - 1 - 4;
+        eventOutputStream.write(new byte[4]);
```
a1exsh, line 29: We could save 2 object allocations here if we just had a (static) 4-byte placeholder array.  Not that important, but rather a low-hanging fruit.

a1exsh:
:+1:

a1exsh:
:+1:

PR #1365, opened 2022-01-31
Use Subscription cache to improve read performance

samizzy:
This PR reverts a previous revert https://github.com/zalando/nakadi/pull/1337.

### DIFF
--- a/api-cursors/src/main/java/org/zalando/nakadi/service/CursorsService.java
+++ b/api-cursors/src/main/java/org/zalando/nakadi/service/CursorsService.java
@@ -142,11 +142,22 @@ public class CursorsService {
             }
         }
     }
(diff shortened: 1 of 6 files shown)


antban:
:+1:

api-cursors/src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -142,11 +142,22 @@ private void validateStreamId(final List<NakadiCursor> cursors,
             }
         }
     }
+    public List<SubscriptionCursorWithoutToken> getSubscriptionCursorsForUpdate(final String subscriptionId)
```
antban, line 145: This method is not needed, as the value to update is stored in zk, so the only case to take it from db is to ensure that subscription still exists. 
As it still adds value (check subscription existence), then let's leave it as it is

samizzy:
:+1:

PR #1369, opened 2022-02-08
Increase idle timeout for Jetty

tor-vs-floki:
Increase Jetty timeout for an experiment


app/src/main/java/org/zalando/nakadi/config/JettyConfig.java
```diff
@@ -16,7 +16,7 @@ public JettyServletWebServerFactory jettyServletWebServerFactory(
             @Value("${server.port:8080}") final String port,
             @Value("${jetty.threadPool.maxThreads:200}") final String maxThreads,
             @Value("${jetty.threadPool.minThreads:8}") final String minThreads,
-            @Value("${jetty.threadPool.idleTimeout:60000}") final String idleTimeout) {
```
a1exsh, line 19: Let's try not to change the default, but find the place where to specify the value externally. ;)

PR #1370, opened 2022-02-10
indicate which exception occured in case of timeout

adyach:
the change brings clarity on which exception occured in case of timeout

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -461,10 +461,11 @@ public class KafkaTopicRepository implements TopicRepository {
                     }
                 }));
             }
-            latch.await(createSendTimeout(), TimeUnit.MILLISECONDS);
-
-            final List<NakadiRecordMetadata> resps =
-                    prepareResponse(nakadiRecords, responses, null);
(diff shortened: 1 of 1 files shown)


a1exsh:
:+1:

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -461,10 +461,11 @@ private void logFailedEvents(final String topicId, final String eventType, final
...
             }
-            latch.await(createSendTimeout(), TimeUnit.MILLISECONDS);
-
-            final List<NakadiRecordMetadata> resps =
-                    prepareResponse(nakadiRecords, responses, null);
+            final boolean recordsPublished = latch.await(createSendTimeout(), TimeUnit.MILLISECONDS);
+            final List<NakadiRecordMetadata> resps = prepareResponse(
+                    nakadiRecords, responses,
+                    recordsPublished ? null : new TimeoutException(
+                            "timeout waiting for events to be sent to kafka"));
```
a1exsh, line 468: Since we are doing it, we could also mention the timeout duration in the message...

adyach: I did it in the first iteration, but decided to remove it, because of the reduced readability of the code and the same value in the message

adyach:
👍

PR #1371, opened 2022-02-10
count down the latch once callback processed

adyach:
# One-line summary
count down the latch once callback processed otherwise event status is not detected


adyach:
👍

PR #1372, opened 2022-02-15
Always set event key for hash partitioning

a1exsh:
1. Correct detection of ordering violations for multi-key hash partitioning.
2. Event key is important for joins in Kafka streams.


core-common/src/main/java/org/zalando/nakadi/partitioning/PartitionStrategy.java
```diff
@@ -13,6 +13,11 @@
     String USER_DEFINED_STRATEGY = "user_defined";
     String RANDOM_STRATEGY = "random";
 
+    default List<String> extractEventKeys(EventType eventType, JSONObject event)
+            throws PartitioningException {
+        throw new PartitioningException("Not supported by partition strategy.");
```
a1exsh, line 18: We should probably even extract it to a different interface, but this will work for now.

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -132,8 +128,8 @@ EventPublishResult processInternal(final String events,
             }
             validateEventOwnership(eventType, batch);
             validate(batch, eventType, delete);
+            assignKeys(batch, eventType);
```
a1exsh, line 131: Changed the order, so that we can later calculate partitions based on the event keys.

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -210,34 +206,24 @@ private void partition(final List<BatchItem> batch, final EventType eventType)
...
-                    try {
-                        final String eventKey = jsonPath.get(partitionKeyField).toString();
-                        item.setEventKey(eventKey);
-                    } catch (final JsonPathAccessException e) {
-                        throw new InvalidPartitionKeyFieldsException(e.getMessage());
-                    }
-                }
-            }
+            final List<String> partitionKeys = partitionResolver.extractEventPartitionKeys(eventType, item.getEvent());
+            item.setEventKey(Arrays.toString(partitionKeys.toArray()));
```
a1exsh, line 226: Anyone knows if the event key interpreted by Kafka Streams?..

core-services/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -73,4 +63,26 @@ public String calculatePartition(final EventType eventType, final JSONObject eve
...
 
+    @Override
+    public String calculatePartition(final EventType eventType, final JSONObject event, final List<String> partitions)
+            throws InvalidPartitionKeyFieldsException {
+
+        final List<String> partitionKeyFields = eventType.getPartitionKeyFields();
+        if (partitionKeyFields.isEmpty()) {
+            throw new RuntimeException("Applying " + this.getClass().getSimpleName() + " although event type " +
+                    "has no partition key fields configured.");
+        }
```
antban, line 74: I think it makes no sense to duplicate this check (especially as it is a RuntimeException)

core-services/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -73,4 +63,26 @@ public String calculatePartition(final EventType eventType, final JSONObject eve
...
+
+        final int hashValue = extractEventKeys(eventType, event).stream()
+                .map(s -> s.hashCode())
```
antban, line 77: This is a bug. `s.hashCode()` should not be used, stringHash probably has some code inside, that is not java version dependent.

a1exsh: Thanks, I have totally overlooked that.  Don't you think our tests should have helped here?

core-services/src/test/java/org/zalando/nakadi/service/publishing/EventPublisherTest.java
```diff
@@ -462,39 +462,33 @@ public void whenSinglePartitioningKeyThenEventKeyIsSet() throws Exception {
...
-
-        publisher.publish(batch.toString(), eventType.getName());
+        Mockito
+                .doReturn(ImmutableList.of("my_key"))
+                .when(partitionResolver)
+                .extractEventPartitionKeys(any(), any());
 
-        final List<BatchItem> publishedBatch = capturePublishedBatch();
-        assertThat(publishedBatch.get(0).getEventKey(), equalTo("my_key"));
+        assertThat(publisher.getEventKeys(eventType, new JSONObject()),
```
a1exsh, line 470: These tests became less useful again...

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -201,7 +197,8 @@ private void partition(final List<BatchItem> batch, final EventType eventType)
         for (final BatchItem item : batch) {
             item.setStep(EventPublishingStep.PARTITIONING);
             try {
-                final String partitionId = partitionResolver.resolvePartition(eventType, item.getEvent());
+                final String partitionId =
```
a1exsh, line 200: This change is going to break partitioning for log-compacted event types, it seems.  Previously the partition was calculated based on `partition_key_fields` and now it is using `partition_compaction_key` (due to the changes in `setEventKey` function below)...

At the same time, all tests passed :(

a1exsh:
:-1:

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
a1exsh: I think in this situation we need the following two attributes instead of one:
```java
String compactionKey;
List<String> partitioningKeys;
```
Either (or both) may be null and both can be set at the same time.

It's a pity of course that we have distinct keys for partitioning and compaction, but it's too late to fix that.

core-services/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -31,37 +31,27 @@ public HashPartitionStrategy(final HashPartitionStrategyCrutch hashPartitioningC
     }
 
     @Override
-    public String calculatePartition(final EventType eventType, final JSONObject event, final List<String> partitions)
+    public List<String> extractEventKeys(final EventType eventType, final JSONObject event)
```
adyach, line 34: basically, the part which gets values from event by partitions key from metadata is not separate function? nice

core-services/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -73,4 +63,20 @@ public String calculatePartition(final EventType eventType, final JSONObject eve
         }
     }
 
+    @Override
+    public String calculatePartition(
+            final JSONObject event, final List<String> eventKeys, final List<String> partitions) {
+
+        final int hashValue = eventKeys.stream()
```
adyach, line 70: this can replaced with 

```
final int hashValue = eventKeys.stream()
            .mapToInt(s -> stringHash.hashCode(s))
            .sum();
```

core-services/src/main/java/org/zalando/nakadi/partitioning/PartitionResolver.java
```diff
@@ -56,19 +56,33 @@ public void validate(final EventTypeBase eventType) throws NoSuchPartitionStrate
...
     }
 
-    public String resolvePartition(final EventType eventType, final JSONObject eventAsJson)
+    public List<String> extractEventPartitionKeys(final EventType eventType, final JSONObject eventAsJson)
             throws PartitioningException {
 
+        return resolveStrategy(eventType).extractEventKeys(eventType, eventAsJson);
+    }
+
+    public String resolvePartition(
```
adyach, line 65: it is strange that `resolvePartition` has `resolveStrategy` inside

adyach: probably partition resolver depends on partition strategy resolver, but not vice versa

adyach: probably, to resolve partition it is required to have strategy resolver, maybe pass it as dependency

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
adyach: > This change is going to break partitioning for log-compacted event types, it seems

it still takes event key for log compacted event types from `partition_compaction_key`, but for hash partitioning it uses event values from partition keys

core-services/src/main/java/org/zalando/nakadi/partitioning/PartitionResolver.java
```diff
@@ -53,29 +62,99 @@ public void validate(final EventTypeBase eventType) throws NoSuchPartitionStrate
...
-
         if (partitionStrategy == null) {
             throw new PartitioningException("Partition Strategy defined for this EventType is not found: " +
                     eventTypeStrategy);
         }
-
         return partitionStrategy;
     }
+
+    public static List<String> getKeyFieldsForExtraction(final EventType eventType) {
```
tor-vs-floki, line 90: This seems to be a concern for Hash partitioning only

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -196,16 +194,31 @@ private List<BatchItemResponse> responses(final List<BatchItem> batch) {
...
+            }
+        }
+    }
+
+    private void partition(final List<BatchItem> batch, final EventType eventType) throws PartitioningException {
 
         final List<String> orderedPartitions = eventTypeCache.getOrderedPartitions(eventType.getName());
 
         for (final BatchItem item : batch) {
             item.setStep(EventPublishingStep.PARTITIONING);
```
antban, line 219: seems to be strange....

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -131,8 +126,11 @@ EventPublishResult processInternal(final String events,
             }
             validateEventOwnership(eventType, batch);
             validate(batch, eventType, delete);
+            if (PartitionStrategy.HASH_STRATEGY.equals(eventType.getPartitionStrategy())) {
+                setPartitionKeys(batch, eventType);
+            }
```
antban, line 131: call with `if PartitionStrategy == something: setPartitionKeys` right before calling `partition` method looks very strange. Does it make sense to put it to `paritition()` method?

a1exsh:
:+1:

core-services/src/main/java/org/zalando/nakadi/partitioning/EventKeyExtractor.java
```diff
@@ -0,0 +1,99 @@
...
+                    .orElse(null);
+        }
+    }
+
+    private static boolean isCompactedEventType(final EventType eventType) {
+        return eventType.getCleanupPolicy() == CleanupPolicy.COMPACT ||
+                eventType.getCleanupPolicy() == CleanupPolicy.COMPACT_AND_DELETE;
+    }
+
+    private static String getEventKeyFromPartitionKeys(final List<String> partitionKeys) {
```
antban, line 64: better `getKafkaKeyFromPartitionKeys`

antban:
:+1:

a1exsh:
:+1:

PR #1373, opened 2022-02-15
Log requested compression of response and compression of request. 

antban:
Add values from request headers to nakadi.access.log event type
- accept_encoding
- content_encoding

These fields may be used to analyze if clients are making use of request compression for publishing and response compression for consumption.


adyach:
there are could be tests requiring fixing as well

antban:
:+1:

core-common/src/test/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializerTest.java
```diff
@@ -105,6 +107,8 @@ private ObjectNode expectedJsonEvent() {
                 .put("app_hashed", "hashed-app")
                 .put("status_code", 201)
                 .put("response_time_ms", 10)
+                .put("accept_encoding", "-")
```
adyach, line 110: why do we need `-`, it eats space. we can have null in there.

antban: In this particular test - it is a test. 
Overall - I am trading stability of schema over disk usage.

adyach: 👍  makes sense to keep the current values as it is!

adyach:
👍

antban:
:+1:

PR #1374, opened 2022-02-18
Avro Schema Registry Poc

samizzy:
Nakadi needs modifications to schema registry to support Avro Schemas. 

This PR contains changes for accepting, storing and evolution of Avro schemas.

### URL
POST `/event-types/{name}/schemas/{version}/compatibility-check` 

### Request Body
It accepts json body containing schema string and type,
for eg: 
``` json
  
 {
    "schema": "{\"type\":\"record\",\"name\":\"Person\",\"fields\":[{\"name\":\"name\",\"type\":\"string\"},{\"name\":\"age\",\"type\":\"int\"},{\"name\":\"address\",\"type\":{\"type\":\"record\",\"name\":\"Address\",\"fields\":[{\"name\":\"street\",\"type\":\"string\"},{\"name\":\"pincode\",\"type\":\"int\"}]}},{\"name\":\"email\",\"type\":\"string\"}]}"
  }
```
### Response Body for 200 status code

Sample response on validation failure,

``` json
{
   "compatible":false,
   "compatibility_mode":"forward",
   "error": “Failed to evolve avro schema due to [{    
             type:READER_FIELD_MISSING_DEFAULT_VALUE,
             location:/fields/0, message:name }]“,
   "validated_against_version": “47”
}

```

Sample response on validation pass,

``` json
{
   "compatible":true,
   "compatibility_mode":"forward",
   "validated_against_version": “47”
}

```
### Response codes
| Http Code | Description                                        |
|-----------|----------------------------------------------------|
| 200       | Validation was performed                           |
| 401       | Unauthenticated                                    |
| 403       | Unauthorised                                       |
| 404       | The version or event-type mentioned does not exist |
| 422       | Unable to parse/process request body                     |

## Create EventType API

Avro event types can now be created by specifying `type` value as `avro_schema` and the avro schema string in `schema` field.

### DIFF
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
@@ -1,6 +1,5 @@
 package org.zalando.nakadi.repository.db;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import org.junit.Before;
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
@@ -164,7 +164,7 @@ public class EventTypeDbRepositoryTest extends AbstractDbRepositoryTest {
 
         repository.saveEventType(eventType);
 
-        eventType.getSchema().setVersion(new JsonVersion("1.1.0"));
+        eventType.getSchema().setVersion("1.1.0");
 
         repository.update(eventType);
 
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
@@ -10,7 +10,6 @@ import org.zalando.nakadi.domain.EventType;
 import org.zalando.nakadi.domain.EventTypeSchema;
 import org.zalando.nakadi.domain.ResourceAuthorization;
 import org.zalando.nakadi.domain.ResourceAuthorizationAttribute;
-import org.zalando.nakadi.domain.JsonVersion;
 import org.zalando.nakadi.exceptions.runtime.DuplicatedEventTypeNameException;
 import org.zalando.nakadi.exceptions.runtime.NoSuchEventTypeException;
 import org.zalando.nakadi.utils.TestUtils;
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
@@ -1,5 +1,6 @@
 package org.zalando.nakadi.repository.db;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.fasterxml.jackson.databind.node.TextNode;
 import org.junit.Before;
@@ -10,7 +11,7 @@ import org.zalando.nakadi.domain.EventType;
 import org.zalando.nakadi.domain.EventTypeSchema;
 import org.zalando.nakadi.domain.ResourceAuthorization;
 import org.zalando.nakadi.domain.ResourceAuthorizationAttribute;
-import org.zalando.nakadi.domain.Version;
+import org.zalando.nakadi.domain.JsonVersion;
 import org.zalando.nakadi.exceptions.runtime.DuplicatedEventTypeNameException;
 import org.zalando.nakadi.exceptions.runtime.NoSuchEventTypeException;
 import org.zalando.nakadi.utils.TestUtils;
@@ -164,7 +165,7 @@ public class EventTypeDbRepositoryTest extends AbstractDbRepositoryTest {
 
         repository.saveEventType(eventType);
 
-        eventType.getSchema().setVersion(new Version("1.1.0"));
+        eventType.getSchema().setVersion(new JsonVersion("1.1.0"));
 
         repository.update(eventType);
 
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
@@ -55,18 +55,20 @@ public class SchemaController {
         return status(HttpStatus.OK).build();
     }
 
-    @RequestMapping(value = "/event-types/{name}/schemas/{version}", method = RequestMethod.PUT)
+    @RequestMapping(value = "/event-types/{name}/schemas/{version}/compatibility-check", method = RequestMethod.POST)
     public ResponseEntity<?> checkCompatibility(@PathVariable("name") final String name,
                                                 @PathVariable("version") final String version,
-                                    @Valid @RequestBody final EventTypeSchemaBase schema,
-                                    final Errors errors) {
+                                                @Valid @RequestBody final EventTypeSchemaBase schema,
+                                                final Errors errors) {
         if (errors.hasErrors()) {
             throw new ValidationException(errors);
         }
 
         final EventType eventType = eventTypeService.get(name);
         if(!schema.getType().equals(eventType.getSchema().getType())){
-            throw new SchemaValidationException("schema type cannot be different");
+            throw new SchemaValidationException(String.format("schema type cannot be " +
+                            "different: expected `%s`, but was `%s`"
+                    , eventType.getSchema().getType(), schema.getType()));
         }
 
         final var toCompareEventTypeSchema = version.equals("latest")?
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
@@ -24,6 +24,7 @@ import org.zalando.nakadi.exceptions.runtime.NoSuchEventTypeException;
 import org.zalando.nakadi.exceptions.runtime.NoSuchSchemaException;
 import org.zalando.nakadi.exceptions.runtime.ValidationException;
 import org.zalando.nakadi.model.CompatibilityResponse;
+import org.zalando.nakadi.model.CompatibilitySchemaRequest;
 import org.zalando.nakadi.service.EventTypeService;
 import org.zalando.nakadi.service.SchemaService;
 
@@ -59,7 +60,7 @@ public class SchemaController {
     @RequestMapping(value = "/event-types/{name}/schemas/{version}/compatibility-check", method = RequestMethod.POST)
     public ResponseEntity<?> checkCompatibility(@PathVariable("name") final String name,
                                                 @PathVariable("version") final String version,
-                                                @Valid @RequestBody final EventTypeSchemaBase schema,
+                                                @Valid @RequestBody final CompatibilitySchemaRequest schema,
                                                 final Errors errors) {
         if (errors.hasErrors()) {
             throw new ValidationException(errors);
@@ -71,7 +72,8 @@ public class SchemaController {
         eventType.setSchema(toCompareEventTypeSchema);
 
         final var newEventTypeBase = new EventTypeBase(eventType);
-        newEventTypeBase.setSchema(schema);
+        newEventTypeBase.setSchema(
+                new EventTypeSchemaBase(eventType.getSchema().getType(), schema.getSchema()));
 
         return compatibilityResponse(eventType, newEventTypeBase);
     }
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
@@ -25,12 +25,10 @@ import org.zalando.nakadi.exceptions.runtime.NoSuchSchemaException;
 import org.zalando.nakadi.exceptions.runtime.ValidationException;
 import org.zalando.nakadi.service.EventTypeService;
 import org.zalando.nakadi.service.SchemaService;
-import org.zalando.problem.Problem;
 
 import javax.validation.Valid;
 
 import static org.springframework.http.ResponseEntity.status;
-import static org.zalando.problem.Status.UNPROCESSABLE_ENTITY;
 
 @RestController
 public class SchemaController {
--- a/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
@@ -16,15 +16,18 @@ import org.zalando.nakadi.domain.EventTypeBase;
 import org.zalando.nakadi.domain.EventTypeSchema;
 import org.zalando.nakadi.domain.EventTypeSchemaBase;
 import org.zalando.nakadi.domain.PaginationWrapper;
-import org.zalando.nakadi.exception.SchemaValidationException;
+import org.zalando.nakadi.exception.SchemaEvolutionException;
 import org.zalando.nakadi.exceptions.runtime.InternalNakadiException;
 import org.zalando.nakadi.exceptions.runtime.InvalidLimitException;
 import org.zalando.nakadi.exceptions.runtime.InvalidVersionNumberException;
 import org.zalando.nakadi.exceptions.runtime.NoSuchEventTypeException;
 import org.zalando.nakadi.exceptions.runtime.NoSuchSchemaException;
 import org.zalando.nakadi.exceptions.runtime.ValidationException;
+import org.zalando.nakadi.model.CompatibilityResponse;
 import org.zalando.nakadi.service.EventTypeService;
 import org.zalando.nakadi.service.SchemaService;
+import org.zalando.problem.Problem;
+import org.zalando.problem.Status;
 
 import javax.validation.Valid;
 
@@ -65,10 +68,14 @@ public class SchemaController {
         }
 
         final EventType eventType = eventTypeService.get(name);
+
         if(!schema.getType().equals(eventType.getSchema().getType())){
-            throw new SchemaValidationException(String.format("schema type cannot be " +
+            final var detail = String.format("schema type cannot be " +
                             "different: expected `%s`, but was `%s`"
-                    , eventType.getSchema().getType(), schema.getType()));
+                    , eventType.getSchema().getType(), schema.getType());
+
+            return status(HttpStatus.CONFLICT).
+                    body(Problem.valueOf(Status.CONFLICT, detail));
         }
 
         final var toCompareEventTypeSchema = version.equals("latest")?
@@ -78,9 +85,20 @@ public class SchemaController {
         final var newEventTypeBase = new EventTypeBase(eventType);
         newEventTypeBase.setSchema(schema);
 
-        schemaService.getValidEvolvedEventType(eventType, newEventTypeBase);
+        return compatibilityResponse(eventType, newEventTypeBase);
+    }
 
-        return ResponseEntity.status(HttpStatus.OK).build();
+    private ResponseEntity<?> compatibilityResponse(final EventType eventType, final EventTypeBase newEventTypeBase) {
+        final var compatibilityMode = newEventTypeBase.getCompatibilityMode();
+        final var version = eventType.getSchema().getVersion().toString();
+        try {
+            schemaService.getValidEvolvedEventType(eventType, newEventTypeBase);
+        } catch (SchemaEvolutionException e) {
+            return status(HttpStatus.OK).
+                    body(new CompatibilityResponse(false, compatibilityMode, e.getMessage(), version));
+        }
+        return status(HttpStatus.OK).
+                body(new CompatibilityResponse(true, compatibilityMode, null, version));
     }
 
     @RequestMapping(value = "/event-types/{name}/schemas", method = RequestMethod.GET)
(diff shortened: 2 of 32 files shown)


api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -53,6 +55,32 @@ public SchemaController(final SchemaService schemaService, final EventTypeServic
...
+        }
+
+        final var toCompareEventTypeSchema = version.equals("latest")?
+                eventType.getSchema(): schemaService.getSchemaVersion(name, version);
+        eventType.setSchema(toCompareEventTypeSchema);
+
+        final var newEventTypeBase = new EventTypeBase(eventType);
+        newEventTypeBase.setSchema(schema);
+
+        schemaService.getValidEvolvedEventType(eventType, newEventTypeBase);
```
adyach, line 79: those spring runtime exception processing is total magic. what happens if they are not compatible?

adyach: what error message and status code will be returned?

api-metastore/src/main/java/org/zalando/nakadi/service/AvroSchemaCompatibility.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Component
+public class AvroSchemaCompatibility {
+
+    private static BiFunction<Schema, Schema, SchemaPairCompatibility> backwardCompatibleFn = (reader, writer) ->
+            SchemaCompatibility.checkReaderWriterCompatibility(reader, writer);
```
adyach, line 25: is avro library thread safe? I expect shema validation can happen from different threads

api-metastore/src/main/java/org/zalando/nakadi/service/AvroSchemaCompatibility.java
```diff
@@ -0,0 +1,153 @@
...
+            case BACKWARD:
+                return backwardCompatibleFn.andThen(this::toAvroIncompatibility).
+                        apply(newSchema, getLastSchema.get());
+
+            case COMPATIBLE:
+                return flatMapToIncompat.apply(
+                        fullyCompatibleFn.apply(newSchema, getLastSchema.get()).stream()
+                );
+
+            case FORWARD_TRANSITIVE:
```
adyach, line 73: I suggest to keep compatibility modes the same as it is for JSON to reduce confusion

api-metastore/src/main/java/org/zalando/nakadi/service/AvroSchemaCompatibility.java
```diff
@@ -0,0 +1,153 @@
...
+                new AvroIncompatibility(incompat.getLocation(), incompat.getMessage(), incompat.getType());
+
+        return schemaPairCompatibility.getResult().getIncompatibilities().stream().
+                map(mapperFn).collect(Collectors.toList());
+    }
+
+    public static class AvroIncompatibility {
+        private final String location;
+        private final String message;
+        private final SchemaCompatibility.SchemaIncompatibilityType mType;
```
adyach, line 104:
```suggestion
        private final SchemaCompatibility.SchemaIncompatibilityType incompatibilityType;
```

api-metastore/src/main/java/org/zalando/nakadi/service/AvroSchemaCompatibility.java
```diff
@@ -0,0 +1,153 @@
...
+import java.util.List;
+import java.util.Objects;
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Component
+public class AvroSchemaCompatibility {
```
adyach, line 22: general comment, it is hard to read the functional style here (maybe it is just me), but it gets to verbose in Java

adyach: for transitive it would be a bit different, but in nutshell what is done:
```
SchemaCompatibility.checkReaderWriterCompatibility(newSchema, getLastSchema.get())
.getResult().getIncompatibilities().
.map(I -> new AvroIncompatibility(incompat.getLocation(), incompat.getMessage(), incompat.getType()))
```

api-metastore/src/main/java/org/zalando/nakadi/validation/schema/CompatibilityModeChangeConstraint.java
```diff
@@ -37,12 +38,14 @@ public CompatibilityModeChangeConstraint(final AdminService adminService,
         final boolean isNakadiAdmin = adminService.isAdmin(AuthorizationService.Operation.WRITE);
         final boolean isEventTypeAdmin = authorizationService
                 .isAuthorized(AuthorizationService.Operation.ADMIN, original.asResource());
-        final boolean isChangeValid = allowedChanges.get(original.getCompatibilityMode())
+        final boolean isChangeValid = original.getSchema().getType() == EventTypeSchemaBase.Type.AVRO_SCHEMA ||
```
adyach, line 41: I do not understand the change here

api-metastore/src/main/java/org/zalando/nakadi/validation/schema/SchemaTypeChangeConstraint.java
```diff
@@ -0,0 +1,18 @@
...
+import org.zalando.nakadi.domain.EventType;
+import org.zalando.nakadi.domain.EventTypeBase;
+
+import java.util.Optional;
+
+public class SchemaTypeChangeConstraint implements SchemaEvolutionConstraint{
+
+    @Override
+    public Optional<SchemaEvolutionIncompatibility> validate(final EventType original, final EventTypeBase eventType) {
+        if(!original.getSchema().getType().equals(eventType.getSchema().getType())){
```
adyach, line 12: how do we plan to migrate from json to avro?

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
```diff
@@ -53,6 +55,11 @@ public EventType saveEventType(final EventTypeBase eventTypeBase) throws Interna
         }
     }
 
+    private String constructVersionString(final EventTypeBase eventTypeBase) {
```
adyach, line 58: would be nice if schema know what type it is then we do not need that method. instead would just create it itself

api-metastore/src/main/java/org/zalando/nakadi/service/AvroSchemaCompatibility.java
samizzy: yes, correct.

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
samizzy: Currently 200 is returned for successful evolution otherwise 422 
An example response is,
```
{
    "title": "Unprocessable Entity",
    "status": 422,
    "detail": "Failed to evolve avro schema due to [{ type:READER_FIELD_MISSING_DEFAULT_VALUE, location:/fields/0, message:zip_code }]"
}
```

api-metastore/src/main/java/org/zalando/nakadi/validation/schema/SchemaTypeChangeConstraint.java
samizzy: I put this in place as I am not sure what things will break on allowing such a change, this is a temporary measure until we figure out things.

api-metastore/src/main/java/org/zalando/nakadi/service/AvroSchemaCompatibility.java
samizzy: I could not find any document stating that `SchemaCompatibility.checkReaderWriterCompatibility` is thread-safe but going through the code of this method I didnt find any state that would be shared for different calls to this method. Every call results in creation of new objects.

api-metastore/src/main/java/org/zalando/nakadi/validation/schema/CompatibilityModeChangeConstraint.java
samizzy: This would mean skip checking of compatibility mode type change when its a Avro type schema, I am not sure if it is desirable but it was done to reduce initial complexity for users.

samizzy: But again the complexity was supposed to be there when the new modes compatibility modes were added, since they are removed now I guess we could remove this change here?

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
samizzy: I have moved to type enum itself since that seemed like the most appropriate place

api-metastore/src/main/java/org/zalando/nakadi/validation/schema/CompatibilityModeChangeConstraint.java
samizzy: Reverted this to how it was originally, so this check will be performed whether its avro or json schema.

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -53,6 +58,49 @@ public SchemaController(final SchemaService schemaService, final EventTypeServic
...
+        return compatibilityResponse(eventType, newEventTypeBase);
+    }
+
+    private ResponseEntity<?> compatibilityResponse(final EventType eventType, final EventTypeBase newEventTypeBase) {
+        final var compatibilityMode = newEventTypeBase.getCompatibilityMode();
+        final var version = eventType.getSchema().getVersion().toString();
+        try {
+            schemaService.getValidEvolvedEventType(eventType, newEventTypeBase);
+        } catch (SchemaEvolutionException e) {
+            return status(HttpStatus.OK).
```
adyach, line 97: maybe it is a matter of taste, but I prefer to use 2 local vars and assign values:
`compatible = true`
`error = null`

and only sigle return at the end

adyach: it will also clarify magic contants in the CompatibilityResponse ctr

api-metastore/src/main/java/org/zalando/nakadi/service/AvroSchemaCompatibility.java
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        final Supplier<Schema> getLastSchema = () -> original.get(original.size() - 1);
+
+        final Function<Stream<SchemaPairCompatibility>, List<AvroIncompatibility>>
+                flatMapToIncompat = inputStream -> inputStream.
+                        map(this::toAvroIncompatibility).
+                        flatMap(Collection::stream).collect(Collectors.toList());
+
+        switch (compatibilityMode) {
```
adyach, line 59: what about `none` type, I would specify it explicitly and throw an exception on unrecognised type

api-metastore/src/main/java/org/zalando/nakadi/service/SchemaEvolutionService.java
```diff
@@ -99,6 +107,48 @@ private void collectErrorMessages(final ValidationException e,
...
+            validateAvroSchema(Collections.singletonList(original.getSchema()),
+                    eventType.getSchema(), compatibilityMode);
+
+        final var level = AvroUtils.getParsedSchema(original.getSchema().getSchema()).
+                equals(AvroUtils.getParsedSchema(eventType.getSchema().getSchema()))? NO_CHANGES : MAJOR;
+
+        return bumpVersion(original, eventType, level);
+    }
+
+    private void validateAvroSchema(final List<? extends EventTypeSchema> original,
```
adyach, line 131: should indicate throwing of `SchemaEvolutionException`

api-metastore/src/main/java/org/zalando/nakadi/service/SchemaEvolutionService.java
```diff
@@ -99,6 +107,48 @@ private void collectErrorMessages(final ValidationException e,
...
+            case JSON_SCHEMA:
+                return evolveJsonSchema(original, eventType);
+            case AVRO_SCHEMA:
+                return evolveAvroSchema(original, eventType);
+            default:
+                throw new RuntimeException("unsupported type for evolving " + eventType.getSchema().getType());
+        }
+    }
+
+    private EventType evolveAvroSchema(final EventType original, final EventTypeBase eventType) {
```
adyach, line 120: should indicate throwing of `SchemaEvolutionException`

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
```diff
@@ -30,6 +30,15 @@ public SchemaRepository(final JdbcTemplate jdbcTemplate, final ObjectMapper obje
                 new SchemaRowMapper());
     }
 
+    public List<EventTypeSchema> listSchemas(final String name, final String version) {
```
adyach, line 33: the method is not used, and there is a similar method

adyach:
there is still schema type is spread across the code (3 places), I would strive to avoid by creating abstraction and keep types in a single place. other than that looks good 👍

adyach:
👍

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
samizzy: > maybe it is a matter of taste, but I prefer to use 2 local vars and assign values: `compatible = true` `error = null`
> 
> and only sigle return at the end

Since our code style requires all variables to be final, its not possible to reassign variables to new values and only have a single return at the end. Something like below is not possible.

``` java
    private ResponseEntity<?> compatibilityResponse(final EventType eventType, final EventTypeBase newEventTypeBase) {
        final var compatibilityMode = newEventTypeBase.getCompatibilityMode();
        final var version = eventType.getSchema().getVersion().toString();
        var compatible = true;
        String errorMsg = null;
        try {
            schemaService.getValidEvolvedEventType(eventType, newEventTypeBase);
        } catch (SchemaEvolutionException e) {
            compatible = false;
            errorMsg = e.getMessage();
        }
        return status(HttpStatus.OK).
                body(new CompatibilityResponse(compatible, compatibilityMode, errorMsg, version));
    }
```

api-metastore/src/main/java/org/zalando/nakadi/service/AvroSchemaCompatibility.java
samizzy: done

api-metastore/src/main/java/org/zalando/nakadi/service/SchemaEvolutionService.java
samizzy: done

api-metastore/src/main/java/org/zalando/nakadi/service/SchemaEvolutionService.java
samizzy: done

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
samizzy: removed, thanks escaped my notice :)

core-common/src/main/java/org/zalando/nakadi/domain/EventTypeSchema.java
```diff
@@ -1,9 +1,20 @@
...
+import static com.fasterxml.jackson.annotation.JsonTypeInfo.Id;
+
 import org.joda.time.DateTime;
 
 public class EventTypeSchema extends EventTypeSchemaBase {
+    @JsonTypeInfo(use = Id.NAME, property = "type", include = As.EXTERNAL_PROPERTY)
+    @JsonSubTypes(value = {
+            @JsonSubTypes.Type(value = JsonVersion.class, name = "JSON_SCHEMA"),
+            @JsonSubTypes.Type(value = AvroVersion.class, name = "AVRO_SCHEMA")
+    })
```
antban, line 15: @samizzy 
Not that I am against fancy version matching, but it seems that it would be much simpler here to use string representation and create actual object to process logic somewhere in code where it is used. 
There following responsibilities are on this class now: 
1. DTO
2. serialization/deserialization of backend logic through dto
3. backend logic of schema evolution

If we leave only DTO part, than the property type will be string, leaving all the problems with mappings behind

antban: I mean make it of type `string`

core-common/src/main/java/org/zalando/nakadi/domain/EventTypeSchemaBase.java
```diff
@@ -17,7 +17,17 @@ public EventTypeSchemaBase(final Type type, final String schema) {
     }
 
     public enum Type {
-        JSON_SCHEMA
+        JSON_SCHEMA("1.0.0"), AVRO_SCHEMA("1");
```
antban, line 20: Are we sharing anywhere semantic versioning?

core-metastore/src/main/java/org/zalando/nakadi/repository/db/EventTypeRepository.java
samizzy: Moved this to respective Version classes now

samizzy:
> there is still schema type is spread across the code (3 places), I would strive to avoid by creating abstraction and keep types in a single place. other than that looks good 👍

Due to recent changes, introduced it in one more place now. I agree it makes sense to refactor the code to reduce this but it would to take some good amount of refactoring. I will make a separate PR to achieves this.

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/EventTypeDbRepositoryTest.java
```diff
@@ -10,7 +10,7 @@
 import org.zalando.nakadi.domain.EventTypeSchema;
 import org.zalando.nakadi.domain.ResourceAuthorization;
 import org.zalando.nakadi.domain.ResourceAuthorizationAttribute;
-import org.zalando.nakadi.domain.Version;
+import org.zalando.nakadi.domain.JsonVersion;
```
antban, line 13: weird...

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -53,6 +58,49 @@ public SchemaController(final SchemaService schemaService, final EventTypeServic
...
     }
 
+    @RequestMapping(value = "/event-types/{name}/schemas/{version}/compatibility-check", method = RequestMethod.POST)
+    public ResponseEntity<?> checkCompatibility(@PathVariable("name") final String name,
+                                                @PathVariable("version") final String version,
+                                                @Valid @RequestBody final EventTypeSchemaBase schema,
+                                                final Errors errors) {
+        if (errors.hasErrors()) {
+            throw new ValidationException(errors);
+        }
```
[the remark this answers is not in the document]
samizzy: tbh its so automagic I am not aware of the role of this line here ;) but we have been using it in many places.

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -53,6 +58,49 @@ public SchemaController(final SchemaService schemaService, final EventTypeServic
...
+
+        final EventType eventType = eventTypeService.get(name);
+
+        if(!schema.getType().equals(eventType.getSchema().getType())){
```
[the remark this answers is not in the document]
samizzy: Yes, because the check for incompatibilities schemaevolution service throws SchemaEvolutionException and that is mapped to 422 status code already. I wanted to respond with 409 (CONFLICT) particularly when using the compatibility api.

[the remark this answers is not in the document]
antban: Well, for that purpose the common approach is to use different kind of exception and map it in controller exception mapper. Thus the logic will be stored at the same place, and controller will be clean.

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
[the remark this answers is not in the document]
antban: Oh, maybe it does make sense to write a test that validation error is processed and do not write this kind of behavior everywhere (as it is useless). 
Usually approach with `Errors` mapping is used when some additional actions should be done (like handle parsing exception, when parsing failed and enriching errors with someth.), but we do not have this requirement here. 
I would suggest to make a test and remove this `errors.hasErrors` calls from everywhere.

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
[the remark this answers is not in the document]
samizzy: As discussed removed 409 altogether

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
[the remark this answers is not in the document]
samizzy: I think this line was placed here to respond with 422, when I remove the `Errors errors` line from the method, then 400 is being returned. I am not sure but maybe its due to this [line](https://github.com/zalando/nakadi/blob/316bf96e2916cd6b030d4dda6b574bc7bb5bafed/core-services/src/main/java/org/zalando/nakadi/controller/advice/NakadiProblemExceptionHandler.java#L73). So I have kept it here and added test cases to ensure ValidationException is being thrown.

core-common/src/main/java/org/zalando/nakadi/domain/EventTypeSchemaBase.java
samizzy: removed version from here

antban:
:+1:

samizzy:
:+1:

PR #1375, opened 2022-03-02
zstd compression for content

adyach:
the request enables zstd compression for content econding


core-common/src/test/java/org/zalando/nakadi/util/CompressionBodyRequestFilterTest.java
```diff
@@ -0,0 +1,21 @@
...
+public class CompressionBodyRequestFilterTest {
+
+    @Test
+    public void testDecompressOneZstdEvent() throws Exception {
+        final CompressionBodyRequestFilter.CompressionServletInputStream stream =
+                new CompressionBodyRequestFilter.CompressionServletInputStream(
+                        CompressionBodyRequestFilterTest.class.getResourceAsStream("event.zstd"));
+
+        final String json = new String(stream.readAllBytes());
+        System.out.println(json);
```
tor-vs-floki, line 16: Leftover?

core-common/build.gradle
```diff
@@ -97,6 +97,8 @@ dependencies {
     compile "org.apache.avro:avro:1.10.2"
     compile "com.fasterxml.jackson.dataformat:jackson-dataformat-avro:2.8.5"
 
+    compile 'com.github.luben:zstd-jni:1.5.0-4'
```
a1exsh, line 100: Is it the same dependency that the kafka client brings with it or a different one?

core-common/src/main/java/org/zalando/nakadi/util/CompressionBodyRequestFilter.java
```diff
@@ -44,18 +45,22 @@ public final void doFilter(final ServletRequest servletRequest, final ServletRes
...
             return;
+        } else if (contentEncodingOpt.isPresent()) {
+            final String contentEncoding = contentEncodingOpt.get();
+            if (contentEncoding.contains("gzip")) {
+                request = new GzipServletRequestWrapper(request);
+            }
+
+            if (contentEncoding.contains("zstd")) {
+                request = new ZstdServletRequestWrapper(request);
+            }
```
antban, line 61: It will be better to use only one request wrapper (using `else if` construction)

core-common/src/main/java/org/zalando/nakadi/util/CompressionBodyRequestFilter.java
```diff
@@ -98,14 +104,31 @@ public BufferedReader getReader() throws IOException {
...
+    }
 
-    private class GzipServletInputStream extends ServletInputStream {
+    static class CompressionServletInputStream extends ServletInputStream {
 
         private final InputStream inputStream;
 
-        GzipServletInputStream(final InputStream inputStream) throws IOException {
+        CompressionServletInputStream(final InputStream inputStream) throws IOException {
             super();
```
antban, line 130: Is it possible to remove this `super()` call ? (just to clean the code)

core-common/src/main/java/org/zalando/nakadi/util/CompressionBodyRequestFilter.java
```diff
@@ -98,14 +104,31 @@ public BufferedReader getReader() throws IOException {
         }
     }
 
+    static class ZstdServletRequestWrapper extends HttpServletRequestWrapper {
```
antban, line 107: Seems that these classes (`GzipServletRequestWrapper` and `ZstdServletRequestWrapper`) are different only by the function that converts input streams one to each other. Does it make sense to simplify now to a single class with this function as a parameter? It will simplify addition of any other content-encodings/

core-common/src/test/java/org/zalando/nakadi/util/CompressionBodyRequestFilterTest.java
```diff
@@ -0,0 +1,21 @@
...
+        final JSONArray array = new JSONArray(json);
+
+        Assert.assertEquals(1, array.length());
+    }
+}
```
antban, line 21: :-(

core-common/build.gradle
adyach: yes

core-common/src/test/java/org/zalando/nakadi/util/CompressionBodyRequestFilterTest.java
```diff
@@ -0,0 +1,26 @@
+package org.zalando.nakadi.util;
+
+import com.github.luben.zstd.ZstdInputStream;
+import org.eclipse.jetty.server.Request;
+import org.json.JSONArray;
+import org.junit.Assert;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+public class CompressionBodyRequestFilterTest {
```
antban, line 10: TBH, after looking precisely what this test is doing - I am not sure if this test is needed at all. You are basically testing library, and hit hit only 1 line of actual code ((

adyach:
👍

PR #1377, opened 2022-03-11
Add more fields to nakadi.access.log event type

a1exsh:
Internal ticket: 963


core-common/src/main/resources/nakadi.access.log.avsc
```diff
@@ -15,6 +15,10 @@
       "name": "query",
       "type": "string"
     },
+    {
```
adyach, line 18: I am not sure that is possible without introduction of versions :(

a1exsh: Is it because the field is added in the middle?  Was it not a concern here: https://github.com/zalando/nakadi/pull/1373/files#diff-c9948493d81e5ba0addbf3d4caf4f33320ba07dd503c7a205ff55063a6e44ac3 ?

adyach: it was, but avro was disabled ;)

antban: with avro it is not possible.

antban:
@a1exsh 
What do you think about adding request size and response size? 
It is very important for requests analysis.

app/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -74,13 +74,13 @@ private AsyncRequestListener(final HttpServletRequest request, final HttpServlet
             this.requestLogInfo = new RequestLogInfo(request, startTime);
 
             if (isAccessLogEnabled()) {
-                logToAccessLog(this.requestLogInfo, HttpStatus.PROCESSING.value(), 0L);
+                logToAccessLog(this.requestLogInfo, HttpStatus.PROCESSING.value(), 0L, 0L);
```
adyach, line 77: magic values, assign to var

app/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -120,41 +120,52 @@ protected void doFilterInternal(final HttpServletRequest request,
...
+                requestLogInfo.contentLength,
+                responseLength);
     }
 
-    private void logToAccessLog(final RequestLogInfo requestLogInfo, final int statusCode, final Long timeSpentMs) {
-        ACCESS_LOGGER.info("{} \"{}{}\" \"{}\" \"{}\" {} {}ms \"{}\" \"{}\" {}B",
+    private void logToAccessLog(final RequestLogInfo requestLogInfo, final int statusCode, final Long responseLength,
+            final Long timeSpentMs) {
+
+        ACCESS_LOGGER.info("{} \"{}{}\" \"{}\" \"{}\" {} {}ms \"{}\" \"{}\" {}B -> {}B",
```
adyach, line 168: do not forget to fix log parser

core-common/src/main/java/org/zalando/nakadi/service/AvroSchema.java
```diff
@@ -7,41 +7,70 @@
...
 
     @Autowired
     public AvroSchema(
             final AvroMapper avroMapper,
             final ObjectMapper objectMapper,
-            @Value("${nakadi.avro.schema.metadata:classpath:metadata.avsc}")
-            final Resource metadataRes,
-            @Value("${nakadi.avro.schema.nakadi-access-log:classpath:nakadi.access.log.avsc}")
-            final Resource nakadiAccessLogRes)
+            @Value("${nakadi.avro.schema.metadata:classpath:event-type-schema/metadata.avsc}")
```
adyach, line 38: lets keep only resource to even type schema

core-common/src/main/java/org/zalando/nakadi/service/AvroSchema.java
```diff
@@ -7,41 +7,70 @@
...
         this.objectMapper = objectMapper;
-        this.metadataSchema = AvroUtils.getParsedSchema(metadataRes.getInputStream());
-        this.nakadiAccessLogSchema = AvroUtils.getParsedSchema(nakadiAccessLogRes.getInputStream());
-    }
+        this.metadataSchema = AvroUtils.getParsedSchema(metadataSchemaRes.getInputStream());
+        this.eventTypeSchema = new HashMap<>();
 
-    public Schema getNakadiAccessLogSchema() {
-        return nakadiAccessLogSchema;
+        for (final String eventTypeName : Set.of("nakadi.access.log")) {
```
adyach, line 48: could we try to load every file from `event-type-schema` folder and load then into tree map

core-services/src/main/java/org/zalando/nakadi/service/timeline/MultiTimelineEventConsumer.java
```diff
@@ -252,8 +252,10 @@ private void electTopicRepositories() throws InvalidCursorException {
                 final TopicRepository repo = entry.getKey();
                 LOG.info("Creating underlying consumer for client id {} and cursors {}",
                         clientId, Arrays.deepToString(entry.getValue().toArray()));
+
+                final String eventTypeName = entry.getValue().get(0).getEventType(); // FIXME
```
adyach, line 256: it looks like there is a better place to get the event type name or Kafka record deserialiser

core-common/src/main/java/org/zalando/nakadi/repository/kafka/AvroDeserializerWithSequenceDecoder.java
```diff
@@ -28,7 +35,11 @@ public AvroDeserializerWithSequenceDecoder(final AvroSchema schemas) {
             metadata.put("received_at", new DateTime(
                     (long) metadata.get("received_at"), DateTimeZone.UTC).toString());
 
-            final GenericRecord event = eventSequenceDecoder.read(envelop.getPayload());
+            final SequenceDecoder eventDecoder = eventSequenceDecoders.computeIfAbsent(
+                    metadata.get("schema_version").toString(),
+                    (v) -> new SequenceDecoder(schemas.getEventTypeSchema(eventTypeName, v)));
```
adyach, line 40: it looks like the sequence decoder does not need to know schema for specific even type. it looks like it would help it to work with all available event type schemas and decide on the fly which one to use.

a1exsh: Well, in the current implementation it does need to know the schema, so that it can initialize the datum reader at construct time.  If we try to make it not depend on the schema, we would just move this complexity from here to that class.

adyach: I mean, you have event type inside metadata

adyach: so it should be sufficient to know how to deserialise inside the deserialiser

adyach: or near it, but the point is that you do not need to pass the event type from outside

core-common/src/main/java/org/zalando/nakadi/service/AvroSchema.java
a1exsh: This will require a bigger change to make this work nicely: e.g. moving these files from resources to container level.  Otherwise, listing resources within a jar within a jar seems to require a lot of effort.

adyach:
👍

PR #1378, opened 2022-03-21
Fix use of logicalType in avro schema

a1exsh:
Internal ticket: 1069


adyach:
👍

core-common/src/main/java/org/zalando/nakadi/service/AvroSchema.java
```diff
@@ -24,12 +24,11 @@
 @Service
 public class AvroSchema {
 
-    public static final byte METADATA_VERSION = 0;
-    public static final Comparator<String> SCHEMA_VERSION_COMPARATOR = Comparator.comparingInt(Integer::parseInt);
+    public static final String METADATA_KEY = "_metadata";
```
adyach, line 27: `_` is that indication of the internal_smth?

adyach: to avoid user event type names `metadata`

a1exsh: sort of. it's a temp. solution anyway and hopefully will be replaced with the proper registry

PR #1379, opened 2022-03-22
Include reason for auth error in response

samizzy:
# One-line summary
Include reason for auth error in response


thomasabraham:
:+1:

PR #1380, opened 2022-03-28
Derive the absolute value at the very end of the operation when doing hash partitioning

tor-vs-floki:
Internal ticket 905


core-services/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -57,7 +57,7 @@ public String calculatePartition(final EventType eventType, final JSONObject eve
                     .sum();
 
 
-            int partitionIndex = abs(hashValue) % partitions.size();
+            int partitionIndex = abs(hashValue % partitions.size());
```
a1exsh, line 60: Cool! :)

a1exsh:
👍

tor-vs-floki:
👍

PR #1383, opened 2022-03-31
Experimenting with a Builder for EventMetadata

thomasabraham:
# Experimenting with a Builder for EventMetadata

With the introduction of Avro, Metadata, among other KPI objects, should be serialized to JSON or Avro based on settings. Currently we have two independent implementation for building a Metadata object, one for JSON and one for Avro. The PR attempt to unify this and have a single Metadata builder object that can build either an Avro version or a JSON version as necessary.

### DIFF
--- a/core-services/src/main/java/org/zalando/nakadi/service/publishing/EventMetadata.java
+++ b/core-services/src/main/java/org/zalando/nakadi/service/publishing/EventMetadata.java
@@ -1,14 +1,27 @@
 package org.zalando.nakadi.service.publishing;
 
+import org.apache.avro.Schema;
+import org.apache.avro.generic.GenericRecord;
+import org.apache.avro.generic.GenericRecordBuilder;
 import org.json.JSONObject;
 import org.springframework.stereotype.Component;
 import org.zalando.nakadi.util.FlowIdUtils;
 import org.zalando.nakadi.util.UUIDGenerator;
 
 import java.time.Instant;
+import java.util.UUID;
 
 @Component
 public class EventMetadata {
+    public static final String METADATA_FIELD = "metadata";
+    public static final String OCCURRED_AT_FIELD = "occurred_at";
+    public static final String EID_FIELD = "eid";
+    public static final String FLOW_ID_FIELD = "flow_id";
+    public static final String RECEIVED_AT_FIELD = "received_at";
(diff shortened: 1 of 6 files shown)


core-services/src/test/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisherTest.java
```diff
@@ -85,7 +86,7 @@ public void testPublishingAccessLogWithAvro() throws IOException {
         final var avroSchema = new AvroSchema(new AvroMapper(), new ObjectMapper(), eventTypeRes);
         final NakadiKpiPublisher publisher = new NakadiKpiPublisher(featureToggleService,
                 jsonProcessor, binaryProcessor, usernameHasher,
-                new EventMetadataTestStub(), new UUIDGenerator(), avroSchema);
+                new EventMetadata(new UUIDGenerator()), new UUIDGenerator(), avroSchema);
```
a1exsh, line 89: Not sure why `UUIDGenerator` two times?

thomasabraham: Good catch, I'll fix that.
It is in test though.

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventMetadata.java
```diff
@@ -1,25 +1,155 @@
...
 import org.springframework.stereotype.Component;
 import org.zalando.nakadi.util.FlowIdUtils;
 import org.zalando.nakadi.util.UUIDGenerator;
 
 import java.time.Instant;
+import java.util.UUID;
 
 @Component
 public class EventMetadata {
+    public static final String METADATA_FIELD = "metadata";
```
a1exsh, line 16: This seems to be out of place.  The rest of field names here are for the metadata object itself, but this one is for the object containing the metadata...

And I'm not sure that we really need these as named constants in the first place (we have tests to verify correctness anyway). ;)

thomasabraham: I would like to keep this as static fields. 
This prevents any future update in one place and not the other.

thomasabraham:
Closing this PR as I realize that we have to figure out how to do metadata enrichment in general before this.

PR #1386, opened 2022-04-01
Introducing KPIEvent and supporting Avro for nakadi.subscription.log

thomasabraham:
# Introducing KPIEvent and supporting Avro for nakadi.subscription.log

> Zalando ticket : team-aruha/issues/984

## Description
A new class called `KPIEvent` is introduced. All KPI events are supposed
to be exted from this. Currently the only KPI event that is published
using this is the `SubscriptionLogEvent`. The `KPIEvent` supports
mapping the classes that extends it directly to `GenericRecord` or
`JSONObject`; as long as the fields are annotated with `@KPIField`
annotation. The `NakadiKpiPublisher` class is also supporting the
publishing of KPI Events now.

## Review
- [ ] Tests
- [ ] Documentation


core-common/src/main/java/org/zalando/nakadi/domain/kpi/KPIEvent.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /*
+        Implementation of this method should return the event-type name of the specific KPI event.
+        This method should not be confused with the getter "getEventType()" in some specific KPI events;
+        which is why this method is named "eventTypeOfThisKPIEvent()
+     */
+    public abstract String eventTypeOfThisKPIEvent();
+
+    private Stream<Field> getKPIFields() {
+        return Arrays.stream(this.getClass().getDeclaredFields())
```
adyach, line 36: that works only with a higher level classes, what if I have more sophisticated structures ?

thomasabraham: Yes, the current implementation is only taking care of the simple structure with one level. 
I'm hoping to take care of the more complex structure once I do have a example for it.

core-common/src/main/java/org/zalando/nakadi/domain/kpi/KPIEvent.java
```diff
@@ -0,0 +1,65 @@
...
+                .filter(field -> field.isAnnotationPresent(KPIField.class));
+    }
+
+    public JSONObject asJsonObject() throws KPIEventMappingException {
+        final var jsonObject = new JSONObject();
+        getKPIFields().forEach(field -> {
+            try {
+                jsonObject.put(field.getAnnotation(KPIField.class).value(), field.get(this));
```
adyach, line 44: Nice approach with annotations! looks very clean in this way. 

Can optimize it even further - cache fields annotations for each class (usually it is done by creating/initializing mappers once per startup) and avoid using clazz.getFields() for each publishing call (there are thousands per second publishes).

Another thing - maybe we can avoid injecting mappers logic into objects (looks like mixing it alltogether, plus inheritance actually removes the beauty of annotations usage, maybe it is also possible to improve).

thomasabraham: Good point, I'll do this.

thomasabraham:
:+1:

core-common/src/main/java/org/zalando/nakadi/domain/kpi/KPIEvent.java
thomasabraham: Done

core-common/src/main/java/org/zalando/nakadi/service/KPIEventMapper.java
```diff
@@ -0,0 +1,109 @@
...
+                .map(field -> findGetterByField(mappedClass, field))
+                .collect(Collectors.toSet());
+    }
+
+    private KPIGetter findGetterByField(final Class<? extends KPIEvent> clazz, final Field field) {
+        final var kpiField = field.getAnnotation(KPIEvent.KPIField.class);
+        final String getterName;
+        if (kpiField.getter().isEmpty()) {
+            final var fieldName = field.getName();
+            getterName = "get" + fieldName.substring(0, 1).toUpperCase() + fieldName.substring(1);
```
adyach, line 61: this is quite dangerous. what about `isRunning()`, upper case... can you just get all the methods. 
for example, Jackson in some places makes [such checks](https://github.com/FasterXML/jackson-databind/blob/39885e702795dc4d57f416b3e44ea1f86d33d5af/src/main/java/com/fasterxml/jackson/databind/introspect/POJOPropertiesCollector.java#L680) to get differentiate getters / setters 

but ultimately it is better to use existing utils to access fields, getters etc. e.g. BeanUtils from Spring or Apache commons. however, I have limited experience writing such logic.

core-common/src/main/java/org/zalando/nakadi/service/KPIEventMapper.java
```diff
@@ -0,0 +1,109 @@
...
+                throw new KPIEventMappingException("Cannot map KPIEvent " + kpiEvent.getClass().getName()
+                        + " to JSON. Unable to read " + kpiGetter.name, iae);
+            }
+        });
+        return jsonObject;
+    }
+
+    private Set<KPIGetter> gettersOf(final Class<? extends KPIEvent> clazz) {
+        if (!this.fieldGetterCache.containsKey(clazz)) {
+            synchronized (this) {
```
adyach, line 101: I am not sure why do you have it? we should make sure that we have loaded all possible kpi events before we start, then there will be no need to cache on the fly

core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
```diff
@@ -49,6 +53,42 @@ protected NakadiKpiPublisher(
         this.eventMetadata = eventMetadata;
         this.uuidGenerator = uuidGenerator;
         this.avroSchema = avroSchema;
+        this.kpiEventMapper = new KPIEventMapper(Set.of(SubscriptionLogEvent.class));
+    }
+
+    public void publish(final Supplier<KPIEvent> kpiEventSupplier) {
```
adyach, line 59: I miss usage of the method

core-common/src/main/java/org/zalando/nakadi/domain/kpi/SubscriptionLogEvent.java
```diff
@@ -0,0 +1,34 @@
+package org.zalando.nakadi.domain.kpi;
+
+import org.zalando.nakadi.config.KPIEventTypes;
+
+public class SubscriptionLogEvent extends KPIEvent {
```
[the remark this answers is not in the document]
thomasabraham: Extending allows it to be passed to the `publish(final Supplier<KPIEvent> kpiEventSupplier)` method.
While annotation is used to indicate the fields.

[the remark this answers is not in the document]
adyach: missed that, all good then

core-common/src/main/java/org/zalando/nakadi/service/KPIEventMapper.java
thomasabraham: I'll continue to improve this. I didn't want to over engineer with all the cases in the first version itself; especially without an example.

core-common/src/main/java/org/zalando/nakadi/service/KPIEventMapper.java
thomasabraham: We plan to load it all during start, but this is there to make the code still work if we forgot to add one.

adyach: can we avoid that logic for now and fail if the code meets unrecognised bean ?

core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
thomasabraham: Good point; forgot to use the new feature.

thomasabraham:
:+1:

core-common/src/main/java/org/zalando/nakadi/service/KPIEventMapper.java
thomasabraham: Done

core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
thomasabraham: This is done now.

adyach:
👍

thomasabraham:
:+1:

PR #1387, opened 2022-04-04
partition as a string

adyach:
partition should be string as it is for JSON schema


a1exsh:
`testAvroEventWasSerialized` needs a fix as well

a1exsh:
:+1:

adyach:
👍

PR #1389, opened 2022-04-07
Make kafka producer retries setting configurable

a1exsh:
Enabling idempotent producer requires `retries` to be set to a value greater
than 0.  The default value is effectively "endless retries", but we might want
to stop trying earlier, so make it configurable and set it to a small value
for now.


a1exsh:
:+1:

adyach:
👍

PR #1390, opened 2022-04-07
Configure new spring version properly

tor-vs-floki:
Config file processing has changed since Spring Boot 2.4 and this PR handles the change accordingly


a1exsh:
:+1:

tor-vs-floki:
👍

PR #1391, opened 2022-04-07
Updated metadata schema and added tests

samizzy:
Internal ticket: 1084


core-common/src/main/resources/event-type-schema/_metadata/_metadata.3.avsc
```diff
@@ -0,0 +1,97 @@
...
+         ],
+         "default":null
+      },
+      {
+         "name":"parent_eids",
+         "type":[
+            "null",
+            {
+            "type":"array",
+            "items":"string"
```
a1exsh, line 84: what about `logicalType`?  it should be `uuid`, as for the `eid` field

core-common/src/main/resources/event-type-schema/_metadata/_metadata.3.avsc
```diff
@@ -0,0 +1,97 @@
+{
+   "name":"Metadata",
```
a1exsh, line 2: The lack of space after `:` makes the whole file look a bit odd.

core-common/src/main/resources/event-type-schema/_metadata/_metadata.3.avsc
```diff
@@ -0,0 +1,97 @@
...
+            }
+         ],
+         "default":null
+      },
+      {
+         "name":"schema_version",
+         "type":"string"
+      },
+      {
+         "name":"published_by",
```
a1exsh, line 45: I think this should also be nullable, as it's set by Nakadi, not by the clients.

core-common/src/test/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializerTest.java
```diff
@@ -63,6 +72,89 @@ public void testDeserializeAvro() throws IOException {
...
+                Byte.parseByte(metadataVersion),
+                getMetadataWriter(metadataVersion, "1", modifyMetadata),
+                eventWriter);
+
+        // try to deserialize that data when we would read Kafka record
+        final byte[] deserializedEvent = deserializer.deserialize(
+                NakadiRecord.Format.AVRO.getFormat(),
+                data
+        );
+    }
```
a1exsh, line 155: Can we add any assertion here?

core-common/src/main/resources/event-type-schema/_metadata/_metadata.3.avsc
```diff
@@ -0,0 +1,106 @@
+{
+  "name": "Metadata",
+  "type": "record",
+  "namespace": "org.zalando.nakadi.generated.avro",
```
samizzy, line 4: Added namespace as well, which will be taken as package when generating stubs

adyach: where usually generated classes go? do they have a separate module?

core-common/src/main/resources/event-type-schema/_metadata/_metadata.3.avsc
```diff
@@ -0,0 +1,106 @@
...
+  "doc": "Event metadata defines data is not related to actual event payload data",
```
adyach, line 5: maybe we can say instead to what that date relates? like `Event metadata defines data about the payload and additional information for Nakadi operations`

core-common/src/main/resources/event-type-schema/_metadata/_metadata.3.avsc
samizzy: They are not generated and they dont have a separate module currently. The temporary solution is user of generated classes will have a copy of the .avsc file and will generate the classes locally.

core-common/src/main/resources/event-type-schema/_metadata/_metadata.3.avsc
samizzy: Alright that sounds better, updated the same

adyach:
👍

adyach:
👍

samizzy:
:+1:

core-common/src/test/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializerTest.java
```diff
@@ -52,17 +62,155 @@ public void testDeserializeAvro() throws IOException {
...
+
+        final JSONObject jsonObject = new JSONObject()
+        .put("flow_id", "hek")
+        .put("partition", 0)
+        .put("received_at", SOME_TIME)
+        .put("published_by", "nakadi-test");
+
+        final var actualJson = getSerializedJsonObject(metadataVersion, jsonObject);
+        final var expectedJson = getExpectedNode0(jsonObject.put("received_at", SOME_TIME_DATE_STRING));
+        Assert.assertTrue(expectedJson.similar(actualJson));
```
a1exsh, line 84: Is `similar` a strong enough test here?  Shouldn't we test for real equality?

samizzy: JSONObject does not ovveridde equals method, so it just checks referential equality. From docs about similar,
```
    /**
     * Determine if two JSONObjects are similar.
     * They must contain the same set of names which must be associated with
     * similar values.
     *
     * @param other The other JSONObject
     * @return true if they are equal
     */

```

a1exsh: I found the docs as well.  But what does "similar values" really mean?  Is it a recursive application of the same method?  What about non-object values?  Is it enough for our test (e.g. do we ignore the actual values)?

samizzy: Yes, its a recursive function as long as its a `JSONObject` or `JSONArray`, but when we get to actual values (base case), the `equals()` method of the value is called. For non - object values (primitives?), the equals method of Primitive Wrapper classes should be getting called. Even though its called similar, it is the strongest equality check offered, perhaps it should just be called equals, although there be some reason they decided to call it similar instead (maybe order of elements is not important in similar, thats what I understand reading the implementation code)

a1exsh: OK, makes sense.  Thank you :)

PR #1395, opened 2022-04-14
Publising Avro API

adyach:
The Nakadi Publishing API is extended to be able to accept Avro binary payloads.

Request:
POST /event-types/{name}/events HTTP/1.1
…
Content-Type: application/avro-binary
Content-Encoding: gzip
Content-Length: 123

<binary payload>

Response format remains unchanged. Meaning the current logic of the response for JSON payloads stays in tact.

Binary payload
On the publishing path HTTP body consists of sequential envelopes (Envelop details https://github.com/zalando/nakadi/pull/1359). The binary representations of the envelopes are joined one after another as a single byte stream. The client serialises envelopes and combines them together into a single request body. Nakadi reads each envelope one after another, deserializing metadata of each envelope and enriching it if necessary.

The current change brings possibility to adopt current publishing logic in separate blocks implementing different steps. The PR is a technical change. The following PR will fill the sense into the empty blocks. At the moment Nakadi does 6 steps:
1. validate event ownership
2. validatation according to event schema
3. partitioning
4. setting event key (log compaction)
5. enrichment
6. submiting to Kafka

The number 1 is implemented partially to show case the usage.

binary publishing batch of events:
Screenshot 2022-04-14 at 17 01 21


core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiBatchMapper.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public List<NakadiRecord> map(final byte[] batch) throws IOException {
+        final List<NakadiRecord> records = new LinkedList<>();
+        final ByteBuffer tmp = ByteBuffer.wrap(batch);
+        while (tmp.hasRemaining()) {
+            final int recordStart = tmp.position();
+            final byte metadataVersion = tmp.get();
+            final int metadataLength = tmp.getInt();
+            final byte[] metadata = new byte[metadataLength];
+            tmp.get(metadata);
```
thomasabraham, line 38: Consolidate this logic.

core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecord.java
```diff
@@ -3,13 +3,50 @@
...
+    }
+
+    @Override
+    public NakadiRecord get() {
+        return this;
+    }
+
+    @Override
+    public Map<String, List<AuthorizationAttribute>> getAuthorization() {
+        return null;
```
thomasabraham, line 47: This probably should not return null.

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
```diff
@@ -38,30 +40,105 @@
...
+            throw new InternalNakadiException("failed to parse batch", e);
+        }
+    }
+
+    private ResponseEntity postBinaryEvents(final String eventTypeName,
+                                            final byte[] batch,
+                                            final HttpServletRequest request,
+                                            final Client client,
+                                            final boolean delete) throws IOException {
+        TracingService.setOperationName("publish_events")
```
thomasabraham, line 99: Maybe use `publish_binary_event` so that we can see how each one perform using the tracing data.

thomasabraham: But this might actually break the checks we have now.
So maybe lets keep it as `publish_events`

a1exsh: We can set "format" attribute on the span if we really want to tell apart binary from json.  Not sure we will ever be looking there, though.

core-common/src/main/java/org/zalando/nakadi/domain/NakadiMetadata.java
```diff
@@ -0,0 +1,11 @@
+package org.zalando.nakadi.domain;
+
+public interface NakadiMetadata {
+
+    String getEid();
+
+    String getPartition();
+
+    String getEventType();
```
thomasabraham, line 9: Adding `metadataVersion` as well can be useful. 

```suggestion
    String getEventType();
    
    byte getMetadataVersion();
```
This way a metadata can also tell which version it is.

core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecordMetadata.java
```diff
@@ -2,35 +2,36 @@
 
 public class NakadiRecordMetadata {
```
thomasabraham, line 3: The `NakadiRecordMetadata` name is very confusing. 
A better name should indicate that it is a response or status of trying to publish an event. 
Something like `EventPublishResult`.

core-services/src/main/java/org/zalando/nakadi/service/publishing/check/Check.java
```diff
@@ -0,0 +1,69 @@
...
+            final List<NakadiRecord> records);
+
+    public abstract Step getCurrentStep();
+
+    protected List<RecordResult> processError(
+            final List<NakadiRecord> records,
+            final NakadiRecord failedRecord,
+            final String error) {
+        return records.stream().map(nakadiRecord -> new RecordResult(
+                failedRecord == nakadiRecord ? Status.FAILED : Status.ABORTED,
```
adyach, line 22: add a comment describing behaviour: list of events, how the error processed

thomasabraham:
:+1:

thomasabraham:
Fix checkstyle checks.

adyach:
👍

thomasabraham:
:+1:

adyach:
👍

PR #1396, opened 2022-04-14
Improving getter inference logic in KPIEventMapper

thomasabraham:
# Improving getter inference logic in KPIEventMapper

> Zalando ticket : team-aruha/issues/984

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/service/KPIEventMapper.java
+++ b/core-common/src/main/java/org/zalando/nakadi/service/KPIEventMapper.java
@@ -1,6 +1,5 @@
 package org.zalando.nakadi.service;
 
-import com.google.common.annotations.VisibleForTesting;
 import org.apache.avro.AvroRuntimeException;
 import org.apache.avro.Schema;
 import org.apache.avro.generic.GenericRecord;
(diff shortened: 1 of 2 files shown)


thomasabraham:
:+1:

core-common/src/main/java/org/zalando/nakadi/service/KPIEventMapper.java
```diff
@@ -45,27 +48,46 @@ public KPIEventMapper(final Set<Class<? extends KPIEvent>> mappedClasses) {
...
+        final var kpiFieldAnnotation = field.getAnnotation(KPIEvent.KPIField.class);
+        final Method getter;
+        if (kpiFieldAnnotation.getter().isEmpty()) {
             final var fieldName = field.getName();
-            getterName = "get" + fieldName.substring(0, 1).toUpperCase() + fieldName.substring(1);
+            final var partialGetterName = fieldName.substring(0, 1).toUpperCase() + fieldName.substring(1);
+            final var getName = "get" + partialGetterName;
+            final var isName = "is" + partialGetterName;
+            if (getterCandidates.containsKey(getName)) {
+                getter = getterCandidates.get(getName);
```
tor-vs-floki, line 77: Does it make sense to check for the `getterCandidate` method return type that matches with the field type?

PR #1397, opened 2022-04-22
Assign partition number to avro metadata

tor-vs-floki:
Internal ticket ARUHA-1038


core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
```diff
@@ -70,61 +74,51 @@ public void publish(final Supplier<KPIEvent> kpiEventSupplier) {
...
-            final String eventType, final Schema schema, final String version, final String user) {
+            final String eventType, final int partition, final Schema schema, final String version, final String user) {
         final long now = System.currentTimeMillis();
         return new GenericRecordBuilder(schema)
                 .set("occurred_at", now)
                 .set("eid", uuidGenerator.randomUUID().toString())
                 .set("flow_id", FlowIdUtils.peek())
                 .set("event_type", eventType)
-                .set("partition", "0") // fixme avro
+                .set("partition", String.valueOf(partition)) // fixme avro
```
a1exsh, line 121:
```suggestion
                .set("partition", String.valueOf(partition))
```

core-services/src/test/java/org/zalando/nakadi/partitioning/MetadataRandomPartitionerTest.java
```diff
@@ -0,0 +1,40 @@
...
+        final var timelineMock = Mockito.mock(Timeline.class);
+        final var eventTypeName = "some-event-type";
+        final var random = Mockito.mock(Random.class);
+
+        when(eventTypeCache.getEventType(eventTypeName)).thenReturn(new EventType());
+        when(timelineService.getTopicRepository(ArgumentMatchers.any(EventType.class))).thenReturn(topicRepositoryMock);
+        when(timelineMock.getTopic()).thenReturn("some-topic-id");
+        when(timelineService.getActiveTimeline(ArgumentMatchers.any(EventType.class))).thenReturn(timelineMock);
+        when(topicRepositoryMock.listPartitionNames(ArgumentMatchers.anyString())).thenReturn(List.of("1", "2", "3"));
+        when(random.nextInt(2)).thenReturn(1);
```
a1exsh, line 34:
```suggestion
        when(random.nextInt(3)).thenReturn(1);
```

Should it be like that?

core-services/src/main/java/org/zalando/nakadi/partitioning/MetadataRandomPartitioner.java
```diff
@@ -0,0 +1,45 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+import org.zalando.nakadi.cache.EventTypeCache;
+import org.zalando.nakadi.service.timeline.TimelineService;
+
+import java.util.List;
+import java.util.Random;
+
+@Component
+public class MetadataRandomPartitioner {
```
adyach, line 12: there is the class https://github.com/zalando/nakadi/blob/master/core-services/src/main/java/org/zalando/nakadi/partitioning/RandomPartitionStrategy.java, can we reuse it ?

core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecord.java
```diff
@@ -82,7 +83,7 @@ public static NakadiRecord fromAvro(final String eventTypeName,
                 }));
         return new NakadiRecord(
                 eventTypeName,
-                null,
+                partition,
```
adyach, line 86: can take it from `metadata` then it won't be possible to mix up partition as a param and partition from metadata

core-services/src/main/java/org/zalando/nakadi/partitioning/MetadataRandomPartitioner.java
tor-vs-floki: I have postponed this decision until the Publishing part for Avro is in place. Suppose we want to reuse those Partition Strategy implementations. In that case, we need to access Event Metadata or Event payload. Having the publishing part in place would make the decision process more straightforward and safer so that we can use all the partition strategies in JSON, Avro, and Metadata event types.

core-services/src/test/java/org/zalando/nakadi/partitioning/MetadataRandomPartitionerTest.java
```diff
@@ -0,0 +1,40 @@
...
+        final var timelineMock = Mockito.mock(Timeline.class);
+        final var eventTypeName = "some-event-type";
+        final var random = Mockito.mock(Random.class);
+
+        when(eventTypeCache.getEventType(eventTypeName)).thenReturn(new EventType());
+        when(timelineService.getTopicRepository(ArgumentMatchers.any(EventType.class))).thenReturn(topicRepositoryMock);
+        when(timelineMock.getTopic()).thenReturn("some-topic-id");
+        when(timelineService.getActiveTimeline(ArgumentMatchers.any(EventType.class))).thenReturn(timelineMock);
+        when(topicRepositoryMock.listPartitionNames(ArgumentMatchers.anyString())).thenReturn(List.of("1", "2", "3"));
+        when(random.nextInt(3)).thenReturn(1);
```
a1exsh, line 34: Now the test should be failing, as it expects 1, but will get 2.

core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
```diff
@@ -70,61 +74,51 @@ public void publish(final Supplier<KPIEvent> kpiEventSupplier) {
                 return;
             }
             final var kpiEvent = kpiEventSupplier.get();
-            final var eventType = kpiEvent.getName();
+            final var eventTypeName = kpiEvent.getName();
 
             if (featureToggleService.isFeatureEnabled(Feature.AVRO_FOR_KPI_EVENTS)) {
+                final var partition = metadataRandomPartitioner.calculatePartition(eventTypeName);
```
a1exsh, line 80: I'd expect the partitioner for binary payloads to work on the same level as we already have for JSON.  That would allow us to reuse exiting code and keep the pieces of code related to partitioning closer together.

tor-vs-floki: I wanted to maintain a smaller scope of the PR and tackle this problem once the Publishing in Avro part is finished.

a1exsh: Hm, maybe we should wait for that change to be merged first and base off it?  Otherwise I fear we'll have to throw away most of the changes here later.

a1exsh:
I recall there were already functions to convert between Nakadi and Kafka partition representation (which are just String <> Integer anyway): https://github.com/zalando/nakadi/blob/2cfbd797de7b82bd23a474a1f32d4510a2ad7bc4/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaCursor.java#L49-L55

I'm not sure if we want to use them here or not. :)

I actually think it's about time we abandon the string notion of "partition names" in Nakadi and just use them as integers.  There are no "partition names" in Kafka.  The concept doesn't really make sense, if you think about it.

core-common/src/main/java/org/zalando/nakadi/partitioning/PartitionStrategy.java
```diff
@@ -13,6 +22,53 @@ public interface PartitionStrategy {
     String USER_DEFINED_STRATEGY = "user_defined";
     String RANDOM_STRATEGY = "random";
 
-    String calculatePartition(EventType eventType, JSONObject event, List<String> partitions)
+    String calculatePartition(EventType eventType, PartitionData event, List<String> partitions)
             throws PartitioningException;
 }
+
+class PartitionData {
```
tor-vs-floki, line 29: I have implemented something roughly and more on a conceptual level. If you see build errors, please ignore them for now.

core-services/src/main/java/org/zalando/nakadi/partitioning/MetadataRandomPartitioner.java
adyach: 👍 then we will have to remove it once publisher is ready to work with partitioning

core-services/src/main/java/org/zalando/nakadi/partitioning/PartitionResolver.java
```diff
@@ -68,7 +69,21 @@ public String resolvePartition(final EventType eventType, final JSONObject event
 
         final List<String> partitions = timelineService.getTopicRepository(eventType)
                 .listPartitionNames(timelineService.getActiveTimeline(eventType).getTopic());
-        return partitionStrategy.calculatePartition(eventType, eventAsJson, partitions);
+        return partitionStrategy.calculatePartition(eventType, PartitionData.fromJson(eventType, eventAsJson), partitions);
     }
 
+    public String resolvePartition(final EventType eventType, final GenericRecord eventAsBinary) {
```
adyach, line 75: I would rather pass as an argument PartitionData to avoid code duplication, and from the calling method create PartitionData

core-services/src/main/java/org/zalando/nakadi/partitioning/RandomPartitionStrategy.java
```diff
@@ -15,7 +15,7 @@ public RandomPartitionStrategy(final Random random) {
     }
 
     @Override
-    public String calculatePartition(final EventType eventType, final JSONObject event, final List<String> partitions) {
+    public String calculatePartition(final EventType eventType, final PartitionData event, final List<String> partitions) {
```
adyach, line 18:
```suggestion
    public String calculatePartition(final EventType eventType, final PartitionData partitionData, final List<String> partitions) {
```

core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
adyach: that code will be moved into publisher, once it is ready

core-services/src/main/java/org/zalando/nakadi/partitioning/UserDefinedPartitionStrategy.java
```diff
@@ -10,10 +10,11 @@
 public class UserDefinedPartitionStrategy implements PartitionStrategy {
 
     @Override
-    public String calculatePartition(final EventType eventType, final JSONObject event, final List<String> partitions)
+    public String calculatePartition(final EventType eventType, final PartitionData event, final List<String> partitions)
             throws PartitioningException {
         try {
-            final String partition = event.getJSONObject("metadata").getString("partition");
+            //final String partition = event.getJSONObject("metadata").getString("partition");
+            final String partition = String.valueOf(event.getPartition());
```
adyach, line 17: `event.getPartition()` returns String

core-services/src/main/java/org/zalando/nakadi/partitioning/MetadataRandomPartitioner.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public MetadataRandomPartitioner(final EventTypeCache eventTypeCache,
+                                     final TimelineService timelineService,
+                                     final Random random) {
+        this.random = random;
+        this.eventTypeCache = eventTypeCache;
+        this.timelineService = timelineService;
+    }
+
+    public int calculatePartition(final String eventTypeName) {
```
adyach, line 33: returning String here will avoid several conversion to String

core-common/src/main/java/org/zalando/nakadi/partitioning/PartitionStrategy.java
```diff
@@ -13,6 +22,53 @@ public interface PartitionStrategy {
...
+                            throw new InvalidPartitionKeyFieldsException(e.getMessage());
+                        }
+                    }))
+                    .map(Try::getOrThrow)
+                    .collect(Collectors.toList());
+        }
+        return new PartitionData(partition, partitionKeys);
+    }
+
+    public static PartitionData fromAvro(final GenericRecord metadata) {
```
adyach, line 68: I would love to see some proxy object which holds necessary metadata fields, that could be used later as well. something in lines with
```
class NakadiMetadata {
private final String partition;
private final List<String> partitionKeys
}
```

adyach: that helps to avoid calling `metadata.get("partition").toString();` etc. every time I need to have partition and also provides a way to replace it in the future with the other type like Proto Buff  etc..

adyach: it can even replace PartitionData, and the functions will work purely on the provided object

core-common/src/main/java/org/zalando/nakadi/partitioning/PartitionStrategy.java
```diff
@@ -13,6 +22,53 @@ public interface PartitionStrategy {
...
+        final String partition = metadata.get("partition").toString();
```
adyach, line 69: can `metadata.get("partition")` return null ?

core-common/src/main/java/org/zalando/nakadi/partitioning/PartitionStrategy.java
```diff
@@ -13,6 +22,53 @@ public interface PartitionStrategy {
...
+        final List<String> partition_keys = (List<String>) Collections.singletonList(metadata.get("partition_keys"));
```
adyach, line 70: also not sure if `Collections.singletonList(metadata.get("partition_keys"))` works as expected, because singeltonlist is a list of a single object, but `metadata.get("partition_keys"))` is an array I suppose

a1exsh:
> I actually think it's about time we abandon the string notion of "partition names" in Nakadi and just use them as integers. There are no "partition names" in Kafka. The concept doesn't really make sense, if you think about it.

See: https://github.com/zalando/nakadi/pull/1398

core-services/src/main/java/org/zalando/nakadi/partitioning/UserDefinedPartitionStrategy.java
```diff
@@ -10,20 +11,19 @@
...
-            final String partition = event.getJSONObject("metadata").getString("partition");
-            if (partitions.contains(partition)) {
-                return partition;
-            } else {
-                throw new PartitioningException("Failed to resolve partition. " +
-                        "Invalid partition specified when publishing event.");
-            }
+        if (Strings.isNullOrEmpty(partitionData.getPartition())) {
+            throw new PartitioningException("Failed to resolve partition. " +
+                    "Failed to get partition from event metadata");
```
adyach, line 18: isn't it to late here to say about metadata? the method works with PartitionData.

core-services/src/main/java/org/zalando/nakadi/partitioning/MetadataRandomPartitioner.java
adyach: can we still for now return String to avoid conversion?

tor-vs-floki: I removed this class completely, because now we can use PartitionResolver for Avro as well

core-services/src/main/java/org/zalando/nakadi/partitioning/UserDefinedPartitionStrategy.java
tor-vs-floki: PartitionData can be missing the partition number since it is only a representation of partition data that comes with the request. 

For instance for publishing requests of event types with a random partition strategy, we don't require the partition numbers to be on the request metadata, and the partition number in metadata is only requested for user-defined partition strategy.

tor-vs-floki: After all, I implemented the logic of extracting the partitioning data into all partition strategies https://github.com/zalando/nakadi/pull/1397/files#diff-e2daf1f8029ae29b6cff742e0ac068578328bef4bebc7e6e28a715a1631d15adR31

core-services/src/main/java/org/zalando/nakadi/partitioning/UserDefinedPartitionStrategy.java
tor-vs-floki: That should be fine because we also validate whether the partition actually exists and the Topic Repository sends back the list of partitions in String[] either way.

core-common/src/main/java/org/zalando/nakadi/partitioning/PartitionStrategy.java
tor-vs-floki: This is already implemented in another PR called NakadiMetadata

core-common/src/main/java/org/zalando/nakadi/partitioning/PartitionStrategy.java
tor-vs-floki: This is already addressed differently

core-common/src/main/java/org/zalando/nakadi/partitioning/PartitionStrategy.java
```diff
@@ -3,16 +3,17 @@
...
     String HASH_STRATEGY = "hash";
     String USER_DEFINED_STRATEGY = "user_defined";
     String RANDOM_STRATEGY = "random";
 
-    String calculatePartition(EventType eventType, JSONObject event, List<String> partitions)
+    String calculatePartition(PartitioningData partitioningData, List<String> partitions)
             throws PartitioningException;
+
+    PartitioningData getDataFromJson(EventType eventType, JSONObject jsonEvent)
+            throws PartitioningException;;
```
thomasabraham, line 18:
```suggestion
            throws PartitioningException;
```

core-common/src/main/java/org/zalando/nakadi/partitioning/PartitionStrategy.java
```diff
@@ -3,16 +3,17 @@
...
 
     String HASH_STRATEGY = "hash";
     String USER_DEFINED_STRATEGY = "user_defined";
     String RANDOM_STRATEGY = "random";
 
-    String calculatePartition(EventType eventType, JSONObject event, List<String> partitions)
+    String calculatePartition(PartitioningData partitioningData, List<String> partitions)
             throws PartitioningException;
+
+    PartitioningData getDataFromJson(EventType eventType, JSONObject jsonEvent)
```
thomasabraham, line 17: Why should `getDataFromJson` be part of `PartitionStrategy` interface? 
If it has to be, then better name it `getPartitionDataFromJson`. 
But I would rather have this as a `fromJson` method in the `PartitionData` class.

core-common/src/main/java/org/zalando/nakadi/partitioning/PartitioningData.java
```diff
@@ -0,0 +1,29 @@
+package org.zalando.nakadi.partitioning;
+
+import java.util.List;
+
+public class PartitioningData {
```
thomasabraham, line 5: Better to have getter and setter of a field together.

core-common/src/main/java/org/zalando/nakadi/partitioning/PartitionStrategy.java
thomasabraham: Ok, I understand that this method is actually calculating the partition when I reached the implementation. 
I think we have to discuss this.

core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
```diff
@@ -103,23 +113,31 @@ public void publish(final Supplier<KPIEvent> kpiEventSupplier) {
...
-                .set("event_type", eventType)
-                .set("partition", "0") // fixme avro
+                .set("event_type", eventTypeName)
                 .set("received_at", now)
-                .set("schema_version", version)
+                .set("schema_version", etSchemaVersion)
                 .set("published_by", user)
                 .build();
+
+        final var genericRecord = new GenericRecordMetadata(metadata, metadataSchemaVersion);
```
thomasabraham, line 135: We have to improve this.

core-services/src/main/java/org/zalando/nakadi/partitioning/MetadataRandomPartitioner.java
tor-vs-floki: Since we extended the scope, I think the comment is resolved with the current implementation

core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
```diff
@@ -103,23 +116,30 @@ public void publish(final Supplier<KPIEvent> kpiEventSupplier) {
...
                 .set("eid", uuidGenerator.randomUUID().toString())
                 .set("flow_id", FlowIdUtils.peek())
-                .set("event_type", eventType)
-                .set("partition", "0") // fixme avro
+                .set("event_type", eventTypeName)
                 .set("received_at", now)
-                .set("schema_version", version)
+                .set("schema_version", etSchemaVersion)
                 .set("published_by", user)
+                .set("partition", randomPartitionStrategy.calculatePartition(partitions))
```
a1exsh, line 139: So we actually bypass the Avro question in the end?

core-services/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -31,32 +31,29 @@ public HashPartitionStrategy(final HashPartitionStrategyCrutch hashPartitioningC
...
+        final var partitioningData = getDataByPartitionKeyFields(eventType, jsonEvent);
+
+        return calculatePartition(partitioningData, partitions);
+    }
+
+    private String calculatePartition(final PartitioningData partitioningData, final List<String> partitions)
             throws InvalidPartitionKeyFieldsException {
-        final List<String> partitionKeyFields = eventType.getPartitionKeyFields();
-        if (partitionKeyFields.isEmpty()) {
+        if (partitioningData.getPartitionKeys().isEmpty()) {
```
a1exsh, line 45: Maybe check for `null` here, even if it's a private method.

tor-vs-floki: We resolved this conversation offline

core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
tor-vs-floki: We resolved this conversation offline

thomasabraham:
:+1:

a1exsh:
:+1:

core-services/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -31,32 +31,28 @@ public HashPartitionStrategy(final HashPartitionStrategyCrutch hashPartitioningC
...
+                                     final List<String> partitions)
+            throws PartitioningException {
+        final var partitioningData = getPartitionKeys(eventType, jsonEvent);
+
+        return calculatePartition(partitioningData, partitions);
+    }
+
+    private String calculatePartition(final PartitioningData partitioningData, final List<String> partitions) {
+        if (partitioningData.getPartitionKeys() == null || partitioningData.getPartitionKeys().isEmpty()) {
+            throw new PartitioningException("Applying " + this.getClass().getSimpleName() + " although event type " +
```
adyach, line 45: what about the message ? it says about event type, but here you do not check for the event type partition keys

core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
```diff
@@ -50,21 +57,26 @@ protected NakadiKpiPublisher(
...
         this.featureToggleService = featureToggleService;
         this.jsonEventsProcessor = jsonEventsProcessor;
         this.binaryEventsProcessor = binaryEventsProcessor;
         this.usernameHasher = usernameHasher;
         this.eventMetadata = eventMetadata;
         this.uuidGenerator = uuidGenerator;
         this.avroSchema = avroSchema;
+        this.nakadiRecordMapper = nakadiRecordMapper;
+        this.eventTypeCache = eventTypeCache;
+        this.randomPartitionStrategy = new RandomPartitionStrategy(new Random());
```
adyach, line 72: 😶‍🌫️

adyach: usually, the approach with the spring to use dependency injection

adyach:
to be honest, I do not see how partitioning is extend to support Avro. the actual change is to provide partition number for internal event types publishing, which is great. however, later on it should be changed to common approach for publishing, because not every internal event type could have random partition strategy. please, follow up.

PR #1398, opened 2022-04-26
[WiP] Optimize partitioning to work with numbers, not names

a1exsh:
The notion of partition "names" was never used beyond string representation of
partition numbers, but lead to inoptimal code in a critical path (sorting the
list of partition "names" on _each_ published event).


core-services/src/main/java/org/zalando/nakadi/partitioning/HashPartitionStrategy.java
```diff
@@ -56,12 +56,8 @@ public String calculatePartition(final EventType eventType, final JSONObject eve
                     .mapToInt(hc -> hc)
                     .sum();
 
-
-            int partitionIndex = abs(hashValue % partitions.size());
-            partitionIndex = hashPartitioningCrutch.adjustPartitionIndex(partitionIndex, partitions.size());
-
-            final List<String> sortedPartitions = partitions.stream().sorted().collect(Collectors.toList());
```
a1exsh, line 63: This was re-calculated for every event. :exploding_head:

core-services/src/main/java/org/zalando/nakadi/partitioning/PartitionResolver.java
```diff
@@ -66,9 +66,10 @@ public String resolvePartition(final EventType eventType, final JSONObject event
                     eventTypeStrategy);
         }
 
-        final List<String> partitions = timelineService.getTopicRepository(eventType)
-                .listPartitionNames(timelineService.getActiveTimeline(eventType).getTopic());
-        return partitionStrategy.calculatePartition(eventType, eventAsJson, partitions);
+        final int partitionsNumber = timelineService.getTopicRepository(eventType)
+                .getPartitionsNumber(timelineService.getActiveTimeline(eventType).getTopic());
```
a1exsh, line 70: Here we are calling `getActiveTimeline()` two times (first inside of `getTopicRepository()`), and that one incurs taking a global lock.  For every published event.

a1exsh:
Closing due to too many conflicts.

PR #1401, opened 2022-04-28
Implement EnrichmentCheck

thomasabraham:
# Implement EnrichmentCheck

> Zalando ticket : team-aruha/issues/1100

## Description
Implement `EnrichmentCheck` to actually enrich the metadata.

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes


core-common/src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
```diff
@@ -39,12 +40,30 @@ public void enrich(final BatchItem batchItem, final EventType eventType) throws
...
+        final var metadata = nakadiRecord.getMetadata();
+        metadata.setPublishedBy(getPublisher());
+        final DateTime dateTime = new DateTime(DateTimeZone.UTC);
+        metadata.setReceivedAt(dateTime.toString());
+        metadata.setEventType(eventType.getName());
+        if ("".equals(metadata.getFlowId())) {
+            metadata.setFlowId(FlowIdUtils.peek());
+        }
+        metadata.setPartition(nakadiRecord.getPartition());
+        metadata.setSchemaVersion(eventType.getSchema().getVersion().toString());
```
a1exsh, line 54: I'd expect the event type schema version to be provided by the client, and not enriched in case of binary record.

core-common/src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
```diff
@@ -39,12 +40,30 @@ public void enrich(final BatchItem batchItem, final EventType eventType) throws
         }
     }
 
+    @Override
+    public void enrich(final NakadiRecord nakadiRecord, final EventType eventType) throws EnrichmentException {
+        final var metadata = nakadiRecord.getMetadata();
+        metadata.setPublishedBy(getPublisher());
+        final DateTime dateTime = new DateTime(DateTimeZone.UTC);
+        metadata.setReceivedAt(dateTime.toString());
```
a1exsh, line 48: I thought `received_at` is not a string value, but `long`?

core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecord.java
```diff
@@ -108,4 +109,12 @@ public NakadiRecord setMetadata(final NakadiMetadata metadata) {
         return this;
     }
 
+    public String getPartition() {
+        return partition;
+    }
+
+    public NakadiRecord setPartition(final String partition) {
+        this.partition = partition;
```
a1exsh, line 117: Can we set partition field of the `metadata` object directly instead?

core-common/src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
thomasabraham: Good point, I'll fix this.

core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecord.java
thomasabraham: The events of event type with category `undefined` don't have a `metadata` AFAIK.
This means that to set partition field of `metadata` object directly, the class responsible for partitioning has to be aware of the category of the event-type.

a1exsh: But it doesn't seem like this method is actually ever called?  Or am I looking in the wrong place?

thomasabraham: No, this is not yet called. 
This part will follow with its own ticket. 
Basically there is a set of `Check` implementations that are applied to the `NakadiRecorrd`. 
Enriching metadata is the one handled in this ticket, assigning correct partition is a different one.

core-services/src/test/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategyTest.java
```diff
@@ -166,4 +183,47 @@ public void setPublisher() throws Exception {
         assertEquals("test-user-123", batch.getEvent().getJSONObject("metadata").getString("published_by"));
     }
 
+    @Test
+    public void testNakadiRecordEnrichment() throws IOException {
+        final var eventType = Mockito.mock(EventType.class);
+        final var nakadiRecord = getTestNakadiRecord();
+
+        strategy.enrich(nakadiRecord, eventType);
+
```
a1exsh, line 192: Assertion missing?

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/BinaryEndToEndAT.java
```diff
@@ -0,0 +1,42 @@
...
+    @Ignore
+    public void testAvroPublishing() throws JsonProcessingException {
+        final var et = EventTypeTestBuilder.builder()
+                .name(TEST_ET_NAME)
+                .category(EventCategory.BUSINESS)
+                .enrichmentStrategies(List.of(EnrichmentStrategyDescriptor.METADATA_ENRICHMENT))
+                .build();
+        NakadiTestUtils.createEventTypeInNakadi(et);
+
+        final byte[] body = Base64.getDecoder().decode(TEST_DATA);
```
adyach, line 32: the endpoint is not suppose to accept base64, only binary payload

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
```diff
@@ -93,12 +97,15 @@ public ResponseEntity postBinaryEvents(@PathVariable final String eventTypeName,
...
-//        }
+        if ("acceptanceTest".equals(profile)) {
+            try {
+                return postBinaryEvents(eventTypeName, batch, request, client, false);
+            } catch (IOException e) {
+                throw new InternalNakadiException("failed to parse batch", e);
+            }
+        } else {
+            return status(HttpStatus.NOT_IMPLEMENTED).body("the method is under development profile = " + profile);
+        }
```
a1exsh, line 108: Maybe let's wait for implementation to be enabled instead of relying on the test profile?

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/BinaryEndToEndAT.java
thomasabraham: Yes, only binary data is send to the endpoint. 
Instead of creating the binary data in the test (using schema) for every test execution, I created it once locally. The binary data that I created locally is encoded in base64 so that it can be included as a string in the test. And this is decoded in the above line so that we send binary data to the endpoint.

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
thomasabraham: I'll disable it for now.

core-common/src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
thomasabraham: changed

core-services/src/test/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategyTest.java
thomasabraham: Added

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
thomasabraham: Disabled.

api-publishing/src/main/java/org/zalando/nakadi/PublishingResultConverter.java
```diff
@@ -60,17 +60,22 @@ private EventPublishingStatus mapPublishingStatus(final Check.Status status) {
...
-                    .setDetail(recordMetadata.getException().getMessage()));
+                    .setDetail((recordMetadata.getException() != null) ?
+                            recordMetadata.getException().getMessage() : ""));
         }
 
-        return new EventPublishResult(status, EventPublishingStep.PUBLISHING, batchItemResponses);
+        final var overallStatus = batchItemResponses.stream()
+                .map(BatchItemResponse::getPublishingStatus)
+                .filter(status -> !EventPublishingStatus.SUBMITTED.equals(status))
+                .findAny().map(status -> EventPublishingStatus.FAILED).orElse(EventPublishingStatus.SUBMITTED);
```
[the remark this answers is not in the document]
adyach: the publishing status is also used in the controller to prepare usage statistic.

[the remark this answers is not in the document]
adyach: so to conclude:
- if there is a failed status in the batch, the publishing result status should be failed
- if there is no failed status, but only aborted the status should be aborted
- otherwise succeeded 

does it make sense ?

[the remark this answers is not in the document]
thomasabraham: Makes sense, I'll update it.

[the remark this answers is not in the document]
adyach: just thought about it, since we separated validation (checks) and publishing, probably your code make sense and works as expected, but I would just fix it for common understanding and logic.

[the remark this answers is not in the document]
thomasabraham: Updated

thomasabraham:
:+1:

PR #1405, opened 2022-05-03
check for the status of the published events and log warning if error

adyach:
# One-line summary

check for the status of the published events and log warning if error


adyach:
👍

tor-vs-floki:
👍

PR #1409, opened 2022-05-09
Extract event owner from avro metadata

a1exsh:
Internal ticket: 1039


core-common/src/main/java/org/zalando/nakadi/domain/EventOwnerHeader.java
```diff
@@ -70,4 +70,9 @@ public int hashCode() {
         result = 31 * result + value.hashCode();
         return result;
     }
+
+    @Override
+    public String toString() {
+        return "{ name: " + this.name + ", value: " + this.value + " }";
```
adyach, line 76: could log sensitive information

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventOwnerExtractor.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.json.JSONObject;
+import org.zalando.nakadi.domain.EventOwnerHeader;
+import org.zalando.nakadi.domain.NakadiMetadata;
+
+public interface EventOwnerExtractor {
+
+    EventOwnerHeader extractEventOwner(JSONObject event);
+
+    EventOwnerHeader extractEventOwner(NakadiMetadata metadata);
```
adyach, line 11: it would be nice to have a single method, but at the same time I realise that `EventOwnerHeader extractEventOwner(NakadiMetadata metadata);` can be later reused to work with other types of the binary.

core-common/src/main/java/org/zalando/nakadi/domain/EventOwnerHeader.java
a1exsh: Without: `java.lang.AssertionError: expected:<org.zalando.nakadi.domain.EventOwnerHeader@aacf320e> but was:<org.zalando.nakadi.domain.EventOwnerHeader@aacf2a8e>` ;)

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventOwnerExtractor.java
a1exsh: Potentially we can expose `metadata` part of a `JSONObject` via some common interface...  Then a single method will be possible.

core-common/src/main/java/org/zalando/nakadi/domain/EventOwnerHeader.java
adyach: could you use equals method in the assertions? i would avoid logging that as much as possible

a1exsh: OK, let's remove it just in case.

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventOwnerExtractorFactory.java
```diff
@@ -24,37 +24,78 @@ public EventOwnerExtractorFactory(final FeatureToggleService featureToggleServic
...
         }
         switch (selector.getType()) {
             case PATH:
-                return createPathExtractor(selector);
+                return createPathExtractor(selector.getName(), selector.getValue());
             case STATIC:
-                return createStaticExtractor(selector);
+                return createStaticExtractor(selector.getName(), selector.getValue());
+            case METADATA:
+                if (!"event_owner".equals(selector.getValue())) {
```
adyach, line 38: I think that check should be earlier, when the event type is created. Otherwise it creates the situation when the user create wrong event owner selector for the event type and can not change it. I suggest to disallow providing value for the metadata selector type.

a1exsh: See: https://github.com/zalando/nakadi/pull/1409/commits/2236d1a99c4ebf8fc33723fb24d9283f9e355616

adyach:
lgtm, tests are failing though. Nakadi API definition can be extend later once publishing works.

a1exsh:
:+1:

adyach:
👍

PR #1410, opened 2022-05-10
combining avro checks and publishing statuses

adyach:
# One-line summary
combining avro checks and publishing statuses


adyach:
👍

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
```diff
@@ -115,6 +128,11 @@ private ResponseEntity postBinaryEvents(final String eventTypeName,
             throw new BlockedException("Application or event type is blocked");
         }
 
+        final EventType eventType = eventTypeCache.getEventType(eventTypeName);
+
+        // throws exception, maybe move it somewhere else
+        authValidator.authorizeEventTypeWrite(eventType);
```
a1exsh, line 134: What happens if it really throws an exception?  Is it handled properly?

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
```diff
@@ -129,15 +147,12 @@ private ResponseEntity postBinaryEvents(final String eventTypeName,
...
-                if (!recordResults.isEmpty()) {
-                    result = publishingResultConverter.mapCheckResultToView(recordResults);
-                } else {
-                    final List<NakadiRecordResult> nakadiRecordsMetadata =
-                            binaryPublisher.publish(eventTypeName, nakadiRecords);
-                    result = publishingResultConverter.mapPublishingResultToView(nakadiRecordsMetadata);
+                final List<NakadiRecordResult> recordResults = binaryPublisher
+                        .publishWithChecks(eventType, nakadiRecords, prePublishingChecks);
+                if (recordResults.isEmpty()) {
+                    throw new RuntimeException("");
```
a1exsh, line 153:
```suggestion
                    throw new InternalNakadiException("unexpected empty record result list");
```

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
adyach: it is handled by handler https://github.com/zalando/nakadi/blob/master/core-services/src/main/java/org/zalando/nakadi/controller/advice/NakadiProblemExceptionHandler.java#L94

app/src/main/java/org/zalando/nakadi/config/NakadiConfig.java
```diff
@@ -47,14 +48,26 @@ public void run() {
         return taskExecutor;
     }
 
+    @Bean
+    @Qualifier("pre-internal-publishing-checks")
```
a1exsh, line 52:
```suggestion
    @Qualifier("internal-publishing-checks")
```

Reads less awkward this way, IMO.  Maybe change the other to just "publishing-checks" as well, there are only checks prior to publishing anyway.

app/src/main/java/org/zalando/nakadi/config/NakadiConfig.java
```diff
@@ -47,14 +48,26 @@ public void run() {
...
+    public List<Check> prePublishingChecks(final EnrichmentCheck enrichmentCheck,
+                                           final PartitionResolver partitionResolver) {
+        return Lists.newArrayList(
+                new PartitioningCheck(partitionResolver),
```
a1exsh, line 56: Can/should we use a shared instance here like we did for the `enrichmentCheck`?

core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecordResult.java
```diff
@@ -3,26 +3,38 @@
...
     public enum Status {
-        SUCCEEDED, FAILED, NOT_ATTEMPTED
+        SUCCEEDED, FAILED, ABORTED
+    }
+
+    public enum Step {
+        NONE,
+        VALIDATION,
+        ENRICHMENT,
+        PARTITIONING,
```
a1exsh, line 13: The actual order of steps is the opposite.  Let's change it to avoid confusion:
```suggestion
        PARTITIONING,
        ENRICHMENT,
```

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -394,8 +394,8 @@ private void logFailedEvents(final String topicId, final String eventType, final
             sb.append(entry.getKey())
                     .append(":[")
                     .append(entry.getValue().stream()
-                     .map(i -> i.getResponse().getPublishingStatus() == EventPublishingStatus.SUBMITTED ? "1" : "0")
-                     .collect(Collectors.joining(", ")))
+                    .map(i -> i.getResponse().getPublishingStatus() == EventPublishingStatus.SUBMITTED ? "1" : "0")
+                    .collect(Collectors.joining(", ")))
```
a1exsh, line 398:
```suggestion
                        .map(i -> i.getResponse().getPublishingStatus() == EventPublishingStatus.SUBMITTED ? "1" : "0")
                        .collect(Collectors.joining(", ")))
```
Here the `.map()` is chained to `.stream()` from the line above, not to `.append()`.

I wish we had some format enforcing :(

adyach: there is format enforcing, but then it goes beyond 120 chars

adyach: and I have a fix for that )

core-services/src/test/java/org/zalando/nakadi/service/publishing/check/EnrichmentCheckTest.java
```diff
@@ -44,7 +44,7 @@ public void testExecuteSuccess() {
         final NakadiRecord nakadiRecord = Mockito.mock(NakadiRecord.class);
         final List<NakadiRecord> nakadiRecords = List.of(nakadiRecord);
 
-        assertNull(enrichmentCheck.execute(eventType, nakadiRecords));
+        assertNotNull(enrichmentCheck.execute(eventType, nakadiRecords));
```
a1exsh, line 47: Should we have a more elaborate assertion here?  Not `null` can still communicate both success (empty list) and failure (non-empty list).

api-publishing/src/main/java/org/zalando/nakadi/EventPublishingController.java
a1exsh: So let's remove the comment if we don't need to move it?

a1exsh:
:+1:

adyach:
👍

PR #1411, opened 2022-05-10
Merge SubscriptionTokenLister with SubscriptionDbRepository

a1exsh:
Fix the problem with using listing token and restricting to readers at the
same time.

> Zalando ticket : 1107


api-metastore/src/test/java/org/zalando/nakadi/controller/EventTypeControllerTest.java
```diff
@@ -566,32 +566,6 @@ public void whenDeleteNoneExistingEventTypeThen404() throws Exception {
                 .andExpect(content().contentType("application/problem+json"));
     }
 
-    @Test
-    public void whenDeleteEventTypeThatHasSubscriptionsThenConflict() throws Exception {
```
a1exsh, line 570: See this test instead: https://github.com/zalando/nakadi/blob/master/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/SubscriptionAT.java#L419

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -152,24 +157,59 @@ public List<Subscription> listSubscriptions(final Set<String> eventTypes, final
...
-    static void applyFilter(
+    private static void applyFilter(
             final Set<String> eventTypes,
             final Optional<String> owningApplication,
             final Optional<AuthorizationAttribute> reader,
+            final StringBuilder queryBuilder,
             final List<String> clauses,
             final List<Object> params) {
+
+        queryBuilder.append("SELECT s_subscription_object FROM zn_data.subscription");
```
antban, line 206: With this approach isn't QueryBuilder a return result of the function?

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -241,4 +292,235 @@ public String toString() {
...
+        final List<Subscription> subscriptionsPlusOne = executeListSubscriptions(
+                queryBuilder, clauses, params, limit + 1, false);
+        // subscriptions are listed in a backwards direction, therefore they should be reversed afterwards
+        Collections.reverse(subscriptionsPlusOne);
+        if (subscriptionsPlusOne.size() > limit) {
+            // we need to remove the first one, as it shows that there is something before.
+            return new ListResult(
+                    subscriptionsPlusOne.subList(1, subscriptionsPlusOne.size()),
+                    new Token(subscriptionsPlusOne.get(limit).getId(), TokenType.FORWARD),
+                    new Token(subscriptionsPlusOne.get(0).getId(), TokenType.BACKWARD)
```
antban, line 312: I think here is a bug. It should be either `BACKWARD_INCL` or `subscriptionsPlusOne.get(1)`

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -241,4 +292,235 @@ public String toString() {
                     '}';
         }
     }
+
+    private ListResult listBackwardToken(
+            final StringBuilder queryBuilder, final List<String> clauses, final List<Object> params,
+            final Token token, final int limit) {
+
+        clauses.add(" s_id " + token.getTokenType().dbClause + " ? ");
+        params.add(token.getSubscriptionId());
```
antban, line 301: maybe these 2 lines should be moved into the calling method (not to repeat yourself)

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
```diff
@@ -121,28 +124,30 @@ public void deleteSubscription(final String id)
...
+            throws ServiceTemporarilyUnavailableException {
+
+        return listSubscriptions(eventTypes, Optional.empty(), Optional.empty(), Optional.empty());
+    }
+
+    /**
+     * Use {@code listSubscriptions} with token instead.
      */
     @Deprecated
     public List<Subscription> listSubscriptions(final Set<String> eventTypes, final Optional<String> owningApplication,
```
antban, line 139: Do I get it correct: this method is `public` + `deprecated` + `listAllSubscriptionsFor` created ? 
Maybe it should not be public or `listSubscriptionsFor` should iterate through token and this method expected to be removed?

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/db/SubscriptionDbRepositoryTest.java
```diff
@@ -219,4 +223,154 @@ private void checkSubscriptionCreatedFromSubscriptionBase(final Subscription sub
...
+
+        final List<Subscription> expectedSubscriptions = testSubscriptions.stream()
+                .filter(sub -> owningApp.equals(sub.getOwningApplication()) && sub.getEventTypes().contains("et1"))
+                .sorted(Comparator.comparing(Subscription::getId))
+                .collect(toList());
+
+        final List<Subscription> subscriptions = repository.listSubscriptions(
+                ImmutableSet.of("et1"), Optional.of(owningApp), Optional.empty(), null, 10).getItems();
+        assertThat(subscriptions, equalTo(expectedSubscriptions));
+    }
```
antban, line 347: What do you think, does it make sense to use table-driven testing for tests of this class?

a1exsh: I don't want to change anything in these tests as long as they work.  I just moved them here from the token lister class, unchanged.

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
a1exsh: We cannot remove it until we removed the `offset`-based pagination :(

The new method is only used internally (never via the API calls) and we will keep that one.

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
a1exsh: Yes, we can rename it to `initQuery` or something and return `StringBuilder` from it.  The whole business with buliding the query from blocks here is quite ugly anyway.

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SubscriptionDbRepository.java
a1exsh: I need to double-check, but I didn't change this line.

a1exsh: I find the logic here to be over-complicated.  Is anyone using the backward listing at all?  We could greatly simplify it if we removed this possibility.,

a1exsh:
:+1:

PR #1414, opened 2022-05-11
Serialize enriched metadata - Part 2

thomasabraham:
# Introduced dedicated `NakadiRecordResource` class

> Zalando ticket : team-aruha/issues/1106

## Review
- [ ] Tests
- [ ] Documentation

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecordResource.java
+++ b/core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecordResource.java
@@ -1,0 +1,49 @@
+package org.zalando.nakadi.domain;
+
+import org.zalando.nakadi.plugin.api.authz.AuthorizationAttribute;
+import org.zalando.nakadi.plugin.api.authz.AuthorizationService;
+import org.zalando.nakadi.plugin.api.authz.Resource;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public class NakadiRecordResource implements Resource<NakadiRecord> {
+    private final NakadiRecord nakadiRecord;
+
+    public NakadiRecordResource(final NakadiRecord nakadiRecord) {
+        this.nakadiRecord = nakadiRecord;
+    }
+
+    @Override
+    public String getName() {
+        return nakadiRecord.getMetadata().getEid();
+    }
+
+    @Override
+    public String getType() {
+        return ResourceImpl.EVENT_RESOURCE;
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecordResource.java
```diff
@@ -0,0 +1,49 @@
...
+public class NakadiRecordResource implements Resource<NakadiRecord> {
+    private final NakadiRecord nakadiRecord;
+
+    public NakadiRecordResource(final NakadiRecord nakadiRecord) {
+        this.nakadiRecord = nakadiRecord;
+    }
+
+    @Override
+    public String getName() {
+        return nakadiRecord.getMetadata().getEid();
```
a1exsh, line 21: Now I recall I wanted to ask about this line :p

adyach: that's copy from [batch item](https://github.com/zalando/nakadi/blob/9f52df9c63136af9255f18e7725c553c4ebf40a5/core-common/src/main/java/org/zalando/nakadi/domain/BatchItem.java#L156).

a1exsh: Can I unsee that, please?.. :see_no_evil: 

That still doesn't answer what's going on there.  Why `eid` makes sense as a "name" of resource in context of Event-based auth?

a1exsh:
:+1:

core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecordResource.java
thomasabraham: AFAIK, we are (mis)using the Authorization concepts of spring security, which works fine for event-types when we originally started using it. An event-type is identified by its name, and the `getName` function in the interface is returns the event-type's name in Nakadi's `ResourceImpl` when used for an event-type; and subscription-id in the case of subscriptions (which are introduced to Nakadi later). See `asResource()` method is `EventType` and `Subscription`. 

Then much later we implemented per event authorization. In order to re-use the existing concepts, plugins and libraries, now we have to represent an event as a `Resource`, just like how we do with `EventType` and `Subscription`.
So the `BatchItem` implements `Resource` in which `getName()` returns `eid` of the event. And naturally this gets copied to the Avro implementation of `BatchItem`, which is `NakadiRecord`.

thomasabraham:
:+1:

PR #1415, opened 2022-05-13
Serialize enriched metadata

thomasabraham:
# Serialize enriched metadata

> Zalando ticket : team-aruha/issues/1106

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes


core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecord.java
```diff
@@ -78,4 +82,15 @@ public NakadiRecord setPartition(final String partition) {
         this.partition = partition;
```
adyach, line 82: probably partition should be part of the metadata

adyach: especially, that we take partition for producerrecord from metadata :)

core-common/src/main/java/org/zalando/nakadi/domain/VersionedAvroSchema.java
```diff
@@ -18,4 +18,8 @@ public Schema getSchema() {
     public String getVersion() {
         return version;
     }
+
+    public byte getVersionAsByte() {
```
adyach, line 22: I think version can be string only

adyach, line 22: I think version can be something different other than 1 byte. it is a string in nature.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -487,7 +482,7 @@ public List<NakadiRecordResult> sendEvents(final String topic,
         } catch (final InterruptException | InterruptedException e) {
             Thread.currentThread().interrupt();
             return prepareResponse(nakadiRecords, responses, e);
-        } catch (final RuntimeException e) {
+        } catch (final RuntimeException | IOException e) {
             kafkaFactory.terminateProducer(producer);
```
adyach, line 486: it looks like IOException is thrown whenever NakadiRecord was failed to be serialised to bytes. In that case there is no need to terminate the producer.

core-common/src/main/java/org/zalando/nakadi/domain/VersionedAvroSchema.java
adyach: discard that, that is metadata version, which is part of the envelop

core-common/src/main/java/org/zalando/nakadi/domain/NakadiRecord.java
thomasabraham: This is done.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
thomasabraham: This is fixed as well.

thomasabraham:
:+1:

core-common/src/main/java/org/zalando/nakadi/repository/kafka/AvroDeserializerWithSequenceDecoder.java
```diff
@@ -39,8 +39,11 @@ public byte[] deserializeAvro(final EnvelopeHolder envelope) throws RuntimeExcep
             metadata.put(NakadiAvroMetadata.OCCURRED_AT, new DateTime(
                     (long) metadata.get(NakadiAvroMetadata.OCCURRED_AT), DateTimeZone.UTC).toString());
 
-            metadata.put(NakadiAvroMetadata.RECEIVED_AT, new DateTime(
-                    (long) metadata.get(NakadiAvroMetadata.RECEIVED_AT), DateTimeZone.UTC).toString());
+            final var receivedAt = metadata.get(NakadiAvroMetadata.RECEIVED_AT);
+            if (receivedAt != null) {
```
adyach, line 43: why that could be the case ?

core-services/src/main/java/org/zalando/nakadi/service/publishing/BinaryEventProcessor.java
```diff
@@ -39,6 +39,11 @@ public BinaryEventProcessor(
         this.binaryEventPublisher = binaryEventPublisher;
         this.eventTypeCache = eventTypeCache;
         this.prePublishingChecks = prePublishingChecks;
+
+        if (prePublishingChecks.isEmpty()) {
+            // Safeguard against silent failure one spring injecting empty list
+            throw new RuntimeException("Checks should not be empty");
```
adyach, line 45:
```suggestion
            throw new RuntimeException("Publishing checks should not be empty");
```

core-common/src/main/java/org/zalando/nakadi/repository/kafka/AvroDeserializerWithSequenceDecoder.java
thomasabraham: The `received_at` is can be null according to the schema. So if the the enrichment fail for some reason, the following line will throw `NullpointerException`. So I decided that it is better to deliver the event without `received_at` timestamp than blocking consumption creating a `NullpointerException` in Nakadi consumption stack.

adyach: > So if the the enrichment fail for some reason

that means event publishing should fail as well.

thomasabraham: The publishing part of Nakadi is supposed to put `received_at` in the metadata. But, there might be a mistake configuration or a bug that result in Nakadi sending an event without `received_at` in the metadata in Kafka. Most likely it will never happen, but still the consuming side of Kafka should handle it without breaking because the schema of metadata allows it.

adyach:
👍

thomasabraham:
:+1:

PR #1416, opened 2022-05-16
Remove check for schema_type change

samizzy:
Allow for change of schema type for a event type


adyach:
however, we should postpone the release, because it is better to test binaries with the new event types, and only then we can allow migration.

adyach:
👍

PR #1417, opened 2022-05-19
Metadata in registry

adyach:
# One-line summary
use schema registry for binary event types


core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -90,8 +90,8 @@ public SchemaService(final SchemaRepository schemaRepository,
                          final EventTypeCache eventTypeCache,
                          final TimelineSync timelineSync,
                          final NakadiSettings nakadiSettings,
-                         final NakadiAuditLogPublisher nakadiAuditLogPublisher,
                          // dirty hack to resolve cycling dep, but they both need each other
+                         @Lazy final NakadiAuditLogPublisher nakadiAuditLogPublisher,
                          @Lazy final NakadiKpiPublisher nakadiKpiPublisher) {
```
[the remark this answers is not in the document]
adyach: funny enough it publishes changes into internal event.type.log event type.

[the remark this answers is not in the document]
a1exsh: Doesn't sound like the right place to do that, IMO.

adyach:
> `BinaryEventPublisherAT` has some failed tests.

Fixed, it needed binary event types

core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -64,6 +74,10 @@ public class SchemaService {
     private final NakadiSettings nakadiSettings;
     private final NakadiAuditLogPublisher nakadiAuditLogPublisher;
     private final NakadiKpiPublisher nakadiKpiPublisher;
+    private final LoadingCache<SchemaId, EventTypeSchema> schemasCache;
+    private final Map<SchemaId, org.apache.avro.Schema> avroSchemasCache;
+
+    private final org.apache.avro.Schema.Parser avroSchemaParser;
```
thomasabraham, line 80: use imports please.

core-services/src/main/java/org/zalando/nakadi/service/SchemaServiceProvider.java
```diff
@@ -0,0 +1,8 @@
+package org.zalando.nakadi.service;
+
+import org.apache.avro.Schema;
+
+public interface SchemaServiceProvider {
```
thomasabraham, line 5: The purpose of this interface is not clear, the `Provider` part indicate that it contains a function like `getSchemaService()` that will return the `SchemaService`. 

Maybe we don't need an interface here, we can directly use `SchemaService` class. But if we do need it, maybe we should name the interface `SchemaService` and the implementations can be things like `EmbededSchemaService`, `PostgresSchemaService`, `ZookeeperSchemaService`, etc. that evolves over time.

adyach: I used interface here to avoid creating mocks for getting schemas in the current tests. The interface allows me to use AvroSchema.java as implementation for TestSchemaServiceProvider proxy. The purpose of the test is only for tests atm. 

The naming can be improved, I see that it can be confusing with kind of factory style, where one provides the impl by having `get()` method. At the same time, `SchemaServiceProvider` has `getAvroSchema`, which **provides** schemas.

thomasabraham: Got it, in this case I suggest `SchemaProviderService`.

thomasabraham:
:+1:

core-services/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
```diff
@@ -11,8 +12,9 @@ public class KafkaRecordDeserializer implements RecordDeserializer {
 
     private final AvroDeserializerWithSequenceDecoder decoder;
 
-    public KafkaRecordDeserializer(final SchemaProviderService schemaService) {
-        this.decoder = new AvroDeserializerWithSequenceDecoder(schemaService);
+    public KafkaRecordDeserializer(final SchemaProviderService schemaService,
```
thomasabraham, line 15: Maybe taking `AvroDeserializerWithSequenceDecoder` as constructor argument is better in this case as it reduces the dependencies on `SchemProviderService` and `LocalSchemRegistry`. 
But its okay to keep this as it is as well.

core-common/src/main/java/org/zalando/nakadi/domain/kpi/AccessLogEvent.java
```diff
@@ -1,8 +1,27 @@
...
 
 public class AccessLogEvent extends KPIEvent {
+
+    private static final String PATH_SCHEMA = "event-type-schema/nakadi.access.log/nakadi.access.log.2.avsc";
+    private static final Schema SCHEMA;
+
+    static {
+        // load latest local schema
+        try {
+            SCHEMA = AvroUtils.getParsedSchema(
```
thomasabraham, line 17: Can we move the static block to `KPIEvent` class as a static method instead of duplicating it, then we can initialize the `SCHEMA` in one line by calling it.

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
```diff
@@ -42,6 +42,15 @@ public EventTypeSchema getSchemaVersion(final String name, final String version)
                     + "\" has no schema with version \"" + version + "\"", e);
         }
     }
+//
+//    public EventTypeSchema getSchemaVersionBySchema(final String name, final String schema) {
+//        return jdbcTemplate.queryForObject(
+//                "SELECT ets_schema_object FROM zn_data.event_type_schema " +
```
thomasabraham, line 48: Forgot to delete this I guess.

core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -324,6 +326,22 @@ public org.apache.avro.Schema getAvroSchema(final String etName,
         });
     }
 
+    @Override
+    public String getAvroSchemaVersion(final String name, final org.apache.avro.Schema schema)
+            throws NoSuchSchemaException {
+        final NameSchema key = new NameSchema(name, schema);
+        return schemaVersionCache.computeIfAbsent(key, (nameSchema) ->
```
thomasabraham, line 333: Note for myself : review this again.

thomasabraham: This is not the best approach, but it will work for now, and we should improve this later. 
One easy optimization is to ensure that the most recently added schema comes first in the list. 

If an exceptional event type have say 50 schema versions for whatever reason, after some time, only the latest schema versions are likely to be used. So we don't want to compare with all the 49 before reaching the 50.

a1exsh: I wonder why are we doing it in the first place...

adyach:
👍

adyach:
👍

PR #1419, opened 2022-05-24
Assert equality of JSONObject::toMap instead of similar()

a1exsh:
1. It seems that "similar" is "almost equal".
2. A failing test report is more helpful.


samizzy:
:+1:

a1exsh:
:+1:

PR #1420, opened 2022-05-25
Nakadi kpi publisher on controller level

thomasabraham:
# Nakadi kpi publisher on controller level

> Zalando ticket : team-aruha/issues/1119

## Description
The `NakadiKpiPublisher` and `NakadiAuditLogPublisher` are services are used from various other services. Using them from some of the services creates cyclic dependencies that needs special attention using `@Lazy` annotation. To avoid this, those usages are being migrated to controller level.


api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -52,9 +63,25 @@ public ResponseEntity<?> create(@PathVariable("name") final String name,
             throw new ValidationException(errors);
         }
 
-        schemaService.addSchema(name, schema);
-
-        return status(HttpStatus.OK).build();
+        final var originalEventType = eventTypeService.fetchFromRepository(name);
+        schemaService.addSchema(originalEventType, schema)
```
a1exsh, line 67: The service could provide both original and updated event type objects in the return value.  This will remove the need for the new method.

At the same time...  I think I made a very similar change in my PR (but for a different reason): https://github.com/zalando/nakadi/pull/1422/files#diff-5a8af8d6f60a128e59e74ed1304d8f16cdcce045cf29b71b84e35190b1530bd2R65

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -52,9 +63,25 @@ public ResponseEntity<?> create(@PathVariable("name") final String name,
...
+
+                    nakadiKpiPublisher.publish(() -> new EventTypeLogEvent()
+                            .setEventType(name)
+                            .setStatus("updated")
+                            .setCategory(originalEventType.getCategory().name())
+                            .setAuthz(originalEventType.getAuthorization() == null ? "disabled" : "enabled")
+                            .setCompatibilityMode(originalEventType.getCompatibilityMode().name()));
+                });
+
+        return status(HttpStatus.OK).build(); // Maybe return different status code when there is no change
```
a1exsh, line 84: Good point, we could use `201 Created` to indicate new schema version was really created.

api-misc/src/main/java/org/zalando/nakadi/controller/SettingsController.java
```diff
@@ -111,7 +116,14 @@ public ResponseEntity<?> updateAdmins(@Valid @RequestBody final ResourceAuthoriz
         if (errors.hasErrors()) {
             throw new ValidationException(errors);
         }
-        adminService.updateAdmins(authz.toPermissionsList(ADMIN_RESOURCE));
+        final var newAdmins = authz.toPermissionsList(ADMIN_RESOURCE);
+        final var oldAdmins = adminService.updateAdmins(newAdmins);
```
a1exsh, line 120: I didn't know this endpoint exists :)  So I was editing this in the DB directly, which is not audited...

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
a1exsh: Looks like it's even documented that way: https://nakadi.io/manual.html#/event-types/name/schemas_post

a1exsh (approved):
:+1:

api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
thomasabraham: Created a new issue for addressing this.

thomasabraham:
:+1:

a1exsh:
:+1:

PR #1421, opened 2022-05-25
Consistent versioning

samizzy:
Change Avro versioning to semantic versioning so that the version numbers are consistent across different schema types.


api-metastore/src/main/java/org/zalando/nakadi/service/SchemaEvolutionService.java
```diff
@@ -109,18 +107,25 @@ private void collectErrorMessages(final ValidationException e,
...
         checkEvolutionIncompatibilities(original, eventType);
         final var version = original.getSchema().getVersion();
+        final var typeChanged = !original.getSchema().getType().
+                equals(eventType.getSchema().getType());
+
+        if(typeChanged) {
+            return bumpVersion(original, eventType, MAJOR, new Version(version));
+        }
+
         switch (eventType.getSchema().getType()) {
```
adyach, line 117: why do we need `switch` if it looks the same for any `type` ?

samizzy: The parameters are same but the methods are different. `evolveJsonSchema()` vs `evolveAvroSchema()`

adyach: ups, missed that

adyach:
👍

samizzy:
:+1:

PR #1422, opened 2022-05-25
Validate JSON schema of events published to Avro topics

a1exsh:
Internal ticket: 1123


api-metastore/src/main/java/org/zalando/nakadi/service/EventTypeService.java
```diff
@@ -586,6 +586,12 @@ public EventType get(final String eventTypeName) throws NoSuchEventTypeException
         return eventType;
     }
 
+    public EventType getNoCache(final String eventTypeName) throws NoSuchEventTypeException, InternalNakadiException {
+        final EventType eventType = eventTypeRepository.findByName(eventTypeName);
+        authorizationValidator.authorizeEventTypeView(eventType);
```
adyach, line 591: that was not there before, why do we need it on that level?

a1exsh: Trying to mimic `get()` a few lines above which has this auth check.

core-common/src/main/java/org/zalando/nakadi/enrichment/Enrichment.java
```diff
@@ -39,10 +39,12 @@ public void validate(final EventTypeBase eventType) throws InvalidEventTypeExcep
...
     }
 
-    public void enrich(final BatchItem batchItem, final EventType eventType) throws EnrichmentException {
+    public void enrich(final BatchItem batchItem, final EventType eventType, final String schemaVersion)
+            throws EnrichmentException {
+
         for (final EnrichmentStrategyDescriptor descriptor : eventType.getEnrichmentStrategies()) {
             final EnrichmentStrategy strategy = getStrategy(descriptor);
-            strategy.enrich(batchItem, eventType);
+            strategy.enrich(batchItem, eventType, schemaVersion);
```
adyach, line 47: it would be better to pass a pair event type -> schema version, that will prevent errors like schema version does not exist for such event type

core-metastore/src/main/java/org/zalando/nakadi/repository/db/SchemaRepository.java
```diff
@@ -43,6 +44,22 @@ public EventTypeSchema getSchemaVersion(final String name, final String version)
         }
     }
 
+    public Optional<EventTypeSchema> getLatestSchemaByType(final String name, final EventTypeSchema.Type schemaType) {
+        //
+        // TODO: actually we should order by version, but that is harder to do in the DB
+        //
+        final String sql = "SELECT ets_schema_object FROM zn_data.event_type_schema " +
+                "WHERE ets_event_type_name = ? AND ets_schema_object ->> 'type' = ? " +
+                "ORDER BY ets_schema_object ->> 'created_at' DESC " +
```
a1exsh, line 53: Btw, `ORDER BY ... created_at` works only as long as we stick to the same time zone and format.  The field itself is treated as string here.

Here and in `SubscriptionDbRepository` we should cast the value to timestamp for ordering (and create the corresponding index to speed it up).

a1exsh:
:+1:

adyach:
👍

PR #1423, opened 2022-05-31
use the latest version for metadata

adyach:
# One-line summary
the metadata used was version 5, so returning it back


adyach:
> Should we rename the file as well: `metadata.1.avsc` => `metadata.5.avsc`?

yep

adyach:
👍

PR #1429, opened 2022-06-14
Fix incorrect schema validation during compatible mode

samizzy:
# One-line summary
Internal ticket no: 1146

## Description
Due to a recent change, schema validation was being perfomed on effective schema accidentally. This change fixes the `effectiveSchema()` method so that it does not mutate input parameters.


a1exsh:
:+1:

samizzy:
:+1:

PR #1436, opened 2022-06-22
Add REST endpoint for getting schema version by schema string

samizzy:
# One-line summary
Internal ticket: 116


api-metastore/src/main/java/org/zalando/nakadi/controller/SchemaController.java
```diff
@@ -161,4 +165,21 @@ public ResponseEntity<?> getSchemaVersion(@PathVariable("name") final String nam
         final EventTypeSchema result = schemaService.getSchemaVersion(name, version);
         return ResponseEntity.status(HttpStatus.OK).body(result);
     }
+
+    @GetMapping(
```
adyach, line 169: I would be useful to think about the API first. For that Nakadi repo has a swagger file. Could you make a proposal first for the API change? In regards to `GET` method with body, usually, for body `POST` request is used and it suits here. We have imagined that API endpoint would look like: `/event-types/{name}/schema/latest`

core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -314,7 +314,28 @@ public String getAvroSchemaVersion(final String name, final org.apache.avro.Sche
...
+    @Override
+    public String getSchemaVersion(final String name, final String schema,
+                                   final EventTypeSchemaBase.Type type)
+            throws NoSuchSchemaException {
+        switch (type) {
+            case AVRO_SCHEMA:
+                return getAvroSchemaVersion(name,  new Parser().parse(schema));
+            case JSON_SCHEMA:
+                final JSONObject jsonSchema = new JSONObject(schema);
+                return schemaRepository.getSchemas(name, 0, 100).stream()
```
a1exsh, line 326: It feels like here and in the avro function, the limit should not be specified — it only makes sense for external requests.

Also, if we only return one object, makes sense to ensure ordering, e.g. to return the latest version that matches the schema.

core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -314,7 +314,28 @@ public String getAvroSchemaVersion(final String name, final org.apache.avro.Sche
...
+                        .filter(ets -> ets.getType() == EventTypeSchemaBase.Type.JSON_SCHEMA)
+                        .filter(ets -> new JSONObject(ets.getSchema()).similar(jsonSchema))
+                        .map(EventTypeSchema::getVersion)
+                        .findFirst()
+                        .orElseThrow(() -> new NoSuchSchemaException(
+                                String.format("schema is not found for %s", name)));
+            default:
+                throw new UnsupportedSchemaTypeException("Unsupported schema type: " + type);
```
a1exsh, line 334: Should be in the `throws` declaration?

docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -574,13 +582,20 @@ paths:
           $ref: '#/definitions/EventTypeSchema'
         required: true
       responses:
+        '200':
+          description: |
+            Returns existing event type schema
```
a1exsh, line 587:
```suggestion
            Existing event type schema, with the version.
```
Should probably also refer to a definition?

api-metastore/src/main/java/org/zalando/nakadi/model/SchemaWrapper.java
```diff
@@ -3,14 +3,14 @@
 import javax.validation.constraints.NotNull;
 import java.util.Objects;
 
-public class CompatibilitySchemaRequest {
+public class SchemaWrapper {
```
a1exsh, line 6: Doesn't look like related to the topic?

core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
samizzy: Created a getAll schemas method that doesnt use limit & offsets, regarding giving the latest version, the db query orders by `created_at desc` and we return the first match in code, so should be the latest.

core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
samizzy: done

a1exsh:
Looks good, but please see the failing tests, they seem to be related to the changes.

a1exsh:
:+1:

samizzy:
:+1:

PR #1437, opened 2022-06-23
Reliably track request and response content length

a1exsh:
Some user agents set the Content-Length header, but not all.  Use request and
response wrapper class together with counting input/output stream wrappers.

### DIFF
--- a/app/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
+++ b/app/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
@@ -1,5 +1,7 @@
 package org.zalando.nakadi.filters;
 
+import com.google.common.io.CountingInputStream;
+import com.google.common.io.CountingOutputStream;
 import com.google.common.net.HttpHeaders;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


a1exsh:
> the PR is transforming from counting actual values (compressed bytes) to non-compressed, is it the intention?

It is not intentional, but I thought it might be useful to measure both compressed and original size of both request and response.
Btw, it doesn't look like we support zstd compression for response — only for request.

antban:
@a1exsh yes, response compression is not supported

a1exsh:
@antban Counting actual bytes on the uncompressed stream is the only reliable way to do it on this level, as the `Content-Length` header is only set by some of the user agents (and they are not required to set it anyway).  If we want to track compressed sizes as well, we should be doing that on the compression filter level.

app/src/main/java/org/zalando/nakadi/filters/LoggingFilter.java
```diff
@@ -179,13 +203,157 @@ private void logToAccessLog(final RequestLogInfo requestLogInfo, final int statu
...
+        }
+    }
+
+    private static class CountingInputStreamWrapper extends ServletInputStream {
+
+        private final ServletInputStream wrappedInputStream;
+        private final CountingInputStream countingInputStream;
+
+        CountingInputStreamWrapper(final ServletInputStream wrappedInputStream) {
+            this.wrappedInputStream = wrappedInputStream;
```
adyach, line 244: do we need because there is not implementation of some of the methods in CountingInputStream ?

a1exsh: Yes, this allows us to delegate calls to `isReady()` and `set*Listener` to the wrapped `ServletInputStream`.
I've copied the code from here initially: https://github.com/zalando/nakadi/blob/master/core-common/src/main/java/org/zalando/nakadi/util/CompressionBodyRequestFilter.java#L119

There we are using a narrower `InputStream` as the parameter type, which doesn't let us use the delegation for these methods.

antban:
:+1:

a1exsh:
:+1:

antban:
:+1:

a1exsh:
:+1:

antban:
:+1:

PR #1438, opened 2022-06-24
set default header

adyach:
# One-line summary
set default application/x-json-stream accept header if users did not supply the header


adyach:
👍

PR #1440, opened 2022-06-27
nakadi schemas retrieval api

adyach:
# One-line summary

nakadi schemas retrieval api

### DIFF
--- a/docs/_data/nakadi-event-bus-api.yaml
+++ b/docs/_data/nakadi-event-bus-api.yaml
@@ -2165,6 +2165,37 @@ paths:
           schema:
             $ref: '#/definitions/Problem'
 
+  /nakadi/avro-schemas/{name}:
+    get:
+      tags:
+        - schema-registry-api
+      description: The API to return Avro schemas to work with Nakadi.
+      parameters:
+        - name: name
+          in: path
+          description: Name of the Avro schema to fetch. Example, batch.publishing, batch.consumption.
+          type: string
+          required: true
+        - name: version
+          in: query
+          description: version of the Avro schema to fetch.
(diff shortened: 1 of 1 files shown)


docs/_data/nakadi-event-bus-api.yaml
```diff
@@ -2165,13 +2165,55 @@ paths:
...
+      tags:
+        - schema-registry-api
+      description: The API to return Avro schemas to work with Nakadi.
+      parameters:
+        - name: name
+          in: path
+          description: Name of the Avro schema to fetch. Example, batch.publishing, batch.consumption.
+          type: string
+          required: true
+        - name: version
```
samizzy, line 2179: 1. Why specify version as a query parameter instead of path?
2. If the api always returns list of objects and version acts as a filter, then maybe makes sense to accept a list of versions in the value?

thomasabraham: Regarding 2, we don't want to make it that complicated without a use case for it. 

If the client know which version they are looking for, then use it to get at most one schema, otherwise, get all versions.
Do we have a use case to get multiple but not all versions?

samizzy: I was just questioning if that is the intended design. If not, taking version as a path parameter like all our other apis would make sense unless there is a reason for this choice.

adyach: > Why specify version as a query parameter instead of path?
I do no see problem with that, it is how filtering is usually applied. From the other hand, the version as a part of the url looks nice, I will prepare the change. thanks pointing out!

PR #1442, opened 2022-06-28
Avro schemas API implementation

thomasabraham:
# Avro schemas API implementation

> Zalando ticket : team-aruha/issues/1129

## Description
Avro schemas API implementation

## Review
- [ ] Tests
- [ ] Documentation


core-services/src/main/java/org/zalando/nakadi/service/NakadiAvroSchemaService.java
```diff
@@ -0,0 +1,48 @@
...
+    @Autowired
+    public NakadiAvroSchemaService(final LocalSchemaRegistry localSchemaRegistry) {
+        this.localSchemaRegistry = localSchemaRegistry;
+    }
+
+    public List<VersionedAvroSchema> getNakadiAvroSchemas(final String schemaName) {
+        final var schemas = localSchemaRegistry.getEventTypeSchemaVersions(schemaName);
+        return schemas.entrySet().stream()
+                .map(entry -> new VersionedAvroSchema(entry.getValue(), entry.getKey()))
+                .collect(Collectors.toList());
```
a1exsh, line 42: The list is probably ordered by version, given the underlying implementation.  Should we make the ordering part of the API contract?

a1exsh:
:+1:

core-services/src/main/java/org/zalando/nakadi/service/NakadiAvroSchemaService.java
a1exsh: Yeah, I'm just wondering how the users are going to work with it and what would be their expectation.

core-services/src/main/java/org/zalando/nakadi/service/NakadiAvroSchemaService.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+@Service
+public class NakadiAvroSchemaService {
+
+    public static final class VersionedAvroSchema {
```
adyach, line 13: I suggest to move it to controller view, because it is clearly a view, which should not belong to service.

thomasabraham:
:+1:

adyach:
👍

PR #1443, opened 2022-06-28
More effecient reading of compressed request body

a1exsh:
- Override block reading methods for efficiency
- Delegate `isReady/isFinished/setReadListener` to the wrapped stream


a1exsh:
:+1:

PR #1449, opened 2022-07-01
Generate kpi event classes

adyach:
# One-line summary

use generated kpi event type classes to send events


core-common/src/main/java/org/zalando/nakadi/mapper/NakadiRecordMapper.java
```diff
@@ -118,11 +118,10 @@ private Envelope mapToEnvelope(final NakadiRecord nakadiRecord) {
...
-    public NakadiRecord fromAvroGenericRecord(final NakadiMetadata metadata,
-                                              final GenericRecord event) throws IOException {
-
+    public NakadiRecord fromAvroRecord(final NakadiMetadata metadata,
+                                       final SpecificRecord event) throws IOException {
         final var payloadOutputStream = new ByteArrayOutputStream();
-        final var eventWriter = new GenericDatumWriter(event.getSchema());
+        final var eventWriter = new SpecificDatumWriter<>(event.getSchema());
         eventWriter.write(event, EncoderFactory.get()
                 .directBinaryEncoder(payloadOutputStream, null));
```
a1exsh, line 126: Not related to this PR: I wonder if we could avoid some of the `new` objects created here.

a1exsh:
:+1:

PR #1451, opened 2022-07-04
Add sidecar profile for feature control via env variables

tor-vs-floki:
Add sidecar profile for feature control via env variables

### DIFF
--- a/api-misc/src/main/java/org/zalando/nakadi/config/RepositoriesConfig.java
+++ b/api-misc/src/main/java/org/zalando/nakadi/config/RepositoriesConfig.java
@@ -25,7 +25,7 @@ public class RepositoriesConfig {
 
     private static final Logger LOG = LoggerFactory.getLogger(RepositoriesConfig.class);
 
-    @Profile({"acceptanceTest", "local"})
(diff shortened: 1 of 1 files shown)


api-misc/src/main/java/org/zalando/nakadi/config/RepositoriesConfig.java
```diff
@@ -25,7 +25,7 @@ public class RepositoriesConfig {
 
     private static final Logger LOG = LoggerFactory.getLogger(RepositoriesConfig.class);
 
-    @Profile({"acceptanceTest", "local"})
+    @Profile({"acceptanceTest", "local", "sidecar"})
```
adyach, line 28: `sidecar` is the mode of execution, but it is very related to the deployment setup. could you achieve the same without adding another profile?

tor-vs-floki: We could work on the naming here, but I'd like to add a new profile as `acceptanceTest` and `local` can be misleading and future changes can break Archiver

tor-vs-floki:
Closing in favour for a better approach

PR #1452, opened 2022-07-04
Local schema service as schema provider service

thomasabraham:
# Local schema service as schema provider service

> Zalando ticket : team-aruha/issues/1128

## Description
The `LocalSchemaRegisrty` implements the `SchemaProviderService`, which is supposed to support different schema implementations.

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes


thomasabraham:
Turns out, the use case for this solution is not there anymore.

PR #1454, opened 2022-07-13
Remove adminClientId, as it was long not used

antban:
There is a weird part of code, that replaced actual(valid) client id with some constant. 
This part of code may cause problems understanding the reported values. 

Therefore the incorrect value is removed from configuration


api-misc/src/test/java/org/zalando/nakadi/controller/StoragesControllerTest.java
```diff
@@ -65,7 +63,7 @@ public void testListStorages() throws Exception {
                 .thenReturn(storages);
         when(adminService.isAdmin(AuthorizationService.Operation.READ)).thenReturn(true);
         mockMvc.perform(get("/storages")
-                .principal(mockPrincipal("org/zalando/nakadi")))
+                        .principal(mockPrincipal("org/zalando/nakadi")))
```
a1exsh, line 66: Do we still check something meaningful here, as we removed the security settings mock above?

antban: will check

antban: there was nothing meaningful there, it was used in many places only because of historical reasons.

PR #1455, opened 2022-07-15
Fix the stream limit

antban:
There is a stream_limit parameter for subscription streaming, that is used to limit the size of the stream. 
The accumulator for this parameter was not always properly accumulated.


antban:
:+1:

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -331,6 +330,7 @@ private void flushData(final EventTypePartition pk, final List<ConsumedEvent> da
             getContext().getKpiCollector().recordBatchSent(pk.getEventType(), batchSizeBytes, data.size());
 
             batchesSent++;
+            sentEvents += data.size();
```
adyach, line 333: do we reset it anywhere ?

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -267,7 +267,6 @@ private void streamToOutput(final boolean streamTimeoutReached) {
                     streamTimeoutReached))) {
                 sentSomething |= !toSend.isEmpty();
                 flushData(e.getKey(), toSend, batchesSent == 0 ? Optional.of("Stream started") : Optional.empty());
-                this.sentEvents += toSend.size();
```
a1exsh, line 270: In this and in the next block, the `messagesAllowedToSend` is tracked differently.  Should we rather use `getMessagesAllowedToSend()` in both cases?

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
adyach: ah, the stream is closed once `sentEvents` reaches certain number, ok

a1exsh:
:+1:

PR #1456, opened 2022-07-15
Use ThreadLocalRandom for partitioning

a1exsh:
The `Random` class is synchronized, but we don't need that for our use case.


adyach:
👍

PR #1458, opened 2022-07-19
Queue publishing of events for the same key in a batch

a1exsh:
When hash partitioning (and/or log compaction) is used this approach ensures
that the order of events is preserved for each key in the face of intermittent
publishing errors.

For now we simply mark any events that were not attempted for publishing as
aborted.  Later this can be improved to implement retry of failed events.

No change of behavior when no event keys are set (random or user-defined
partitioning): the whole batch is submitted as a single chunk.

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/domain/BatchItem.java
+++ b/core-common/src/main/java/org/zalando/nakadi/domain/BatchItem.java
@@ -65,7 +65,6 @@ public class BatchItem implements Resource<BatchItem> {
     private String[] injectionValues;
     private final List<Integer> skipCharacters;
     private String partition;
-    private String brokerId;
     private String eventKey;
     private List<String> partitionKeys;
     private int eventSize;
@@ -114,15 +113,6 @@ public class BatchItem implements Resource<BatchItem> {
 
     public void setPartition(final String partition) {
         this.partition = partition;
-    }
-
-    @Nullable
-    public String getBrokerId() {
-        return brokerId;
-    }
-
-    public void setBrokerId(final String brokerId) {
-        this.brokerId = brokerId;
     }
 
     @Nullable
--- a/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -299,6 +299,11 @@ public class KafkaTopicRepository implements TopicRepository {
     public void syncPostBatch(
             final String topicId, final List<BatchItem> batch, final String eventType, final boolean delete)
             throws EventPublishingException {
+
+        final List<BatchItem> itemsLeft = new ArrayList<>(batch);
+        final List<BatchItem> chunkItems = new ArrayList<>();
+        final Set<String> chunkKeys = new HashSet<>();
+
         final Producer<byte[], byte[]> producer = kafkaFactory.takeProducer();
         try {
             final Map<String, String> partitionToBroker = producer.partitionsFor(topicId).stream()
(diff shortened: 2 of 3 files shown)


core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -313,66 +318,108 @@ public void syncPostBatch(
...
-                    item.updateStatusAndDetail(EventPublishingStatus.FAILED,
-                            String.format("No leader for partition: %s, topic: %s.", item.getPartition(), topicId));
-                    LOG.error("Failed to publish to kafka. No leader for ({}:{}).",
-                            topicId, item.getPartition());
-                    continue;
+            long timeoutMs = createSendTimeout();
+            while (true) {
+                final long chunkStartTime = System.currentTimeMillis();
+
+                for (int i = 0; i < itemsLeft.size(); ) {
```
a1exsh, line 325: Potential `O(N²)` if all events are for the same key...

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -313,33 +314,59 @@ public void syncPostBatch(
...
+                if (itemKey == null) {
+                    // items without a key go out of order
+                    futures.add(publishItem(producer, topicId, eventType, item, delete));
+                } else {
+                    // items for a single key are composed together for sequential publishing
+                    CompletableFuture<Exception> fut = futuresByKey.get(itemKey);
+                    if (fut == null) {
+                        fut = publishItem(producer, topicId, eventType, item, delete);
+                    } else {
+                        fut = fut.thenCompose((e) -> {
```
a1exsh, line 339: Maybe we want to use `thenComposeAsync()` here, as `publishItem()` may block on the next `producer.send()` call...

core-common/src/main/java/org/zalando/nakadi/domain/BatchItem.java
```diff
@@ -65,7 +65,6 @@ public static EmptyInjectionConfiguration build(final int position, final boolea
     private String[] injectionValues;
     private final List<Integer> skipCharacters;
     private String partition;
-    private String brokerId;
```
adyach, line 68: looks like unrelated change

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -299,41 +300,55 @@ public boolean topicExists(final String topic) throws TopicRepositoryException {
...
+                                        return sendItem.apply(item);
+                                    }
+                                });
+                    }
+                    futuresByKey.put(itemKey, fut);
                 }
-                item.setStep(EventPublishingStep.PUBLISHING);
-                sendFutures.put(item, publishItem(producer, topicId, eventType, item, delete));
             }
+            // we will wait for all futures together
```
a1exsh, line 341: Btw, there is now enough code shared between the JSON and binary send methods, to think about extracting it into a common method, IMO.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -299,41 +300,55 @@ public boolean topicExists(final String topic) throws TopicRepositoryException {
     public void syncPostBatch(
             final String topicId, final List<BatchItem> batch, final String eventType, final boolean delete)
             throws EventPublishingException {
+
         final Producer<byte[], byte[]> producer = kafkaFactory.takeProducer();
         try {
-            final Map<String, String> partitionToBroker = producer.partitionsFor(topicId).stream()
```
adyach, line 304: I do not understand how this is related to the change

adyach, line 304: it is not clear how this is related to the change

a1exsh: It's becomes harder to recall what was the reason to change this part, but it looks like in the context of the bigger change we don't have to track this anymore.  Esp. since we intend to enable retries, the item may be sent to a different broker in a repeated attempt, so it's no longer a 1:1 relation.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -299,41 +300,55 @@ public boolean topicExists(final String topic) throws TopicRepositoryException {
...
+
             for (final BatchItem item : batch) {
-                final String brokerId = item.getBrokerId();
-                if (brokerId == null) {
-                    item.updateStatusAndDetail(EventPublishingStatus.FAILED,
-                            String.format("No leader for partition: %s, topic: %s.", item.getPartition(), topicId));
-                    LOG.error("Failed to publish to kafka. No leader for ({}:{}).",
-                            topicId, item.getPartition());
-                    continue;
+                final String itemKey = item.getEventKey();
```
adyach, line 318: it feels like that logic should be one level higher and the grouping should be done not by the event id, but rather per partition, this is what matters for publishing. It sounds logically more correct. probably, it will change the method signature 🤷

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
adyach: I am not against removing it, just maybe a separate PR that it does not affect actual change

a1exsh:
After measuring the expected performance impact decided not to go this way.

PR #1464, opened 2022-08-24
Reference base image directly

tor-vs-floki:
There is no need to have the base image as a variable

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
-ARG BASE_IMAGE
-FROM ${BASE_IMAGE}
+FROM container-registry.zalando.net/library/eclipse-temurin-11-jre:latest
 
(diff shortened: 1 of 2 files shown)


tor-vs-floki:
👍

Dockerfile
```diff
@@ -1,5 +1,4 @@
-ARG BASE_IMAGE
-FROM ${BASE_IMAGE}
+FROM container-registry.zalando.net/library/eclipse-temurin-11-jre:latest
```
antban, line 1: It makes sense to base on JDK, as sometimes we need to take a memory dump, and with JDK installed it is much simpler

antban:
:+1:

tor-vs-floki:
👍

PR #1466, opened 2022-09-12
Correct use of tracing error tag

samizzy:
# One-line summary
Use correct tracing tags for errors.

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/service/TracingService.java
+++ b/core-common/src/main/java/org/zalando/nakadi/service/TracingService.java
@@ -23,6 +23,7 @@ public class TracingService {
 
     private static final long BUCKET_5_KB = 5000L;
     private static final long BUCKET_50_KB = 50000L;
+    public static final String ERROR_DESCRIPTION = "error.description";
 
     public static String getSLOBucketName(final long batchSize) {
         if (batchSize > BUCKET_50_KB) {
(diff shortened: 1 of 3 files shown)


core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -259,9 +258,8 @@ private void validate(final List<BatchItem> batch, final EventType eventType, fi
                 } catch (final EventValidationException e) {
                     item.updateStatusAndDetail(EventPublishingStatus.FAILED, e.getMessage());
                     if (eventType.getCategory() != EventCategory.UNDEFINED) {
-                        TracingService.log(ImmutableMap.of(
-                                "event.id", item.getEvent().getJSONObject("metadata").getString("eid"),
-                                "error", e.getMessage()));
+                        TracingService.setTag("event.id", item.getEvent().getJSONObject("metadata").getString("eid"));
```
adyach, line 261: it should be a log not a tag

core-common/src/main/java/org/zalando/nakadi/service/TracingService.java
```diff
@@ -94,7 +95,7 @@ public static Span setErrorFlag(final Span span) {
 
     public static void logError(final String error) {
         if (error != null) {
-            getActiveSpan().log(ImmutableMap.of("error.description", error));
+            getActiveSpan().log(ImmutableMap.of(ERROR_DESCRIPTION, error));
```
adyach, line 98: should we mark it as error as well ?

adyach:
👍

samizzy:
:+1:

PR #1467, opened 2022-09-20
Update partition count logic

samizzy:
# One-line summary
Remove partition count logic that depended on messagePerMinute and messageSize from default statistics

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/repository/kafka/PartitionsCalculator.java
+++ b/core-common/src/main/java/org/zalando/nakadi/repository/kafka/PartitionsCalculator.java
@@ -14,7 +14,7 @@ public class PartitionsCalculator {
     }
 
     public int getBestPartitionsCount(final EventTypeStatistics stat) {
-        if (null == stat || stat.getReadParallelism() == 0 && stat.getWriteParallelism() == 0) {
+        if (null == stat || stat.getReadParallelism() <= 0 && stat.getWriteParallelism() <= 0) {
             return defaultPartitionCount;
         }
         final int maxPartitionsDueParallelism = Math.max(stat.getReadParallelism(), stat.getWriteParallelism());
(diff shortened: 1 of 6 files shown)


core-common/src/main/java/org/zalando/nakadi/repository/kafka/PartitionsCalculator.java
```diff
@@ -1,188 +1,30 @@
...
         this.defaultPartitionCount = defaultPartitionCount;
         this.maxPartitionCount = maxPartitionCount;
-        for (final SpeedStatistics ss : instanceInfo.getStats()) {
-            stats.put(ss.getMessageSize(), ss.getSpeed());
-        }
     }
 
     public int getBestPartitionsCount(final EventTypeStatistics stat) {
-        if (null == stat) {
+        if (null == stat || stat.getReadParallelism() == 0 && stat.getWriteParallelism() == 0) {
```
tor-vs-floki, line 17: `readParallelism` and `writeParallelism` are already annotated as :

```
    @NotNull
    @Min(value = 1, message = "at least one reader expected")
    private Integer readParallelism;
    @NotNull
    @Min(value = 1, message = "at least one writer expected")
    private Integer writeParallelism;
    ```

samizzy: thats true but I thought since the class is named `PartitionsCalculator.java` and the method is public I would have these additional checks. I updated the check with `<=` in a new change to accomodate for more.

tor-vs-floki: That is true, and if we don't run validator explicitly we don't gain protection by using `@Min` annotation.

tor-vs-floki:
👍

samizzy:
:+1:

PR #1469, opened 2022-10-26
Fix logging leak

antban:
# One-line summary
There is incorrect logging library usage - loggers are created for random names

## Description
The loggers are supposed to be created per class or per other static entity. 
Right now Nakadi is creating loggers for random names. 
Logging libraries to not have an ability to remove loggers by names that are not in use anymore, and effectively this may result in OOM after some time. 

This PR replaces the way how the subscription id, stream id and flow Id are passed and used in logging purposes. 

1. Nakadi usage of NDC is replaces with usage of MDC
2. Subscription id and stream id are migrated to MDC instead of logger name
3. A special filter for logger records is added to print only subscription ids we are interested in.

## Deployment Notes
Logging configuration should be reworked


core-common/src/test/java/org/zalando/nakadi/util/FlowIdRequestFilterTest.java
```diff
@@ -0,0 +1,24 @@
+package org.zalando.nakadi.util;
+
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+
+public class FlowIdRequestFilterTest {
```
adyach, line 8: what do we test here?

core-common/src/main/java/org/zalando/nakadi/util/MDCUtils.java
```diff
@@ -0,0 +1,114 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
+
+import javax.annotation.Nullable;
+import java.io.Closeable;
+import java.util.HashMap;
+import java.util.Map;
+
+public final class MDCUtils {
```
adyach, line 12: several concerns: 
- it is not clear why you need that code and can not use `MDC` class directly or via proxy
- the easy way to create just necessary resource like subscription id / stream / flow would be easier to read the code, instead you create the god method, which can do anything, is there a need in the future?
- the class creates hashmap per request plus MDC hashmap

antban: > it is not clear why that code is needed and can not use MDC class directly or via proxy

There is a lot of places that are using the same names and the same functionality, That is the reason why the code is concentrated in a single place

> the easy way to create just necessary resource like subscription id / stream / flow would be easier to read the code, instead it has a god method, which can do anything, is there a need in the future?

In order to push/pop several values and properly remove the values in case of unknown exceptions - there is a complex procedure to roll back.

> the class creates hashmap per request plus MDC hashmap

Thanks for spotting! I will do the optimization for the most obvious part - for 1 variable (flow id).

antban: After discussion - the complex procedure to roll back was removed and simplified. The second comment is obsolete now as well

antban:
@adyach @tor-vs-floki Please, take another look

core-common/src/main/java/org/zalando/nakadi/util/MDCUtils.java
```diff
@@ -0,0 +1,90 @@
...
+        final String oldSubscriptionId = MDC.get(SUBSCRIPTION_ID);
+        final String oldStreamID = MDC.get(SUBSCRIPTION_STREAM_ID);
+
+        MDC.put(SUBSCRIPTION_ID, maskNull(subscriptionId));
+        MDC.put(SUBSCRIPTION_STREAM_ID, maskNull(streamId));
+
+        return () -> {
+            restoreValue(SUBSCRIPTION_STREAM_ID, oldStreamID);
+            restoreValue(SUBSCRIPTION_ID, oldSubscriptionId);
+        };
```
tor-vs-floki, line 63:
```suggestion
        final MDCUtils.CloseableNoEx subscriptionIdCloseable = withSingleMDCValue(SUBSCRIPTION_ID, subscriptionId);
        final MDCUtils.CloseableNoEx resetStreamIdFn = withSingleMDCValue(SUBSCRIPTION_STREAM_ID, streamId);

        return () -> {
            subscriptionIdCloseable.close();
            resetStreamIdFn.close();
        };
```

core-services/src/main/java/org/zalando/nakadi/service/publishing/NakadiKpiPublisher.java
```diff
@@ -95,7 +89,7 @@ private NakadiMetadata buildMetadata(final String eventTypeName,
         metadata.setEid(uuidGenerator.randomUUID().toString());
         metadata.setEventType(eventTypeName);
         metadata.setSchemaVersion(eventVersion);
-        metadata.setFlowId(FlowIdUtils.peek());
+        metadata.setFlowId(MDCUtils.getFlowId());
```
tor-vs-floki, line 92: NakadiKpiPublisher is now coupled with NDCUtils aka Log utilities

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventMetadata.java
```diff
@@ -20,6 +20,6 @@ public JSONObject addTo(final JSONObject event) {
         return event.put("metadata", new JSONObject()
                 .put("occurred_at", Instant.now())
                 .put("eid", uuidGenerator.randomUUID())
-                .put("flow_id", FlowIdUtils.peek()));
+                .put("flow_id", MDCUtils.getFlowId()));
```
tor-vs-floki, line 23: EventMetadata is now coupled with NDCUtils aka Log utilities

adyach:
👍

core-common/src/main/java/org/zalando/nakadi/enrichment/MetadataEnrichmentStrategy.java
```diff
@@ -52,7 +52,7 @@ public void enrich(final NakadiRecord nakadiRecord, final EventType eventType) t
         metadata.setPublishedBy(getPublisher());
         metadata.setReceivedAt(Instant.now());
         if (metadata.getFlowId() == null || metadata.getFlowId().isEmpty()) {
-            metadata.setFlowId(FlowIdUtils.peek());
+            metadata.setFlowId(MDCUtils.getFlowId());
```
tor-vs-floki, line 55: MDCUtils spilled here as well

app/src/test/java/org/zalando/nakadi/util/SubscriptionIdLogFilterTest.java
```diff
@@ -0,0 +1,28 @@
...
+            // Should be logged
+            LOG.info("With wrong subscription id, but high level");
+        }
+
+        try (var ignore = MDCUtils.withSubscriptionId("00000000-0000-0000-0000-000000000000")) {
+            // Should be logged
+            LOG.debug("With correct subscription id");
+        }
+    }
+}
```
tor-vs-floki, line 28: Do we need to commit this test?

antban:
:+1:

tor-vs-floki:
👍

a1exsh:
Noice!

Looks a bit involved though.  Have we looked at [putClosable](https://www.slf4j.org/api/org/slf4j/MDC.html#putCloseable-java.lang.String-java.lang.String-), for example?

PR #1470, opened 2022-10-27
Quickfix for an edge case where repartition breaks

tor-vs-floki:
When there are multiple active timelines across different Kafka clusters, repartitioning breaks.

### DIFF
--- a/core-services/src/test/java/org/zalando/nakadi/service/timeline/TimelineServiceTest.java
+++ b/core-services/src/test/java/org/zalando/nakadi/service/timeline/TimelineServiceTest.java
@@ -1,6 +1,7 @@
 package org.zalando.nakadi.service.timeline;
 
 import com.google.common.collect.ImmutableList;
+import org.assertj.core.util.Lists;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


antban:
:+1:

antban:
Agree with the test part

core-services/src/test/java/org/zalando/nakadi/service/timeline/TimelineServiceTest.java
```diff
@@ -177,4 +178,44 @@ public void whenCreateTimelineForCompactedEventTypeThenException() throws Except
...
+                .thenReturn(topicRepository1);
+        when(topicRepositoryHolder.getTopicRepository(s2))
+                .thenReturn(topicRepository2);
+
+        timelineService.updateTimeLineForRepartition(eventType, 2);
+
+        Mockito.verify(topicRepository1, Mockito.times(1)).repartition("topic2", 2);
+        Mockito.verify(topicRepository2, Mockito.times(1)).repartition("topic3", 2);
+
+        Assert.assertEquals("[1, -1]", t2.getLatestPosition().toDebugString());
```
adyach, line 218: it is not clear why you check latest position, if objection is to repartition. maybe check number of partitions before and after, all necessary calls were made. another option is to go with acceptance test with two storages, which is longer, but reliable.

tor-vs-floki: It is `TimelineService` responsibility to add the initial offset to each timeline for a newly added partition. Any active timeline that is not the latest one should hold the offsets of the prior Kafka topic.

eg:
 **before** repartition(topic with 1 partition), any active timeline should have the latest offset(s) from the time of timeline therefore `latestPosition=[1]`

**after** repartition(increase partition count by 1), all active timelines(beside the most recent one) should be patched with an initial offset eg `latestPosition=[1,-1]`

Not sure about the acceptance test, if you are referring to integration tests against an actual running system then we have to spawn 2 Kafka clusters, which I am not sure about it.

adyach: the acceptance could you the same kafka, just create a separate storage, but lets proceed with what you have, the test covers the part was fixed 👍

adyach:
👍

tor-vs-floki:
👍

PR #1471, opened 2022-11-08
Remove disk stats collection job from nakadi

antban:
The reason is that the job is collecting the data only for a kafka instance that is run by bubuku.
In this kind of approach the stats collection job should be externalized from nakadi (nakadi is relying on kafka, not bubuku).


adyach:
👍

a1exsh:
:+1:

PR #1472, opened 2022-11-08
fix log message for commits in the future

ePaul:
# One-line summary

Extend log message to include the offset just compared.

## Description

@tanvir002700 noted in an internal chat a situation (running Nakadi locally)
where this error message is raised. When trying to look into it, I noticed that the
offset printed as part of the error message is not too helpful:
Nakadi is comparing `offset` and `sentOffset`, but was logging `sentOffset` and `commitOffset`.

This PR adds `offset` to the log message, hopefully making it easier to debug problems when this occurs.

## Review
- [x] Tests
- [x] Documentation

## Deployment Notes

### DIFF
--- a/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
+++ b/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
@@ -186,7 +186,7 @@ class PartitionData {
     CommitResult onCommitOffset(final NakadiCursor offset) {
         boolean seekKafka = false;
         if (comparator.compare(offset, sentOffset) > 0) {
-            LOG.error("Commit in future: new: {}, sent: {}, committed: {}. Will skip sending obsolete data", offset, sentOffset,
+            LOG.error("Commit in future: received: {}, sent: {}, committed: {}. Will skip sending obsolete data", offset, sentOffset,
                     commitOffset);
             seekKafka = true;
             sentOffset = offset;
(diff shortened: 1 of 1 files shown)


ePaul:
I didn't specifically check whether any tests are affected – looking for "will skip sending obsolete data" didn't show other locations than the one modified here, so I suppose there are no tests for this. Same for documentation.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java
```diff
@@ -186,7 +186,7 @@ private CommitResult(final boolean seekOnKafka, final long committedCount) {
     CommitResult onCommitOffset(final NakadiCursor offset) {
         boolean seekKafka = false;
         if (comparator.compare(offset, sentOffset) > 0) {
-            LOG.error("Commit in future: current: {}, committed {} will skip sending obsolete data", sentOffset,
+            LOG.error("Commit in future: new: {}, sent: {}, committed: {}. Will skip sending obsolete data", offset, sentOffset,
```
adyach, line 189: what does `new` means here ?

ePaul: It's the offset which is currently being tried to commit. I felt that would be clearer than "current", but feel free to suggest a better wording.

antban, line 189:
```suggestion
            LOG.error("Commit in future: received: {}, sent: {}, committed: {}. Will skip sending obsolete data", offset, sentOffset,
```

antban: Maybe like this?

ePaul: I've applied @antban's suggestion ("received").

antban:
:+1:

ePaul:
:+1:

a1exsh:
Hm: `[ant:checkstyle] [ERROR] /workspace/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/PartitionData.java:189: Line is longer than 120 characters (found 133). [LineLength]`

Wonder how it passed the build earlier? :)

PR #1478, opened 2022-12-16
metrics to measure paritioning keys per batch

adyach:
# One-line summary

Measure how many partitioning keys are published per batch. It should help to understand the latencies in case Nakadi publishes events by keys in order (sequential)


a1exsh:
:+1:

adyach:
👍

PR #1484, opened 2023-01-05
Revert "metrics to measure paritioning keys per batch (#1478)"

adyach:
This reverts commit 88ce0b1963336c3b7bb5f96b3e249db0921b9bb5 and 980f874af13fa1e796224d5863e460d91def0140, because test of the metrics was done and they are not needed anymore


adyach:
👍

core-services/src/main/java/org/zalando/nakadi/service/publishing/EventPublisher.java
```diff
@@ -293,7 +274,7 @@ private void validate(final List<BatchItem> batch, final EventType eventType, fi
 
     private void submit(
             final List<BatchItem> batch, final EventType eventType, final boolean delete)
-            throws EventPublishingException, InternalNakadiException {
+        throws EventPublishingException, InternalNakadiException {
```
a1exsh, line 277: Is that your IDE?

a1exsh:
:+1:

PR #1485, opened 2023-01-06
respond 'bad request' if content is not gzip

adyach:
at the moment nakadi throws 503 in case content encoding is set to gzip and the content is not gzip format. the commit changes the behaviour that nakadi responds with 400 in case of the content is not gzip format. it is harder to achieve for zstd because zstdinput stream does not validate the content on the start, the desicion is to keep it (the pubslishing will fail and user connection will be abrupted)


a1exsh:
:+1:

adyach:
👍

a1exsh:
:+1:

PR #1486, opened 2023-01-09
track publishing latency

adyach:
the commit adds ability to track latency on publishing path in more granular way. that should help with debugging latency issues.


a1exsh:
:+1:

adyach:
👍

PR #1487, opened 2023-01-13
measure wait for the batch sent to kafka

adyach:
the change should help to analyse how long it takes for kafka to process the batch and understand if latency comes from the producer


adyach:
👍

adyach:
👍

a1exsh:
:+1:

PR #1489, opened 2023-01-17
Don't create extra KafkaLocationManager

a1exsh:
Seems like an oversight that we were creating two of them per topic repository.


a1exsh:
:+1:

adyach:
👍

PR #1488, opened 2023-01-16
Multiple active Kafka producers

a1exsh:
A Kafka producer is single-threaded.  With our setup it makes sense to have more than one of them active, for better utilization of resources.

Internal ticket: 1298

PR #1490, opened 2023-01-17
Dedicate a Kafka producer to each of the publishing SLO buckets

a1exsh:
This is the evolution of #1488, where we already had multiple producers.

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
+++ b/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
@@ -13,7 +13,9 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nullable;
 import java.io.Closeable;
+import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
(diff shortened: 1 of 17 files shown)


core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -13,56 +13,48 @@
...
-            this.activeProducers.add(null);
-        }
-
-        this.activeProducerCounter = new AtomicLong(0);
+        this.activeProducerByKey = new ConcurrentHashMap<>();
     }
 
     @Nullable
-    private Producer<byte[], byte[]> takeUnderLock(final int index, final boolean canCreate) {
+    private Producer<byte[], byte[]> takeUnderLock(final String key, final boolean canCreate) {
```
adyach, line 44:
```suggestion
    private Producer<byte[], byte[]> takeUnderLock(final String producerId, final boolean canCreate) {
```

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -13,56 +13,48 @@
...
         this.useCount = new ConcurrentHashMap<>();
         this.rwLock = new ReentrantReadWriteLock();
-
-        this.activeProducers = new ArrayList<>(numActiveProducers);
-        for (int i = 0; i < numActiveProducers; ++i) {
-            this.activeProducers.add(null);
-        }
-
-        this.activeProducerCounter = new AtomicLong(0);
+        this.activeProducerByKey = new ConcurrentHashMap<>();
```
adyach, line 40: why should it be concurrent if we change it under write lock ?

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -78,29 +70,31 @@ private Producer<byte[], byte[]> takeUnderLock(final int index, final boolean ca
...
      */
-    public void releaseProducer(final Producer<byte[], byte[]> producer) {
+    public void releaseProducer(final String key, final Producer<byte[], byte[]> producer) {
         final AtomicInteger counter = useCount.get(producer);
         if (counter != null && 0 == counter.decrementAndGet()) {
             final boolean deleteProducer;
             rwLock.readLock().lock();
             try {
-                deleteProducer = !activeProducers.contains(producer);
+                deleteProducer = activeProducerByKey.get(key) != producer;
```
adyach, line 97: if use `containsValue` then `key` is not needed

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -128,21 +126,24 @@ public void releaseProducer(final Producer<byte[], byte[]> producer) {
...
-        LOG.info("Received signal to terminate producer " + producer);
+    public void terminateProducer(final String key, final Producer<byte[], byte[]> producer) {
+        LOG.info("Received signal to terminate producer for key {}: {} ", key, producer);
         rwLock.writeLock().lock();
         try {
-            final int index = activeProducers.indexOf(producer);
-            if (index >= 0) {
-                activeProducers.set(index, null);
+            if (activeProducerByKey.containsKey(key)) {
+                activeProducerByKey.remove(key);
```
adyach, line 134: may use only remove, it returns `null if the map contained no mapping for the key`, otherwise it is double contains

adyach: and since it is not that often operation maybe we can just remove by value, that would anticipate changing the method signature

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -230,9 +232,13 @@ public void repartition(final String topic, final int partitionsNumber) throws C
             if (!Boolean.TRUE.equals(areNewPartitionsAdded)) {
                 throw new TopicConfigException(String.format("Failed to repartition topic to %s", partitionsNumber));
             }
-            final Producer<byte[], byte[]> producer = kafkaFactory.takeProducer();
-            kafkaFactory.terminateProducer(producer);
-            kafkaFactory.releaseProducer(producer);
+            // Man, WTF!?..
```
adyach, line 235: ¯\_(ツ)_/¯

a1exsh:
:+1:

PR #1496, opened 2023-02-08
remove excessive check for the leader

adyach:
at the moment nakadi manually checks for the leader, that's not necessary because kafka producer does it for us. it also helps to keep the code clean and prepare for the future changes.

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -1,5 +1,6 @@
 package org.zalando.nakadi.repository.kafka;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import io.opentracing.Tracer;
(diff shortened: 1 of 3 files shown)


tor-vs-floki:
👍

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -306,31 +305,11 @@ public void syncPostBatch(
...
         final Producer<byte[], byte[]> producer = kafkaFactory.takeProducer();
         try {
-            final Map<String, String> partitionToBroker = producer.partitionsFor(topicId).stream()
-                    .filter(partitionInfo -> partitionInfo.leader() != null)
-                    .collect(
-                            Collectors.toMap(
-                                    p -> String.valueOf(p.partition()),
-                                    p -> p.leader().idString() + "_" + p.leader().host()));
-            batch.forEach(item -> {
-                Preconditions.checkNotNull(
```
a1exsh, line 316: I think checking for partition is actually useful.  If we don't specify it, Kafka will do its own partitioning.

adyach: yes, I cut too much..

a1exsh:
:+1:

adyach:
👍

PR #1498, opened 2023-02-10
use producer per topic partition hash

adyach:
take producer based on the topic partition hashing that helps to distribute paratitions evenly between producers


a1exsh:
:+1:

adyach:
👍

PR #1499, opened 2023-02-20
Count unique `event-type:partition` instances

a1exsh:
We need to see how it grows over time to better understand the publisher performance degradation.


adyach:
👍

a1exsh:
:+1:

adyach:
👍

PR #1502, opened 2023-03-17
Add more detail to schema validation errors

samizzy:
### One-line summary
Its helpful for users if they see more meaningful errors that can point to the problem in their schema.


burimshala:
👍

samizzy:
:+1:

tor-vs-floki:
👍

PR #1505, opened 2023-03-29
more logs for consumer usage in time lag calculation

adyach:
# One-line summary

more logs for consumer usage in time lag calculation


a1exsh:
:+1:

adyach:
👍

PR #1509, opened 2023-04-03
Validate datalake annotations

tor-vs-floki:
Validate some annotations according to their format and dependence

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/annotations/validation/AnnotationKey.java
+++ b/core-common/src/main/java/org/zalando/nakadi/annotations/validation/AnnotationKey.java
@@ -24,7 +24,7 @@ import java.lang.annotation.Target;
         message = "Key name should start and end with a letter or a digit")
 @Constraint(validatedBy = {})
 public @interface AnnotationKey {
-    String message() default "{org.zalando.nakadi.annotations.validation.AnnotationKey.message}}";
+    String message() default "{org.zalando.nakadi.annotations.validation.AnnotationKey.message}";
 
     Class<?>[] groups() default {};
 
--- a/core-common/src/test/java/org/zalando/nakadi/annotations/validation/DataLakeAnnotationsTest.java
+++ b/core-common/src/test/java/org/zalando/nakadi/annotations/validation/DataLakeAnnotationsTest.java
@@ -61,4 +61,18 @@ public class DataLakeAnnotationsTest {
         final Set<ConstraintViolation<TestClass>> result = validator.validate(new TestClass(annotations));
         assertTrue("Retention period and reason exist correctly", result.isEmpty());
     }
+
+    @Test
+    public void itWorksWithOtherAnnotations() {
+        final var annotations = Map.of("some-annotation", "some-value");
+        final Set<ConstraintViolation<TestClass>> result = validator.validate(new TestClass(annotations));
+        assertTrue("Should not throw errors when few annotations are passed, " +
+                "but none of them belongs to data lake", result.isEmpty());
+    }
+
+    @Test
+    public void itWorksWithoutAnnotations(){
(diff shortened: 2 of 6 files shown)


core-common/src/test/java/org/zalando/nakadi/annotations/validation/DataLakeAnnotationsTest.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    @Test
+    public void whenRetentionPeriodFormatIsWrongThenFail() {
+        final var annotations = Map.of(
+                DataLakeAnnotationValidator.RETENTION_PERIOD_ANNOTATION, "1 airplane",
+                DataLakeAnnotationValidator.RETENTION_REASON_ANNOTATION, "I need my data"
+        );
+        final Set<ConstraintViolation<TestClass>> result = validator.validate(new TestClass(annotations));
+        assertFalse("Missing retention reason is treated valid", result.isEmpty());
```
a1exsh, line 52: Copy-paste from above.  Can/should we check the actual validation result, not just that it's not empty?

core-common/src/main/java/org/zalando/nakadi/annotations/validation/AnnotationKey.java
```diff
@@ -24,7 +24,7 @@
         message = "Key name should start and end with a letter or a digit")
 @Constraint(validatedBy = {})
 public @interface AnnotationKey {
-    String message() default "{org.zalando.nakadi.annotations.validation.AnnotationKey.message}}";
+    String message() default "{org.zalando.nakadi.annotations.validation.AnnotationKey.message}";
```
a1exsh, line 27: Oh, wow — how did it work before? :)

core-common/src/test/java/org/zalando/nakadi/annotations/validation/DataLakeAnnotationsTest.java
tor-vs-floki: That is a good tip. I've provided better messages from the validator and extra use cases as well.

core-common/src/main/java/org/zalando/nakadi/annotations/validation/AnnotationKey.java
tor-vs-floki: Apparently this code is never reached as message is overwrite by Pattern and NotEmpty annotations

core-common/src/test/java/org/zalando/nakadi/annotations/validation/DataLakeAnnotationsTest.java
```diff
@@ -39,7 +42,8 @@ public void whenRetentionPeriodThenRetentionReasonRequired() {
                 DataLakeAnnotationValidator.RETENTION_PERIOD_ANNOTATION, "1 day"
         );
         final Set<ConstraintViolation<TestClass>> result = validator.validate(new TestClass(annotations));
-        assertFalse("Missing retention reason is treated valid", result.isEmpty());
+        assertTrue(result.stream().anyMatch(r -> r.getMessage().equals("Retention reason is required, when " +
```
a1exsh, line 40: Ideally, a failing assertion would report the actual vs. expected values, but given the data types here I see it may be not easy to express.

a1exsh:
:+1:

core-common/src/test/java/org/zalando/nakadi/annotations/validation/DataLakeAnnotationsTest.java
tor-vs-floki: I failed to find an assert utility that out of the box could inspect a set and test for a value while giving clear instruction whether something is found or nothing is found at all.

tor-vs-floki:
👍

PR #1510, opened 2023-04-13
Persist annotations/labels on subscription update

samizzy:
# One-line summary

In subscription update API, when performing update if the new subscription had annotation/label value set then they were ignored. This PR fixes this and adds a test case to verify.


tor-vs-floki:
👍

samizzy:
:+1:

PR #1511, opened 2023-04-20
Record Kafka consumer create rate

a1exsh:
Related to on-going investigation of publishing latency.


core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaFactory.java
```diff
@@ -151,6 +157,10 @@ public void terminateProducer(final Producer<byte[], byte[]> producer) {
     }
 
     public Consumer<byte[], byte[]> getConsumer(final String clientId) {
+
+        return getConsumer();
```
a1exsh, line 161: First we want to see the rate before introducing the pool — will change it back in the next PR.

adyach:
👍

a1exsh:
:+1:

PR #1514, opened 2023-04-24
Use Kafka consumer pool always, if it is there

a1exsh:
1. For misc we always want to use the consumer pool to reduce the rate of creating consumers.
2. For the actual consumption stack we will be creating consumers normally.


a1exsh:
:+1:

adyach:
it is unexpected that pool can be `0` size, but making decision to create consumer or take from the pool at the timelag checker level requires a lot of code changes, os we agreed to remake it if it has positive effect.

adyach:
👍

PR #1521, opened 2023-05-11
Remove reassign() from TopicRepository interface

a1exsh:
There is no reason for this method to be there.


adyach:
@a1exsh can you scope the changes only related to the claimed in the description and move others in another PR? thanks

PR #1523, opened 2023-05-24
POC for Selective consumption for a  subscription id 

samizzy:
## Description
The PR allows a publisher to publish events to a given event type using the standard publishing API and allows only particular consumers with a specific subscription id to consume these events. The publishing api has been slightly modified to accept a query param `subscription-id` for this purpose.

In case of consumers using the deprecated consumption api they dont receive any events that are published using the above API.

A few sentences describing the overall goals of the pull request's
commits.

## Review
- [ X ] Tests 
- [ X ] Documentation

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/domain/HeaderTag.java
+++ b/core-common/src/main/java/org/zalando/nakadi/domain/HeaderTag.java
@@ -9,14 +9,14 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public enum HeaderTag {
-    SUBSCRIPTION_ID;
+    CONSUMER_SUBSCRIPTION_ID;
 
     private static final Map<String, HeaderTag> STRING_TO_ENUM = HeaderTag.
             stream().
             collect(Collectors.toMap(HeaderTag::name, Function.identity()));
 
-    public static Optional<HeaderTag> fromString(final String consumerTag){
-        return Optional.ofNullable(STRING_TO_ENUM.get(consumerTag));
+    public static Optional<HeaderTag> fromString(final String headerTag){
+        return Optional.ofNullable(STRING_TO_ENUM.get(headerTag.toUpperCase()));
     }
 
     private static Stream<HeaderTag> stream() {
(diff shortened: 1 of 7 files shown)


a1exsh:
:+1:

core-common/src/main/java/org/zalando/nakadi/domain/ConsumerTag.java
```diff
@@ -0,0 +1,21 @@
+package org.zalando.nakadi.domain;
+
+
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+public enum ConsumerTag {
+    SUBSCRIPTION_ID;
```
1u0, line 9: **Meta:** I think it would be more general to support other (potential future) use cases in this enum (and the `KafkaConsumerTagSerializer` helper class) not limited to "consumer tag" semantics but to general "message tag/header".

With that in mind, maybe rename the enum (and the helper class) to smt. like `(Message)HeaderTag`? The `SUBSCRIPTION_ID` would become more specific then, like `CONSUMER_SUBSCRIPTION_ID`.

core-common/src/main/java/org/zalando/nakadi/domain/HeaderTag.java
```diff
@@ -0,0 +1,25 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public enum HeaderTag {
+    CONSUMER_SUBSCRIPTION_ID;
+
+    private static final Map<String, HeaderTag> STRING_TO_ENUM = HeaderTag.
+            stream().
```
1u0, line 15: Minor: this can be inlined (the `stream()` method won't be used then)
```suggestion
    private static final Map<String, HeaderTag> STRING_TO_ENUM = Arrays.stream(values()).
```

1u0 (commented):
:+1:

samizzy:
:+1:

a1exsh:
:+1:

PR #1520, opened 2023-05-10
multi-timeline consumer pool 

adyach:
# One-line summary
the commit creates a pool of nakadi multi-timeline consumers to avoid excessive heap allocations on Kafka

that was especially possible because of prio refactoring https://github.com/zalando/nakadi/pull/1516 and https://github.com/zalando/nakadi/pull/1517

PR #1524, opened 2023-05-24
Revert "multi-timeline consumer pool"

a1exsh:
Reverts zalando/nakadi#1520

The solution is incomplete due to consumers being created often in the middle layer for checking earlies/newest offsets within partition.  We will have to find a different approach, but for now stick with the consumer pool at the bottom layer that covers every use.


adyach:
👍

a1exsh:
:+1:

PR #1525, opened 2023-05-24
Implement optional settings that enable Nakadi security

tor-vs-floki:
Introduce `SECURITY_PROTOCOL`, `SASL_MECHANISM`, `KAFKA_USERNAME` and `KAFKA_PASSWORD` Nakadi settings that will allow Nakadi to configure Kafka clients with security SASL Plain username + password.

Nakadi discovers the list of Brokers by looking in all registered storages entries in Zookeeper, and iterates `/brokers/ids` to discover every broker IP address and Port. This PR introduces `PREFERRED_LISTENER_PORT` setting that allow Nakadi to connect to different listeners in Kafka(previously would connect to the port available in Zookeeper entry, which is the first port of all available listeners, and also an internal Zookeeper structure).

### DIFF
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
@@ -77,6 +77,8 @@ public class KafkaRepositoryAT extends BaseAT {
     private static final long DEFAULT_CURATOR_MAX_LIFETIME_MS = 1000;
     private static final long DEFAULT_CURATOR_ROTATION_MS = 10000;
     private static final int TCP_SEND_BUFFER_SIZE = 128 * 1024;
+    private static final int PLAIN_TEXT_SERVER_PORT = 9092;
+    private static final int SASL_SERVER_PORT = 9093;
 
     private NakadiSettings nakadiSettings;
     private KafkaSettings kafkaSettings;
@@ -113,7 +115,9 @@ public class KafkaRepositoryAT extends BaseAT {
         kafkaSettings = new KafkaSettings(KAFKA_RETRIES, KAFKA_REQUEST_TIMEOUT, KAFKA_BATCH_SIZE, KAFKA_BUFFER_MEMORY,
                 KAFKA_LINGER_MS, KAFKA_MAX_IN_FLIGHT_REQUESTS_PER_CONNECTION, KAFKA_ENABLE_AUTO_COMMIT,
                 KAFKA_MAX_REQUEST_SIZE, KAFKA_DELIVERY_TIMEOUT, KAFKA_MAX_BLOCK_TIMEOUT, "",
-                KAFKA_COMPRESSION_TYPE, TCP_SEND_BUFFER_SIZE);
+                KAFKA_COMPRESSION_TYPE, TCP_SEND_BUFFER_SIZE, Optional.of(9093),
+                Optional.of("SASL_PLAINTEXT"), Optional.of("PLAIN"), Optional.of("nakadi"),
+                Optional.of("nakadi_password"));
         kafkaHelper = new KafkaTestHelper(KAFKA_URL);
         defaultTopicConfig = new NakadiTopicConfig(DEFAULT_PARTITION_COUNT, DEFAULT_CLEANUP_POLICY,
                 Optional.of(DEFAULT_RETENTION_TIME));
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
@@ -77,8 +77,6 @@ public class KafkaRepositoryAT extends BaseAT {
     private static final long DEFAULT_CURATOR_MAX_LIFETIME_MS = 1000;
     private static final long DEFAULT_CURATOR_ROTATION_MS = 10000;
     private static final int TCP_SEND_BUFFER_SIZE = 128 * 1024;
-    private static final int PLAIN_TEXT_SERVER_PORT = 9092;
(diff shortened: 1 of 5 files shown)


acceptance-test/src/acceptance-test/java/org/zalando/nakadi/repository/kafka/KafkaRepositoryAT.java
```diff
@@ -77,6 +77,8 @@ public class KafkaRepositoryAT extends BaseAT {
     private static final long DEFAULT_CURATOR_MAX_LIFETIME_MS = 1000;
     private static final long DEFAULT_CURATOR_ROTATION_MS = 10000;
     private static final int TCP_SEND_BUFFER_SIZE = 128 * 1024;
+    private static final int PLAIN_TEXT_SERVER_PORT = 9092;
```
adyach, line 80: it looks like it is not used

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -35,11 +36,31 @@ public KafkaLocationManager(final ZooKeeperHolder zkFactory, final KafkaSettings
...
         this.scheduledExecutor = Executors.newSingleThreadScheduledExecutor();
         this.scheduledExecutor.scheduleAtFixedRate(() -> updateBootstrapServersSafe(false), 1, 1, TimeUnit.MINUTES);
     }
 
+    private void applySecurityProperties() {
+        if (this.kafkaSettings.getSecurityProtocol().isPresent()) {
+            this.kafkaProperties.put(CommonClientConfigs.SECURITY_PROTOCOL_CONFIG,
+                    this.kafkaSettings.getSecurityProtocol().get());
+        }
+        if (this.kafkaSettings.getSaslMechanism().isPresent()) {
```
adyach, line 50: I assume the configuration should be applied as a whole (protocol, mechanism, users, passwords) or nothing should be applied. otherwise it will be in inconsistent state.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -35,11 +36,31 @@ public KafkaLocationManager(final ZooKeeperHolder zkFactory, final KafkaSettings
...
+                    this.kafkaSettings.getSaslMechanism().get());
+        }
+        if (this.kafkaSettings.getKafkaUsername().isPresent() &&
+                this.kafkaSettings.getKafkaPassword().isPresent()) {
+            this.kafkaProperties.put(SaslConfigs.SASL_JAAS_CONFIG,
+                    "org.apache.kafka.common.security.plain.PlainLoginModule required " +
+                            "username=" + this.kafkaSettings.getKafkaUsername().get()
+                            + " password=\"" + this.kafkaSettings.getKafkaPassword().get() + "\";");
+        }
+        LOG.info("Kafka security settings" + this.kafkaProperties.toString());
```
adyach, line 61: please, do not set user/password to the logs

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaSettings.java
```diff
@@ -26,7 +28,11 @@ public class KafkaSettings {
     private final String clientRack;
     private final String compressionType;
     private final int socketSendBufferBytes;
-
+    private final Optional<Integer> preferredListenerPort;
```
adyach, line 31: maybe lets call it sslPort, I assume that the whole purpose of the change and it is correlated with other params.

adyach: or some other name which indicates relation to ssl config

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -68,7 +89,12 @@ private void updateBootstrapServers(final boolean createWatcher) {
             }
             for (final String brokerId : childrenBuilder.forPath(BROKERS_IDS_PATH)) {
                 final byte[] brokerData = curator.getData().forPath(BROKERS_IDS_PATH + "/" + brokerId);
-                brokers.add(Broker.fromByteJson(brokerData));
+                if (this.kafkaSettings.getPreferredListenerPort().isPresent()) {
```
adyach, line 92: to avoid duplication and make it even better to read we could do:
`brokers.add(Broker.fromByteJson(brokerData, kafkaSettings.getPreferredListenerPort()));`

and  `Broker.fromByteJson` looks like 
```
        static Broker fromByteJson(final byte[] data, final Optional<Integer> portOpt)
....
            return new Broker(host, portOpt.orElse(json.getInt("port")));
        }
```

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
tor-vs-floki: I like the idea that uncompleted security configuration wont break Nakadi, and in this case I will be verbal in logs to explain that a missing piece of configuration wont enable security to instruct the engineer that some part of configuration is missing.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
tor-vs-floki: Ouch

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaSettings.java
tor-vs-floki: I hope that this configuration value will be used only for a listener that has some security, but in the notion of configuring the Kafka client it is still only a port, therefore I think it is more precise to not specify what kind of security is configured on the Broker side

adyach:
👍

tor-vs-floki:
👍

PR #1528, opened 2023-06-14
Update zookeeper client

1u0:
# One-line summary

## Description
Update `org.apache.zookeeper` dependency to `3.6.1` (to reduce noisy logging) which includes fix for:
https://issues.apache.org/jira/browse/ZOOKEEPER-4376

## Deployment Notes
This change should be compatible with the Zookeeper server `3.6.1`.

### DIFF
--- a/api-metastore/src/main/java/org/zalando/nakadi/model/AuthorizationAttributeQueryParser.java
+++ b/api-metastore/src/main/java/org/zalando/nakadi/model/AuthorizationAttributeQueryParser.java
@@ -1,6 +1,6 @@
 package org.zalando.nakadi.model;
 
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
 import org.zalando.nakadi.domain.ResourceAuthorizationAttribute;
 import org.zalando.nakadi.plugin.api.authz.AuthorizationAttribute;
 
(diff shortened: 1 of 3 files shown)


api-metastore/src/main/java/org/zalando/nakadi/model/AuthorizationAttributeQueryParser.java
```diff
@@ -1,6 +1,6 @@
 package org.zalando.nakadi.model;
 
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
```
a1exsh, line 3: Is it a compile- or a run-time error if we don't change this?

a1exsh: I notice we are already using a mix of the two (in different sub-projects):
```
import org.apache.commons.lang.StringUtils;
import org.apache.commons.lang3.StringUtils;
```

That's not good.  We should fix it, but I'd prefer to do that separately from bumping ZooKeeper version.

a1exsh: Ah, I see now that these are the only two uses of `lang` that are left there. OK.

a1exsh:
:+1:

api-metastore/src/main/java/org/zalando/nakadi/model/AuthorizationAttributeQueryParser.java
1u0: That would be compile error, so the first commit is actually per-requisite for the version bump.
The updated Zookeeper client version is not using the old `org.apache.commons` lib anymore. The Nakadi project was using it as transitive dependency.

samizzy:
:+1:

rishiraj88:
Good PR.

PR #1529, opened 2023-06-16
Add logging for ISO-4217 currency format checks

1u0:
# One-line summary

## Description
This PR adds logging for ISO-4217 (currencies) code checks during publishing events according to their event schemas.
Note: the validation is not enforced (for now), but just logging for observation for now.
To avoid too much logging (in case of violations), one can use `log4j` filtering (for example using [BurstFilter](https://logging.apache.org/log4j/2.x/manual/filters.html#burstfilter)).

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/validation/ISO4217CurrencyValidator.java
+++ b/core-common/src/main/java/org/zalando/nakadi/validation/ISO4217CurrencyValidator.java
@@ -1,0 +1,40 @@
+package org.zalando.nakadi.validation;
+
+import org.everit.json.schema.FormatValidator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Currency;
+import java.util.Optional;
+
+public class ISO4217CurrencyValidator implements FormatValidator {
+    private static final Logger LOG = LoggerFactory.getLogger(ISO4217CurrencyValidator.class);
+    private final String eventTypeName;
+    private final String formatName;
+
+    ISO4217CurrencyValidator(final String eventTypeName) {
+        this(eventTypeName, "iso-4217");
+    }
+
+    ISO4217CurrencyValidator(final String eventTypeName, final String formatName) {
+        this.eventTypeName = eventTypeName;
+        this.formatName = formatName;
+    }
+
+    @Override
+    public Optional<String> validate(final String value) {
+        try {
+            Currency.getInstance(value);
+        } catch (IllegalArgumentException ex) {
+            // Note: not returning this as validation error (yet), but just log it.
+            LOG.warn("Currency format violation (format_name={}): [event_type={}, value={}]",
+                    formatName, eventTypeName, value);
+        }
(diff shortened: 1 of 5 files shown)


core-common/src/main/java/org/zalando/nakadi/validation/ISO4217CurrencyValidator.java
```diff
@@ -0,0 +1,40 @@
+package org.zalando.nakadi.validation;
+
+import org.everit.json.schema.FormatValidator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.Currency;
+import java.util.Optional;
+
+public class ISO4217CurrencyValidator implements FormatValidator {
```
adyach, line 10: lets be explicit that it is not currency validation, but rather currency code validation
```suggestion
public class ISO4217CurrencyCodeValidator implements FormatValidator {
```

core-common/src/main/java/org/zalando/nakadi/validation/ISO4217CurrencyValidator.java
```diff
@@ -0,0 +1,40 @@
...
+    private static final Logger LOG = LoggerFactory.getLogger(ISO4217CurrencyValidator.class);
+    private final String eventTypeName;
+    private final String formatName;
+
+    ISO4217CurrencyValidator(final String eventTypeName) {
+        this(eventTypeName, "iso-4217");
+    }
+
+    ISO4217CurrencyValidator(final String eventTypeName, final String formatName) {
```
adyach, line 19: why do we need to pass format name when we have named validator as `ISO4217`CurrencyValidator. Lets keep it in the name of the class only.

adyach:
I am fine to go just with the validation of the current code, but it would be great to test that this validator is picked up and takes correct value from the JSON object.

core-common/src/main/java/org/zalando/nakadi/validation/ISO4217CurrencyValidator.java
1u0: Because there are at least two format names that seen in practice: `iso-4217` and `ISO-4217`. Hence I'm registering both of them (the `org.everit.json` lib, used for schema validation in the project is case sensitive for format names).

I'm open where to keep the `ISO-4217` constant string.
However, this is not the "approved" format name according to https://opensource.zalando.com/restful-api-guidelines/#238. For now, I hope this is rather a transitional period to have one validator under different format names.

In the future, either Nakadi should normalize the format names (transform to lower case) under the hood (when creating validation objects per event type schema), or alternatively, this can be fixed at data level (by updating event types schemas in the schema registry).
The system also could be more stricter to what it accepts as format names (i.e., validate more stricter the event type schemas).

tor-vs-floki:
👍

samizzy:
:+1:

PR #1530, opened 2023-06-27
Relax Kafka producer reset cases

a1exsh:
Neither `NotLeaderOrFollower` nor `Timeout` is a good reason for resetting the producer.

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
+++ b/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
@@ -18,7 +18,6 @@ import org.apache.kafka.common.TopicPartition;
 import org.apache.kafka.common.config.ConfigResource;
 import org.apache.kafka.common.errors.InterruptException;
 import org.apache.kafka.common.errors.NetworkException;
-import org.apache.kafka.common.errors.NotLeaderForPartitionException;
 import org.apache.kafka.common.errors.TopicExistsException;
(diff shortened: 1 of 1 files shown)


Jan-M:
:+1:

1u0:
:+1:

1u0:
:+1:

1u0:
:+1:

1u0:
:+1:

1u0:
:+1:

1u0:
:+1:

adyach:
I agree those exceptions should not recreate producer, but it is not clear how you decided on what should be left. 
For example `NotLeaderForPartitionException`, `NetworkException` do not have to trigger producer recreation either. Those are transient errors, producer will refresh metadata and the issue will be resolved. 
The left exception is only `UnknownServerException. The name suggest (and code as well) that it is not producer side error which means there is no reason to recreate the producer, it simply won't help, because the error is on the server side.

I do not completely remember why the list of the exceptions is like that, that's why I suggest to remove that section completely and validate that it works as expected.

a1exsh:
Eventually we may remove it completely, as I also don't see any specific reason for closing exiting producer.  The documentation mentions one or two cases where this may be necessary, when working with transactions or using idempotent producer, which is not the case for us.

At the same time, I suppose the special handling there for a reason, and for now I have removed the ones that I could test locally and have seen that the producer can be used further after simply catching one of them.

a1exsh:
:+1:

PR #1534, opened 2023-07-05
evict cache value only if load is successful

adyach:
at the moment if value was evicted and load from database fails cache wont have a value, that will create unavailability on the publishing path. the commit changes that behaviour to have value ecivted from the cache only in case the value was successfully loaded


a1exsh:
:+1:

Jan-M:
:+1:

PR #1536, opened 2023-07-10
Do not close producers on specific errors

a1exsh:
Not removing producers means we do not explicitly close them, so we don't need all the locking code around take/terminate/release in `KafkaFactory`.

Note the change in `KafkaTopicRepository::listPartitionNames`: use admin client instead of some producer, because we need the up-to-date result, not some cached value.  This call is not used in hot path, as the partitions list is storead in the `EventTypeCache`.  Also when invalidating that cache entry we need to fetch an up-to-date value.

### DIFF
--- a/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/utils/NakadiTestUtils.java
+++ b/acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/utils/NakadiTestUtils.java
@@ -111,7 +111,9 @@ public class NakadiTestUtils {
         given()
                 .body("[" + events + "]")
                 .contentType(JSON)
-                .post(path);
+                .post(path)
+                .then()
+                .statusCode(HttpStatus.SC_OK);
     }
 
     public static void repartitionEventType(final EventType eventType, final int partitionsNumber)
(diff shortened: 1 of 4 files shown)


acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/utils/NakadiTestUtils.java
```diff
@@ -111,7 +111,9 @@ public static void processEvents(final String path, final int count, final IntFu
         given()
                 .body("[" + events + "]")
                 .contentType(JSON)
-                .post(path);
+                .post(path)
+                .then()
+                .statusCode(HttpStatus.SC_OK);
```
a1exsh, line 116: Not ignoring the status code here helps to spot wrong expectations in the tests instead of trying to figure out why we are consuming less events that we have (presumably) published.

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaTopicRepository.java
```diff
@@ -227,9 +214,6 @@ public void repartition(final String topic, final int partitionsNumber) throws C
             if (!Boolean.TRUE.equals(areNewPartitionsAdded)) {
                 throw new TopicConfigException(String.format("Failed to repartition topic to %s", partitionsNumber));
             }
-            final Producer<byte[], byte[]> producer = kafkaFactory.takeProducer(topic);
-            kafkaFactory.terminateProducer(producer);
-            kafkaFactory.releaseProducer(producer);
```
a1exsh, line 232: This code was here only to make re-partitioning tests pass.

The proper fix is not to use a producer when listing partitions.

adyach:
👍

PR #1537, opened 2023-07-11
allow fallback to remote on IO error

samizzy:
## Description
In Nakadi, there is logic to retry load token info first from local token and then fallback to remote token info on failure. This logic does not account for when the local token info fails with timeout/IO exception. This PR introduces fix so that it accounts for IO exception and retries remote token info.

## Review
- [x] Tests

### DIFF
--- a/app/src/main/java/org/zalando/nakadi/config/AuthenticationConfig.java
+++ b/app/src/main/java/org/zalando/nakadi/config/AuthenticationConfig.java
@@ -22,6 +22,7 @@ import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
 import org.springframework.security.oauth2.provider.token.ResourceServerTokenServices;
 import org.springframework.util.Assert;
 import org.springframework.web.client.HttpStatusCodeException;
+import org.springframework.web.client.ResourceAccessException;
 import org.springframework.web.client.RestTemplate;
 import org.zalando.nakadi.exceptions.runtime.ServiceTemporarilyUnavailableException;
 import org.zalando.nakadi.service.FeatureToggleService;
@@ -154,7 +155,7 @@ public class AuthenticationConfig {
                     .header(HttpHeaders.AUTHORIZATION, String.format(BEARER_TOKEN_TEMPLATE, accessToken)).build();
             try {
                 return restTemplate.exchange(entity, TOKENINFO_MAP).getBody();
-            } catch (HttpStatusCodeException e) {
+            } catch (HttpStatusCodeException | ResourceAccessException e ) {
                 throw new ServiceTemporarilyUnavailableException("Unable to validate token", e);
             } catch (Exception e) {
                 LOGGER.warn("Unable to get authorisation from tokeninfo", e);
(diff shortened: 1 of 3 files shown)


adyach:
👍 lets find other way to achieve the same

samizzy:
@adyach Open to suggestions, do you have something in mind?
Just for context, the interface `TokenInfoRequestExecutor` (snippet below) doesnt allow to throw exceptions and every implementation class returns the error with key `error`. So there is no possibility to catch IO exceptions as such.  The caller of this method `getMap` is `TokenInfoResourceServerTokenServices#loadAuthentication` which throws all error messages with `InvalidTokenException`


```
public interface TokenInfoRequestExecutor {

    Map<String, Object> getMap(final String accessToken);

}
```

adyach:
from the quick look there are many options:
1) TokenInfoResourceServerTokenServices is a part of the Zalando libraries, it could be a way to improve overall
2) You can always extend TokenInfoResourceServerTokenServices and change the logic (yes, some duplication of the code)
3) There is also extension NakadiTokenInfoRequestExecutor which overrides `getMap`, if there is not other option I would go  with adding more sounding error description to catch.
...

adyach:
and to the why I think we should do it in the other way:
1. the error message can change in the future, and the team maintaining Nakadi will never know that part of code does not work anymore.
2. `IO error` can be due to different reasons, one of them is timeout. It potentially servers for other errors which you are not aware of. This will increase the load on other endpoints.
3. This code snippet to check the error does not look like production grade code, it is more of the hack or test code, which I expect to go away after some testing

```
if(!ex.getMessage().contains("I/O error")) {
     throw ex;
 }
```

samizzy:
> from the quick look there are many options:
> 
> 1. TokenInfoResourceServerTokenServices is a part of the Zalando libraries, it could be a way to improve overall
> 2. You can always extend TokenInfoResourceServerTokenServices and change the logic (yes, some duplication of the code)
> 3. There is also extension NakadiTokenInfoRequestExecutor which overrides `getMap`, if there is not other option I would go  with adding more sounding error description to catch.
>    ...

I have gone with suggestion 3 here, although its still catching general IOException.

app/src/main/java/org/zalando/nakadi/config/AuthenticationConfig.java
```diff
@@ -154,7 +155,7 @@ public Map<String, Object> getMap(final String accessToken) {
                     .header(HttpHeaders.AUTHORIZATION, String.format(BEARER_TOKEN_TEMPLATE, accessToken)).build();
             try {
                 return restTemplate.exchange(entity, TOKENINFO_MAP).getBody();
-            } catch (HttpStatusCodeException e) {
+            } catch (HttpStatusCodeException | ResourceAccessException e ) {
```
samizzy, line 158: from what I can see, `ResourceAccessException` is thrown when `RestTemplate` encounters a IOException

app/src/main/java/org/zalando/nakadi/config/NakadiResourceServerTokenServices.java
```diff
@@ -42,9 +42,11 @@ public OAuth2Authentication loadAuthentication(final String accessToken) throws
             if (!featureToggleService.isFeatureEnabled(Feature.REMOTE_TOKENINFO)) {
                 try {
                     return localService.loadAuthentication(accessToken);
-                } catch (final OAuth2Exception ex) {
+                }
```
adyach, line 45: lets keep formatting the same across the file (in the other part it uses different)

app/src/main/java/org/zalando/nakadi/config/AuthenticationConfig.java
adyach: is there any test to see that it works as expected? current test does not represent that.

adyach:
some minor comments to fix otherwise better than before 👍

adyach:
👍

samizzy:
:+1:

PR #1538, opened 2023-07-13
Fix retry in `listPartitionNames`

a1exsh:
When a topic is just created, describing it immediately may fail with `UnknownTopicOrPartitionException` (wrapped in an `ExecutionException` due to use of futures).


adyach:
it looks fine, maybe we should check if some parts of the code relay on the order which is provided by producer partitions method, because that can be changed with admin client approach.

adyach:
👍

a1exsh:
:+1:

PR #1539, opened 2023-07-21
[Preview] Handle `OutOfOrderSequenceException` in Kafka producer

a1exsh:
We've removed the special exception handling recently, but if we are going to enable retries (and use idempotent producer to avoid ordering violations), we need to be prepared to handle this exception.


a1exsh:
After thinking longer, it looks like we do not need special handling of this error after all, since we are not calling `Producer#send()` again with the same records, so no reordering should happen.

PR #1540, opened 2023-07-24
Use idempotent producer once again to enable safe retries

a1exsh:
The main objective is to enable retries, which we can only safely do when using the idempotence feature.

This change only enables us to configure idempotent producer with retries, but doesn't change the current default settings.

### DIFF
--- a/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
+++ b/core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
@@ -144,8 +144,9 @@ public class KafkaLocationManager {
         producerProps.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG,
                 "org.apache.kafka.common.serialization.ByteArraySerializer");
         producerProps.put(ProducerConfig.ACKS_CONFIG, "all");
-        producerProps.put(ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG, true);
 
+        producerProps.put(ProducerConfig.RETRIES_CONFIG, kafkaSettings.getRetries());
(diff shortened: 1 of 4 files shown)


core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
```diff
@@ -144,9 +144,8 @@ public Properties getKafkaProducerProperties() {
         producerProps.put(ProducerConfig.VALUE_SERIALIZER_CLASS_CONFIG,
                 "org.apache.kafka.common.serialization.ByteArraySerializer");
         producerProps.put(ProducerConfig.ACKS_CONFIG, "all");
-        producerProps.put(ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG, false);
+        producerProps.put(ProducerConfig.ENABLE_IDEMPOTENCE_CONFIG, true);
 
-        producerProps.put(ProducerConfig.RETRIES_CONFIG, kafkaSettings.getRetries());
```
1u0, line 149: Side note: I think as an alternative, those (and other) configs could made configurable (and optional, so that if not set externally, they won't be set here and the Kafka client would use the default values). This way would allow to experiment with settings at deployment time wo/ rebuilding the project.

1u0:
:+1:

core-common/src/main/java/org/zalando/nakadi/repository/kafka/KafkaLocationManager.java
a1exsh: You are right.  I'm jumping ahead of myself a bit here: the idea is to reduce the temptation to tune these parameters, once they are tested, but we haven't tested them yet.

a1exsh:
:+1:

1u0:
:+1:

PR #1541, opened 2023-07-26
Enforce alphabetic ISO-4217 currency format validation

1u0:
# One-line summary

Changes `ISO4217CurrencyCodeValidator` to a proper schema (format) validator (instead of logging invalid values) for alphabetic ISO 4217 currency codes.

This is follow up of #1529.

## Description
Goal: reject publishing events with event type schemas' containing `iso-4217` format.

## Review
- [x] Tests

## Deployment Notes

The feature is not feature toggled (there is some inverse dependency of the `core-common` and the sub-project containing `FeatureToggleService`. For simplicity of implementation, this change assumes that there are no publishers violating this (new) restriction.

### DIFF
--- a/core-common/src/test/java/org/zalando/nakadi/validation/ISO4217CurrencyCodeValidatorTest.java
+++ b/core-common/src/test/java/org/zalando/nakadi/validation/ISO4217CurrencyCodeValidatorTest.java
@@ -4,6 +4,7 @@ import org.junit.Test;
 import org.zalando.nakadi.utils.IsOptional;
 
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.containsString;
 
 public class ISO4217CurrencyCodeValidatorTest {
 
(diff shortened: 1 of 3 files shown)


core-common/src/test/java/org/zalando/nakadi/validation/ISO4217CurrencyCodeValidatorTest.java
```diff
@@ -31,10 +32,14 @@ public void testValidator() {
                 "SEK",
         };
 
-        final var validator = new ISO4217CurrencyCodeValidator("an-event-type");
+        final var validator = new ISO4217CurrencyCodeValidator();
         for (final String value : invalidValues) {
             // For invalid values, the validator should return empty result.
```
a1exsh, line 37: Looks like the comment should be removed now?

1u0: I've missed that line.

a1exsh:
:+1:

1u0:
:+1:

Jan-M:
:+1:

PR #1542, opened 2023-07-27
Validate `event_owner_selector` path refers to a field in the schema

a1exsh:
- Validate that the field is present when creating/updating event types
- Log a warning if the field is still missing or `null` when publishing events

Internal ticket: 1431

### DIFF
--- a/core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
+++ b/core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
@@ -239,7 +239,6 @@ public class SchemaService implements SchemaProviderService {
         final EventOwnerSelector eventOwnerSelector = eventType.getEventOwnerSelector();
         if (eventOwnerSelector != null) {
             if (eventOwnerSelector.getType() == EventOwnerSelector.Type.PATH) {
-                // TODO: validate that it's a required property!
                 validateFieldsInSchema("event_owner_selector.value", List.of(eventOwnerSelector.getValue()),
                         effectiveSchema);
             }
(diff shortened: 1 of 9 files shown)


core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -232,14 +233,23 @@ private void validateJsonTypeSchema(final EventTypeBase eventType, final String
             throw new SchemaValidationException("\"metadata\" property is reserved");
         }
 
+        final JSONObject effectiveSchemaAsJson = jsonSchemaEnrichment.effectiveSchema(eventType, eventTypeSchema);
+        final Schema effectiveSchema = SchemaLoader.load(effectiveSchemaAsJson);
+
+        final EventOwnerSelector eventOwnerSelector = eventType.getEventOwnerSelector();
+        if (eventOwnerSelector != null) {
+            if (eventOwnerSelector.getType() == EventOwnerSelector.Type.PATH) {
+                validateFieldsInSchema("event_owner_selector.value", List.of(eventOwnerSelector.getValue()),
```
samizzy, line 242: Since we are not blocking publishing for existing event type, does it also make sense to not block updating for existing event types? maybe we could introduce this for newly created event types and/or event types which are already valid should not become invalid.

a1exsh: In general the migration can be more gradual, but here we know that only 4 event types need to be updated — that should be pretty straightforward.

samizzy:
:+1:

a1exsh:
:+1:

PR #1545, opened 2023-07-31
Return client error on unknown meta schema

1u0:
## Description

Change JSON event-type creation/validation to throw client error (`422`) instead of server error (`5xx`) on unsupported/unknown meta-schema URLs (`"$schema": "..."` in schema-JSON).

## Review
- [x] Tests

### DIFF
--- a/core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
+++ b/core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
@@ -1,5 +1,6 @@
 package org.zalando.nakadi.service;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.cache.CacheLoader;
 import com.google.common.cache.LoadingCache;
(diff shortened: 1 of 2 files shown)


core-services/src/main/java/org/zalando/nakadi/service/SchemaService.java
```diff
@@ -288,9 +293,10 @@ public InputStream get(final String ref) throws SchemaValidationException {
         }
     }
 
-    public static void isStrictlyValidJson(final String jsonInString) throws InvalidEventTypeException {
+    @VisibleForTesting
+    public static JSONObject parseJsonSchema(final String jsonInString) throws InvalidEventTypeException {
```
a1exsh, line 297:
```suggestion
    private static JSONObject parseJsonSchema(final String jsonInString) throws InvalidEventTypeException {
```
Did you mean to make it private?

1u0: > 1. Did you mean to make it private?

This is still called in tests. At most I have changed it to package private (otherwise, tests needs to be rewritten too).


> 2. Looks like `throws` part is outdated.

Good point!

a1exsh:
@1u0 needs a checkstyle fix (unused import of `InvalidEventTypeException`) and then it's good to go.

a1exsh:
:+1:

1u0:
:+1:

PR #1546, opened 2023-08-07
simplify logic to verify subscription tag

adyach:
# One-line summary

simplify logic to verify subscription tag

### DIFF
--- a/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
+++ b/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
@@ -36,6 +36,8 @@ import java.io.IOException;
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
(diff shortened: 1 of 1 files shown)


api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
```diff
@@ -36,6 +36,8 @@
 import java.util.Collection;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Objects;
+import java.util.Optional;
```
a1exsh, line 40:
```suggestion
```
Checkstyle thinks these are unused.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
```diff
@@ -277,19 +279,12 @@ public boolean isSubscriptionConsumptionBlocked() {
     }
 
     public boolean isConsumptionBlocked(final ConsumedEvent event) {
```
a1exsh, line 281: The new code seems incorrect:

1. We should check `isConsumptionBlocked(event)`, even if the event has the matching subscription id tag.
2. If I read it correctly, the new code has the opposite effect — events with subscription id tag are given out to all subscriptions, except the one which should have seen it.

It may be helpful to name the function differently to make it more clear what it should accomplish:
```suggestion
    public boolean shouldSkipEvent(final ConsumedEvent event) {
```

I'm also noticing now that the low-level consumer should check more than just presence of any custom tags to decide if the event should be skipped (namely, it should check the presence of the subscription id tag).

PR #1547, opened 2023-08-08
add test case for selective consumption wrt avro

samizzy:
## Description
Added some test cases to verify that selective consumption of published events also works when using Avro for publishing.


adyach:
👍

samizzy:
:+1:

PR #1550, opened 2023-08-31
Add code for skipping events that do not match intended event type

samizzy:
Add functionality to skip events while consuming that do match the event types found in subscription.

### DIFF
--- a/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
+++ b/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
@@ -280,14 +280,21 @@ public class StreamingContext implements SubscriptionStreamer {
     }
 
     public boolean isConsumptionBlocked(final ConsumedEvent event) {
-        final var metadataEventTypeName = kafkaRecordDeserializer.getEventTypeName(event.getEvent());
-        if(!event.getPosition().getEventType().contains(metadataEventTypeName)){
-            LOG.warn("Found unexpected event for event type: {} " +
-                            "but expected {} having offset {}. Event timestamp: {}, Source topic:  {}",
-                    metadataEventTypeName, event.getPosition().getEventType(), event.getPosition(),
-                    event.getTimestamp(),
-                    event.getPosition().getTimeline().getEventType());
-            return true;
+
+        try {
+            final String metadataEventTypeName = kafkaRecordDeserializer.getEventTypeName(event.getEvent());
+            if(!event.getPosition().getEventType().contains(metadataEventTypeName)){
+                LOG.warn("Found unexpected event for event type: {} " +
+                                "but expected {} having offset {}. Event timestamp: {}, Source topic:  {}",
+                        metadataEventTypeName, event.getPosition().getEventType(), event.getPosition(),
+                        event.getTimestamp(),
+                        event.getPosition().getTimeline().getEventType());
+                return true;
+            }
+        } catch (final IOException e) {
+            //this case shouldnt happen
+            LOG.error("Failed to parse metadata", e);
+            return false;
         }
 
         if (event.getConsumerTags().isEmpty()) {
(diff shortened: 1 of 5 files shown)


api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
```diff
@@ -277,10 +280,17 @@ public boolean isSubscriptionConsumptionBlocked() {
     }
 
     public boolean isConsumptionBlocked(final ConsumedEvent event) {
+        final var metadataEventTypeName = kafkaRecordDeserializer.getEventTypeName(event.getEvent());
+        if(!subscription.getEventTypes().contains(metadataEventTypeName)){
```
adyach, line 284: ConsumedEvent has cursor, cursor should have event type. lets use it.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
```diff
@@ -277,10 +280,17 @@ public boolean isSubscriptionConsumptionBlocked() {
     }
 
     public boolean isConsumptionBlocked(final ConsumedEvent event) {
+        final var metadataEventTypeName = kafkaRecordDeserializer.getEventTypeName(event.getEvent());
```
adyach, line 283: would it be possible to utilise parser from BatchItem to parse only event_type from metadata?

samizzy: BatchItem unfortunately does not offer such functionality, it either parsers strictly or non strictly from what I see.

samizzy: Since org.json lib didnt have such feature, I have used Jackson instead. Will only parse metadata

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/StreamingContext.java
```diff
@@ -277,10 +284,28 @@ public boolean isSubscriptionConsumptionBlocked() {
...
+                                    "but expected {} having offset {}. Event timestamp: {}, Source topic:  {}",
+                            metadataEventTypeName, event.getPosition().getEventType(), event.getPosition(),
+                            event.getTimestamp(),
+                            event.getPosition().getTimeline().getEventType());
+                    return true;
+                }
+            } catch (final IOException e) {
+                //this case shouldnt happen
+                LOG.error("Failed to parse metadata", e);
+                return false;
```
a1exsh, line 302: I think it's more appropriate to let this exception propagate and be handled elsewhere.  Neither `true` nor `false` are satisfying here as return value.

a1exsh:
:+1:

core-services/src/main/java/org/zalando/nakadi/repository/kafka/KafkaRecordDeserializer.java
```diff
@@ -93,4 +116,10 @@ private JSONObject mapToJson(final Metadata metadata) {
         }
         return metadataObj;
     }
+
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    private static class MetadataHolder {
+        @JsonSetter("metadata")
+        private JsonNode metadata;
+    }
```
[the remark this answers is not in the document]
fghibellini: I tried to look into the jackson databind source and it would take me too long to figure out exactly how the parsing works but [this piece of code](https://github.com/FasterXML/jackson-databind/blob/9d18eb057b692819b7dc40f87740cbc8707f89a2/src/main/java/com/fasterxml/jackson/databind/deser/std/StdNodeBasedDeserializer.java#L98) doesn't give me much hope that it bears any benefit compared to `.readTree()`.

If we strictly wanted to be sure that we don't allocate or parse things unnecessarily we could use the Jackson streaming API which is used by databind under the hood ([example code](http://www.cowtowncoder.com/blog/archives/2009/01/entry_132.html)), but I think it's a bit of an overkill.

fghibellini:
👍

1u0:
:+1:

Jan-M:
:+1:

PR #1553, opened 2023-10-04
Add `number_of_events` tag to `publish_events` spans

a1exsh:
To allow the publishers to learn about their batch size distribution.

Using the same name as in `nakadi.batch.published` event: https://github.com/zalando/nakadi/blob/master/core-common/src/main/resources/avro-schema/nakadi.batch.published.avsc#L24

### DIFF
--- a/core-services/src/main/java/org/zalando/nakadi/service/publishing/BinaryEventPublisher.java
+++ b/core-services/src/main/java/org/zalando/nakadi/service/publishing/BinaryEventPublisher.java
@@ -82,6 +82,9 @@ public class BinaryEventPublisher {
                                                      final List<NakadiRecord> records,
                                                      final List<Check> checks,
                                                      final Map<HeaderTag, String> consumerTags) {
+
+        TracingService.setTag("number_of_events", String.valueOf(records.size()));
+
         for (final Check check : checks) {
             final List<NakadiRecordResult> res = check.execute(eventType, records);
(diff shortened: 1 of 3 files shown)


core-services/src/main/java/org/zalando/nakadi/service/publishing/BinaryEventPublisher.java
```diff
@@ -82,6 +82,9 @@ private List<NakadiRecordResult> processInternal(final EventType eventType,
                                                      final List<NakadiRecord> records,
                                                      final List<Check> checks,
                                                      final Map<HeaderTag, String> consumerTags) {
+
+        TracingService.setTag("number_of_events", String.valueOf(records.size()));
```
1u0, line 86: There is `.setTag(String, Number)` method on `Span` - this might result in less memory footprint to store this tag data.

```suggestion
        TracingService.getActiveSpan().setTag("number_of_events", records.size());
```

a1exsh: Good point.  Actually, the value has to be numeric for users to be able to use it in a query later.  I just overlooked the Span's methods.

1u0:
:+1:

a1exsh:
:+1:

1u0:
:+1:

PR #1555, opened 2023-10-06
Move setting of `number_of_events` tag to controller

a1exsh:
Turns out we don't always have the span in the binary publisher (due to it being used for internal KPI events publishing).


1u0:
:+1:

a1exsh:
:+1:

PR #1556, opened 2023-10-11
store failed commit on exit

adyach:
the fix ensures the closing state does not lose failed commits in case the stream is closed before commit timeout reached


adyach:
👍

PR #1558, opened 2023-10-12
Add periodic exclusive job to attach DLQ redrive event type

a1exsh:
Internal ticket: 1517

### DIFF
--- a/app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
+++ b/app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
@@ -131,14 +131,14 @@ public class DlqRedriveEventTypeAttachmentJob {
                 .updateTopology(topology -> addPartitionsIfMissing(topology.getPartitions(), unassignedDlqPartitions));
     }
 
-    private static Partition[] addPartitionsIfMissing(
+    static Partition[] addPartitionsIfMissing(
             final Partition[] initialPartitions,
             final List<Partition> toAddPartitions) {
 
         final Map<EventTypePartition, Partition> assignment =
                 Arrays.stream(initialPartitions).collect(Collectors.toMap(k -> k.getKey(), v -> v));
         return
-                Stream.concat(assignment.entrySet().stream(),
+                Stream.concat(assignment.values().stream(),
                         toAddPartitions.stream().filter(p -> !assignment.containsKey(p.getKey())))
                 .toArray(Partition[]::new);
     }
--- a/app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
+++ b/app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
@@ -23,6 +23,7 @@ import org.zalando.nakadi.view.Cursor;
 import org.zalando.nakadi.service.subscription.zk.SubscriptionClientFactory;
 import org.zalando.nakadi.service.subscription.zk.ZkSubscriptionClient;
 
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
```diff
@@ -0,0 +1,161 @@
...
+        // idempotent, does not overwrite existing offsets
+        client.createOffsetZNodes(
+                unassignedDlqPartitions.stream()
+                .map(p -> p.getPartition())
+                .map(p -> new Cursor(p, Cursor.BEFORE_OLDEST_OFFSET))
+                .map(c -> cursorConverter.convert(dlqRedriveEventTypeName, c))
+                .map(nc -> cursorConverter.convertToNoToken(nc))
+                .collect(Collectors.toList()));
+
+        client.updateTopology(topology -> selectMissingPartitions(topology.getPartitions(), unassignedDlqPartitions));
```
a1exsh, line 148: If `selectMissingPartitions` returns an empty array, no actual update is done in ZK.

app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
```diff
@@ -0,0 +1,161 @@
...
+
+        final List<String> orderedPartitions = eventTypeCache.getOrderedPartitions(dlqRedriveEventTypeName);
+
+        final List<NakadiCursor> beginCursors = orderedPartitions.stream()
+                .map(p -> new Cursor(p, Cursor.BEFORE_OLDEST_OFFSET))
+                .map(c -> cursorConverter.convert(dlqRedriveEventTypeName, c))
+                .collect(Collectors.toList());
+
+        // it's inefficient to re-consume from begin, but we expect this event type to be empty most of the time anyway
+        final HighLevelConsumer consumer = timelineService.createEventConsumer(JOB_NAME + "-job", beginCursors);
```
adyach, line 95: it leaked, need to close it after usage, please

app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
```diff
@@ -0,0 +1,161 @@
...
+
+        final Set<String> subscriptionIds = new HashSet<>();
+        while (true) {
```
adyach, line 98: I would not do it and rather rely on scheduling mechanism to execute it again. alternatively, you could limit execution by time.

app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
```diff
@@ -0,0 +1,161 @@
...
+            final List<ConsumedEvent> events = consumer.readEvents();
```
adyach, line 99: it will read from begin over and over again because it does not commit offsets. basically, this loop will run forever. it looks like you need to track offsets. I think it is better to reconsider the job approach.

app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    private void addDlqPartitionsToSubscription(
+            final Subscription subscription,
+            final List<Partition> unassignedDlqPartitions) throws Exception {
+
+        final ZkSubscriptionClient client = subscriptionClientFactory.createClient(subscription);
+
+        // idempotent, does not overwrite existing offsets
+        client.createOffsetZNodes(
```
adyach, line 140: maybe we would like to aggregate it under one method in zookeeper client, it looks too low level for a job

app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
a1exsh: > it will read from begin over and over again because it does not commit offsets.

That's the point here, but most of the time the topic is completely empty, so why bother?

> basically, this loop will run forever. it looks like you need to track offsets. I think it is better to reconsider the job approach.

It will not run forever, it will only consume from begin till end on every job run.

adyach: ah, actually, I am wrong, it should work just fine because consumer will manage it.

app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
```diff
@@ -0,0 +1,169 @@
...
+        final List<String> orderedPartitions = eventTypeCache.getOrderedPartitions(dlqRedriveEventTypeName);
+
+        // it's inefficient to re-consume from begin, but we expect this event type to be empty most of the time anyway
+        final List<NakadiCursor> beginCursors = orderedPartitions.stream()
+                .map(p -> new Cursor(p, Cursor.BEFORE_OLDEST_OFFSET))
+                .map(c -> cursorConverter.convert(dlqRedriveEventTypeName, c))
+                .collect(Collectors.toList());
+
+        try (HighLevelConsumer consumer = timelineService.createEventConsumer(JOB_NAME + "-job", beginCursors)) {
+            while (true) {
```
adyach, line 105: it looks like if there events constantly the loop never ends. I would rather rely on scheduling mechanism, but then you have to track and commit offsets. current approach is not sustainable for the future, but I think assumption here is that the topic is almost all the time empty.

a1exsh: yes, the idea is that it almost never has to do actual work.  by the time consuming constantly from begin becomes a problem, we will not have this temporary approach and rather always consume from the DQL redrive, I think

adyach:
👍

PR #1559, opened 2023-10-12
Add support submitting DLQ refeed offsets for subscription api

1u0:
# One-line summary

 * Add support submitting DLQ refeed offsets for subscription api
 * Allows to add new event types to subscription's topology (stored in Zookeeper)

Note: this is shared code for verified PoC (#1554) and the planned rollout approach #1558.

### DIFF
--- a/api-cursors/src/main/java/org/zalando/nakadi/service/CursorsService.java
+++ b/api-cursors/src/main/java/org/zalando/nakadi/service/CursorsService.java
@@ -85,7 +85,11 @@ public class CursorsService {
             final Subscription subscription = subscriptionCache.getSubscription(subscriptionId);
             authorizationValidator.authorizeSubscriptionView(subscription);
             authorizationValidator.authorizeSubscriptionCommit(subscription);
-            validateSubscriptionCommitCursors(subscription, cursors);
+            // Note: not checking directly here if cursors belong to the subscription.
+            // Such check happens indirectly (in validateStreamId) against subscription's topology.
+            // The subscription's topology might contain additional implicit event types (like DLQ event type),
+            // for which we allow committing (but not resetting) offsets.
+            checkCursorsStorageAvailability(cursors);
             try (ZkSubscriptionClient zkClient = zkSubscriptionFactory.createClient(subscription)) {
                 validateStreamId(cursors, streamId, zkClient, subscriptionId);
                 return zkClient.commitOffsets(
@@ -238,11 +242,8 @@ public class CursorsService {
         }
     }
 
-    private void validateSubscriptionCommitCursors(final Subscription subscription,
-                                                   final List<NakadiCursor> cursors)
+    private void checkCursorsStorageAvailability(final List<NakadiCursor> cursors)
             throws UnableProcessException {
-        validateCursorsBelongToSubscription(subscription, cursors);
-
         cursors.forEach(cursor -> {
             try {
                 cursor.checkStorageAvailability();
--- a/core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionClient.java
+++ b/core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionClient.java
@@ -1,10 +1,8 @@
 package org.zalando.nakadi.service.subscription.zk;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import java.util.function.Function;
 import org.apache.commons.codec.binary.Hex;
 import org.zalando.nakadi.domain.EventTypePartition;
-import org.zalando.nakadi.exceptions.runtime.NakadiBaseException;
 import org.zalando.nakadi.exceptions.runtime.NakadiRuntimeException;
 import org.zalando.nakadi.exceptions.runtime.OperationTimeoutException;
 import org.zalando.nakadi.exceptions.runtime.ServiceTemporarilyUnavailableException;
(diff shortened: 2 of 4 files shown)


core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionClient.java
```diff
@@ -202,21 +203,18 @@ public Partition[] getPartitions() {
...
+                        .findFirst();
+                if (selected.isPresent()) {
+                    resultPartitions.set(selected.getAsInt(), newValue);
+                } else {
+                    resultPartitions.add(newValue);
                 }
-                resultPartitions[selectedIdx] = newValue;
             }
-            return new Topology(resultPartitions, Optional.ofNullable(version).orElse(0) + 1);
+            return new Topology(resultPartitions.toArray(new Partition[0]), Optional.ofNullable(version).orElse(0) + 1);
```
a1exsh, line 217:
```suggestion
            return new Topology(resultPartitions.toArray(Partition[]::new), Optional.ofNullable(version).orElse(0) + 1);
```

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionClient.java
```diff
@@ -202,21 +203,18 @@ public Partition[] getPartitions() {
         }
 
         public Topology withUpdatedPartitions(final Partition[] partitions) {
```
a1exsh, line 205:
```suggestion
        public Topology withUpdatedPartitions(final Partition[] newPartitions) {
```
It's confusing when the parameter hides the class field.

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionClient.java
```diff
@@ -202,21 +203,18 @@ public Partition[] getPartitions() {
...
-            final Partition[] resultPartitions = Arrays.copyOf(this.partitions, this.partitions.length);
+            final var resultPartitions = new ArrayList<>(Arrays.asList(this.partitions));
             for (final Partition newValue : partitions) {
```
a1exsh, line 207: Since we are changing the implementation here, I'd rather go with making two maps and doing something like: `oldPartitions.addAll(newPartitions)`.

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionClient.java
adyach, line 205: @1u0 could you elaborate on why you change the implementation ?

adyach: it does not look like related to the change set

api-cursors/src/main/java/org/zalando/nakadi/service/CursorsService.java
```diff
@@ -254,11 +264,16 @@ private void validateSubscriptionCommitCursors(final Subscription subscription,
     }
 
     private void validateCursorsBelongToSubscription(final Subscription subscription,
-                                                     final List<NakadiCursor> cursors)
+                                                     final List<NakadiCursor> cursors,
+                                                     final Collection<String> additionalAllowedEventTypes)
             throws UnableProcessException {
-        final List<String> wrongEventTypes = cursors.stream()
+        var stream = cursors.stream()
```
adyach, line 270: It looks like having the event type checked all the time will reduce amount of code. I would just add it by default and check it for every subscription in `validateCursorsBelongToSubscription()` method. It will be smaller change like that.

```
  final List<String> wrongEventTypes = cursors.stream().
  .filter(et -> !subscription.getEventTypes().contains(et))
  .filter(et -> !et.equals(dlqEventType))
```

The actual correct way to do it would be to check the topology from the zookeeper. That can be done by refactoring validateStreamId: separating the logic of `getTopology()` into different method. Once you have a topology you can execute necessary check like if dlqEventType in topology then enrich Subscription with dlqEventType by creating ephemeral subscription and reusing the logic `validateCursorsBelongToSubscription`, it will be easier to understand / read after a couple of months.

1u0: +1 for checking cursors against the subscription's topology (vs. list of event types in subscription).

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionClient.java
1u0: Such method call would effectively delegate to `.toArray(new Partition[0])`.
What would be the benefit of such indirection?

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/ZkSubscriptionClient.java
1u0: > * Allows to add new event types to subscription's topology (stored in Zookeeper)
> Note: this is shared code for verified PoC (https://github.com/zalando/nakadi/pull/1554) and the planned rollout approach https://github.com/zalando/nakadi/pull/1558.

It's part of shared change that also needed for #1558.

api-cursors/src/main/java/org/zalando/nakadi/service/CursorsService.java
adyach: are you going to add that ?

1u0: This is already added in the latest PR update.

adyach:
👍

1u0:
:+1:

PR #1561, opened 2023-10-16
Add unprocessable event policy on consumption

a1exsh:
This allows the user to choose between the current implementation of skipping the unprocessable events, or sending those events to the dead letter queue.


api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -370,6 +382,49 @@ private void streamToOutput(final boolean streamTimeoutReached) {
...
+                        new JSONObject()
+                        .put("eid", getContext().getUuidGenerator().randomUUID().toString())
+                        .put("occurred_at", Instant.now().toString()));
+
+        final JSONArray deadLetterBatch =
+                new JSONArray().put(deadLetter);
+
+        final EventPublishResult result;
+        try {
+            result = getContext().getEventPublisher().publishInternal(
```
samizzy, line 416: Does it retry publishing internally on failure?

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
```diff
@@ -751,22 +817,37 @@ public void whenIsLookingForDeadLetterAndSendAllEventsOneByOneThenBackToNormalBa
...
+    public void shouldSkipDeadLetterAndConsumptionToBeContinuedAndDeadLetterFoundInTheQueueLater()
+            throws IOException, InterruptedException {
+
         final EventType eventType = NakadiTestUtils.createBusinessEventTypeWithPartitions(4);
 
         publishBusinessEventWithUserDefinedPartition(eventType.getName(),
-                50, i -> String.format("{\"foo\":\"bar%d\"}", i), i -> String.valueOf(i % 4));
+                50, i -> String.format("bar%d", i), i -> String.valueOf(i % 4));
+
+        final String poisonPillValue = "bar10";
```
samizzy, line 828: trying to understand why `bar10` specifically is the poison pill and not the values before it?

samizzy (commented):
One other thing to take care of is, either to have this dlq event type created or somehow disallow `DEAD_LETTER_QUEUE` policy now.

a1exsh:
> One other thing to take care of is, either to have this dlq event type created or somehow disallow `DEAD_LETTER_QUEUE` policy now.

Right, this remains to be followed-up on.  I tend to keep the policy allowed even if the event type is not present — that's not harmful, as long as no one actually depends on it working.

core-common/src/main/java/org/zalando/nakadi/annotations/validation/DeadLetterAnnotationValidator.java
```diff
@@ -10,20 +12,34 @@ public class DeadLetterAnnotationValidator implements
...
     public boolean isValid(final Map<String, String> annotations, final ConstraintValidatorContext context) {
         if (annotations == null) {
             return true;
         }
+        if (!isValidMaxEventSendCount(annotations, context)) {
+            return false;
+        }
+        if (!isValidUnprocessableEventPolicy(annotations, context)) {
+            return false;
+        }
```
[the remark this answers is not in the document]
a1exsh: > validation if only one key is null can be done in following way: `...`

Yes, but I prefer my way ;)

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
a1exsh: That depends on the configuration.  Currently — no.

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/HilaAT.java
a1exsh: > trying to understand why `bar10` specifically is the poison pill and not the values before it?

We just want some value in the middle of the stream.  We may even select it randomly, but that'd make it harder to debug if the test stops working for some reason.

samizzy: I took a look again and now understand the logic.

samizzy:
:+1:

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -300,11 +307,16 @@ private void streamToOutput(final boolean streamTimeoutReached) {
...
+
                     final ConsumedEvent failedEvent = toSend.remove(0);
-                    // todo: skip or to dlq
+
                     LOG.warn("Skipping event {} from partition {} due to failed commits count {}",
                             failedEvent.getPosition(), etp, partition.getFailedCommitsCount());
 
+                    if (getContext().getUnprocessableEventPolicy() == UnprocessableEventPolicy.DEAD_LETTER_QUEUE) {
+                        sendToDeadLetterQueue(failedEvent, partition.getFailedCommitsCount());
+                    }
```
1u0, line 318: This (and `sendToDeadLetterQueue(...)`) mostly interact with context's fields.
What about instead delegate this implementation to the context?

I.e., smt. like (here):
```
getContext().processFailedEvent(failedEvent, partition.getFailedCommitsCount());
```

And the rest of the code would be implementation details in the `StreamingContext`.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -370,6 +382,49 @@ private void streamToOutput(final boolean streamTimeoutReached) {
         }
     }
 
+    private void sendToDeadLetterQueue(final ConsumedEvent event, final int failedCommitsCount) {
+        // TODO: how do we handle AVRO here?
+
+        final String failedEventString = new String(event.getEvent(), StandardCharsets.UTF_8);
+        final JSONObject failedEvent = new JSONObject(failedEventString);
```
1u0, line 389: IIUC, this is basically de-serializes the failed event.
Alternatively, the failed event could be stored as string (or byte array? relevant to my next comment).
For JSON type tha would involve only JSON-escaping that string, which should be more efficient than de-serialization, I think.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
a1exsh: Yeah.  Or we may have something like a dedicated service for publishing to DLQ.  That's up to some refactoring.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
a1exsh: You are right, but efficiency should not concern us for this use case, IMO. :)

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -370,6 +382,49 @@ private void streamToOutput(final boolean streamTimeoutReached) {
...
+                        .put("eid", getContext().getUuidGenerator().randomUUID().toString())
+                        .put("occurred_at", Instant.now().toString()));
+
+        final JSONArray deadLetterBatch =
+                new JSONArray().put(deadLetter);
+
+        final EventPublishResult result;
+        try {
+            result = getContext().getEventPublisher().publishInternal(
+                    deadLetterBatch.toString(), getContext().getDeadLetterQueueEventTypeName(), null);
```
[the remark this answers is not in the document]
a1exsh: Avro is a big question mark wrt. to DLQ now.

a1exsh:
@samizzy @1u0 I've included the DLQ store event type definition in the app resources — that enables local testing and makes the code shorter as well.  Please have another look.

samizzy:
:+1:

a1exsh:
:+1:

acceptance-test/src/acceptance-test/java/org/zalando/nakadi/webservice/hila/StreamBatch.java
```diff
@@ -1,45 +1,65 @@
...
-import static java.util.Collections.unmodifiableList;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.hasProperty;
+import static org.hamcrest.Matchers.isA;
+import static uk.co.datumedge.hamcrest.json.SameJSONAs.sameJSONObjectAs;
 
 @Immutable
```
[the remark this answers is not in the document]
a1exsh: > Minor: remove this annotation (now, the objects stored in the `events` list are mutable).

Does the annotation actually promise that much (deep immutability)?  It was not holding to that promise in the first place anyway.

PR #1566, opened 2023-10-31
Feed events one by one until the end of problematic batch

a1exsh:
The current behaviour that Auto DLQ goes back to normal mode once it found one bad event.  It can happen that there are more events after that bad one. Current behaviour will create longer bad events identification time.

Internal ticket: 1537


adyach:
👍

PR #1563, opened 2023-10-18
Big cleanup of Hamcrest imports

a1exsh:
It got messy over the time, mainly due to auto-import suggestions from IDEs.

PR #1567, opened 2023-10-31
Add unprocessable event policy on consumption

a1exsh:
#1561
#1563


api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -370,6 +382,49 @@ private void streamToOutput(final boolean streamTimeoutReached) {
...
+        final EventPublishResult result;
+        try {
+            result = getContext().getEventPublisher().publishInternal(
+                    deadLetterBatch.toString(), getContext().getDeadLetterQueueEventTypeName(), null);
+        } catch (final Exception e) {
+            throw new DeadLetterQueueStoreException("Failed to store an event to the dead letter queue", e);
+        }
+
+        if (result.getStatus() != EventPublishingStatus.SUBMITTED) {
+            throw new DeadLetterQueueStoreException("Failed to store an event to the dead letter queue: " +
```
adyach, line 423: it looks like it blocks consumption of the event type in case there an issue with completely unrelated event type. we need to rethink it and move into direction of non blocking processing.

adyach:
@a1exsh were those changes reviewed? I went quickly though the change, one comment I have we can address in the follow up.

a1exsh:
> @a1exsh were those changes reviewed?

You should see that both linked PRs were merged after a review.

api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
a1exsh: > it looks like it blocks consumption of the event type in case there an issue with completely unrelated event type

Yes, but only if Auto-DLQ is used and there actually some events that need to be stored, the likelihood of that coinciding with any issues with the DLQ store event type is pretty slim.

> we need to rethink it and move into direction of non blocking processing.

I don't think that's possible at all.  The whole point is to avoid data loss.

adyach:
👍

PR #1572, opened 2023-11-24
Reset DLQ state on cursor reset

fghibellini:
# One-line summary

Zalando ticket : ARUHA-1531

## Description
If a user resets the cursors of a subscription the DLQ state of any of the updated partitions should be reset to the default state.

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes


adyach:
@fghibellini does it face the correct upstream branch ?

PR #1573, opened 2023-11-27
[Fix] DLQ mode batches sent

samizzy:
# One-line summary

This PR makes sure that in DLQ mode, Nakadi sends only a single batch of one event and only uptil dead letter offset.

## Description
Problems:
1. In DLQ mode, Nakadi sends single event batches but doesnt wait for commit for the sent event and sends more single event batches until it reaches configured batch size during stream connection.
2. In DLQ mode, when consumer commits a single batch event offset, Nakadi sends more single event batches than expected dead letter offset (due to flushes).

### DIFF
--- a/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
+++ b/api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
@@ -348,6 +348,11 @@ class StreamingState extends State {
                 messagesAllowedForPartition -= toSend.size();
                 LOG.error("messagesAllowedToSend after reduction is: {}", messagesAllowedToSend);
                 LOG.error("messagesAllowedForPartition after reduction is: {}", messagesAllowedForPartition);
+
+                //break after sending 1 event, as there is nothing more to send until commit arrives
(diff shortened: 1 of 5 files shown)


api-consumption/src/main/java/org/zalando/nakadi/service/subscription/state/StreamingState.java
```diff
@@ -298,15 +298,14 @@ private void streamToOutput(final boolean streamTimeoutReached) {
                         currentTimeMillis,
                         Math.min(getBatchLimitEvents(), messagesAllowedForPartition),
                         getParameters().batchTimeoutMillis,
-                        streamTimeoutReached);
-
+                        streamTimeoutReached,
+                        inDlqMode(partition));
```
a1exsh, line 302: It looks more promising to make the decision in `getMessagesAllowedToSend()`, by passing the current partition there.  It should then return either 1 or 0 for any partition that is currently in the auto-DLQ search mode.  Wdyt?

samizzy: moved logic to `getMessagesAllowedForPartition`

a1exsh:
:+1:

PR #1579, opened 2023-12-05
Workaround: reset subscriptions' running sessions on attaching DLQ pa…

1u0:
# One-line summary

## Description

This change resets (stops) subscription's running sessions upon attaching to the subscription the DLQ partitions. This would force a consuming application to re-connect to subscription (create new sessions) where they would be able to observe (consume) partitions of DLQ into addition to partitions from normal event types.

Note: this is a workaround (quick) solution. A better approach would be to rebalance partitions among running session without stopping them.

### DIFF
--- a/app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
+++ b/app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
@@ -7,6 +7,7 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Service;
 import org.zalando.nakadi.cache.EventTypeCache;
+import org.zalando.nakadi.config.NakadiSettings;
 import org.zalando.nakadi.domain.ConsumedEvent;
 import org.zalando.nakadi.domain.EventTypePartition;
 import org.zalando.nakadi.domain.Feature;
@@ -29,6 +30,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 @Service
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
```diff
@@ -153,7 +158,19 @@ private void addDlqPartitionsToSubscription(
...
+        client.updateTopology(topology -> {
+                final var newPartitions = selectMissingPartitions(topology.getPartitions(), unassignedDlqPartitions);
+                if (!hasNewPartitions[0] && newPartitions.length != 0) {
+                    hasNewPartitions[0] = true;
+                }
+                return newPartitions;
+        });
+        if (hasNewPartitions[0]) {
+            client.closeSubscriptionStreams(
+                    () -> LOG.info("subscription streams were closed, after repartitioning"),
```
adyach, line 171: let add subscription id into the message

adyach:
👍

app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
```diff
@@ -153,7 +158,23 @@ private void addDlqPartitionsToSubscription(
                 .map(nc -> cursorConverter.convertToNoToken(nc))
                 .collect(Collectors.toList()));
 
-        client.updateTopology(topology -> selectMissingPartitions(topology.getPartitions(), unassignedDlqPartitions));
+        final boolean[] hasNewPartitions = {false};
+        client.updateTopology(topology -> {
+                final var newPartitions = selectMissingPartitions(topology.getPartitions(), unassignedDlqPartitions);
+                if (!hasNewPartitions[0] && newPartitions.length != 0) {
```
a1exsh, line 164: `hasNewPartitions[0]` — isn't it always `false` at this point?

PR #1581, opened 2023-12-13
Rebalance subscription's sessions on DLQ event type attachment to topology

1u0:
## Description

This change rebalances subscription's topology among current sessions dynamically, upon attaching the DLQ event type to the subscription's topology.

This is considered to be more graceful behavior (towards consuming applications), compared to the current (workaround, #1579) approach where the sessions got stopped (forcing consuming applications to reconnect).

### DIFF
--- a/app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
+++ b/app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
@@ -146,30 +146,31 @@ public class DlqRedriveEventTypeAttachmentJob {
             final Subscription subscription,
             final List<Partition> unassignedDlqPartitions) throws Exception {
 
-        final ZkSubscriptionClient client = subscriptionClientFactory.createClient(subscription);
+        try (ZkSubscriptionClient client = subscriptionClientFactory.createClient(subscription)) {
 
-        // idempotent, does not overwrite existing offsets
-        client.createOffsetZNodes(
-                unassignedDlqPartitions.stream()
-                .map(p -> p.getPartition())
-                .map(p -> new Cursor(p, Cursor.BEFORE_OLDEST_OFFSET))
-                .map(c -> cursorConverter.convert(dlqRedriveEventTypeName, c))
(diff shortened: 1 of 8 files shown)


core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/SubscriptionRebalancer.java
```diff
@@ -1,4 +1,4 @@
-package org.zalando.nakadi.service.subscription;
```
a1exsh, line 1: Is it necessary to move it to a different package, or it's more of a style thing?

app/src/main/java/org/zalando/nakadi/service/job/DlqRedriveEventTypeAttachmentJob.java
```diff
@@ -167,13 +166,10 @@ private void addDlqPartitionsToSubscription(
...
-        // shortcut to enable new partitions for streaming, otherwise they will stay unassigned. 
-        // the better way is to rebalance them.
         if (hasNewPartitions[0]) {
-            client.closeSubscriptionStreams(
-                    () -> LOG.info("Subscription `{}` streams were closed due to addition of " +
-                                   "Nakadi DLQ Event Type", subscription.getId()),
-                    TimeUnit.SECONDS.toMillis(nakadiSettings.getMaxCommitTimeout()));
+            LOG.info("Rebalancing `{}` subscription's sessions due to addition of " +
+                    "Nakadi DLQ Event Type", subscription.getId());
+            client.rebalanceSessions();
```
a1exsh, line 172: Btw, don't we need to close `client` properly? :zipper_mouth_face:

core-common/src/main/java/org/zalando/nakadi/service/subscription/zk/SubscriptionRebalancer.java
1u0: This class is now used only from one place (not counting the class's test).
It was slightly relocated with this change in order to keep it package private (by moving it to the package where it's used).

a1exsh:
:+1:

adyach:
👍

PR #1588, opened 2024-01-27
Use opensource docker registry

roskenet:
# Use Zalando opensource docker registry

Starting a nakadi local environment as documented in the README fails with 403 because of the registry in the Dockerfile.
Use Zalando's opensource registry instead.


a1exsh:
@roskenet have you tried this: `pierone login --url container-registry.zalando.net` ?

roskenet:
@a1exsh Then we shouldn't opensource it.

a1exsh:
> @a1exsh Then we shouldn't opensource it. ;-) It doesn't start up for a mortal Non-Zalando-Dev.

True, and it was the plan since 3 years now anyway.  The announcement will follow.

PR #1591, opened 2024-02-12
Log debug topic id header for misplaced events

a1exsh:
1. Reuse `HeaderTag` for adding the debug header.
2. Use `EnumMap` for slightly more efficiency.
3. Log it when we detect misplaced event.


a1exsh:
:+1:

1u0:
:+1:

PR #1592, opened 2024-02-22
Mention that event_stream.read|write scopes are not used

ePaul:
# One-line summary

Mention that event_stream.read|write scopes are not used

> Zalando ticket : –

## Description

This adds two lines to the API definition indicating that the nakadi.event_stream.read|write scopes are not used in Zalando's main Nakadi installation.
(That's the case since mid 2017.)

If you think we should instead remove all traces of these scopes from the whole document, I can do that too.

## Review
- [ ] Tests
- [ ] Documentation

## Deployment Notes
This doesn't change any behavior, just the documentation.


a1exsh:
Thanks @ePaul , will update internally.

